PR #3, opened 2020-06-26
Hide cd bm-inventory error

iranzo:
Hide cd bm-inventory error if folder is not yet existing


iranzo:
@tsorya is there anything else to get this merged? 

Thanks!

PR #6, opened 2020-06-28
README fixes

rollandf:
- Fix Clone command to new repository location

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ On the bare metal host:
 ```bash
 dnf install -y git make
 cd /home/test # don't do it on /root it will breaks build image mounts and fail to run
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,6 +3,7 @@ This project deploys the OpenShift Assisted Installer in Minikube and spawns lib
 
 # Prerequisites
 - CentOS 8 or RHEL 8 host
+- Skipper https://github.com/stratoscale/skipper
```
tsorya, line 6: skipper will installed in create_full_environment, not sure we need to ask it as Prerequisites. 
If we are adding it here we need to add all the packages we install in create_full_environment

PR #11, opened 2020-06-30
Support Route53 DNS env variables

danielerez:
Added support for ROUTE53_SECRET and BASE_DNS_DOMAINS env vars.
* ROUTE53_SECRET: base64 of ./aws/credentials file.
* BASE_DNS_DOMAINS: should be in the following format,
name1:id1/provider1,name2:id2/provider2


danielerez:
@tsorya @romfreiman @ori-amizur @filanov

Makefile
```diff
@@ -8,6 +8,7 @@ NUM_WORKERS := $(or $(NUM_WORKERS),0)
 STORAGE_POOL_PATH := $(or $(STORAGE_POOL_PATH), $(PWD)/storage_pool)
 SSH_PUB_KEY := $(or $(SSH_PUB_KEY),$(shell cat ssh_key/key.pub))
 PULL_SECRET :=  $(or $(PULL_SECRET), $(shell if ! [ -z "${PULL_SECRET_FILE}" ];then cat ${PULL_SECRET_FILE};fi))
+ROUTE53_SECRET := $(or $(ROUTE53_SECRET), "")
```
tsorya, line 11: what will happened with it? how it will be use?
just to understand, don't see any line that uses it,

danielerez: oh sorry, the context is in the inventory PR: https://github.com/filanov/bm-inventory/pull/317

tsorya: got it

romfreiman:
I'm missing the validation phase that it's actually works. Also, we have to skip the dns part of terrafrom, right?

danielerez:
> I'm missing the validation phase that it's actually works. Also, we have to skip the dns part of terrafrom, right?

yeah, handling on consecutive PRs.

scripts/assisted_deployment.sh
```diff
@@ -5,6 +5,11 @@ function destroy_all() {
 }
 
 function set_dns() {
```
romfreiman, line 7: will the dns be deleted on destroy all? I mean if we deploy once with NM configuration but then want to deploy without it?
@tsorya @YuviGold

PR #13, opened 2020-07-01
Add Install Guide

iranzo:
Add Install guide

### DIFF
--- a/GUIDE.md
+++ b/GUIDE.md
@@ -1,0 +1,165 @@
+# Assisted Installer Unofficial Install Guide
+
+## Prerequisites
+
+1. Install CentOS8/RHEL8 on the assisted installer host<sup id="a1">[1](#f1)</sup>
+   - This host will run minikube and the UI for deploying OpenShift on Bare Metal
+1. Setup DHCP/DNS records for the following OpenShift nodes and VIPs. List includes
+   - Master nodes
+   - Worker nodes
+   - API VIP
+   - Ingress VIP
+     > NOTE: dnsmasq can be used to setup [DHCP](https://openshift-kni.github.io/baremetal-deploy/4.4/Deployment.html#creating-dhcp-reservations-using-dnsmasq-option2_ipi-install-prerequisites) and [DNS](https://openshift-kni.github.io/baremetal-deploy/4.4/Deployment.html#creating-dns-records-using-dnsmasq-option2_ipi-install-prerequisites). Out of scope for this document but I've included links on how to do it.
+1. Ensure these values are properly set via nslookup and dig commands from your assisted installer host
+1. Download your Pull Secret from https://cloud.redhat.com/openshift/install/metal/user-provisioned
+
+## Deployment
+
+Procedure
+
+As `$USER` user with sudo privileges ,
+
+1.  Generate an SSH key if not already available (`ssh_key/key`)
+
+        [$USER@assisted_installer ~]# ssh-keygen -t rsa -f ~/.ssh/id_rsa -P ''
+
+1.  Install git and make on your CentOS assisted installer host if not already present
+
+        [$USER@assisted_installer ~]# dnf install -y make git
+
+1.  Create a directory under `~/` labeled `assisted-installer` and `git clone` the `test-infra` git repository
+
+        [$USER@assisted_installer ~]# mkdir ~/assisted-installer
+        [$USER@assisted installer assisted-installer]# cd ~/assisted-installer/
+        [$USER@assisted_installer assisted-installer]# git clone https://github.com/openshift/assisted-test-infra
+        [$USER@assisted_installer assisted-installer]# cd assisted-test-infra
+
+1.  Setup the assisted installer's environment by running the `install_environment.sh` script
+
+        [$USER@assisted_installer assisted-test-infra]# ./scripts/install_environment.sh
+
+1.  Once complete, run the `create_full_environment` using `make`
+
+        [$USER@assisted_installer assisted-test-infra]# make create_full_environment
+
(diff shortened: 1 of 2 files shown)


iranzo:
@rlopez133 @DirectedSoul1 @romfreiman 

Can you give it a look to compare about the original document to see if there's something to fix?

Thx!

GUIDE.md
```diff
@@ -0,0 +1,167 @@
+# Assisted Installer Unofficial Install Guide
+
+> DISCLAIMER: These are just steps we have used to do the deployments. They are by no means "official" ; it is just our team's experience in how we've been able to deploy and we are providing a reproducible method. Input on making this better is welcomed and encouraged!
+
```
rlopez133, line 4: We should remove this disclaimer.

GUIDE.md
```diff
@@ -0,0 +1,167 @@
...
+        [$USER@assisted_installer ~]# for i in <master0-idrac-ip> <master1-idrac-ip> <master2-idrac-ip>; do podman run --net=host idracbootfromiso -r $i -u <idrac-user> -p "<idrac-pw>" -i http://       <host-ip>:8080/live.iso; done
+
+    > NOTE1: Ensure to include the proper host **iDRAC IPs**, **iDRAC user** and **iDRAC password**. This for loop assumes iDRAC user and iDRAC password are the same, if different adjust the shell command accordingly.
+
+
+    > NOTE2: Also make sure to **remove/eject the older ISO** which were mounted using `racadm` command as shown below
+
+        sshpass -p '*****' ssh root@<iDRAC-IP> racadm remoteimage -d
+
+1.  The `live.iso` should reboot the nodes and boot them into a Fedora Live image. Once it has done this, shortly you will notice the nodes becoming discoverable for your cluster via the `http://<host-ip>:6008/clusters/<cluster-id>` dashboard.
```
rlopez133, line 136: If we get the Ansible playbook merged, this will all change.

iranzo: it can be amended later :)

DirectedSoul1: yeah, also we can keep this guide and provide steps for ansible in a sub-section? just a suggestion.

GUIDE.md
```diff
@@ -0,0 +1,165 @@
...
+
+As `$USER` user with sudo privileges ,
+
+1.  Generate an SSH key
+
+        [$USER@assisted_installer ~]# ssh-keygen -t rsa -f ~/.ssh/id_rsa -P ''
+
+1.  Install git and make on your CentOS assisted installer host if not already present
+
+        [$USER@assisted_installer ~]# dnf install -y make git
```
tsorya, line 28: lets add make here

GUIDE.md
```diff
@@ -0,0 +1,165 @@
...
+1.  Create a directory under `~/` labeled `assisted-installer` and `git clone` the `test-infra` git repository
+
+        [$USER@assisted_installer ~]# mkdir ~/assisted-installer
+        [$USER@assisted installer assisted-installer]# cd ~/assisted-installer/
+        [$USER@assisted_installer assisted-installer]# git clone https://github.com/openshift/assisted-test-infra
+        [$USER@assisted_installer assisted-installer]# cd assisted-test-infra
+
+1.  Setup the assisted installer's environment by running the `install_environment.sh` script
+
+        [$USER@assisted_installer assisted-test-infra]# ./scripts/install_environment.sh
```
tsorya, line 39: no need. just run "make create_full_environment" and move make to the line where git is installed.

GUIDE.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Deployment
+
+Procedure
+
+As `$USER` user with sudo privileges ,
+
+1.  Generate an SSH key
+
+        [$USER@assisted_installer ~]# ssh-keygen -t rsa -f ~/.ssh/id_rsa -P ''
```
tsorya, line 24: in test-infra u have have folder with ssh keys. so u can just use them
ssh_key/key

GUIDE.md
```diff
@@ -0,0 +1,165 @@
...
+        [$USER@assisted_installer assisted-test-infra]# make run
+        .
+        .
+        .
+
+
+        deploy_ui.sh: OCP METAL UI can be reached at http://<host-ip>:6008
+        deploy_ui.sh: Done
+
+1. Once the UI has finished deploying, on a browser, access it via your host's IP and use port 6008, URL will be something like `http://<host-ip>:6008`
```
tsorya, line 60: all this can be done by running "make download_iso"

GUIDE.md
iranzo: `make` is already there, something I'm missing?

GUIDE.md
iranzo: so after the install make, git, just run 'make create_full_environment' right?

GUIDE.md
tsorya: my fault

GUIDE.md
tsorya: right

GUIDE.md
rlopez133: @tsorya while this is true, I think the real question is what do you want the experience to be like? Would you prefer people use the UI or use the make download_iso?

GUIDE.md
rlopez133: @tsorya @iranzo However,  if you have a folder with ssh keys, there is an assumption that ssh keys exist. Likely a better approach here would be to say generate an ssh key if not available.

GUIDE.md
tsorya: u are right but i think we will have 2 types of users so we can at least describe the second option?

tsorya:
looks great. thanks

PR #14, opened 2020-07-01
Do not use chunked read for S3 downloads

avishayt:
This is not compatible with setting content-length


romfreiman:
@tsorya

romfreiman:
Does it work?

tsorya:
@romfreiman yap, tried

PR #17, opened 2020-07-02
start_discovery: handle managed-dns-domains

danielerez:
In order to ensure registration of unique DNS records,
the cluster name should be randomly generated.
Hence, passed --managed-dns-domains to start_discovery.py
as an indication. Also, parsed the base domain from
args.managed_dns_domains map.

### DIFF
--- a/discovery-infra/start_discovery.py
+++ b/discovery-infra/start_discovery.py
@@ -221,6 +221,9 @@ def nodes_flow(client, cluster_name, cluster):
 def main():
     client = None
     cluster = {}
+    if args.managed_dns_domains != ":":
+        args.cluster_name = ""
+        args.base_dns_domain = args.managed_dns_domains.split(':')[0]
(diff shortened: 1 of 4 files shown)


danielerez:
@tsorya @romfreiman

discovery-infra/start_discovery.py
```diff
@@ -311,6 +314,9 @@ def main():
         type=str,
         default="redhat.com",
     )
+    parser.add_argument(
+        "-mD", "--managed-dns-domains", help="Managed DNS domains", type=str, default=""
```
romfreiman, line 318: cam u maybe write more descriptive help here. Not sure that I undestand what should be provided. Even give an example

danielerez: sure, added explanation and example, also updated the readme in: https://github.com/openshift/assisted-test-infra/pull/16

danielerez: added DNS validation in: https://github.com/openshift/assisted-test-infra/pull/21

PR #20, opened 2020-07-02
Support docker build multistage

YuviGold:


Dockerfile.test-infra
```diff
@@ -1,3 +1,4 @@
+FROM quay.io/ocpmetal/bm-inventory:latest AS inventory
```
ronniel1, line 1: This should not always be "latest"
The tag should be passed as a m argument

tsorya: this should be param

tsorya: look here https://github.com/openshift/assisted-test-infra/pull/12/files

YuviGold: Fixed.

romfreiman: @ronniel1 @tsorya is this something that should be addressed now or can wait?

filanov:
Worked for me.
But it wasn't easy. i had to build the image again (it wasn't in the documentation).
So please update the documentation.

PR #21, opened 2020-07-05
validate DNS domains on cluster install

danielerez:
After cluster installation, validate DNS domain resolvability of api. and ingress.apps. addresses.

### DIFF
--- a/discovery-infra/start_discovery.py
+++ b/discovery-infra/start_discovery.py
@@ -6,6 +6,7 @@ import ipaddress
 import json
 import os
 import pprint
+import socket
 import time
 import uuid
 from distutils.dir_util import copy_tree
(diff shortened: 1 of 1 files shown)


discovery-infra/start_discovery.py
```diff
@@ -169,6 +170,25 @@ def _create_node_details(cluster_name):
...
+def validate_dns(client, cluster_id):
+    if not args.managed_dns_domains:
+        # 'set_dns' (using dnsmasq) is invoked after nodes_flow
+        return
+
+    cluster = client.cluster_get(cluster_id)
+    api_address = "api.{}.{}".format(cluster.name, cluster.base_dns_domain)
+    ingress_address = "ingress.apps.{}.{}".format(cluster.name, cluster.base_dns_domain)
+    try:
+        api_vip = socket.gethostbyname(api_address)
```
romfreiman, line 182: mm do u think it's enough? I guess even libvirt will return same results, no?
Maybe there is a python package that can tell u who did the resoltion? (instead of parsing dig)

danielerez: > mm do u think it's enough? I guess even libvirt will return same results, no?
> Maybe there is a python package that can tell u who did the resoltion? (instead of parsing dig)

Not sure it's crucial as set_dns is invoked only after this flow, so this indication should be good enough in this point. 
Any way, I'll try getting dns authority using [http://www.dnspython.org/](dnspython) in a separate PR to not block this.

discovery-infra/start_discovery.py
```diff
@@ -169,6 +170,25 @@ def _create_node_details(cluster_name):
...
+        ingress_vip = socket.gethostbyname(ingress_address)
+        if api_vip == cluster.api_vip and ingress_vip == cluster.ingress_vip:
+            log.info("DNS domains are resolvable")
+        else:
+            raise Exception
```
tsorya, line 187: Exception("DNS domains are not resolvable") 
i except, lets add log.exception with relevant message

tsorya: and raise

danielerez: added

discovery-infra/start_discovery.py
```diff
@@ -169,6 +170,25 @@ def _create_node_details(cluster_name):
...
+    ingress_address = "ingress.apps.{}.{}".format(cluster.name, cluster.base_dns_domain)
+    try:
+        api_vip = socket.gethostbyname(api_address)
+        ingress_vip = socket.gethostbyname(ingress_address)
+        if api_vip != cluster.api_vip or ingress_vip != cluster.ingress_vip:
+            raise Exception("DNS domains are not resolvable")
+
+        log.info("DNS domains are resolvable")
+    except:
+        raise Exception("Failed to resolve DNS domains")
```
tsorya, line 189: `log.exception("Failed to resolve DNS domains")
raise
`

tsorya:
@danielerez is still relevant pr?

danielerez:
> @danielerez is still relevant pr?

nope, already handled by https://github.com/openshift/assisted-test-infra/pull/22
closing.

PR #22, opened 2020-07-06
validate DNS authority using dnspython

danielerez:
Used dns.resolver (dnspython) to validate DNS authority and resolvability.

Note: added 'dnspython' package to requirements, so a new image of test-infra should be pushed to quay.

### DIFF
--- a/discovery-infra/start_discovery.py
+++ b/discovery-infra/start_discovery.py
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 
 import argparse
+import dns.resolver
 import ipaddress
 import json
 import os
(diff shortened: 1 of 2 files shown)


danielerez:
@romfreiman @tsorya

discovery-infra/start_discovery.py
```diff
@@ -169,6 +170,31 @@ def _create_node_details(cluster_name):
...
 
 
+def validate_dns(client, cluster_id):
+    if args.managed_dns_domains == ":":
+        # 'set_dns' (using dnsmasq) is invoked after nodes_flow
+        return
+
+    cluster = client.cluster_get(cluster_id)
+    api_address = "api.{}.{}".format(cluster.name, cluster.base_dns_domain)
+    ingress_address = "ingress.apps.{}.{}".format(cluster.name, cluster.base_dns_domain)
```
tsorya, line 180: lets add log with dns domain name.
Something like , validating domains ... exists.

requirements.txt
```diff
@@ -8,3 +8,4 @@ requests==2.20.0
 retry==0.9.2
 tqdm==4.46.0
 waiting==1.4.1
+dnspython==1.16.0
```
tsorya, line 11: we will have lint validation and it verifies that packages in requirements are in alphabetical order

discovery-infra/start_discovery.py
danielerez: done

requirements.txt
danielerez: done

PR #26, opened 2020-07-07
update configmap: empty default for base domains

danielerez:
BASE_DNS_DOMAINS env var is defaulted to an empty string
    ('""') in Makefile if unset. Thus, handles accordingly in
    update_bm_inventory_cm when added to the configmap.


discovery-infra/update_bm_inventory_cm.py
```diff
@@ -21,7 +21,7 @@
     ("INVENTORY_PORT", ""),
     ("AGENT_DOCKER_IMAGE", ""),
     ("KUBECONFIG_GENERATE_IMAGE", ""),
-    ("BASE_DNS_DOMAINS", ":"),
+    ("BASE_DNS_DOMAINS"),
```
tsorya, line 24: u must have have 2 params here, it will try to evn_data = os.getenv(env[0], env[1])
and fail if it is not set

danielerez: It actually doesn't fail as it treats env as a string :) but pushed a new patch with a proper fix.

discovery-infra/update_bm_inventory_cm.py
```diff
@@ -36,8 +36,11 @@ def get_relevant_envs():
     data = {}
     for env in ENVS:
         evn_data = os.getenv(env[0], env[1])
-        if evn_data:
-            data[env[0]] = os.getenv(env[0], env[1])
+        # Set value as empty if variable is an empty string (e.g. defaulted in Makefile)
+        if evn_data == '""':
```
tsorya, line 40: it is not good. Why u need to set it? if not set just don't set it and inventory will use it's default

danielerez: The env var is already set in the Makefile, so os.getenv gets the value.

PR #32, opened 2020-07-09
Increased node registration timeout

slintes:
When having many nodes (>4), VM startup can take longer than expected, and the installation runs into a timeout while waiting for nodes being registered.
With this increased timeout I was able to process 3 masters + 3 workers.


discovery-infra/consts.py
```diff
@@ -7,7 +7,7 @@
 IMAGE_PATH = "%s/installer-image.iso" % IMAGE_FOLDER
 STORAGE_PATH = "/var/lib/libvirt/openshift-images"
 SSH_KEY = "ssh_key/key.pub"
-NODES_REGISTERED_TIMEOUT = 180
+NODES_REGISTERED_TIMEOUT = 600
```
tsorya, line 10: maybe we can add it as NODES_REGISTERED_TIMEOUT * number of nodes in place where we use it?

slintes: Good question, not sure though if startup time scales that way. Probably depends on several factors, mostly what the host's CPU can deliver. I did not do any measurement, 600 is a more or less random number, thinking that waiting 10 minutes instead of 3 does not hurt that much in case there is a real problem.

PR #33, opened 2020-07-09
Fixed worker IP address when using non default network CIDR

slintes:
Worker nodes were assigned an IP address from the default 192.168.126.0/24 range even though the network cidr was set to another value, which resulted in an error because virsh reported empty host names for DHCP leases (what? see https://bugzilla.redhat.com/show_bug.cgi?id=137504), which failed node registration (even though the nodes showed up in the UI with `localhost` hostname.
Not sure if my fix is nice enough code wise, I'm not familiar with Python, but it worked :)


tsorya:
thanks

PR #35, opened 2020-07-12
discovery: custom cluster name when domain specified

danielerez:
Handle custom cluster name when 'base-dns-domains' is specified
(cluster_name + random_postfix).


tsorya:
@danielerez it will break set_dns 
https://github.com/openshift/assisted-test-infra/blob/master/scripts/assisted_deployment.sh#L7

tsorya:
we need to handle it before pushing

danielerez:
> set_dns will not work

right, fixed here: https://github.com/openshift/assisted-test-infra/pull/37

PR #37, opened 2020-07-12
set_dns: skip when base_dns_domains is set

danielerez:
Skip dnsmasq flow when BASE_DNS_DOMAINS env var is set.

### DIFF
--- a/scripts/assisted_deployment.sh
+++ b/scripts/assisted_deployment.sh
@@ -5,7 +5,7 @@ function destroy_all() {
 }
 
 function set_dns() {
-    if [ $BASE_DNS_DOMAINS != ":" ]; then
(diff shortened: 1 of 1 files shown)


tsorya:
can u explain a little more? How it will work if base domain not set?

tsorya:
i think i got it but can u add this pull request to previous one? and just open one with 2 commits? will be easier to follow

danielerez:
> i think i got it but can u add this pull request to previous one? and just open one with 2 commits? will be easier to follow

I'm just checking whether the env var was defaulted to an empty string in Makefile (i.e. was not defined).
But sure, added to https://github.com/openshift/assisted-test-infra/pull/35.

PR #43, opened 2020-07-16
Remove OS from line as we can do either RHEL or CentOS

iranzo:
Minor edits to GUIDE.md


iranzo:
@tsorya can I get /lgtm again? I've rebased against latest master

PR #46, opened 2020-07-16
Use Precommit GitHub Action

iranzo:
Uses pre-commit github action as additional test for validating PR's

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,14 @@
+name: pre-commit
+
+on:
+  pull_request:
+  push:
+    branches: [master]
+
+jobs:
+  pre-commit:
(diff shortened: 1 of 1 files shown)


tsorya:
@YuviGold i think we have it in prow right?

iranzo:
> @YuviGold i think we have it in prow right?

Hi, I was talking with @YuviGold  on slack regarding this as the instance seems to be failing in some tests while setting it up (check my prior PR's)

iranzo:
For example: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/origin-ci-test/pr-logs/pull/openshift_assisted-test-infra/45/pull-ci-openshift-assisted-test-infra-master-lint/1283741441767510016

PR #54, opened 2020-07-22
Add troubleshooting section to the install guide

empovit:


romfreiman:
u cannot link internal doc into opensource project I think

PR #58, opened 2020-07-27
MGMT-1231 Accept custom namespace

RazRegev:
Accept NAMESPACE argument in all entrypoints.
If not received try to get it from env or use
'assisted-installer' as default.


GUIDE.md
```diff
@@ -225,9 +225,9 @@ There are issues installing BM Inventory.
 
 **Solution**
 
-1. Run `kubectl get pods -n assisted-installer` and look for the _bm-inventory_ pod name.
+1. Run `kubectl get pods -n <namespace>` and look for the _bm-inventory_ pod name.
```
tsorya, line 228: specify default one as example

discovery-infra/utils.py
```diff
@@ -41,14 +41,14 @@ def run_command_with_output(command):
...
+def get_service_url_with_retries(service_name, namespace):
+    return get_service_url(service_name, namespace)
 
 
-def get_service_url(service_name):
+def get_service_url(service_name, namespace):
     try:
         log.info("Getting inventory URL")
-        cmd = "minikube service %s --url -n assisted-installer" % service_name
+        cmd = f"minikube -n {namespace} service {service_name} --url"
```
tsorya, line 51: nice

tsorya:
@RazRegev u need to added to skipper.yaml too

RazRegev:
> @RazRegev u need to added to skipper.yaml too

done

discovery-infra/bm_inventory_api.py
```diff
@@ -149,13 +149,13 @@ def install_cluster(self, cluster_id):
         return self.client.install_cluster(cluster_id=cluster_id)
 
 
-def create_client(inventory_url=None, wait_for_url=True):
+def create_client(namespace, inventory_url=None, wait_for_url=True):
```
YuviGold, line 152: `create_client` parameters order is `(namespace, service, ...)` while `utils.get_service_url_with_retries` and `utils.get_service_url` is the opposite.
Please choose one order.

RazRegev: Why this is relevant? I would understand the logic if they were received the same arguments, but they aren't related in any way. The reason create client receive namespace at the 1st arg is because the other have defaults, and this doesn't. I don't think we need to align all functions receive namespace to be at 1st place cause of that.

RazRegev:
@YuviGold can you approve again? I forced pushed after lint

discovery-infra/bm_inventory_api.py
tsorya: i thikn we can remove defaults but as u wish

PR #62, opened 2020-07-29
Support custom MTU for nodes libvirt network

masayag:
Use `NETWORK_MTU=9001 make deploy_nodes` in order to create libvirt network with MTU 9001:

```
→ sudo virsh net-dumpxml test-infra-net
<network connections='3'>
  <name>test-infra-net</name>
  <mtu size='9001'/>
  ...
</network>
```

The created nodes will have their interfaces configured with network's MTU.


masayag:
@tsorya Could you run CI job that uses the `NETWORK_MTU=1495 make ...` variable ?

PR #66, opened 2020-07-30
MGMT-1655 - add additional network interface to master and worker VMs

yevgeny-shnaidman:


discovery-infra/delete_nodes.py
```diff
@@ -43,7 +43,8 @@ def delete_nodes(tfvars):
             virsh_cleanup.DEFAULT_SKIP_LIST,
             [
                 tfvars.get("cluster_name", consts.TEST_INFRA),
-                tfvars.get("libvirt_network_name", consts.TEST_INFRA),
+                tfvars.get("libvirt_network_name", consts.TEST_INFRA_NETWORK),
+                tfvars.get("libvirt_network_name_2", consts.TEST_INFRA_NETWORK_2),
```
tsorya, line 47: not sure _2 is a good name, can't we have something more descriptive ? :)

tsorya:
@RazRegev please go through.

eranco74:
Why?
Our installation doesn't require a second interface.

discovery-infra/delete_nodes.py
yevgeny-shnaidman: i am opened to suggestions

yevgeny-shnaidman:
@eranco74 2 reasons:
1) 99.99 % of HW server have 2 and more network interfaces, and we want to be a close as possible to that scenario. for example,if we had 2 interface we could have debugged ETCD issue on VMs and not real Baremetal environment
2) until BMH team finish with their side of updating BMO to run without provisioning interface ( which will probably be pushed to 4.7), we need additional interface to be able to reach "green" cluster installation

discovery-infra/delete_nodes.py
YuviGold: I don't like it either.. Can it be an array?

yevgeny-shnaidman: @YuviGold what do you mean by array? We will still need explicit names for each one

tsorya: it can be array in terraform but it will be much more complicated to implement i think

YuviGold: > @YuviGold what do you mean by array? We will still need explicit names for each one

You said you want to support multiple networks

yevgeny-shnaidman: after discussion with @ronniel1  i am cancelling this pull request, and  the second interface will be added by @RazRegev as part of his task

yevgeny-shnaidman:
after discussion with @ronniel1 i am cancelling this pull request, and the second interface will be added by @RazRegev as part of his task

PR #67, opened 2020-07-30
MGMT-1231: deploy_ui.sh pull image before generating ui yaml

RazRegev:
Since the namespace argument was added, we should use the
up to date latest image.

### DIFF
--- a/scripts/deploy_ui.sh
+++ b/scripts/deploy_ui.sh
@@ -27,6 +27,7 @@ mkdir -p build
 
 print_log "Starting ui"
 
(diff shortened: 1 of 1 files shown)


scripts/deploy_ui.sh
```diff
@@ -27,6 +27,7 @@ mkdir -p build
 
 print_log "Starting ui"
 
+${CONTAINER_COMMAND} pull quay.io/ocpmetal/ocp-metal-ui:latest
```
tsorya, line 30: why?

eranco74: because the host might have an older image (with the same tag) and it will not pull the updated one

PR #71, opened 2020-07-31
Handle assisted-service user auth

carbonin:
The goal of this PR is to allow auth the be disabled by default while also creating a way for a user to pass an offline token which will be used to fetch jwt tokens for api authentication when interacting with assisted-service.

assisted-service PR implementing api auth: https://github.com/openshift/assisted-service/pull/45
Requires https://github.com/openshift/assisted-service/pull/44 for key handling in swagger client


carbonin:
Should I also make the `ENABLE_AUTH` change in `update_bm_inventory_cm.py`?

tsorya:
@carbonin by default there will be no auth?
Will u add support to test-infra to be able to work with auth enabled?

carbonin:
> by default there will be no auth?

I was just trying to keep the current behavior so nothing breaks when https://github.com/filanov/bm-inventory/pull/363 is merged. I can definitely require auth by default if that's what we want here.

> Will u add support to test-infra to be able to work with auth enabled?

Yup, still working on that will un-WIP when that's ready.

discovery-infra/assisted_service_api.py
```diff
@@ -14,9 +18,51 @@ def __init__(self, inventory_url):
         self.inventory_url = inventory_url
         configs = Configuration()
         configs.host = self.inventory_url + "/api/assisted-install/v1"
+        if 'OFFLINE_TOKEN' in os.environ:
```
tsorya, line 21: not sure why we need "if" here. looks like if there is no token function will just return

discovery-infra/assisted_service_api.py
```diff
@@ -14,9 +18,51 @@ def __init__(self, inventory_url):
...
+
+                expires_on = json.loads(base64.b64decode(segment))['exp']
+
+                # if this key doesn't expire or if it has more than 30 seconds left, don't refresh
+                remaining = expires_on - time.time()
+                if expires_on == 0 or remaining > 30:
+                    return
+
+            # fetch new key if expired or not set yet
+            token_url = "https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token"
```
nmagnezi, line 47: This should probably get an env variable with this url as a default value

discovery-infra/assisted_service_api.py
```diff
@@ -14,9 +18,51 @@ def __init__(self, inventory_url):
...
+            params = {
+                "client_id":     "cloud-services",
+                "client_secret": offline_token,
```
nmagnezi, line 50: I noticed that in [pyocm](https://github.com/TheDiemer/pyocm/blob/master/pyocm/pyocm.py#L36-L39) they don't use `client_secret`, why is the delta?

discovery-infra/assisted_service_api.py
```diff
@@ -14,9 +18,51 @@ def __init__(self, inventory_url):
...
+                "grant_type":    "refresh_token",
+                "refresh_token": offline_token,
+            }
+
+            log.info("Refreshing API key")
+            response = requests.post(token_url, data = params)
```
nmagnezi, line 56: should we also catch exceptions?

discovery-infra/assisted_service_api.py
carbonin: Great question. I took this from what I saw in `ocm token -v10` I'll try without the secret. It did seem strange to specify it twice.

discovery-infra/assisted_service_api.py
carbonin: Sorry, I'm not super familiar with python yet. What exceptions could this raise? None of the docs/examples I was reading mentioned that.

discovery-infra/assisted_service_api.py
carbonin: Yup seems to work without the secret for me using `curl` :+1: good catch.

README.md
```diff
@@ -81,6 +81,9 @@ Check the [Install Guide](GUIDE.md) for installation instructions.
 | CONNECTIVITY_CHECK_IMAGE | connectivity-check image to use, will update assisted-service config map with given value                                           |
 | HARDWARE_INFO_IMAGE      | hardware-info image to use, will update assisted-service config map with given value                                                |
 | INVENTORY_IMAGE          | assisted-service image to be updated in assisted-service config map with given value                                                    |
+| ENABLE_AUTH              | configure assisted-service to authenticate API requests, default: false                                                             |
+| OFFLINE_TOKEN            | token used to fetch JWT tokens for assisted-service authentication                                                                  |
```
rollandf, line 85: Maybe add the URL from where to get the token: https://cloud.redhat.com/openshift/token

skipper.yaml
```diff
@@ -47,3 +47,5 @@ env:
   CONNECTIVITY_CHECK_IMAGE: $CONNECTIVITY_CHECK_IMAGE
   HARDWARE_INFO_IMAGE: $HARDWARE_INFO_IMAGE
   INVENTORY_IMAGE: $INVENTORY_IMAGE
+  OFFLINE_TOKEN: $OFFLINE_TOKEN
+  SSO_URL: $SSO_URL
```
rollandf, line 51: ENABLE_AUTH ?

carbonin:
/retest

Going to keep doing this until I at least see the same error twice :sweat_smile:

carbonin:
Does anyone know what this failure is or how I might resolve it?
```
fatal: [localhost]: FAILED! => {"changed": false, "msg": "ERROR: Packet setup failed."}
```

carbonin:
Different error this time, although it took over 4 hours to run ...
```
2020/08/10 14:55:25 Build src succeeded after 21s
2020/08/10 14:55:25 Building assisted-test-infra
2020/08/10 14:58:31 Build assisted-test-infra succeeded after 3m6s
2020/08/10 14:58:31 Tagging assisted-test-infra into stable
2020/08/10 14:58:32 Create release image registry.svc.ci.openshift.org/ci-op-fg7dyyn3/release:latest
{"component":"entrypoint","file":"prow/entrypoint/run.go:165","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Process did not finish before 4h0m0s timeout","severity":"error","time":"2020-08-10T18:54:40Z"}
time="2020-08-10T18:54:40Z" level=info msg="Received signal." signal=interrupt
2020/08/10 18:54:40 error: Process interrupted with signal interrupt, cancelling execution...
2020/08/10 18:54:40 cleanup: Deleting release pod release-latest
2020/08/10 18:56:00 error: unable to signal to artifacts container to terminate in pod release-latest, triggering deletion: could not run remote command: unable to upgrade connection: container not found ("artifacts")
2020/08/10 18:56:00 error: unable to retrieve artifacts from pod release-latest: could not read gzipped artifacts: unable to upgrade connection: container not found ("artifacts")
2020/08/10 18:56:35 No custom metadata found and prow metadata already exists. Not updating the metadata.
2020/08/10 18:56:35 Ran for 4h1m54s
error: some steps failed:
  * could not run steps: execution cancelled
  * could not run steps: step [release:latest] failed: release "release-latest" failed: the pod ci-op-fg7dyyn3/release-latest was deleted without completing after 3h58m3s (failed containers: )
time="2020-08-10T18:56:35Z" level=info msg="Reporting job state 'failed' with reason 'executing_graph:interrupted'"
time="2020-08-10T18:56:35Z" level=info msg="Reporting job state 'failed' with reason 'executing_graph:step_failed:assembling_release:creating_release:running_pod'"
{"component":"entrypoint","file":"prow/entrypoint/run.go:247","func":"k8s.io/test-infra/prow/entrypoint.gracefullyTerminate","level":"error","msg":"Process gracefully exited before 30m0s grace period","severity":"error","time":"2020-08-10T18:56:35Z"}
```

:confused:

carbonin:
@tsorya can you take another look at this one? The tests went green (finally).

discovery-infra/assisted_service_api.py
tsorya: i think u can catch all exception and do log.exception with nice message

discovery-infra/assisted_service_api.py
```diff
@@ -14,9 +18,52 @@ def __init__(self, inventory_url):
         self.inventory_url = inventory_url
         configs = Configuration()
         configs.host = self.inventory_url + "/api/assisted-install/v1"
+        self.set_config_auth(configs)
+
         self.api = ApiClient(configuration=configs)
         self.client = api.InstallerApi(api_client=self.api)
 
+    def set_config_auth(self, c):
+        offline_token = os.environ.get('OFFLINE_TOKEN', "")
```
tsorya, line 27: till now all params where passed through command line args but it is ok i think. Just u need to verify what value with u will have here. U don't need to set "" as default value

discovery-infra/assisted_service_api.py
```diff
@@ -14,9 +18,52 @@ def __init__(self, inventory_url):
...
+        if offline_token == "": return
```
tsorya, line 28: if not offline_token

tsorya: lets have log here

discovery-infra/assisted_service_api.py
tsorya: on the other hand, what to do if exception occurs? exit the test?

discovery-infra/assisted_service_api.py
```diff
@@ -14,9 +18,52 @@ def __init__(self, inventory_url):
...
+                expires_on = json.loads(base64.b64decode(segment))['exp']
+
+                # if this key doesn't expire or if it has more than 30 seconds left, don't refresh
+                remaining = expires_on - time.time()
+                if expires_on == 0 or remaining > 30:
+                    return
+
+            # fetch new key if expired or not set yet
+            token_url = "https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token"
+            env_url = os.environ.get("SSO_URL", "")
```
tsorya, line 47: token_url = os.environ.get("SSO_URL", "https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token")

tsorya: u can set default SSO_URL in Makefile

discovery-infra/assisted_service_api.py
```diff
@@ -14,9 +18,52 @@ def __init__(self, inventory_url):
...
+            params = {
+                "client_id":     "cloud-services",
+                "grant_type":    "refresh_token",
+                "refresh_token": offline_token,
+            }
+
+            log.info("Refreshing API key")
+            response = requests.post(token_url, data = params)
+            if not response.ok:
+                log.error("Failed to fetch JWT token: %: %" % response.reason, response.text)
```
tsorya, line 59: can we continue to run with this error?

discovery-infra/assisted_service_api.py
tsorya: i think just set it default

carbonin: Yeah, setting the default in the Makefile would work for this one. Generally the weirdness around how I'm checking for env vars (and checking for empty strings specifically) is because of how skipper automatically sets env vars in `skipper.yaml` to `""` if they were unset by the user.

I call this out in the commit message of 9b7414aa495f9c6c91eb6f10816f7615a3948576.

discovery-infra/assisted_service_api.py
carbonin: Yeah, if we fail here then the test will just fail. I mean we could continue and let auth fail, but I think that would be harder to debug.

We know here that the test will not succeed so better to throw an exception or log that we failed to fetch the token and fail here.

discovery-infra/assisted_service_api.py
carbonin: Probably not. Same argument as above.

discovery-infra/assisted_service_api.py
carbonin: Same thing here with skipper, it's setting the value to `""` if the user leaves it unset so I'm trying to collapse the actually unset case and "skipper unset" case into one condition by setting the default to `""`

discovery-infra/assisted_service_api.py
tsorya: so i think u need to throw exception here

tsorya: u have such functionality in responce. check and throw exception on error

discovery-infra/assisted_service_api.py
tsorya: ok

discovery-infra/assisted_service_api.py
tsorya: agree, so lets throw

scripts/deploy_assisted_service.sh
```diff
@@ -9,12 +9,13 @@ export NAMESPACE=${NAMESPACE:-assisted-installer}
...
 export SERVICE_BASE_URL="http://${SERVICE_URL}:${SERVICE_PORT}"
+export ENABLE_AUTH=${ENABLE_AUTH:-false}
 
 mkdir -p build
 
 print_log "Updating assisted_service params"
-skipper run discovery-infra/update_assisted_service_cm.py
-skipper run "make -C assisted-service/ deploy-all" ${SKIPPER_PARAMS} DEPLOY_TAG=${DEPLOY_TAG} NAMESPACE=${NAMESPACE}
+skipper run discovery-infra/update_assisted_service_cm.py ENABLE_AUTH=${ENABLE_AUTH}
+skipper run "make -C assisted-service/ deploy-all" ${SKIPPER_PARAMS} DEPLOY_TAG=${DEPLOY_TAG} NAMESPACE=${NAMESPACE} ENABLE_AUTH=${ENABLE_AUTH}
```
carbonin, line 18: Also I'm thinking that these changes might not be needed?

Do we have a use case for deploying assisted-service inside minikube and also configuring for auth against the red hat sso? Wouldn't we just use the instance on cloud.redhat.com at that point?

If we do want this available, then I'll need another PR to add the credentials assisted-service will use to do the reverse lookup of user info in https://github.com/openshift/assisted-service/pull/48 (which most people won't have access to)

discovery-infra/assisted_service_api.py
carbonin: This one resolved @tsorya ?

carbonin:
@tsorya any other comments on this one?

PR #72, opened 2020-08-01
MGMT-1668: Automatic port allocation per namespace

RazRegev:
MGMT-1668: Automatic port allocation per namespace

We want to be able to run multiple namespaces on same hosts.
Currently it is possible, but the external ports of the uis
and inventories of each namespace overrides the connection
of each other (6000 for inventory, 6008 to ui).

In order to be able to expose all ports, added bash logic
that:
1. checks for the next free port for a service, by a given
start port.
2. xinetd files and services names will be unique also by
namespaces, so services from different namespaces won't
override each other.
3. deletes previews allocated port per namespace (if exist).
In case we deploy same namespace twice, we want to reuse the
same port.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,9 +8,17 @@ PULL_PARAM=$(shell if [ "${CONTAINER_COMMAND}" = "podman" ];then echo "--pull-al
 
 SKIPPER_PARAMS ?= -i
 
+# global env
+NAMESPACE := $(or $(NAMESPACE),assisted-installer)
+HOST_IP := $(or $(HOST_IP),$(shell bash scripts/utils.sh get_main_ip))
+
 # bm-inventory
 BMI_BRANCH := $(or $(BMI_BRANCH), "master")
 SERVICE := $(or $(SERVICE), quay.io/ocpmetal/bm-inventory:latest)
+INVENTORY_START_PORT := $(or $(INVENTORY_START_PORT),6000)
+
+# ocp-metal-ui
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -135,7 +142,7 @@ set_dns:
 	scripts/assisted_deployment.sh set_dns
 
 deploy_ui: start_minikube
-	DEPLOY_TAG=$(DEPLOY_TAG) scripts/deploy_ui.sh
+	DEPLOY_TAG=$(DEPLOY_TAG) UI_PORT=$(shell bash scripts/utils.sh get_service_port ocp-metal-ui $(NAMESPACE) $(HOST_IP) $(UI_START_PORT)) scripts/deploy_ui.sh
```
tsorya, line 145: why get_service_port is not a part of deploy ui or deploy_bm ... ?

scripts/utils.sh
```diff
@@ -50,6 +50,41 @@ function get_main_ip() {
...
+    service=$1
+    ns=$2
+    if sudo test -f /etc/xinetd.d/$service-$ns; then
+        sudo systemctl stop xinetd
+        sudo rm /etc/xinetd.d/$service-$ns --force
+        sudo systemctl start xinetd
+    fi
+}
+
+function first_unreachable_port() {
```
tsorya, line 74: Maybe better to have namespace_<int> -> 6000+same <int> ?
will be much easier.

PR #75, opened 2020-08-03
MGMT-1715 Replace host and port with a single URL parameter

empovit:


tsorya:
@empovit u need to change it also here https://github.com/openshift/assisted-test-infra/blob/master/discovery-infra/update_bm_inventory_cm.py#L30

empovit:
@tsorya yes, thank you! I've overlooked a few places that need to be changed. Will update the PR soon (it's still WIP).

PR #76, opened 2020-08-03
Fixes on issues found in rhel8.2 system

tsorya:


scripts/install_environment.sh
```diff
@@ -60,6 +60,9 @@ function install_runtime_container() {
     else
         echo "docker or podman is already installed"
     fi
+
+    # logging with general user to overcome docker hub rate limit issue
+    ${CONTAINER_COMMAND} login --username ocpmetal --password ocpmetal777 docker.io
```
eranco74, line 65: This won't do nothing

PR #78, opened 2020-08-04
[MGMT-766][MGMT-1108] Provide HTTP/S Proxy for OCP cluster installation

masayag:
The environment variables HTTP_PROXY_URL / HTTPS_PROXY_URL / NO_PROXY could be
provided in order to set the Cluster Proxy on the installed cluster.

The HTTP Proxy will be used as a replacement of the deprecated PROXY_URL
variable for communication of the agent with the assisted-installer service.

Depends on https://github.com/openshift/assisted-service/pull/41


masayag:
> This is already supported in assisted-service?
> What is NO_PROXY use for?

Taken from step 3 of https://docs.openshift.com/container-platform/4.5/networking/enable-cluster-wide-proxy.html#nw-proxy-configure-object_config-cluster-wide-proxy:

> A comma-separated list of destination domain names, domains, IP addresses or other network CIDRs to exclude proxying. Preface a domain with . to include all subdomains of that domain. Use * to bypass proxy for all destinations. Note that if you scale up workers not included in networking.machineNetwork[].cidr from the installation configuration, you must add them to this list to prevent connection issues.

Makefile
```diff
@@ -209,7 +211,11 @@ delete_all_virsh_resources: destroy_nodes delete_minikube
 #######
 
 _download_iso:
+<<<<<<< HEAD
```
tsorya, line 214: please fix rebase issues

Makefile
```diff
@@ -36,8 +36,10 @@ NETWORK_CIDR := $(or $(NETWORK_CIDR),"192.168.126.0/24")
 NETWORK_NAME := $(or $(NETWORK_NAME), test-infra-net)
 NETWORK_BRIDGE := $(or $(NETWORK_BRIDGE), tt0)
 NETWORK_MTU := $(or $(NETWORK_MTU), 1500)
-PROXY_URL := $(or $(PROXY_URL), "")
 RUN_WITH_VIPS := $(or $(RUN_WITH_VIPS), "yes")
+HTTP_PROXY_URL := $(or $(HTTP_PROXY_URL), "")
```
tsorya, line 40: there is a possibility to have those params as OC_PARAMS env?
Not critical just can be nice

ansible-bm-install/inventory/hosts_without_os.sample
```diff
@@ -23,9 +23,6 @@ openshift_version="4.5"
 # (Optional) Assign specific Cluster ID
 #cluster_id=""
 
-# (Optional) Assign Proxy URL
```
tsorya, line 26: please talk to @rlopez133 regarding this, Just update him for the change

Makefile
masayag: @tsorya by looking at the purpose of OC_PARAMS - they seems to designed for the `oc` tool which isn't the case for HTTP_PROXY_URL. HTTP_PROXY_URL will be used to configure how agents should access the service and how to define the Proxy on the installed cluster. I don't OC_PARAMS is a good fit for them.

PR #80, opened 2020-08-06
Update README with service repository variables

masayag:


YuviGold:
Thanks! Missed that.

PR #81, opened 2020-08-06
Fixing bug with not setting SERVICE_BASE_URL in assisted configmap

tsorya:


scripts/deploy_assisted_service.sh
```diff
@@ -8,6 +8,7 @@ export SERVICE_NAME=assisted-service
 export NAMESPACE=${NAMESPACE:-assisted-installer}
 export SERVICE_URL=$(get_main_ip)
 export SERVICE_PORT=${SERVICE_PORT:-6000}
+export SERVICE_BASE_URL="http://${SERVICE_URL}:${SERVICE_PORT}"
```
romfreiman, line 11: Why http?

tsorya: cause it is still http

tsorya: when we will not support it lets move to https @empovit ?

empovit: Some environments can't have HTTPS. SSL termination at ingress doesn't work in minikube where port forwarding is involved, we already discussed it. I will add a flag when HTTPS support is ready.

PR #83, opened 2020-08-11
Bugfix installing assisted-service tar instead of searching setup.py

YuviGold:


iranzo:
@tsorya It might need to force merge as this 'fixes' the current CI is that right?

tsorya:
sounds like

tsorya:
not sure though, it looks like we have another problem

PR #84, opened 2020-08-11
Raz/mgmt 1765 deploy oc

RazRegev:


Dockerfile.test-infra
```diff
@@ -22,5 +22,7 @@ RUN pip3 install --no-cache-dir -r /tmp/requirements.txt && \
 # TODO: Remove once OpenShift CI will be upgraded to 4.2 (see https://access.redhat.com/articles/4859371)
 RUN chmod uga+w /etc/passwd
 
+RUN curl -Lo oc.tar.gz https://mirror.openshift.com/pub/openshift-v4/clients/oc/${OPENSHIFT_VERSION:-4.5}/linux/oc.tar.gz && tar -C /usr/local/bin -xf oc.tar.gz && rm -f oc.tar.gz
```
tsorya, line 25: no need right?

Makefile
```diff
@@ -49,6 +50,18 @@ DEPLOY_TAG := $(or $(DEPLOY_TAG), "")
 IMAGE_NAME=test-infra
 IMAGE_REG_NAME=quay.io/itsoiref/$(IMAGE_NAME)
 
+# oc deploy
+OC_MODE := $(or $(OC_MODE),"")
```
tsorya, line 54: i think u can set this flag if u get oc_user, what do u think?

Makefile
```diff
@@ -49,6 +50,18 @@ DEPLOY_TAG := $(or $(DEPLOY_TAG), "")
...
+ifneq ($(OC_MODE),)
+        OC_MODE_FLAG = --oc-mode
+else
+        OC_MODE_FLAG = ""
+endif
+
+OC_USER := $(or $(OC_USER),"")
```
tsorya, line 61: maybe add it as ADDITIONAL_PARAMS?

Dockerfile.test-infra
RazRegev: still needed :(

Makefile
RazRegev: thought that but preferred to be more explicit, maybe we'll decide to change user field to token

discovery-infra/oc_utils.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def oc_iter_urls_by_service(service, scheme):
+    s = oc.selector('routes', field_selectors={'spec.to.name': service})
+    for route in s.objects():
+        yield scheme + '://' + route.model['spec']['host']
+
+
+def is_assisted_service_reachable(url):
+    r = requests.get(url + '/health')
```
YuviGold, line 66: Add timeout, otherwise it would block

Dockerfile.test-infra
YuviGold: The image should be passed by an environment variable like in dev-scripts to support Prow different images
Look in here:
https://github.com/openshift-metal3/dev-scripts/blob/master/03_build_installer.sh#L11
https://github.com/openshift-metal3/dev-scripts/blob/master/ocp_install_env.sh#L24

Makefile
```diff
@@ -162,16 +173,16 @@ install_cluster:
...
+	discovery-infra/start_discovery.py -i $(ISO) -n $(NUM_MASTERS) -p $(STORAGE_POOL_PATH) -k '$(SSH_PUB_KEY)' -md $(MASTER_DISK) -wd $(WORKER_DISK) -mm $(MASTER_MEMORY) -wm $(WORKER_MEMORY) -nw $(NUM_WORKERS) -ps '$(PULL_SECRET)' -bd $(BASE_DOMAIN) -cN $(CLUSTER_NAME) -vN $(NETWORK_CIDR) -nN $(NETWORK_NAME) -nB $(NETWORK_BRIDGE) -nM $(NETWORK_MTU) -ov $(OPENSHIFT_VERSION) -rv $(RUN_WITH_VIPS) -iU $(REMOTE_SERVICE_URL) -id $(CLUSTER_ID) -mD $(BASE_DNS_DOMAINS) -ns $(NAMESPACE) --service-name $(SERVICE_NAME)  -ocu $(OC_USER) -ocp $(OC_PASS) -ocs $(OC_SCHEME) $(ADDITIONAL_PARAMS)
 
 deploy_nodes_with_install:
-	skipper make _deploy_nodes NAMESPACE=$(NAMESPACE) ADDITIONAL_PARAMS=-in $(SKIPPER_PARAMS)
+	skipper make _deploy_nodes NAMESPACE=$(NAMESPACE) ADDITIONAL_PARAMS='${ADDITIONAL_PARAMS}\ -in' $(SKIPPER_PARAMS)
```
YuviGold, line 179: `\` is a mistake?

discovery-infra/oc_utils.py
```diff
@@ -0,0 +1,68 @@
...
+
+        raise RuntimeError(f'could not find any reachable url for: {service}')
+
+    return geturl()
+
+
+def oc_iter_urls_by_service(service, scheme):
+    s = oc.selector('routes', field_selectors={'spec.to.name': service})
+    for route in s.objects():
+        yield scheme + '://' + route.model['spec']['host']
```
YuviGold, line 62: Make an f-string for readability

discovery-infra/oc_utils.py
```diff
@@ -0,0 +1,68 @@
...
+def oc_set_project(project):
+    log.info('Setting default oc project (namespace) to: %s', project)
+    oc.set_default_project(project)
+
+
+def oc_get_service_url(service='assisted-service', scheme='http', retries=5):
+    @retry(tries=retries, delay=3, backoff=2)
+    def geturl():
+        for url in oc_iter_urls_by_service(service, scheme):
+            if is_assisted_service_reachable(url):
```
YuviGold, line 51: Maybe use the same method like in https://github.com/openshift/assisted-service/blob/master/tools/wait_for_assisted_service.py#L39
Should catch `(requests.exceptions.ConnectionError, requests.exceptions.ReadTimeout)`

Dockerfile.test-infra
tsorya: openshift version is not enough? I think it is enough

Makefile
```diff
@@ -49,6 +50,16 @@ DEPLOY_TAG := $(or $(DEPLOY_TAG), "")
 IMAGE_NAME=test-infra
 IMAGE_REG_NAME=quay.io/itsoiref/$(IMAGE_NAME)
 
+# oc deploy
+OC_MODE := $(or $(OC_MODE),)
+ifneq ($(OC_MODE),)
+        ADDITIONAL_PARAMS = --oc-mode
```
tsorya, line 56: i meant why not to add all the OC params here? just build relevant string here

tsorya: or even call it OC_PARAMS

Makefile
RazRegev: nope, its just concatenates space to the variable value

discovery-infra/oc_utils.py
RazRegev: can't see why this is way is preferred

tsorya: i like waiting.wait i think it is more readable than internal function with retry as decorator. It is overkill.

RazRegev: hope that it's better now

Dockerfile.test-infra
```diff
@@ -22,5 +22,8 @@ RUN pip3 install --no-cache-dir -r /tmp/requirements.txt && \
 # TODO: Remove once OpenShift CI will be upgraded to 4.2 (see https://access.redhat.com/articles/4859371)
 RUN chmod uga+w /etc/passwd
 
+ARG OPENSHIFT_RELEASE_IMAGE=https://mirror.openshift.com/pub/openshift-v4/clients/oc/{OPENSHIFT_VERSION:-4.5}/linux/oc.tar.gz
```
YuviGold, line 25: Unfortunately I don't it is supported as well on Prow :(
Please transfer this oc extraction part to `scripts/install_environment.sh` script. So the argument will be passed on container runtime and not on container build time

discovery-infra/oc_utils.py
```diff
@@ -0,0 +1,169 @@
...
+
+def _load_kubeconfig():
+    config_path = os.environ.get('KUBECONFIG', consts.DEFAULT_KUBECONFIG_PATH)
+    with open(config_path) as fp:
+        return yaml.safe_load(fp)
+
+
+def _iterate_clusters(config):
+    clusters = config.get('clusters', [])
+    if len(clusters) == 0:
```
YuviGold, line 115: `if not clusters`

discovery-infra/oc_utils.py
```diff
@@ -0,0 +1,169 @@
...
+        '--token', token,
+        '--server', server,
+        '--insecure-skip-tls-verify'
+    ]
+
+
+def _get_token_and_server_from_kubeconfig():
+    config = _load_kubeconfig()
+    for cluster in _iterate_clusters(config):
+        server = _get_server_from_cluster(cluster)
```
YuviGold, line 90: A more pythonic way would be 
```python
    for cluster in _iterate_clusters(config):
         try:
             server = _get_server_from_cluster(cluster)
             user = _get_user_by_cluster(config, cluster)
             token = _get_token_by_user(config, user)
             return token, server
        except KeyError:
            continue
...

def _get_server_from_cluster(cluster):
    if f'.{consts.OC_DOMAIN}' not in cluster['cluster']['server']:
        raise KeyError(f'.{consts.OC_DOMAIN}')

    return cluster['cluster']['server']
```

discovery-infra/oc_utils.py
YuviGold: I don't see you are catching specific exceptions or at all

RazRegev: I didn't want to catch any exception here

discovery-infra/oc_utils.py
RazRegev: did that for explicitly

discovery-infra/oc_utils.py
RazRegev: I don't agree so much. 
each function does it's own checks and returns nothing if failed to find what is was looking for.
does not necessarily KeyError, so it will force me to raise exceptions which is less elegant why and also not truth in all the cases. Even in your example you raised key error (where is should be value error instead).

discovery-infra/oc_utils.py
RazRegev: but now when you say that I want to change it

Dockerfile.test-infra
tsorya: not sure he can move it to scripts/install_environment.sh cause he needs it as part of image. 
but maybe we can mount it?
@RazRegev ?

Makefile
```diff
@@ -43,12 +43,27 @@ SSH_PUB_KEY := $(or $(SSH_PUB_KEY),$(shell cat ssh_key/key.pub))
 PULL_SECRET :=  $(or $(PULL_SECRET), $(shell if ! [ -z "${PULL_SECRET_FILE}" ];then cat ${PULL_SECRET_FILE};fi))
 ROUTE53_SECRET := $(or $(ROUTE53_SECRET), "")
 
+# build
+OPENSHIFT_VERSION := $(or $(OPENSHIFT_VERSION),4.5)
+OPENSHIFT_RELEASE_IMAGE := $(or $(OPENSHIFT_RELEASE_IMAGE),https://mirror.openshift.com/pub/openshift-v4/clients/oc/$(OPENSHIFT_VERSION)/linux/oc.tar.gz)
```
tsorya, line 48: i don't understand why we need release image as param. It is always the same.

discovery-infra/oc_utils.py
```diff
@@ -0,0 +1,177 @@
...
+            return u['user']['token']
+
+
+def oc_set_project(project):
+    log.info('Setting default oc project (namespace) to: %s', project)
+    oc.set_default_project(project)
+
+
+def oc_get_service_url(service='assisted-service', scheme='http', retries=5):
+    return retry(
```
tsorya, line 148: not sure why to make have a retry here and not wait :) but if u insist

Dockerfile.test-infra
tsorya: @YuviGold can u validate we can live with this change in Prow or not?

Makefile
```diff
@@ -43,12 +43,27 @@ SSH_PUB_KEY := $(or $(SSH_PUB_KEY),$(shell cat ssh_key/key.pub))
...
 
+# oc deploy
+KUBECONFIG := $(or $(KUBECONFIG),${HOME}/.kube/config)
+ifneq ($(or $(OC_MODE),),)
+        OC_USER := $(or $(OC_USER),"")
+        OC_PASS := $(or $(OC_PASS),"")
+        OC_TOKEN := $(or $(OC_TOKEN),"")
+        OC_SERVER := $(or $(OC_SERVER),https://api.ocp.prod.psi.redhat.com:6443)
+        OC_SCHEME := $(or $(OC_SCHEME),http)
+        OC_PARAMS = --oc-mode -ocu $(OC_USER) -ocp $(OC_PASS) -oct $(OC_TOKEN) -ocs $(OC_SERVER) --oc-scheme $(OC_SCHEME)
```
tsorya, line 64: like :+1:

Dockerfile.test-infra
RazRegev: I will change that but I don't understand what is the problem with it

PR #85, opened 2020-08-16
Raz/mgmt 1763

RazRegev:


discovery-infra/consts.py
```diff
@@ -14,11 +14,11 @@
 NUMBER_OF_MASTERS = 3
 TEST_INFRA = "test-infra"
 CLUSTER = "%s-cluster" % TEST_INFRA
-CLUSTER_PREFIX = "%s-" % CLUSTER
+CLUSTER_PREFIX = "%s" % CLUSTER
 TEST_NETWORK = "%s-net" % TEST_INFRA
```
tsorya, line 18: or delete it or use it. I like TEST_NETWORK as name more cause we have many networks here :)

discovery-infra/delete_nodes.py
```diff
@@ -12,63 +15,113 @@
...
+
+def delete_nodes(cluster_name, namespace, tf_folder, tfvars):
+    """ Runs terraform destroy and then cleans it with virsh cleanup to delete
+        everything relevant.
+    """
+    @utils.on_exception(
+        message='Failed to run terraform delete',
+        silent=True
+    )
+    def _try_to_delete_nodes():
```
tsorya, line 42: lets not have it as internal function

discovery-infra/delete_nodes.py
```diff
@@ -12,63 +15,113 @@
...
+            '-auto-approve '
+            '-input=false '
+            '-state=terraform.tfstate ' 
+            '-state-out=terraform.tfstate ' 
+            '-var-file=terraform.tfvars.json'
         )
 
+    _try_to_delete_nodes()
+
+    default_network_name = consts.NETWORK_NAME_PREFIX + namespace
```
tsorya, line 56: no need for this env

discovery-infra/delete_nodes.py
```diff
@@ -12,63 +15,113 @@
...
+
+
+@utils.on_exception(message='Failed to delete nodes', silent=True)
+def delete_cluster(cluster_name, namespace):
+    log.info(
+        'Deleting cluster: %s in namespace: %s',
+        cluster_name, namespace
+    )
+
+    tf_folder = os.path.join(consts.TF_FOLDER, f'{cluster_name}:{namespace}')
```
tsorya, line 105: why we have  ":" in folder name? it looks strange just do __ for example

discovery-infra/install_cluster.py
```diff
@@ -11,6 +11,15 @@
 from logger import log
 
 
+def _get_tf_folder():
+    cluster_name = args.cluster_name or consts.CLUSTER_PREFIX
+    cluster_name += f'-{args.namespace}'
+    return os.path.join(
+        consts.TF_FOLDER,
+        f'{cluster_name}__{args.namespace}'
```
tsorya, line 19: missing something. One place u have cluster_name:namespace and here {cluster_name}__{args.namespace}
?

discovery-infra/install_cluster.py
```diff
@@ -155,6 +166,12 @@ def main():
         type=str,
         default='assisted-service'
     )
+    parser.add_argument(
```
tsorya, line 169: what will happened if we have couple of clusters with the same name?

discovery-infra/oc_utils.py
```diff
@@ -35,57 +36,130 @@ def extend_parser_with_oc_arguments(parser):
...
-            f'/apis/route.openshift.io/v1/namespaces/{namespace}/routes',
-            method='GET',
-            query_params=[('fieldSelector', f'spec.to.name={service}')],
-            response_type='V1ResourceQuotaList',
-            auth_settings=self.config.auth_settings(),
-            _return_http_data_only=True
+        self.__verify_ssl = verify_ssl
+
+
+class OpenShiftAPIClient(ApiClient):
```
tsorya, line 76: not sure i understand why we need this class?

discovery-infra/start_discovery.py
```diff
@@ -27,61 +29,110 @@ def _create_ip_address_list(node_count, starting_ip_addr):
...
+def _iterate_network_ips(machine_cidr, start_ip=None):
+    for ip in IPNetwork(machine_cidr).iter_hosts():
+        if start_ip and start_ip > ip:
+            continue
+
+        yield str(ip.ipv4())
+
+
+def _is_alive(ip):
+    return '64 bytes from' in utils.run_command(f'/bin/ping -c1 -w1 -i0.2 {ip}')
```
tsorya, line 93: why we need "64 bytes from" ? it exits with error if ip is not reachable

discovery-infra/delete_nodes.py
```diff
@@ -12,63 +15,113 @@
...
+def delete_clusters_from_all_namespaces():
+    for dirname in os.listdir(consts.TF_FOLDER):
+        res = _get_name_and_namespace_from_dirname(dirname)
+        if not res:
+            continue
+        name, namespace = res
+        delete_cluster(name, namespace)
+
+
+def _get_name_and_namespace_from_dirname(dirname):
```
tsorya, line 87: lets have one function in utils or new common file that handles all things relevant to this parsing and folder creation

discovery-infra/install_cluster.py
YuviGold: I guess cluster-id should be preffered

discovery-infra/utils.py
```diff
@@ -230,11 +250,16 @@ def file_exists(file_path):
...
-def recreate_folder(folder):
+def recreate_folder(folder, on_create=None):
     if os.path.exists(folder):
         shutil.rmtree(folder)
+
     os.makedirs(folder, exist_ok=True)
-    run_command("chmod ugo+rx %s" % folder)
+    run_command("chmod -R ugo+rx %s" % folder)
+
+    if on_create is None:
```
YuviGold, line 260: I find this feature a little bit redundant. It can run after this function has finished and the folder has created.

discovery-infra/oc_utils.py
tsorya: not sure why we need 2 classes to create client while overwriting base classes of them and all this for only one use in small function?
as for me it looks overcomplicated for our case. (overall written good, just not sure why all this code is needed)
@YuviGold

discovery-infra/oc_utils.py
```diff
@@ -35,57 +36,130 @@ def extend_parser_with_oc_arguments(parser):
...
+
+    def call_api(
+            self,
```
tsorya, line 79: sure all can be kwargs, they are not interesting params as you use only auth_settings

discovery-infra/start_discovery.py
```diff
@@ -27,61 +29,110 @@ def _create_ip_address_list(node_count, starting_ip_addr):
...
-    )
     tfvars["libvirt_storage_pool_path"] = storage_path
     tfvars.update(nodes_details)
 
-    with open(consts.TFVARS_JSON_FILE, "w") as _file:
+    with open(tfvars_json_file, "w") as _file:
         json.dump(tfvars, _file)
 
 
+def _find_free_ips(machine_cidr, start_ip, num_required_ips):
```
tsorya, line 62: why we need to find _find_free_ips in subnet? we can't have 2 clusters in the same subnet so why it is interesting?

tsorya: u need to validate subnet is free and not used and not ips

tsorya: even worth if one of the ips is not free it will be a mistake to try to create cluster u will not know what happened to your cluster on failure in this case.

discovery-infra/start_discovery.py
```diff
@@ -166,6 +213,45 @@ def _cluster_create_params():
...
+    i = 0
+    bridge = f'tt{i}'
+    while _is_bridge_exist(bridge):
+        i += 1
+        bridge = f'tt{i}'
+    return bridge
+
+
+def _is_bridge_exist(bridge):
+    def raise_error_if_occurred(cmd, out, err):
```
YuviGold, line 239: 1. Using the same name is confusing.
2. I don't see a reason for this `callback` to exist. Why not normal try-except with raises?
```python
    try:    
        cmd_output = utils.run_command(
            f'/usr/sbin/ip link show {bridge}',
            shell=False
        )
    except RuntimeError as ex:
        if 'does not exist' in ex.message:
            is_exist = False
        else:
            raise

    is_exist = True
    ....
```

3. After giving it a thought - Why do you even check the output and not the returned value? It would be much simpler and still accurate

discovery-infra/start_discovery.py
```diff
@@ -27,61 +29,110 @@ def _create_ip_address_list(node_count, starting_ip_addr):
...
     tfvars["libvirt_storage_pool_path"] = storage_path
     tfvars.update(nodes_details)
 
-    with open(consts.TFVARS_JSON_FILE, "w") as _file:
+    with open(tfvars_json_file, "w") as _file:
         json.dump(tfvars, _file)
 
 
+def _find_free_ips(machine_cidr, start_ip, num_required_ips):
+    if 0 >= num_required_ips:
```
YuviGold, line 63: Did you mean `<=`?

discovery-infra/start_discovery.py
```diff
@@ -166,6 +213,45 @@ def _cluster_create_params():
...
 
+def _find_free_cidr_machine(start_cidr_machine):
+    cidr_machine = IPNetwork(start_cidr_machine)
+    while _is_cidr_machine_exist(str(cidr_machine)):
+        cidr_machine += 1
+    return str(cidr_machine)
+
+
+def _is_cidr_machine_exist(cidr_machine):
+    cmd_output = utils.run_command(f'/usr/sbin/ip route show {cidr_machine}')
```
tsorya, line 224: it is ok but if you have 192.168.122.0/24 already used and user ask for 192.168.122.0/25 u will not fail but u can't really use it

discovery-infra/start_discovery.py
```diff
@@ -166,6 +213,45 @@ def _cluster_create_params():
...
+def _find_free_network_bridge():
+    i = 0
+    bridge = f'tt{i}'
+    while _is_bridge_exist(bridge):
+        i += 1
+        bridge = f'tt{i}'
+    return bridge
+
+
+def _is_bridge_exist(bridge):
```
tsorya, line 238: why not to get full list and check if string in it?

discovery-infra/start_discovery.py
```diff
@@ -27,61 +29,110 @@ def _create_ip_address_list(node_count, starting_ip_addr):
...
+        return []
+
+    free_ips = []
+    for ip in _iterate_network_ips(machine_cidr, start_ip):
+        if _is_alive(ip):
+            continue
+
+        free_ips.append(ip)
+        if len(free_ips) == num_required_ips:
+            break
```
YuviGold, line 73: Use `for-else` to indicate if a break occured.
```python
    for ip in _iterate_network_ips(machine_cidr, start_ip):
        ...
    else:
        raise RuntimeError(
            f'found {len(free_ips)}/{num_required_ips} free ips '
            f'in cidr {machine_cidr}'
        )
```

discovery-infra/start_discovery.py
```diff
@@ -215,17 +301,29 @@ def validate_dns(client, cluster_id):
 
 # Create vms from downloaded iso that will connect to assisted-service and register
 # If install cluster is set , it will run install cluster command and wait till all nodes will be in installing status
-def nodes_flow(client, cluster_name, cluster):
+def nodes_flow(client, cluster_name, cluster, image_path):
     nodes_details = _create_node_details(cluster_name)
     if cluster:
         nodes_details["cluster_inventory_id"] = cluster.id
+
+    tf_folder = os.path.join(
```
tsorya, line 309: as i said earlier lets have one function that returns tf-folder

discovery-infra/start_discovery.py
```diff
@@ -166,6 +213,45 @@ def _cluster_create_params():
     return params
 
 
+def _find_free_cidr_machine(start_cidr_machine):
+    cidr_machine = IPNetwork(start_cidr_machine)
+    while _is_cidr_machine_exist(str(cidr_machine)):
+        cidr_machine += 1
```
tsorya, line 219: the next one will be 192.168.128.0/24?

discovery-infra/delete_nodes.py
```diff
@@ -12,63 +15,113 @@
...
-        except:
-            log.exception("Failed to delete nodes")
+        _delete_virsh_resources()
+        return
+
+    if args.namespace == 'all':
+        delete_clusters_from_all_namespaces()
+        return
+
+    cluster_name = args.cluster_name or consts.CLUSTER_PREFIX
```
YuviGold, line 122: Try to create strings assignment as simple as possible with the right templates, to avoid multiple lines.
```python
cluster_name = f'{args.cluster_name or consts.CLUSTER_PREFIX}-{args.namespace}'
```

discovery-infra/start_discovery.py
```diff
@@ -395,10 +507,7 @@ def main():
         "--vm-network-cidr",
         help="Vm network cidr",
         type=str,
-        default="192.168.126.0/24",
-    )
-    parser.add_argument(
-        "-nN", "--network-name", help="Network name", type=str, default="test-infra-net"
+        required=False,
```
tsorya, line 510: u don't need to set required=False it is default option. 
all params that has -- are optional

discovery-infra/utils.py
```diff
@@ -18,17 +19,35 @@
 conn = libvirt.open("qemu:///system")
 
 
-def run_command(command, shell=False):
+def run_command(command, shell=False, callback=None):
```
tsorya, line 22: don't understand the usage of callback in sync function. U are not using it in async manner, why it is needed?

YuviGold:
Need to create the code a little bit simpler. The callbacks aren't necessary.
/hold

discovery-infra/utils.py
tsorya: if u don't want to raise exception just pass flag for it check=False i think and make it default true

tsorya:
very nice job!!!

discovery-infra/install_cluster.py
RazRegev: I started with '__' as you see in this commit but I wanted a separator that cant be passed as part of the namespace or cluster name.

discovery-infra/oc_utils.py
RazRegev: This is how I should have done it in the first place. 
The greatest benefit of it is that we can use this class as a regular kubernetes api client and have same interface for local and remote clients.

RazRegev: And it is more logically correct I think

discovery-infra/start_discovery.py
RazRegev: no, just don't continue to the for loop if no ips were required

RazRegev: never mind this function will be deleted anyway

discovery-infra/oc_utils.py
RazRegev: I can't relay on kwargs, this function must get args (so it can be also used by kubernetes library as a replacment for APIClient). And I need to be sure I catch this argument :(

discovery-infra/start_discovery.py
RazRegev: yap

discovery-infra/start_discovery.py
RazRegev: don't know why I did that

discovery-infra/install_cluster.py
RazRegev: this is problem doesn't necessarily related to this pr since this question is relevant also for current state.

RazRegev: I will fix this in a different pull request

discovery-infra/install_cluster.py
```diff
@@ -155,6 +158,12 @@ def main():
...
+        '-cn',
+        '--cluster-name',
```
tsorya, line 163: still don't understand why we need name

discovery-infra/oc_utils.py
```diff
@@ -35,57 +36,130 @@ def extend_parser_with_oc_arguments(parser):
...
+            files=None,
+            response_type=None,
+            auth_settings=None,
+            async_req=None,
+            _return_http_data_only=None,
+            collection_formats=None,
+            _preload_content=True,
+            _request_timeout=None
+    ):
+        if auth_settings == ['BearerToken']:
```
tsorya, line 96: all u change is auth_settings = self.configuration.auth_settings() right?
so if it is part of client already just use auth_settings=client.configuration.auth_settings()
and willbe no need to add this class

discovery-infra/start_discovery.py
```diff
@@ -47,41 +53,63 @@ def fill_tfvars(image_path, storage_path, master_count, nodes_details):
         + 10
         + int(tfvars["master_count"])
     )
+    master_count = min(master_count, consts.NUMBER_OF_MASTERS)
     tfvars["image_path"] = image_path
-    tfvars["master_count"] = min(master_count, consts.NUMBER_OF_MASTERS)
-    tfvars["libvirt_master_ips"] = _create_ip_address_list(
+    tfvars["master_count"] = master_count
+    tfvars['libvirt_master_ips'] = _create_ip_address_list(
```
tsorya, line 59: lets have "libvirt_master_ips" as all others

discovery-infra/start_discovery.py
```diff
@@ -47,41 +53,63 @@ def fill_tfvars(image_path, storage_path, master_count, nodes_details):
...
         min(master_count, consts.NUMBER_OF_MASTERS), starting_ip_addr=master_starting_ip
     )
-    tfvars["api_vip"] = _get_vips_ips()[0]
-    tfvars["libvirt_worker_ips"] = _create_ip_address_list(
+    tfvars['libvirt_worker_ips'] = _create_ip_address_list(
```
tsorya, line 62: same or change others

discovery-infra/start_discovery.py
```diff
@@ -47,41 +53,63 @@ def fill_tfvars(image_path, storage_path, master_count, nodes_details):
         + 10
         + int(tfvars["master_count"])
     )
+    master_count = min(master_count, consts.NUMBER_OF_MASTERS)
     tfvars["image_path"] = image_path
-    tfvars["master_count"] = min(master_count, consts.NUMBER_OF_MASTERS)
-    tfvars["libvirt_master_ips"] = _create_ip_address_list(
+    tfvars["master_count"] = master_count
+    tfvars['libvirt_master_ips'] = _create_ip_address_list(
         min(master_count, consts.NUMBER_OF_MASTERS), starting_ip_addr=master_starting_ip
```
tsorya, line 60: u already have master_count no need to for min here

discovery-infra/start_discovery.py
```diff
@@ -166,6 +194,38 @@ def _cluster_create_params():
...
+    if str(cidr_machine) in routes:
+        return True
+    return _ping(str(cidr_machine.ip + 1))
+
+
+def _ping(ip):
+    try:
+        utils.run_command(f'ping -c1 -w1 -i0.2 {ip}')
+    except RuntimeError as e:
+        if str(e).endswith('exited with code: 1'):
```
tsorya, line 215: why endswith? shy just not to check return code?

discovery-infra/utils.py
```diff
@@ -23,12 +24,30 @@ def run_command(command, shell=False):
...
+    def _io_buffer_to_str(buf):
+        if hasattr(buf, 'read'):
+            buf = buf.read().decode()
+        return buf
+
+    out = _io_buffer_to_str(process.stdout).strip()
+    err = _io_buffer_to_str(process.stderr).strip()
+
+    if err:
+        raise RuntimeError(
```
tsorya, line 41: and if u have something in stderr but command returned 0 ? why to fail? we can't know how others implement there code, u can rely only on exit code

tsorya: if u want custom exception please create one

tsorya: or as i said previously add flag that says something like verify raise_on_failure and set it to check=<this_flag>
and always return output, errorcode i wourld even send stderr to output so if u have error u can just print output

discovery-infra/utils.py
```diff
@@ -230,11 +250,16 @@ def file_exists(file_path):
...
     if os.path.exists(folder):
         shutil.rmtree(folder)
+
     os.makedirs(folder, exist_ok=True)
-    run_command("chmod ugo+rx %s" % folder)
+    run_command("chmod -R ugo+rx %s" % folder)
+
+    if on_create is None:
+        return
+    return on_create(folder)
```
tsorya, line 262: to say the truth i agree with @YuviGold why to have callback if u need it only once? u really see we someday will use it in another func?

tsorya:
the question is if we need to handle parallel execution. I think we will need so think about adding some sort of lock on resources

discovery-infra/install_cluster.py
RazRegev: for cases you want to create two clusters that talk with the same assisted-service.

PR #87, opened 2020-08-17
Raz/mgmt 1668

RazRegev:


Makefile
```diff
@@ -13,6 +13,11 @@ SERVICE_BRANCH := $(or $(SERVICE_BRANCH), "master")
 SERVICE_REPO := $(or $(SERVICE_REPO), "https://github.com/openshift/assisted-service")
 SERVICE := $(or $(SERVICE), quay.io/ocpmetal/assisted-service:latest)
 SERVICE_NAME := $(or $(SERVICE_NAME),assisted-service)
+SERVICE_START_PORT := $(or $(SERVICE_START_PORT),6000)
```
tsorya, line 16: not sure if we need those values in Makefile

Makefile
```diff
@@ -13,6 +13,11 @@ SERVICE_BRANCH := $(or $(SERVICE_BRANCH), "master")
...
+
+# ui
+UI_SERVICE_NAME := $(or $(UI_SERVICE_NAME),ocp-metal-ui)
```
tsorya, line 19: same

Makefile
tsorya: not sure if someone needs to change starting port

scripts/utils.sh
```diff
@@ -14,11 +14,48 @@ function url_reachable() {
     return $?
 }
 
+function search_for_next_free_port() {
+    service=$1
+    namespace=$2
+    port=$3
+    delete_xinetd_files_by_substr $service:$namespace:
```
tsorya, line 21: can u please explain the logic of delete_xinetd_files_by_substr

tsorya: i mean why it is part of search_for_next_free_port

RazRegev: u are right thee is no reason it will be in this function

PR #89, opened 2020-08-18
Refresh token early to be resilient to time skew

carbonin:
When testing this, the jenkins worker node was 3-5 minutes behind
the correct time. This meant that we thought the token was still
valid when it was not.


rollandf (approved):
:+1:

PR #90, opened 2020-08-20
Add OCM integration parameters

rollandf:
In order to be able to test OCM integration, parameters of the SSO Service Account and OCM URL are needed.


rollandf:
@danielerez @carbonin

PR #94, opened 2020-08-24
Rename NO_PROXY to NO_PROXY_VALUES

masayag:
In order to distinct between NO_PROXY env value that might be set for
the server that runs the test-assisted-infra to the desired envrironment
variable that should be propagated to assisted service, the NO_PROXY env
variable should be renamed.


YuviGold (approved):
LGTM

PR #100, opened 2020-08-30
Fix skipper parameters location

YuviGold:


rollandf:
Just out of curiosity, the parameters are not passed if they are not located right after `make`?

YuviGold:
> Just out of curiosity, the parameters are not passed if they are not located right after `make`?

It depends how `skipper` has been implemented.
I didn't even check it. But in my POV these parameters should get first 

```bash
➜  ~ skipper make --help
Usage: skipper make [OPTIONS] [MAKE_PARAMS]...

  Execute makefile target(s)

Options:
  -i, --interactive  Interactive mode
  -n, --name TEXT    Container name
  -e, --env TEXT     Environment variables to pass the container
  -f TEXT            Makefile to use
  -c, --cache        Use cache image
  --help             Show this message and exit.
```

Makefile
```diff
@@ -125,15 +125,15 @@ copy_terraform_files:
...
+	skipper make $(SKIPPER_PARAMS) _run_terraform
 
 _run_terraform:
 		cd build/terraform/ && \
 		terraform init -plugin-dir=/root/.terraform.d/plugins/ && \
 		terraform apply -auto-approve -input=false -state=terraform.tfstate -state-out=terraform.tfstate -var-file=terraform.tfvars.json
 
 destroy_terraform:
-	skipper make _destroy_terraform $(SKIPPER_PARAMS)
+	skipper make $(SKIPPER_PARAMS) _destroy_terraform
```
tsorya, line 136: nope,  it not right. For some reason skipper wants them after command

tsorya:
@YuviGold can you point me to the problem u have with skipper params? What are u trying to fix?

Makefile
YuviGold: Why do you think so?

PR #103, opened 2020-09-01
Check PULL_SECRET env variable set before deployment

nmagnezi:
Currently, it takes some time for execution to fail for a missing PULL_SECRET.
With this change, test-infra will immediately fail for the lack of this prerequisite.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,7 +149,7 @@ run_full_flow: run deploy_nodes set_dns
 
 redeploy_all: destroy run_full_flow
 
-run_full_flow_with_install: run deploy_nodes_with_install set_dns
+run_full_flow_with_install: validate_pull_secret_variable run deploy_nodes_with_install set_dns
 
 redeploy_all_with_install: destroy  run_full_flow_with_install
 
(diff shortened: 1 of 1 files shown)


YuviGold:
I don't think that the Makefile should enforce this behavior
I can see this is already checked in [start_discovery.py](https://github.com/openshift/assisted-test-infra/blob/master/discovery-infra/start_discovery.py#L491)

```python
    if not args.pull_secret and args.install_cluster:
        raise Exception("Can't install cluster without pull secret, please provide one")
```

/hold

nmagnezi:
@eranco74 @YuviGold

I'm aware of that validation, but having it validated in this stage creates a poor user (developer) experience:

Say someone forgot to set `$PULL_SECRET` and triggered: `make run_full_flow_with_install`

The result would be that it fails after a minute and a half (while people know it takes time to complete and look elsewhere), rather than failing immediately:
```
# time make run_full_flow_with_install
scripts/run_minikube.sh
Configuring minikube...

<trimmed>

  File "discovery-infra/start_discovery.py", line 476, in <module>
    raise Exception("Can't install cluster without pull secret, please provide one")
Exception: Can't install cluster without pull secret, please provide one
make: *** [Makefile:184: _deploy_nodes] Error 1
make: *** [Makefile:187: deploy_nodes_with_install] Error 2

real	1m28.139s
user	0m2.337s
sys	0m0.940s
```

YuviGold:
Maybe there is a way we can check for a variable existence like bash `set -o nounset`
What do you think? @tsorya

tsorya:
@nmagnezi i tend with @eranco74 and @YuviGold .
What will happened when u start running without pull secret:
1. start minikube
2. deploy service
3. fail on start_discovery.py

and it is not a big deal cause 1,2 are not influenced by pull secret. So when u will see the error and export pull secret the rerun will not restart all the flow and will start with 3

nmagnezi:
> @nmagnezi i tend to agree with @eranco74 and @YuviGold .
> What will happened when u start running without pull secret:
> 
>     1. start minikube
> 
>     2. deploy service
> 
>     3. fail on start_discovery.py
> 
> 
> and it is not a big deal cause 1,2 are not influenced by pull secret. So when u will see the error and export pull secret the rerun will not restart all the flow and will start with 3

I understand, but the point of this pull request is to avoid a case in which a prerequisite is not met, but only checked after some time, while the developer's attention is elsewhere. 

The current check happens too late in the process.

PR #106, opened 2020-09-02
Raz/mgmt 1668 pord forwarding by namespace

RazRegev:


scripts/utils.sh
```diff
@@ -5,6 +5,22 @@ set -o nounset
 export KUBECONFIG=${KUBECONFIG:-$HOME/.kube/config}
 export NAMESPACE=${NAMESPACE:-assisted-installer}
 
+function get_namespace_index() {
+    namespace=$1
+    oc_flag=${2:-}
+
+    index=$(python3 scripts/indexer.py --action set --namespace $namespace $oc_flag)
```
tsorya, line 12: it is not running inside image? are u sure will have all packages that u need?

PR #107, opened 2020-09-02
MGMT-1883 - Update the install config before installing the cluster

carbonin:
This issues an update which should have no effect as it matches the
current default, but it will test the API and that the override logic
doesn't throw any errors.

@romfreiman is this what you had in mind?


carbonin:
I wasn't sure how customizable we wanted this to be in this repo ... Should I allow the json string to be provided by the user? Or are we okay with something hard-coded for now?

romfreiman:
Exactly that. I wanted to be sure that we'll not break this API

PR #111, opened 2020-09-03
Remove multistage, install SERVICE on bring_assisted_service target

YuviGold:


YuviGold:
/hold
Need to insert the client inside the container

eranco74 (commented):
/lgtm

eranco74:
> Need to insert the client inside the container
If the test-infra image require the client pip package let's add it to the requirements.txt

YuviGold:
> > Need to insert the client inside the container

> If the test-infra image require the client pip package let's add it to the requirements.txt

@eranco74 
The problem is that we don't upload a python package to any registry

eranco74:
> > > Need to insert the client inside the container
> 
> > If the test-infra image require the client pip package let's add it to the requirements.txt
> 
> @eranco74
> The problem is that we don't upload a python package to any registry

@YuviGold, @filanov is there a ticket for publishing the assisted-service pip package?

filanov:
> > > > Need to insert the client inside the container
> > 
> > 
> > > If the test-infra image require the client pip package let's add it to the requirements.txt
> > 
> > 
> > @eranco74
> > The problem is that we don't upload a python package to any registry
> 
> @YuviGold, @filanov is there a ticket for publishing the assisted-service pip package?

I remember that we talked about it but i didn't find any related ticket.

YuviGold:
> > > > Need to insert the client inside the container
> > 
> > 
> > > If the test-infra image require the client pip package let's add it to the requirements.txt
> > 
> > 
> > @eranco74
> > The problem is that we don't upload a python package to any registry
> 
> @YuviGold, @filanov is there a ticket for publishing the assisted-service pip package?

In your opinion do we want to upload the package?

filanov:
> > > > > Need to insert the client inside the container
> > > 
> > > 
> > > > If the test-infra image require the client pip package let's add it to the requirements.txt
> > > 
> > > 
> > > @eranco74
> > > The problem is that we don't upload a python package to any registry
> > 
> > 
> > @YuviGold, @filanov is there a ticket for publishing the assisted-service pip package?
> 
> In your opinion do we want to upload the package?

I'm not sure how other services are doing it, QE, test-infra and other people that want to use our python client so if it can help someone i don't mind to publish a client.
The issue is that besides test-infra i'm not sure that other clients will take the right version.

YuviGold:
assisted-service-client is required for everywhere with skipper. This solution won't work.

PR #116, opened 2020-09-05
MGMT-1931 - default openshift version supported is now 4.6

yevgeny-shnaidman:


YuviGold:
@yevgeny-shnaidman Both CIs failed on Installation TO. Maybe the new version installation takes more time?

yevgeny-shnaidman:
@YuviGold no, this commit is part of general commits for 4.6. We will first push all the rest of the repositories with 4.6 commits, and after that Ci should path

PR #127, opened 2020-09-09
Add a call to the new endpoint for updating the discovery ignition

carbonin:
~~Requires https://github.com/openshift/assisted-service/pull/310~~
Merged


carbonin:
Added the change from https://github.com/openshift/assisted-test-infra/pull/166 just to make sure this is passing tests

YuviGold (approved):
I don't like that it is inserted as part of the happy-workflow, when the function name `override_ignition` doesn't tell if this step is mandatory for the workflow success.
With that being said, assisted-test-infra is messy anyhow. So this isn't a battle worth fighting for.

When the system tests infrastructure will be available, please move this code section as a test.

/approve
/lgtm

PR #133, opened 2020-09-13
Raz/mgmt 1655 alternative network for vms

RazRegev:


YuviGold:
Why do we need every function and variable to be cloned with "secondary"?
/hold

terraform_files/terraform.tfvars.json
```diff
@@ -2,20 +2,25 @@
   "cluster_name": "test-infra-cluster",
   "cluster_domain": "redhat.com",
   "machine_cidr": "192.168.126.0/24",
+  "provisioning_cidr": "192.168.140.0/24",
   "master_count": "3",
   "worker_count": "1",
   "image_path": "tmp",
   "libvirt_uri": "qemu+tcp://192.168.122.1/system",
   "libvirt_network_if": "tt0",
+  "libvirt_network_if_2": "tt1",
```
tsorya, line 11: maybe stt1? it will not break you with index?

RazRegev: fixed

discovery-infra/start_discovery.py
```diff
@@ -259,6 +303,8 @@ def nodes_flow(client, cluster_name, cluster, image_path):
     if cluster:
         nodes_details["cluster_inventory_id"] = cluster.id
 
+    nodes_details.update(_secondary_network_details())
```
yevgeny-shnaidman, line 306: why not to create second network in the _create_node_details function. Second network is not going to be optional, this is our default configuration

RazRegev: correct, yet we need to support multiple environments, so we need to generate the parameters for each environment to avoid conflicts.

RazRegev: I realized in delay what you meant, fixed

PR #137, opened 2020-09-14
Raz/test parallel ci

RazRegev:


discovery-infra/virsh_cleanup.py
```diff
@@ -76,9 +78,10 @@ def clean_networks(skip_list, resource_filter):
 
 
 def clean_virsh_resources(skip_list, resource_filter):
-    clean_domains(skip_list, resource_filter)
-    clean_pools(skip_list, resource_filter)
-    clean_networks(skip_list, resource_filter)
+    with utils.file_lock_context():
```
tsorya, line 81: why we need lock here?

discovery-infra/utils.py
```diff
@@ -368,3 +366,31 @@ def wrapped(*args, **kwargs):
...
+
+
+@contextmanager
+def file_lock_context(filepath='/tmp/discovery-infra.lock', timeout=300):
+    logging.getLogger('filelock').setLevel(logging.INFO)
+
+    lock = filelock.FileLock(filepath, timeout)
+    try:
+        lock.acquire()
+    except filelock.Timeout:
```
tsorya, line 378: can u explain it? if somebody took lock we will delete it?

discovery-infra/utils.py
```diff
@@ -368,3 +366,31 @@ def wrapped(*args, **kwargs):
...
+        log.info(
+            'Deleting lock file: %s '
+            'since it exceeded timeout of: %d seconds',
+            filepath, timeout
+        )
+        os.unlink(filepath)
+        lock.acquire()
+
+    try:
+        yield
```
tsorya, line 388: if action requires more than 30 seconds we will get an exception ?

RazRegev: 5 minutes not 30 seconds, and not raise an exception but override the lock.
Do you think I should increase the timeout?

RazRegev: 5 minutes, and not raise exception but override existing lock (see previews answer)

discovery-infra/utils.py
RazRegev: if acquiring the lock is passed a certain timeout that we know it is not reasonable to wait for that long, we delete the lock file and acquire it. this for cases user aborts CI while lock is acquired. in such cases the lock is not release and we need to override it

discovery-infra/virsh_cleanup.py
RazRegev: because libvirt command keeps failing when executing them in parallel.

PR #138, opened 2020-09-15
changed downloaded iso path to be under /tmp/images with a unique name

RazRegev:


RazRegev:
opened https://github.com/openshift/assisted-test-infra/pull/139

PR #139, opened 2020-09-15
MGMT-2186: changed downloaded iso path to be under /tmp/images with a…

RazRegev:
… unique name

### DIFF
--- a/discovery-infra/utils.py
+++ b/discovery-infra/utils.py
@@ -240,10 +240,14 @@ def file_exists(file_path):
     return Path(file_path).exists()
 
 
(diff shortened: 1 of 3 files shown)


tsorya (commented):
/lgtm

discovery-infra/utils.py
```diff
@@ -240,6 +240,14 @@ def file_exists(file_path):
     return Path(file_path).exists()
 
 
+def create_folder(folder, with_chmod=True):
```
tsorya, line 243: same code as here recreate_folder? just add here flag "recreate if exists" and delete recreate_folder

PR #140, opened 2020-09-15
Add support for Dell legacy iDRAC

dav1x:
-Detects version of iDRAC based on system generation 12, 13, 14 etc.
-Based on version uses redfish API or racadm mount container image


dav1x:
PTAL @rlopez133

PR #144, opened 2020-09-16
sorts table of environment variables in README

mhrivnak:
I found that the table is long enough to make it a cumbersome trying to find a
specific variable name, so I sorted it.


mhrivnak:
Rebasing this is going to take way longer than the 5s it took me to sort the list to begin with. I'll submit a new PR.

PR #147, opened 2020-09-17
Adding VMware support to Assisted installer infra via Ansible playbooks

dav1x:
1. Uploads discovery image
2. Creates VM with discovery images
3. Re-arranged layout of plays so plays would run less per host and maintained original structure.


dav1x:
PTAL @rlopez133

dav1x:
Any update to this? Need the Ansible updates for a blog post I'm working on.

PR #152, opened 2020-09-23
Makefile: Moved tf directory deletion from 'destroy' to 'destroy_nodes'

RazRegev:
Since it make more sense that it will be done there.


YuviGold (approved):
LGTM

PR #153, opened 2020-09-23
MGMT-2080 - add support for day2 flow

yevgeny-shnaidman:


Makefile
```diff
@@ -42,6 +42,10 @@ HTTPS_PROXY_URL := $(or $(HTTPS_PROXY_URL), "")
 NO_PROXY_VALUES := $(or $(NO_PROXY_VALUES), "")
 VIP_DHCP_ALLOCATION := $(or $(VIP_DHCP_ALLOCATION),yes)
 
+#day2 params
+API_VIP_IP := $(or $(API_VIP_IP),"")
```
tsorya, line 46: why not to take them from cluster?

yevgeny-shnaidman: the whole point of day2 is that we don't have an access to the OCP cluster. The way i want to run this is that we first run OCP cluster full install and then we run a target of day2.

tsorya: right. But we have. So why test will not just take from cluster. I think it is bad idea to ask user to remember his vips.
Not need in test env

Makefile
```diff
@@ -205,6 +213,15 @@ deploy_nodes_with_install:
 deploy_nodes:
 	skipper make $(SKIPPER_PARAMS) _deploy_nodes NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE) $(OC_FLAG)) NAMESPACE=$(NAMESPACE)
 
+_deploy_day2_nodes:
+	discovery-infra/start_discovery.py -i $(ISO) -k '$(SSH_PUB_KEY)' -nw $(NUM_WORKERS) -ps '$(PULL_SECRET)' -bd $(BASE_DOMAIN) -cN $(CLUSTER_NAME) -ov $(OPENSHIFT_VERSION) -id $(CLUSTER_ID) -ns $(NAMESPACE) -pX $(HTTP_PROXY_URL) -sX $(HTTPS_PROXY_URL) -nX $(NO_PROXY_VALUES) --service-name $(SERVICE_NAME) --profile $(PROFILE) --ns-index $(NAMESPACE_INDEX) --day2-cluster yes -avi $(API_VIP_IP) -avd $(API_VIP_DNSNAME) $(OC_PARAMS) $(KEEP_ISO_FLAG) $(ADDITIONAL_PARAMS)
```
tsorya, line 217: what is different from deploy nodes? 
i think u need only to update ADDITIONAL_PARAMS

Makefile
tsorya: but maybe i am wrong

discovery-infra/assisted_service_api.py
```diff
@@ -129,6 +137,14 @@ def set_hosts_roles(self, cluster_id, hosts_with_roles):
             cluster_id=cluster_id, cluster_update_params=hosts
         )
 
+    def set_pull_secret(self, cluster_id, pull_secret):
+        log.info(
+            "Setting pull secret for cluster %s", cluster_id
+        )
+        update_params = models.ClusterUpdateParams(pull_secret=pull_secret)
+        return self.client.update_cluster(
```
tsorya, line 145: where we are updating cluster? in assisted-service? so it is the same cluster we can take all data from it?

discovery-infra/consts.py
```diff
@@ -4,6 +4,9 @@
 TF_FOLDER = "build/terraform"
 TFVARS_JSON_NAME = "terraform.tfvars.json"
 IMAGE_FOLDER = "/tmp/images"
+TF_MAIN_JSON_NAME = "main.tf"
+BASE_IMAGE_FOLDER = "/tmp/images"
```
tsorya, line 8: we have it already

discovery-infra/consts.py
```diff
@@ -4,6 +4,9 @@
...
+IMAGE_NAME = "installer-image.iso"
```
tsorya, line 9: not sure it is always the same name right now. @RazRegev ?

discovery-infra/start_discovery.py
```diff
@@ -291,6 +303,84 @@ def validate_dns(client, cluster_id):
         raise e
 
 
+def configure_terraform_workers_nodes(tf_folder, num_worker_nodes):
```
tsorya, line 306: it is relevant only for 2_day right? maybe better to create python file dedicated for it?

tsorya: and what with second interface? @RazRegev ?

discovery-infra/start_discovery.py
```diff
@@ -291,6 +303,84 @@ def validate_dns(client, cluster_id):
...
+        worker_starting_ip = ipaddress.ip_address(master_end_ip)
+    else:
+        worker_starting_ip = ipaddress.ip_address(tfvars['libvirt_worker_ips'][0])
+    worker_ips_list = workers_ip_list + _create_ip_address_list(num_worker_nodes, worker_starting_ip + 1)
+    tfvars['libvirt_worker_ips'] = worker_ips_list
+    utils.set_tfvars(tf_folder, tfvars)
+
+
+def configure_terraform_api_dns(tf_folder, api_vip_ip, api_vip_dnsname):
+    main_str = utils.get_tf_main(tf_folder)
```
tsorya, line 323: can't we have dns as part of tfvars.json?

tsorya: and just do the same as with all other params?

discovery-infra/start_discovery.py
```diff
@@ -291,6 +303,84 @@ def validate_dns(client, cluster_id):
...
+    dns_config = "hosts  {\n      ip = \"%s\"\n      hostname = \"%s\"\n    }\n    " % (api_vip_ip, api_vip_dnsname)
+    main_str = main_str[:start_index] + dns_config + main_str[start_index:]
+    utils.set_tf_main(tf_folder, main_str)
+
+def get_network_nodes_from_terraform(tf_folder):
+    tfvars = utils.get_tfvars(tf_folder)
+    return tfvars['libvirt_network_name'], tfvars['master_count'] + tfvars['worker_count']
+
+
+def apply_day2_tf_configuration(cluster_name, num_worker_nodes, api_vip_ip, api_vip_dnsname):
```
tsorya, line 334: lets move all day2 to another file

discovery-infra/start_discovery.py
```diff
@@ -291,6 +303,84 @@ def validate_dns(client, cluster_id):
...
+    tf_folder = utils.get_tf_folder(cluster_name, args.namespace)
+    configure_terraform_workers_nodes(tf_folder, num_worker_nodes)
+    configure_terraform_api_dns(tf_folder, api_vip_ip, api_vip_dnsname)
+    return get_network_nodes_from_terraform(tf_folder)
+
+
+def day2_nodes_flow(client, cluster_name, cluster, image_path, num_worker_nodes, api_vip_ip, api_vip_dnsname):
+    tf_network_name, total_num_nodes = apply_day2_tf_configuration(cluster_name, num_worker_nodes, api_vip_ip, api_vip_dnsname)
+    utils.run_command(f'make _apply_terraform CLUSTER_NAME={cluster_name}')
```
tsorya, line 343: lets have utils.run_command(f'make _apply_terraform CLUSTER_NAME={cluster_name}') in utils and use it from 2 places.
i think u must have filelock on it @RazRegev ?

discovery-infra/start_discovery.py
```diff
@@ -291,6 +303,84 @@ def validate_dns(client, cluster_id):
...
+    time.sleep(5)
```
tsorya, line 344: and why sleeps?

tsorya: use wait for something please

discovery-infra/start_discovery.py
```diff
@@ -291,6 +303,84 @@ def validate_dns(client, cluster_id):
...
+        cluster_id=cluster.id,
+        nodes_count=num_worker_nodes,
+        statuses=[
+            consts.NodesStatus.KNOWN
+        ],
+        interval=30,
+    )
+
+    if args.install_cluster:
+        time.sleep(10)
```
tsorya, line 370: same

discovery-infra/start_discovery.py
```diff
@@ -357,13 +447,33 @@ def nodes_flow(client, cluster_name, cluster, image_path):
             validate_dns(client, cluster.id)
 
 
-def main():
-    client = None
-    cluster = {}
+def execute_day2_flow(image_folder, internal_cluster_name, num_workers, api_vip_ip, api_vip_dnsname):
```
tsorya, line 450: @RazRegev can u verify it will work with all latest changes ?

discovery-infra/start_discovery.py
```diff
@@ -357,13 +447,33 @@ def nodes_flow(client, cluster_name, cluster, image_path):
...
+    cluster_id = str(uuid.uuid4())
+    random_postfix = cluster_id[:8]
+    ui_cluster_name = internal_cluster_name + f'-{random_postfix}'
+    cluster = client.create_day2_cluster(
+            ui_cluster_name, cluster_id, **_day2_cluster_create_params()
+            )
 
-    internal_cluster_name = f'{args.cluster_name or consts.CLUSTER_PREFIX}-{args.namespace}'
-    log.info('Cluster name: %s', internal_cluster_name)
+    set_cluster_pull_secret(client, cluster_id, args.pull_secret)
```
tsorya, line 462: can't we pass pull secret on create?

discovery-infra/utils.py
```diff
@@ -113,6 +113,16 @@ def are_all_libvirt_nodes_in_cluster_hosts(client, cluster_id, network_name):
     )
 
 
+def are_libvirt_nodes_in_cluster_hosts(client, cluster_id, num_nodes):
+    hosts_macs = client.get_hosts_id_with_macs(cluster_id)
+    num_macs = len([mac for mac in hosts_macs if mac != ''])
+    return num_macs >= num_nodes
+
+
+def get_cluster_hosts_macs(client, cluster_id):
```
tsorya, line 122: if u add this one why not to use it in are_libvirt_nodes_in_cluster_hosts?

discovery-infra/utils.py
```diff
@@ -113,6 +113,16 @@ def are_all_libvirt_nodes_in_cluster_hosts(client, cluster_id, network_name):
     )
 
 
+def are_libvirt_nodes_in_cluster_hosts(client, cluster_id, num_nodes):
```
tsorya, line 116: u don't need to check that those are hosts that u expect them to be there?

discovery-infra/day2.py
```diff
@@ -0,0 +1,106 @@
+import os
+import utils
+import waiting
+import uuid
+import assisted_service_api
+import consts
+
+
+def execute_day2_flow(internal_cluster_name, num_workers, api_vip_ip, api_vip_dnsname):
+    utils.recreate_folder(consts.IMAGE_FOLDER, force_recreate=False)
```
RazRegev, line 16: you created an image folder at main of start_discovery.py, I think we should receive it as an argument

discovery-infra/day2.py
```diff
@@ -0,0 +1,106 @@
...
+
+    image_path = os.path.join(
+            consts.IMAGE_FOLDER,
+            f'{args.namespace}-installer-image.iso'
+        )
+    client.generate_and_download_image(
+            cluster_id=cluster.id,
+            image_path=image_path,
+            ssh_key=args.ssh_key,
+            )
```
RazRegev, line 37: nit: many of the function argument lines should be shifted a tab left

discovery-infra/day2.py
```diff
@@ -0,0 +1,106 @@
...
+
+    day2_nodes_flow(client, internal_cluster_name, cluster, image_path, num_workers, api_vip_ip, api_vip_dnsname)
+
+
+def day2_nodes_flow(client, cluster_name, cluster, image_path, num_worker_nodes, api_vip_ip, api_vip_dnsname):
+    tf_network_name, total_num_nodes = apply_day2_tf_configuration(cluster_name, num_worker_nodes, api_vip_ip, api_vip_dnsname)
+    utils.run_command(f'make _apply_terraform CLUSTER_NAME={cluster_name}')
+    time.sleep(5)
```
RazRegev, line 48: again why sleep? `utils.run_command()` is blocking, if that's why you added sleep here

RazRegev (commented):
sorry, I went by commits, I see you change things between commits.
can you squash changes so I could review the relevant commits?
also I will try to test it to see if it doesn't break the parallel environments feature and the 2'nd network

discovery-infra/day2.py
tsorya: lets have no sleeps and move them to waits

discovery-infra/day2.py
```diff
@@ -0,0 +1,134 @@
...
+        cluster_id=cluster.id,
+        nodes_count=num_worker_nodes,
+        statuses=[
+            consts.NodesStatus.KNOWN
+        ],
+        interval=30,
+    )
+
+    if args.install_cluster:
+        time.sleep(10)
```
tsorya, line 74: same here

discovery-infra/assisted_service_api.py
yevgeny-shnaidman: this is the day2 cluster, not the installed cluster

discovery-infra/start_discovery.py
yevgeny-shnaidman: @RazRegev why do we need the filelock? each terraform is in a different directory defined by cluster and namespace

discovery-infra/consts.py
tsorya: still the same IMAGE_FOLDER==BASE_IMAGE_FOLDER

discovery-infra/start_discovery.py
yevgeny-shnaidman: fixed

discovery-infra/day2.py
```diff
@@ -0,0 +1,134 @@
...
+    tfvars = utils.get_tfvars(tf_folder)
+    num_workers = tfvars['worker_count']
+    num_workers = num_workers + num_worker_nodes
+    tfvars['worker_count'] = num_workers
+    set_workers_ips_by_type(tfvars, num_worker_nodes, 'libvirt_master_ips', 'libvirt_worker_ips')
+    set_workers_ips_by_type(tfvars, num_worker_nodes, 'libvirt_secondary_master_ips', 'libvirt_secondary_worker_ips')
+    utils.set_tfvars(tf_folder, tfvars)
+
+
+def configure_terraform_api_dns(tf_folder, api_vip_ip, api_vip_dnsname):
```
tsorya, line 105: asked previously but can't find the answer. Can't we add dns as part of tfvars?

discovery-infra/start_discovery.py
yevgeny-shnaidman: no, we day2 register API does not require secret, user will pass it as part of create/download ISO

discovery-infra/utils.py
yevgeny-shnaidman: it can be problematic, since i don't know how to differentiate the worker nodes: some of them might be in installed cluster and not in the day2. i think that if we wanted to boot n workers and we see n workers registered in the cluster - then it is a good enough check for me

discovery-infra/day2.py
yevgeny-shnaidman: i would prefer not to: we don't need it for day1, and day2 is eventually a temp solution. I think it's better to update the main on the run

discovery-infra/start_discovery.py
```diff
@@ -291,6 +303,84 @@ def validate_dns(client, cluster_id):
         raise e
 
 
+def configure_terraform_workers_nodes(tf_folder, num_worker_nodes):
+    tfvars = utils.get_tfvars(tf_folder)
+    num_workers = tfvars['worker_count']
+    num_workers = num_workers + num_worker_nodes
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

discovery-infra/start_discovery.py
```diff
@@ -291,6 +303,84 @@ def validate_dns(client, cluster_id):
...
+    tfvars['worker_count'] = num_workers
+    master_end_ip = tfvars['libvirt_master_ips'][-1]
```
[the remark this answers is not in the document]
yevgeny-shnaidman: i think yes

discovery-infra/utils.py
```diff
@@ -126,6 +136,31 @@ def get_tfvars(tf_folder):
     return tfvars
 
 
+def set_tfvars(tf_folder, tfvars_json):
+    tf_json_file = os.path.join(tf_folder, consts.TFVARS_JSON_NAME)
+    if not os.path.exists(tf_json_file):
+        raise Exception(f'{tf_json_file} does not exists')
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

discovery-infra/start_discovery.py
```diff
@@ -235,6 +239,14 @@ def _cluster_create_params():
     return params
 
 
+def _day2_cluster_create_params():
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

discovery-infra/start_discovery.py
yevgeny-shnaidman: aw we agreed, changing dns by changing the api-vip and the cluster-name in the tfvars

PR #155, opened 2020-09-24
Fix overriding SERVICE_BASE_URL

YuviGold:


RazRegev (commented):
if you provide option to override it here so please add it also to deploy_ui.sh

YuviGold:
> if you provide option to override it here so please add it also to deploy_ui.sh

@RazRegev 
I don't see a reason why it should be overridden

PR #158, opened 2020-10-02
MGMT-1803 Add on-prem to test infra

rwsu:
Added support for testing the e2e flow using onprem. Introduces
two additional environment variables:

DEPLOY_TARGET - specifies where assisted-service is deployed. minikube
or podman-localhost. In the future there could be a podman-liveiso option.

ASSISTED_SERVICE_HOST - specifies the FQDN or IP address where assisted-service
when DEPLOY_TARGET is podman-localhost. This differs from minikube where
the service url can be discovered.


rwsu:
For on-prem to work in the e2e flow, the container image
quay.io/ocpmetal/assisted-service-onprem:latest needs to be published.
It is currently not available.

PR #159, opened 2020-10-03
MGMT-2242: fixed set dns for cluster name

RazRegev:
Added namespace to cluster name in set_dns and removed ui_cluseter_name.

### DIFF
--- a/discovery-infra/start_discovery.py
+++ b/discovery-infra/start_discovery.py
@@ -353,7 +353,7 @@ def nodes_flow(client, cluster_name, cluster, image_path):
             validate_dns(client, cluster.id)
 
 
-def execute_day1_flow(internal_cluster_name):
+def execute_day1_flow(cluster_name):
     client = None
     cluster = {}
     if args.managed_dns_domains:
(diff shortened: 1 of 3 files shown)


discovery-infra/start_discovery.py
```diff
@@ -377,12 +377,9 @@ def execute_day1_flow(internal_cluster_name):
         if args.cluster_id:
             cluster = client.cluster_get(cluster_id=args.cluster_id)
         else:
-            random_postfix = str(uuid.uuid4())[:8]
-            ui_cluster_name = internal_cluster_name + f'-{random_postfix}'
-            log.info('Cluster name on UI: %s', ui_cluster_name)
-
             cluster = client.create_cluster(
-                ui_cluster_name, ssh_public_key=args.ssh_key, **_cluster_create_params()
+                cluster_name,
```
yevgeny-shnaidman, line 381: it means that now all the cluster will be created with the same name. Do we have anywhere a code that uses cluster name to get/find/delete a cluster?

YuviGold (approved):
LGTM

PR #163, opened 2020-10-07
MGMT-2345 Download cluster logs, events and host logs

YuviGold:


discovery-infra/install_cluster.py
```diff
@@ -64,7 +64,9 @@ def wait_till_installed(client, cluster, timeout=60 * 60 * 2):
             timeout=consts.CLUSTER_INSTALLATION_TIMEOUT,
         )
     finally:
-        download_logs_from_all_hosts(client=client, cluster_id=cluster.id)
+        output_folder = f'build/{cluster.id}'
+        utils.recreate_folder(output_folder)
+        download_logs_from_all_hosts(client=client, cluster_id=cluster.id, output_folder=output_folder)
```
tsorya, line 69: U must verify that u get right number of logs inside tar.

Makefile
```diff
@@ -75,6 +75,14 @@ ifdef KEEP_ISO
     KEEP_ISO_FLAG = --keep-iso
 endif
 
+ifdef FORCE
```
tsorya, line 78: why we need this flag? There is no reason for it.

Makefile
```diff
@@ -75,6 +75,14 @@ ifdef KEEP_ISO
...
+	FORCE_FLAG = --force
+endif
+
+ifdef DEST
+	DEST_FLAG = --dest $(DEST)
```
tsorya, line 83: lets call it LOGS_DEST

discovery-infra/download_logs.py
```diff
@@ -0,0 +1,57 @@
...
+def donwload_logs(client: InventoryClient, cluster_id: str, dest: str, force: bool):
+    output_folder = os.path.join(dest, cluster_id)
+
+    if os.path.isdir(output_folder) and not force:
+        log.info(f"Directory {output_folder} already exist. Skipping")
+        return
+
+    recreate_folder(output_folder, force_recreate=force)
+
+    download_logs_from_all_hosts(client, cluster_id, output_folder)
```
tsorya, line 41: why u need it? your are downloading all the logs in the next line

discovery-infra/download_logs.py
```diff
@@ -0,0 +1,57 @@
...
+        if not clusters:
+            log.info('No clusters were found')
+            return
+
+        for cluster in clusters:
+            if cluster['status'] in FAILED_STATUSES:
+                donwload_logs(client, cluster['id'], args.dest, args.force)
+
+
+def donwload_logs(client: InventoryClient, cluster_id: str, dest: str, force: bool):
```
tsorya, line 32: really? types? for what? all the code we don't have and here you add it? It will not allow to run same code on older python version and many times we need due to old OS

tsorya (changes requested):
Validation of downloaded tar.gz is a must

discovery-infra/install_cluster.py
tsorya: my fault

discovery-infra/download_logs.py
```diff
@@ -0,0 +1,89 @@
...
+            if should_download_logs(cluster):
+                download_logs(client, cluster['id'], args.dest)
+
+
+def should_download_logs(cluster: dict):
+    return cluster['status'] in FAILED_STATUSES
+
+
+def download_logs(client: InventoryClient, cluster_id: str, dest: str):
+    if glob(f'{dest}/*{cluster_id}'):
```
tsorya, line 39: i think u can add timestamp to the folder name and always download the files

discovery-infra/download_logs.py
```diff
@@ -0,0 +1,89 @@
...
+        json.dump(d, metadata_file, indent=4)
+
+
+def get_ui_url_from_api_url(api_url: str):
+    known_urls = {
+        'https://api.openshift.com/': 'https://cloud.redhat.com/openshift/assisted-installer',
+        'https://api.stage.openshift.com/': 'https://qaprodauth.cloud.redhat.com/openshift',
+    }
+
+    for k, v in known_urls.items():
```
tsorya, line 70: just do known_urls[api_url] , u will get key error

tsorya: no need for all this code

discovery-infra/download_logs.py
```diff
@@ -0,0 +1,89 @@
...
+def download_logs(client: InventoryClient, cluster_id: str, dest: str):
+    if glob(f'{dest}/*{cluster_id}'):
+        log.info(f"Skipping. The logs of cluster {cluster_id} already exist under '{dest}' directory.")
+        return
+
+    output_folder = os.path.join(dest, f'{date.today()}_{cluster_id}')
+    recreate_folder(output_folder)
+
+    write_metadata_file(client.inventory_url, cluster_id, os.path.join(output_folder, 'metdata.json'))
+    client.download_cluster_events(cluster_id, os.path.join(output_folder, f'cluster_{cluster_id}_events.json'))
```
tsorya, line 47: and now if events will fail, u don't want to continue to logs?

romfreiman: yes u do

discovery-infra/download_logs.py
YuviGold: I thought about that. But I want to include also cases when the URL isn't same. Like 
```
https://api.stage.openshift.com/
https://api.stage.openshift.com
api.stage.openshift.com/
```

tsorya: got it

discovery-infra/download_logs.py
```diff
@@ -0,0 +1,92 @@
...
+def download_logs(client: InventoryClient, cluster: dict, dest: str):
+    if glob(f"{dest}/*{cluster['id']}"):
+        log.info(f"Skipping. The logs of cluster {cluster['id']} already exist under '{dest}' directory.")
+        return
+
+    output_folder = os.path.join(dest, f"{date.today()}_{cluster['id']}")
+    recreate_folder(output_folder)
+
+    write_metadata_file(client, cluster, os.path.join(output_folder, 'metdata.json'))
+    client.download_cluster_events(cluster['id'], os.path.join(output_folder, f"cluster_{cluster['id']}_events.json"))
```
tsorya, line 48: i am sure i asked this question already. Don't u want to download logs if events fails?

discovery-infra/download_logs.py
```diff
@@ -0,0 +1,103 @@
...
+        started_at = started_at.strftime(TIME_FORMAT)
+
+    return os.path.join(dest, f"{started_at}_{cluster['id']}")
+
+
+def write_metadata_file(client: InventoryClient, cluster: dict, file_name: str):
+    d = {'cluster': cluster}
+
+    try:
+        d['link'] = f"{get_ui_url_from_api_url(client.inventory_url)}/clusters/{cluster['id']}"
```
avishayt, line 70: @romfreiman No you're behind :)

discovery-infra/install_cluster.py
```diff
@@ -112,12 +114,10 @@ def run_install_flow(client, cluster_id, kubeconfig_path, pull_secret):
...
-    output_folder = f'build/{cluster_id}'
-    utils.recreate_folder(output_folder)
+def download_logs_from_all_hosts(client, cluster_id, output_folder):
     hosts = client.get_cluster_hosts(cluster_id=cluster_id)
     for host in hosts:
-        output_file = os.path.join(output_folder, f'{host["id"]}_logs.tar.gz')
+        output_file = os.path.join(output_folder, f'host_{host["id"]}.tar.gz')
         waiting.wait(
             lambda: client.download_host_logs(cluster_id=cluster_id,
                                               host_id=host["id"],
```
avishayt, line 123: Cluster logs already contain host logs - not needed IMO
@tsorya

tsorya: right missed it

tsorya: but then we need to validate in installation flow that all relevant logs are here.

PR #164, opened 2020-10-07
sorts table of environment variables in README

mhrivnak:
I found that the table is long enough to make it a cumbersome trying to find a
specific variable name, so I sorted it.

Replaces #144


YuviGold (approved):
/approve

PR #165, opened 2020-10-07
MGMT-2293 - install day1 and day2 in one command

yevgeny-shnaidman:


discovery-infra/start_discovery.py
```diff
@@ -604,11 +610,13 @@ def main():
...
-        nargs='?',
-        const=False,
-        default=False,
-        help="day 2 cluster creation"
+        help="day2 cluster",
+        action="store_true",
+    )
+    parser.add_argument(
+        "--day1-cluster",
+        help="day1 cluster",
```
tsorya, line 618: i don't understand this flag.

tsorya: why else was not enough?

Makefile
```diff
@@ -215,22 +216,26 @@ install_cluster:
...
 
 deploy_nodes_with_install:
-	skipper make $(SKIPPER_PARAMS) _deploy_nodes NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE) $(OC_FLAG)) NAMESPACE=$(NAMESPACE) ADDITIONAL_PARAMS=-in $(SKIPPER_PARAMS)
+	skipper make $(SKIPPER_PARAMS) _deploy_nodes NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE) $(OC_FLAG)) NAMESPACE=$(NAMESPACE) ADDITIONAL_PARAMS=-in $(SKIPPER_PARAMS) DAY1_PARAMS=--day1-cluster
```
tsorya, line 222: why it can't be part of Additional params?

Makefile
```diff
@@ -215,22 +216,26 @@ install_cluster:
...
 	discovery-infra/start_discovery.py -i $(ISO) -k '$(SSH_PUB_KEY)' -nw $(NUM_WORKERS) -ps '$(PULL_SECRET)' -bd $(BASE_DOMAIN) -cN $(CLUSTER_NAME) -ov $(OPENSHIFT_VERSION) -id $(CLUSTER_ID) -ns $(NAMESPACE) -pX $(HTTP_PROXY_URL) -sX $(HTTPS_PROXY_URL) -nX $(NO_PROXY_VALUES) --service-name $(SERVICE_NAME) --profile $(PROFILE) --ns-index $(NAMESPACE_INDEX) --day2-cluster yes -avi $(API_VIP_IP) -avd $(API_VIP_DNSNAME) $(OC_PARAMS) $(KEEP_ISO_FLAG) $(ADDITIONAL_PARAMS)
 
 deploy_day2_nodes:
-	skipper make $(SKIPPER_PARAMS) _deploy_day2_nodes NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE) $(OC_FLAG)) NAMESPACE=$(NAMESPACE)
+	skipper make $(SKIPPER_PARAMS) _deploy_nodes NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE) $(OC_FLAG)) NAMESPACE=$(NAMESPACE) $(SKIPPER_PARAMS) DAY2_PARAMS=--day2-cluster
+
+deploy_day2_nodes_with_install:
+	skipper make $(SKIPPER_PARAMS) _deploy_nodes NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE) $(OC_FLAG)) NAMESPACE=$(NAMESPACE) $(SKIPPER_PARAMS) ADDITIONAL_PARAMS=-in DAY2_PARAMS=--day2-cluster
+
+install_day1_and_day2:
```
tsorya, line 236: maybe better just to run regular deploy with some env? @YuviGold what do u think?

Makefile
tsorya: i think it can be, not sure why we need 2 more os envs

discovery-infra/start_discovery.py
yevgeny-shnaidman: we need to be able to run 3 scenarios with the same target (deploy_nodes):
1) run day1 only
2) run day2 only
3) run day one and then day2
so, day1-cluster means that we need to run day1 installation

Makefile
yevgeny-shnaidman: if you can tell me how to set Additional params with multiple values - i will do it :)

Makefile
yevgeny-shnaidman: i think that adding a new target is more clear to the user. Otherwise he somehow needs to know to define that env variable

PR #166, opened 2020-10-08
Pull centos from registry.centos.org

carbonin:
We were getting errors in CI related to dockerhub rate limiting


carbonin:
Marking as WIP as we're discussing where to pull this image from in https://github.com/openshift/assisted-installer-agent/pull/48

YuviGold (approved):

PR #168, opened 2020-10-11
First Drop off of qe testing in test infra.

nshidlin:
Changes:
adding pytest in requirments.txt
adding discovery-infra/tests dir containing tests and pytest.ini for test markers
adding test_cancel_and_reset test
adding  methods to cancel and reset cluster install in assisted_service_api
fixing bug in get_hosts_in_statuses from assisted_service_api
adding node_controller interface and implementation for QE VM nodes


discovery-infra/consts.py
```diff
@@ -37,9 +37,13 @@ class NodesStatus:
     ERROR = "error"
     PENDING_FOR_INPUT = "pending-for-input"
     DAY2_INSTALLED = "added-to-existing-cluster"
+    RESETING_PENDING_USER_ACTION = "resetting-pending-user-action"
 
 
 class ClusterStatus:
+    INSUFFICIENT = "insufficient"
```
tsorya, line 44: maybe we can take from assisted-service python client and remove those consts?

discovery-infra/node_controlers/qe_vm_controler.py
```diff
@@ -0,0 +1,94 @@
...
+        for node in nodes.keys():
+            self.format_node_disk(node)
+
+        return
+
+    def get_ingress_and_api_vips(self):
+
+        return {"api_vip":"192.168.123.5", "ingress_vip":"192.168.123.10"}
+
+def create_controler():
```
tsorya, line 92: u don't need this one.

discovery-infra/utils.py
```diff
@@ -255,6 +282,14 @@ def wait_till_cluster_is_in_status(
         log.info("Cluster: %s", client.cluster_get(cluster_id))
         raise
 
+def is_cluster_in_status(client, cluster_id, statuses):
```
tsorya, line 285: move line 276 to work with this function

discovery-infra/utils.py
```diff
@@ -436,3 +471,15 @@ def get_network_leases(network_name):
 
 def create_ip_address_list(node_count, starting_ip_addr):
     return [str(ipaddress.ip_address(starting_ip_addr) + i) for i in range(node_count)]
+
+def set_hosts_roles_based_on_requested_name(client, cluster_id):
```
tsorya, line 475: we have set_hosts_roles in start_discovery. Lets have one function in utils that is doing it.

skipper.yaml
```diff
@@ -66,4 +66,4 @@ env:
   API_VIP_DNSNAME: $API_VIP_DNSNAME
   DEPLOY_TARGET: $DEPLOY_TARGET
   ASSISTED_SERVICE_HOST: $ASSISTED_SERVICE_HOST
-  LOGS_DEST: $LOGS_DEST
```
tsorya, line 69: we need this one. Why to remove?

discovery-infra/node_controlers/qe_vm_controler.py
```diff
@@ -0,0 +1,94 @@
...
+        for domain in domains:
+            domain_name = domain.name()
+            if ("master" in domain_name) or ("worker" in domain_name):
+                domain_state = "running" if domain.isActive() else "shut_off"
+                nodes[domain_name] = domain_state
+
+        return nodes
+    
+    def shutdown_node(self, node_name):
+
```
tsorya, line 30: not sure why we need spaces in those functions? Not a big deal but interesting why?

discovery-infra/node_controlers/qe_vm_controler.py
```diff
@@ -0,0 +1,88 @@
...
+    def restart_node(self, node_name):
+        node = self.libvirt_connection.lookupByName(node_name)
+
+        node.destroy()
+        node.create()
+
+        return
+
+    def format_node_disk(self, node_name):
+        command = f"qemu-img info /var/lib/libvirt/images/linchpin/{node_name}.qcow2 | grep 'virtual size'"
```
ronniel1, line 67: Don't we need to start the VM with the ISO?

discovery-infra/tests/test_cancel_and_reset.py
```diff
@@ -0,0 +1,113 @@
...
+
+nodes_count = num_masters + num_workers
+
+api_client = assisted_service_api.create_client(remote_service_url)
+controler = node_controlers.qe_vm_controler.create_controler()
+
+@pytest.mark.regression
+def test_cancel_reset_after_node_boot():
+
+# Define new cluster
```
ronniel1, line 30: Shouldn't all tests have setup(), and teardown ()functions?
I also assume that at least for teardown() we should have some common, default, cleanup functionality

discovery-infra/node_controlers/qe_vm_controler.py
nshidlin: Right. The boot order for the VMs is set to hd then cdrom. In the case of a reset we need to format the hds so the node will boot from the iso

discovery-infra/node_controlers/qe_vm_controler.py
nshidlin: removed

skipper.yaml
nshidlin: Must have slipped during rebase. Added back

discovery-infra/tests/test_cancel_and_reset.py
nshidlin: Adding, Moved the PR to draft. Will reopen when setup() and teardown() is in place

requirements.txt
```diff
@@ -11,3 +11,6 @@ retry==0.9.2
 tqdm==4.46.0
 waiting==1.4.1
 netaddr==0.8.0
+kubernetes==11.0.0
+filelock
```
YuviGold, line 15: Please freeze versions

discovery-infra/node_controlers/node_controler.py
```diff
@@ -0,0 +1,132 @@
+class NodeControler(object):
```
YuviGold, line 1: The docstrings for all these functions don't provide any extra info.
Use meaningful function names, parameters, and use type hints.
e.g.
Replace
```python
    def shutdown_node(self, node_name):
        """
        Shuts down node
        Parameters:
        node_name (str): name of node to shutdown
        Returns:
        None
        """
```
With
```python
    def shutdown_node(self, node_name: str) -> None:
       raise NotImplementedError
```

discovery-infra/node_controlers/node_controler.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def __init__(self):
```
YuviGold, line 3: Remove empty `__init__` and `__del__` and keep the default ctor&dtor.

discovery-infra/node_controlers/qe_vm_controler.py
```diff
@@ -0,0 +1,90 @@
...
+                nodes[domain_name] = domain_state
+
+        return nodes
+    
+    def shutdown_node(self, node_name):
+        node = self.libvirt_connection.lookupByName(node_name)
+
+        node.destroy()
+
+        return
```
YuviGold, line 33: Remove empty `return` from functions

discovery-infra/node_controlers/qe_vm_controler.py
```diff
@@ -0,0 +1,90 @@
...
+    def __del__(self):
+        self.libvirt_connection.close()
+
+    def list_nodes(self):
+        nodes = {}
+        domains = self.libvirt_connection.listAllDomains()
+
+        for domain in domains:
+            domain_name = domain.name()
+            if ("master" in domain_name) or ("worker" in domain_name):
```
YuviGold, line 22: Use constants

discovery-infra/node_controlers/qe_vm_controler.py
```diff
@@ -0,0 +1,90 @@
...
+
+        for node in nodes.keys():
+            self.start_node(node)
+
+        return
+
+    def restart_node(self, node_name):
+        node = self.libvirt_connection.lookupByName(node_name)
+
+        node.destroy()
```
YuviGold, line 61: Can `node.destroy()` fail? Maybe try-catch?

discovery-infra/tests/base_cancel_reset.py
```diff
@@ -0,0 +1,43 @@
...
+offline_token = os.environ.get('OFFLINE_TOKEN')
+openshift_version = os.environ.get('OPENSHIFT_VERSION')
+base_domain = os.environ.get('BASE_DOMAIN')
+iso_download_path = os.environ.get('ISO')
+num_masters = int(os.environ.get('NUM_MASTERS'))
+num_workers = int(os.environ.get('NUM_WORKERS'))
+
+nodes_count = num_masters + num_workers
+
+class BaseCancelReset(object):
```
YuviGold, line 25: From python3+ no need to inherit `object` explicitly, while all classes inherit `object` by default.

discovery-infra/tests/base_cancel_reset.py
```diff
@@ -0,0 +1,43 @@
+import os
+import pytest
+
+import assisted_service_api
+import utils
+import consts
+
+if os.environ.get('NODE_ENV') == 'QE_VM':
```
YuviGold: You made the same condition inside `node_controller` fixture. You can remove it from here

discovery-infra/tests/base_cancel_reset.py
```diff
@@ -0,0 +1,43 @@
...
+base_domain = os.environ.get('BASE_DOMAIN')
+iso_download_path = os.environ.get('ISO')
+num_masters = int(os.environ.get('NUM_MASTERS'))
+num_workers = int(os.environ.get('NUM_WORKERS'))
+
+nodes_count = num_masters + num_workers
+
+class BaseCancelReset(object):
+    @pytest.fixture
+    def node_controler(self):
```
YuviGold, line 27: There's no need to create class for the fixtures. They can be public.
IMO, it's best practice to import everything on top of the module instead of inside functions.
```python
from assisted_service_api import create_client
from node_controlers.qe_vm_controler import QeVMController

@pytest.fixture
def node_controler():
    if os.environ.get('NODE_ENV') == 'QE_VM':
        yield QeVMController()
    else:
        raise NotImplementedError()

@pytest.fixture
def api_client():
    yield create_client()
```

discovery-infra/tests/base_cancel_reset.py
```diff
@@ -0,0 +1,43 @@
...
+
+import assisted_service_api
+import utils
+import consts
+
+if os.environ.get('NODE_ENV') == 'QE_VM':
+    #TODO change import to "as"
+    import node_controlers.qe_vm_controler 
+
+ssh_public_key = os.environ.get('SSH_PUB_KEY')
```
YuviGold: UPPER_SNAKE_CASE for constant globals.

discovery-infra/tests/test_cancel_and_reset.py
```diff
@@ -0,0 +1,115 @@
...
+num_masters = int(os.environ.get('NUM_MASTERS'))
+num_workers = int(os.environ.get('NUM_WORKERS'))
+
+nodes_count = num_masters + num_workers
+
+# api_client = assisted_service_api.create_client(remote_service_url)
+# controler = node_controlers.qe_vm_controler.create_controler()
+from tests.base_cancel_reset import BaseCancelReset
+
+class TestCancelReset(BaseCancelReset):
```
YuviGold, line 28: 1. Split into helper functions for readability.
2. I don't see a reason to create a class if you don't hold members or utilizes class operations.

pseudo:

```python
def test_reboot():
    cluster = _new_cluster()
    _generate_and_download_iso(cluster)
    ....

@pytest.fixture()
def _new_cluster()
    return api_client.create_cluster(
            cluster_name, 
            ssh_public_key=ssh_public_key, 
            pull_secret=pull_secret, 
            openshift_version=openshift_version, 
            base_dns_domain=base_domain, 
            vip_dhcp_allocation=False
        )

@pytest.fixture()
def _generate_and_download_iso(cluster):
    api_client.generate_and_download_image(
        cluster_id=cluster.id, 
        ssh_key=ssh_public_key, 
        image_path=iso_download_path
    )
```

YuviGold (changes requested):

discovery-infra/node_controlers/qe_vm_controler.py
nshidlin: removed

discovery-infra/utils.py
nshidlin: changed

discovery-infra/tests/test_cancel_and_reset.py
nshidlin: added, setup and tear down in the fixtures

requirements.txt
nshidlin: done

discovery-infra/node_controlers/node_controler.py
nshidlin: done

discovery-infra/node_controlers/node_controler.py
nshidlin: done

discovery-infra/node_controlers/qe_vm_controler.py
nshidlin: done

discovery-infra/node_controlers/qe_vm_controler.py
nshidlin: done

discovery-infra/tests/base_cancel_reset.py
nshidlin: done

discovery-infra/tests/base_cancel_reset.py
nshidlin: done

discovery-infra/tests/base_cancel_reset.py
nshidlin: The class (BaseTest) now holds the fixtures, and contains helper functions so that any new test plan can be a class that inherits from BaseTest

discovery-infra/tests/base_cancel_reset.py
nshidlin: moved these vars into tests/conftest.py

discovery-infra/tests/test_cancel_and_reset.py
nshidlin: moved to helper functions in BaseTest class

discovery-infra/node_controlers/qe_vm_controler.py
tsorya: maybe there is even another option? as for example to kill qemu process?

discovery-infra/tests/base_test.py
```diff
@@ -0,0 +1,148 @@
...
+from tests.conftest import EnvVariable
+
+if os.environ.get('NODE_ENV') == 'QE_VM':
+    from node_controlers.qe_vm_controler import QeVmController as node_controler
+else:
+    raise NotImplementedError
+
+class BaseTest:
+    @pytest.fixture()
+    def api_client(self, request):
```
tsorya, line 16: why not to add api client to init?
much easier will be to use it. Not sure we we need it as fixture. same with controller

tsorya: in test u will just use it as self.api_client.do_smth()

tsorya: and then all the functions here will not need to get it from outside, i think it is much more easier

discovery-infra/node_controlers/qe_vm_controler.py
nshidlin: @YuviGold node.destroy() can fail and raise an exception which will exit the test. I think this should would this way, wdty?
@tsorya this seems to aggressive to me. The libvirt python does this through the libvirt api

discovery-infra/utils.py
```diff
@@ -255,6 +286,14 @@ def wait_till_cluster_is_in_status(
         log.info("Cluster: %s", client.cluster_get(cluster_id))
         raise
 
+def is_cluster_in_status(client, cluster_id, statuses):
+    log.info("Is cluster %s in status %s", cluster_id, statuses)
+    try:
+        return client.cluster_get(cluster_id).status in statuses
+    except:
+        log.info("Cluster: %s", client.cluster_get(cluster_id))
```
tsorya, line 294: all u do in try is to get cluster. So if it fails there u want one more time to run it?
I think u need something like this .
log.exception("Failed to get cluster")

discovery-infra/tests/base_test.py
nshidlin: pytest classes cannot have an __init__ or __del__ method other than the defaults

tsorya: hmm. right everything works with fixtures

discovery-infra/node_controlers/qe_vm_controler.py
tsorya: i know but it will really painful to fail the test cause for some reason libvirt failed to stop vm.

tsorya: but ok for the start, will handle those cases one day

discovery-infra/utils.py
nshidlin: So, log the exception and try once more? sure

discovery-infra/consts.py
nshidlin: from what i see in the generated client there is no way get the list of cluster/host states

discovery-infra/utils.py
nshidlin: done

discovery-infra/node_controlers/qe_vm_controler.py
nshidlin: Added a check to to see that the node is running before shutdown and vice-versa

discovery-infra/utils.py
nshidlin: set_hosts_roles in start_discovery is tightly dependent on the teraform node implementation. once the node_controller for teraform nodes is implemented we can move to use this wdyt?

discovery-infra/consts.py
YuviGold: swagger python client isn't generating them nicely :( 
```python
    @status.setter
    def status(self, status):
        """Sets the status of this Cluster.

        Status of the OpenShift cluster.  # noqa: E501

        :param status: The status of this Cluster.  # noqa: E501
        :type: str
        """
        if status is None:
            raise ValueError("Invalid value for `status`, must not be `None`")  # noqa: E501
        allowed_values = ["insufficient", "ready", "error", "preparing-for-installation", "pending-for-input", "installing", "finalizing", "installed"]  # noqa: E501
        if status not in allowed_values:
            raise ValueError(
                "Invalid value for `status` ({0}), must be one of {1}"  # noqa: E501
                .format(status, allowed_values)
            )

        self._status = status
```

discovery-infra/tests/conftest.py
```diff
@@ -0,0 +1,16 @@
+import pytest
+import os
+
+class EnvVariable:
```
YuviGold, line 4: You don't need a class

discovery-infra/tests/conftest.py
```diff
@@ -0,0 +1,16 @@
...
+    PULL_SECRET = os.environ.get('PULL_SECRET').strip()
+    CLUSTER_NAME = os.environ.get('CLUSTER_NAME').strip()
+    OFFLINE_TOKEN = os.environ.get('OFFLINE_TOKEN').strip()
+    OPENSHIFT_VERSION = os.environ.get('OPENSHIFT_VERSION').strip()
+    BASE_DOMAIN = os.environ.get('BASE_DOMAIN').strip()
+    ISO_DOWNLOAD_PATH = os.environ.get('ISO').strip()
+    NUM_MASTERS = int(os.environ.get('NUM_MASTERS').strip())
+    NUM_WORKERS = int(os.environ.get('NUM_WORKERS').strip())
+    NUM_NODES = NUM_WORKERS + NUM_MASTERS
+    VIP_DHCP_ALLOCATION = os.environ.get('VIP_DHCP_ALLOCATION').strip() == 'True'
```
YuviGold, line 16: `distutils.util.strtobool(os.environ.get('VIP_DHCP_ALLOCATION').strip())`

discovery-infra/node_controlers/node_controler.py
```diff
@@ -0,0 +1,27 @@
+class NodeControler(object):    
+    def list_nodes(self) -> dict:
```
YuviGold, line 2: Use [`typing.Dict`](https://docs.python.org/3/library/typing.html#typing.Dict) with key-value

discovery-infra/node_controlers/node_controler.py
```diff
@@ -0,0 +1,27 @@
+class NodeControler(object):
```
YuviGold, line 1: No need to inherit from `object`

discovery-infra/tests/base_test.py
```diff
@@ -0,0 +1,148 @@
...
+        assert utils.is_cluster_in_status(
+            client=api_client,
+            cluster_id=cluster_id, 
+            statuses=[consts.ClusterStatus.CANCELLED]
+        )
+    def reset_cluster_install(self, cluster_id, api_client):
+        api_client.reset_cluster_install(cluster_id=cluster_id)
+
+    def assert_cluster_is_in_insufficient_status(self, cluster_id, api_client):
+        assert utils.is_cluster_in_status(
```
YuviGold, line 114: `assert` shouldn't be inside the helper functions. Just return the value and `assert` in the calling test

discovery-infra/tests/base_test.py
```diff
@@ -0,0 +1,148 @@
...
+            controler.start_node(node_name)
+    
+    def wait_untill_cluster_is_ready_for_install(self, cluster_id, api_client):
+        utils.wait_till_cluster_is_in_status(
+            client=api_client, 
+            cluster_id=cluster_id,
+            statuses= [consts.ClusterStatus.READY]
+        )
+
+    def wait_for_cluster_to_install(self, cluster_id, api_client):
```
YuviGold, line 138: Support providing a timeout parameter with `consts.CLUSTER_INSTALLATION_TIMEOUT` as default
```python
def wait_for_cluster_to_install(self, cluster_id, api_client, timeout=consts.CLUSTER_INSTALLATION_TIMEOUT):
```

discovery-infra/tests/conftest.py
nshidlin: done

discovery-infra/tests/conftest.py
nshidlin: util.strtobool returns 0/1 used: bool(util.strtobool(os.environ.get('VIP_DHCP_ALLOCATION'))) to match expected value in client

discovery-infra/node_controlers/node_controler.py
nshidlin: done

discovery-infra/node_controlers/node_controler.py
nshidlin: done

discovery-infra/tests/base_test.py

discovery-infra/tests/base_test.py

YuviGold (approved):
LGTM

Makefile
```diff
@@ -294,3 +294,9 @@ lint:
 
 _lint:
 	pre-commit run --all-files
+
+test:
+	skipper make _test
```
tsorya, line 299: we need an option to run specific test, possible? lets just open ticket for that

YuviGold (approved):

PR #170, opened 2020-10-12
Added deploy-target argument to deploy_nodes and install_cluster

RazRegev:
Since it is required for creating the assisted-service client in local mode.

  File "/home/raz/go/src/assisted-test-infra/discovery-infra/utils.py", line 396, in wrapped
    return fn(*args, **kwargs)
  File "discovery-infra/delete_nodes.py", line 29, in try_to_delete_cluster
    url=utils.get_assisted_service_url_by_args(args=args, wait=False)
  File "/home/raz/go/src/assisted-test-infra/discovery-infra/utils.py", line 301, in get_assisted_service_url_by_args
    kwargs['deploy_target'] = args.deploy_target
AttributeError: 'Namespace' object has no attribute 'deploy_target'

### DIFF
--- a/discovery-infra/delete_nodes.py
+++ b/discovery-infra/delete_nodes.py
@@ -172,6 +172,12 @@ if __name__ == "__main__":
         type=str,
         default='assisted-installer'
     )
+    parser.add_argument(
+        '--deploy-target',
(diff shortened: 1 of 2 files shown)


YuviGold (approved):
LGTM

PR #174, opened 2020-10-13
added API_VIP_CONNECTIVITY_CHECK_IMAGE env var

danielerez:
Added API_VIP_CONNECTIVITY_CHECK_IMAGE env var to skipper.yaml and configmap.


danielerez:
@tsorya @yevgeny-shnaidman

eranco74:
@danielerez note that if this ENV override is already supported in assisted-service (seems to be) there is no need adding it here.

danielerez:
> @danielerez note that if this ENV override is already supported in assisted-service (seems to be) there is no need adding it here.

also for deploying 

> @danielerez note that if this ENV override is already supported in assisted-service (seems to be) there is no need adding it here.

also if deploying the service using test-infra? cause it doesn't seem to propagate without this.

skipper.yaml
```diff
@@ -46,6 +46,7 @@ env:
   NAMESPACE: $NAMESPACE
   IMAGE_BUILDER: $IMAGE_BUILDER
   CONNECTIVITY_CHECK_IMAGE: $CONNECTIVITY_CHECK_IMAGE
+  API_VIP_CONNECTIVITY_CHECK_IMAGE: $API_VIP_CONNECTIVITY_CHECK_IMAGE
```
tsorya, line 49: it is enough to set it here, no need for change in update_assisted_service_cm.py

danielerez: ok, removed from cm.

PR #177, opened 2020-10-15
Added new test cancel reset test after boot

lalon4:
also aligned to PEP8


discovery-infra/consts.py
```diff
@@ -48,3 +48,13 @@ class ClusterStatus:
     ERROR = "error"
     PENDING_FOR_INPUT = "pending-for-input"
     CANCELLED = "cancelled"
+
+
+class HostsProgressStages:
```
YuviGold, line 53: Use [Enum classes](https://docs.python.org/3/library/enum.html)

discovery-infra/utils.py
```diff
@@ -267,6 +267,47 @@ def wait_till_at_least_one_host_is_in_status(
...
+
+
+def are_host_progress_in_stage(hosts, stages, nodes_count=1):
+    log.info("Checking hosts installation stage")
+    hosts_in_stage = [host for host in hosts
+                      if (host["progress"]["current_stage"]) in stages]
+    if len(hosts_in_stage) >= nodes_count:
+        return True
+    host_info = [(host["id"], (host["progress"]["current_stage"])) for host in hosts]
+    log.info(
```
YuviGold, line 301: `nodes_count` shouldn't be an integer with `%d`?
Anyway, use[ f-string](https://www.geeksforgeeks.org/formatted-string-literals-f-strings-python/) for readability.

discovery-infra/utils.py
```diff
@@ -267,6 +267,47 @@ def wait_till_at_least_one_host_is_in_status(
...
+    except:
+        hosts = client.get_cluster_hosts(cluster_id)
+        log.info("All nodes: %s", hosts)
+        raise
+
+
+def are_host_progress_in_stage(hosts, stages, nodes_count=1):
+    log.info("Checking hosts installation stage")
+    hosts_in_stage = [host for host in hosts
+                      if (host["progress"]["current_stage"]) in stages]
```
YuviGold, line 297: `stages` is a list of strings, you don't need to put the value inside of parenthesis, it's confusing (thought it is a list of tuples).
```python
 if host["progress"]["current_stage"] in stages]
```

discovery-infra/utils.py
```diff
@@ -267,6 +267,47 @@ def wait_till_at_least_one_host_is_in_status(
...
+                client.get_cluster_hosts(cluster_id),
+                stages,
+            ),
+            timeout_seconds=timeout,
+            sleep_seconds=interval,
+            waiting_for="Node to be in of the stage %s" % stages,
+        )
+    except:
+        hosts = client.get_cluster_hosts(cluster_id)
+        log.info("All nodes: %s", hosts)
```
YuviGold, line 290: use `log.error` to indicate an error
Don't need all the nodes information, just their stages.
```python
log.error(f"All nodes stages: {[host['progress']['current_stage'] for host in hosts]} when waited for {stages}")
```

discovery-infra/tests/base_test.py
```diff
@@ -1,32 +1,33 @@
...
 from tests.conftest import env_variables
 
 if os.environ.get('NODE_ENV') == 'QE_VM':
-    from node_controlers.qe_vm_controler import QeVmController as node_controler
+    from node_controlers.qe_vm_controler import QeVmController as nodeController
 else:
     raise NotImplementedError
 
+
 class BaseTest:
```
YuviGold, line 15: Now that the functions are static methods I don't see a reason to have a class.
All these methods can be public as well without a class.

lalon4:
created new PR https://github.com/openshift/assisted-test-infra/pull/180

PR #193, opened 2020-10-21
Added too to create jira ticket for triaging failed clusters.

ronniel1:
Also added anther tool to add "signarures" to the triaging tickets.
Currently there is only one such signature - add hosts information.
More work should be done in order to be able to easily extend the
signatures, and also to add the signatures when creating the tickets

### DIFF
--- a/discovery-infra/add_triage_signature.py
+++ b/discovery-infra/add_triage_signature.py
@@ -112,11 +112,9 @@ def get_credentials_from_netrc(server, netrc_file=DEFAULT_NETRC_FILE):
     return username, password
 
 
-def get_jira_client(netrc_file=DEFAULT_NETRC_FILE):
-    username, password = get_credentials_from_netrc(urlparse(JIRA_SERVER).hostname, netrc_file)
+def get_jira_client(username, password):
     logger.info("log-in with username: %s", username)
     return jira.JIRA(JIRA_SERVER, basic_auth=(username, password))
-
 
 ############################
 # Signature runner functionality
@@ -159,7 +157,15 @@ def get_all_triage_tickets(jclient):
 
 
 def main(args):
-    jclient = get_jira_client()
+    if args.user_password is None:
+        username, password = get_credentials_from_netrc(urlparse(JIRA_SERVER).hostname, args.netrc)
+    else:
+        try:
+            [username, password] = args.user_password.split(":", 1)
+        except:
+            logger.error("Failed to parse user:password")
+
+    jclient = get_jira_client(username, password)
 
     if args.all_issues:
         issues = get_all_triage_tickets(jclient)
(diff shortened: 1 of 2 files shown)


discovery-infra/add_triage_signature.py
```diff
@@ -0,0 +1,211 @@
...
+    @staticmethod
+    def _generate_table_for_report(hosts):
+        return tabulate(hosts, headers="keys", tablefmt="jira") + "\n"
+
+    @staticmethod
+    def _logs_url_fixup(url):
+        '''
+        the log server has two formats for the url
+        - URL for the UI  - http://assisted-logs-collector.usersys.redhat.com/#/2020-10-15_19:10:06_347ce6e8-bb4d-4751-825f-5e92e24da0d9/
+        - URL for the API - http://assisted-logs-collector.usersys.redhat.com/files/2020-10-15_19:10:06_347ce6e8-bb4d-4751-825f-5e92e24da0d9/
```
YuviGold, line 63: Cool! I didn't know it has this feature.

discovery-infra/add_triage_signature.py
```diff
@@ -0,0 +1,211 @@
...
+        for comment in comments:
+            if self._identifing_string in comment.body:
+                return comment
+        return None
+
+    def _update_triaging_ticket(self, key, comment, should_update=False):
+        report = "\n"
+        report += self._identifing_string
+        report += comment
+        comment = self._find_signature_comment(key)
```
YuviGold, line 44: Maybe change to another variable name different than the parameter, it's confusing

discovery-infra/add_triage_signature.py
```diff
@@ -0,0 +1,211 @@
...
+    return jira.JIRA(JIRA_SERVER, basic_auth=(username, password))
+
+
+############################
+# Signature runner functionality
+############################
+LOGS_URL_FROM_DESCRIPTION = re.compile(r".*logs:\* \[(http.*)\]")
+
+'''
+def get_all_triage_tickets(jclient):
```
YuviGold, line 127: Can be removed, defined below

discovery-infra/create_triage_tickets.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+
+def get_credentials_from_netrc(server, netrc_file=DEFAULT_NETRC_FILE):
+    cred = netrc.netrc(os.path.expanduser(netrc_file))
+    username, _, password = cred.authenticators(server)
+    return username, password
+
+
+def get_jira_client(netrc_file):
```
YuviGold, line 49: Can you unite the utils? `get_credentials_from_netrc` , `get_jira_client`, `get_all_triage_tickets` are defined twice on both the modules.

ronniel1: You are right
There is more effort that needs to be done here.
This is far from complete, but it works now
Since I don;'t have much time to work on it, I want to start using this now, and we'll assign someone to continue with it
Also need to work on adding more signature tools

YuviGold (approved):
LGTM

PR #195, opened 2020-10-22
MGMT-2462 AI on OCP cluster

danielerez:
- Added 'deploy-on-ocp-cluster' target: automates the deployemnt of
  service and UI on the test OCP cluster (using build/kubeconfig by default).
- Added 'ocp.py': for updating /etc/hosts.
- Added 'ocp.sh': for deploying service/ui using assited-service (invoked with 'skipper run').

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,6 +81,7 @@ OCM_BASE_URL := $(or $(OCM_BASE_URL), https://api-integration.6943.hive-integrat
 # minikube profile is used to manage multiple minikube instances
 PROFILE := $(or $(PROFILE),minikube)
 DEPLOY_TARGET := $(or $(DEPLOY_TARGET),minikube)
+OCP_KUBECONFIG := $(or $(OCP_KUBECONFIG),build/kubeconfig)
 
 .EXPORT_ALL_VARIABLES:
 
@@ -198,6 +199,22 @@ kill_port_forwardings:
 
 kill_all_port_forwardings:
 	scripts/utils.sh kill_port_forwardings '$(SERVICE_NAME) $(UI_SERVICE_NAME)'
+
+deploy_on_ocp_cluster:
+	# service
+	DEPLOY_TARGET=ocp NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE)) \
+		DEPLOY_TAG=$(DEPLOY_TAG) DEPLOY_MANIFEST_TAG=$(DEPLOY_MANIFEST_TAG) OCP_KUBECONFIG=$(OCP_KUBECONFIG) \
+		CLUSTER_NAME=$(CLUSTER_NAME) CLUSTER_ID=$(CLUSTER_ID) PROFILE=$(PROFILE) \
+		scripts/deploy_assisted_service.sh
+
+	# UI
+	DEPLOY_TARGET=ocp NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE)) \
+		DEPLOY_TAG=$(DEPLOY_TAG) DEPLOY_MANIFEST_TAG=$(DEPLOY_MANIFEST_TAG) OCP_KUBECONFIG=$(OCP_KUBECONFIG) \
+		CLUSTER_NAME=$(CLUSTER_NAME) CLUSTER_ID=$(CLUSTER_ID) PROFILE=$(PROFILE) \
+		scripts/deploy_ui.sh
+
+config_etc_hosts_for_ocp_cluster:
+	discovery-infra/ocp.py --config-etc-hosts -cn $(CLUSTER_NAME) -id $(CLUSTER_ID) -ns $(NAMESPACE) --service-name $(SERVICE_NAME) --profile $(PROFILE) --deploy-target ocp $(ADDITIONAL_PARAMS)
 
 ###########
 # Cluster #
--- a/discovery-infra/ocp.py
+++ b/discovery-infra/ocp.py
@@ -1,0 +1,85 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+
+import argparse
+import day2
+import utils
+import oc_utils
+import assisted_service_api
+import consts
+import sys
+
+
+def get_ocp_cluster(args):
+    if not args.cluster_id:
+        cluster_name = f'{args.cluster_name or consts.CLUSTER_PREFIX}-{args.namespace}'
+        tf_folder = utils.get_tf_folder(cluster_name, args.namespace)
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -199,6 +200,22 @@ kill_port_forwardings:
 kill_all_port_forwardings:
 	scripts/utils.sh kill_port_forwardings '$(SERVICE_NAME) $(UI_SERVICE_NAME)'
 
+deploy_service_on_cluster:
```
yevgeny-shnaidman, line 203: lets rename it to "deploy_service_on_ocp_cluster

scripts/day2.sh
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+function deploy_day2_service() {
```
yevgeny-shnaidman, line 4: let's rename deploy_ocp_service and the file should be ocp.sh

discovery-infra/day2.py
```diff
@@ -215,3 +222,79 @@ def _day2_cluster_create_params(openshift_version, api_vip_dnsname):
         "api_vip_dnsname": api_vip_dnsname,
     }
     return params
+
+def get_ocp_cluster(args):
```
yevgeny-shnaidman, line 226: lets keep day2.py for day2 on clud operations, and move your need code  to a new file ( for example ocp.py

Makefile
danielerez: done

scripts/day2.sh
danielerez: done

discovery-infra/day2.py
danielerez: done

Makefile
```diff
@@ -199,6 +200,22 @@ kill_port_forwardings:
...
+		scripts/deploy_assisted_service.sh
+
+	# UI
+	DEPLOY_TARGET=ocp NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE)) \
+		DEPLOY_TAG=$(DEPLOY_TAG) DEPLOY_MANIFEST_TAG=$(DEPLOY_MANIFEST_TAG) SERVICE_KUBECONFIG=$(SERVICE_KUBECONFIG) \
+		CLUSTER_NAME=$(CLUSTER_NAME) CLUSTER_ID=$(CLUSTER_ID) PROFILE=$(PROFILE) \
+		scripts/deploy_ui.sh
+
+config_etc_hosts_for_ocp_cluster:
+	discovery-infra/ocp.py --config-etc-hosts -cn $(CLUSTER_NAME) -id $(CLUSTER_ID) -ns $(NAMESPACE) --service-name $(SERVICE_NAME) --profile $(PROFILE) --deploy-target ocp $(ADDITIONAL_PARAMS)
```
yevgeny-shnaidman, line 217: i think we can move it under deploy_service_on_ocp_cluster,no?

yevgeny-shnaidman, line 217: do we need all those parameters? Can we use just cluster_id?

danielerez: no, since we use it in ocp.sh, as it's invoked using skipper.

danielerez: need those for get_assisted_service_url_by_args, as we create an assisted-service client for getting the cluster.

discovery-infra/ocp.py
```diff
@@ -0,0 +1,85 @@
...
+        '--profile',
+        help='Minikube profile for assisted-installer deployment',
+        type=str,
+        default='assisted-installer'
+    )
+    parser.add_argument(
+        '--deploy-target',
+        help='Where assisted-service is deployed',
+        type=str,
+        default='minikube'
```
RazRegev, line 71: shouldn't it be 'ocp'?

discovery-infra/ocp.py
```diff
@@ -0,0 +1,85 @@
...
+        sys.stdout.write(ocp_cluster.api_vip)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="OCP")
+    parser.add_argument(
+        "-id", "--cluster-id",
+        help="Cluster id to install",
+        type=str,
+        default=None
```
RazRegev, line 40: don't think you need to explicitly add `default=None `

discovery-infra/ocp.py
```diff
@@ -0,0 +1,85 @@
...
+    client = assisted_service_api.create_client(
+            url=utils.get_assisted_service_url_by_args(args=args)
+    )
+    return client.cluster_get(cluster_id=args.cluster_id)
+
+def main(args):
+    if args.config_etc_hosts:
+        ocp_cluster = get_ocp_cluster(args)
+        ocp_cluster_name = ocp_cluster.name
+        ocp_api_vip_dnsname = "api." + ocp_cluster_name + "." + ocp_cluster.base_dns_domain
```
RazRegev, line 27: nit: ` f"api.{ocp_cluster_name}.{ocp_cluster.base_dns_domain}`

scripts/deploy_assisted_service.sh
```diff
@@ -11,13 +11,25 @@ export SERVICE_PORT=$(( 6000 + $NAMESPACE_INDEX ))
 export SERVICE_BASE_URL=${SERVICE_BASE_URL:-"http://${SERVICE_URL}:${SERVICE_PORT}"}
 export EXTERNAL_PORT=${EXTERNAL_PORT:-y}
 export PROFILE=${PROFILE:-assisted-installer}
+export OCP_SERVICE_PORT=$(( 7000 + $NAMESPACE_INDEX ))
```
RazRegev, line 14: not sure I understand, why should it run with another  port then the default assisted-service port 6000?
as I understand they are not overriding each other

discovery-infra/ocp.py
danielerez: good call, better to change the default.

discovery-infra/ocp.py
danielerez: for sure? copied it from: https://github.com/openshift/assisted-test-infra/blob/a425c840c727b4e3be1c2341e8b70930b10cb216/discovery-infra/install_cluster.py#L158
maybe it's defaulted to an empty string if not specified? I'll give it a try

discovery-infra/ocp.py
danielerez: done

scripts/deploy_assisted_service.sh
danielerez: we should have only one per cluster, but 6000 is already used by the 'parent' service. so in order to reach both with same hostname, we use different ports.

danielerez:
@tsorya PTAL

RazRegev (approved):

PR #210, opened 2020-11-01
adding set_correct_boot_order_to_all_nodes as a setup before each test run

lalon4:


discovery-infra/tests/base_test.py
```diff
@@ -21,6 +21,7 @@ class BaseTest:
     @pytest.fixture(scope="function")
     def node_controller(self, setup_node_controller):
         controller = setup_node_controller
+        controller.set_correct_boot_order_to_all_nodes()
```
tsorya, line 24: why it is here? it must be part of teardown no?

tsorya: talked

PR #211, opened 2020-11-01
MGMT-2703 - optionally wait for CVO to become available after install…

yevgeny-shnaidman:
…ation

### DIFF
--- a/discovery-infra/start_discovery.py
+++ b/discovery-infra/start_discovery.py
@@ -347,6 +347,10 @@ def nodes_flow(client, cluster_name, cluster, image_path):
             )
             # Validate DNS domains resolvability
             validate_dns(client, cluster.id)
+            if args.wait_for_cvo:
+                cluster_info = client.cluster_get(cluster.id)
+                utils.config_etc_hosts(cluster_info)
+                utils.wait_for_cvo_available()
 
 
 def override_ignition(client, cluster_id):
@@ -537,6 +541,12 @@ if __name__ == "__main__":
         "-in",
         "--install-cluster",
         help="Install cluster, will take latest id",
+        action="store_true",
+    )
+    parser.add_argument(
+        "-cvo",
+        "--wait-for-cvo",
+        help="Wait for CVO available after cluster installation",
         action="store_true",
     )
     parser.add_argument(
--- a/discovery-infra/utils.py
+++ b/discovery-infra/utils.py
@@ -102,6 +102,29 @@ def get_libvirt_nodes_mac_role_ip_and_name(network_name):
         raise
 
 
+def wait_for_cvo_available():
+    waiting.wait(
+        lambda: is_cvo_available(),
+        timeout_seconds=3600,
+        sleep_seconds=20,
+        waiting_for="CVO to become available",
+    )
+
+
+def is_cvo_available():
+    try:
+        res = subprocess.check_output("kubectl --kubeconfig=build/kubeconfig get clusterversion -o json", shell=True)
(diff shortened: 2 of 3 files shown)


discovery-infra/utils.py
```diff
@@ -525,3 +549,15 @@ def set_hosts_roles_based_on_requested_name(client, cluster_id):
     client.set_hosts_roles(cluster_id=cluster_id, hosts_with_roles=hosts_with_roles)
 
 
+def config_etc_hosts(cluster):
```
tsorya, line 552: u don't need it. With latest commit from @YuviGold  u have dns already

discovery-infra/start_discovery.py
```diff
@@ -347,6 +347,10 @@ def nodes_flow(client, cluster_name, cluster, image_path):
             )
             # Validate DNS domains resolvability
             validate_dns(client, cluster.id)
+            if args.wait_for_cvo:
+                cluster_info = client.cluster_get(cluster.id)
+                utils.config_etc_hosts(cluster_info)
```
romfreiman, line 352: why isnt it the default? Shouldnt we always do some verification on the installed OCP?

discovery-infra/utils.py
```diff
@@ -102,6 +102,29 @@ def get_libvirt_nodes_mac_role_ip_and_name(network_name):
...
+        sleep_seconds=20,
+        waiting_for="CVO to become available",
+    )
+
+
+def is_cvo_available():
+    try:
+        res = subprocess.check_output("kubectl --kubeconfig=build/kubeconfig get clusterversion -o json", shell=True)
+        conditions = json.loads(res)['items'][0]['status']['conditions']
+        for condition in conditions:
```
romfreiman, line 118: maybe we should print the output here? Otherwise it will be hard to debug, no?

romfreiman:
You can adddress my comments once it's merged.

@yevgeny-shnaidman @YuviGold how will prow execute this?

discovery-infra/start_discovery.py
yevgeny-shnaidman: i think that right now most of the people want to check the usual install. I have added a new target for those that want to wait for CVO and debug.That way we don't interfere with CI and PRs

discovery-infra/utils.py
yevgeny-shnaidman: the main debugging will be done by the must-gather.  right now, even when we have issues with etcd, what we will see in the conditions is: multiple errors with api server endpoint oath. It does not tell much

tsorya: u can look on failure status, u will all those errors

discovery-infra/start_discovery.py
romfreiman: sorry if I was not clear - I specifically addressed  utils.config_etc_hosts(cluster_info), not the whole function

yevgeny-shnaidman: aaa-a-a. 
1) right now we need it only for CVO
2) this whole issue should be solved by @YuviGold commit of set_dns. unfortunately, on some servers ( for example HW Lab) this cause issues with resolv.conf, which causes the whole flow to fail (@tsorya wa trying to look into it). I prefer to push it with changes to etc/hosts and not to wait till the set_dns is foixed. Once the fix is in place, i will simply remove etc/hosts configurations

discovery-infra/utils.py
tsorya: i think we miss logs here

PR #212, opened 2020-11-01
MGMT-2137 Add public registries variable

empovit:


YuviGold (approved):
LGTM

carbonin:
This looks like it's still using the release source from before https://github.com/openshift/release/pull/13339
Going to close/reopen to see if that gets the new source

PR #213, opened 2020-11-01
adding cancel reset test

lalon4:


discovery-infra/tests/test_cancel_and_reset.py
```diff
@@ -90,3 +90,48 @@ def test_cancel_reset_after_node_boot(self, api_client, node_controller, cluster
...
+        # Cancel cluster install once cluster installation start
+        self.wait_for_installing_in_progress(cluster_id=cluster_id, api_client=api_client)
+        self.cancel_cluster_install(cluster_id=cluster_id, api_client=api_client)
+        assert self.is_cluster_in_cancelled_status(
+            cluster_id=cluster_id,
+            api_client=api_client
+        )
+        # Shutdown one node
+        nodes = node_controller.list_nodes()
+        node = list(nodes.values())[0]
```
lalon4, line 116: this one depending on https://github.com/openshift/assisted-test-infra/pull/214/files
will fix it once it merged

discovery-infra/tests/test_cancel_and_reset.py
```diff
@@ -90,3 +90,48 @@ def test_cancel_reset_after_node_boot(self, api_client, node_controller, cluster
...
+        node.shutdown()
+        # Reset cluster install
+        self.reset_cluster_install(cluster_id=cluster_id, api_client=api_client)
+        assert self.is_cluster_in_insufficient_status(
```
nshidlin, line 120: Isn't the cluster in insufficient status regardless of whether or not a node is shutdown? Is there a way to see why it's in that status and assert on that?

PR #221, opened 2020-11-02
Adding negative user and agent auth test

nshidlin:
Changes:
Adding x-secret-key to api client along with agentauth methods
Allowing multiple api-clients with different auth creds in a test
Adding negative user and agent auth test


discovery-infra/tests/test_auth.py
```diff
@@ -0,0 +1,282 @@
...
+from tests.conftest import env_variables
+from assisted_service_client.rest import ApiException
+from test_infra import assisted_service_api, consts, utils
+
+def random_name():
+    return ''.join(random.choice(ascii_lowercase) for i in range(10))
+
+
+@pytest.fixture()
+def api_client():
```
nshidlin, line 16: Subsequent PR will unify this with fixture from conftest and update existing tests

discovery-infra/tests/test_auth.py
```diff
@@ -0,0 +1,282 @@
...
+                utils.get_env('PROFILE'), utils.get_env('NAMESPACE'), 'assisted-service', utils.get_env('DEPLOY_TARGET'))
+        
+        return assisted_service_api.create_client(url, offline_token, **kwargs)
+
+
+    yield get_api_client
+
+@pytest.mark.skipif(not env_variables['offline_token'], reason="not cloud env")
+class TestAuth(BaseTest):
+    @pytest.fixture()
```
nshidlin, line 31: Subsequent PR will unify this with fixture from BaseTest and update existing tests

discovery-infra/test_infra/assisted_service_api.py
```diff
@@ -13,19 +13,19 @@
 
 
 class InventoryClient(object):
-    def __init__(self, inventory_url):
+    def __init__(self, inventory_url, offline_token, pull_secret):
```
tsorya, line 16: we can have client without offline token and pull secret right? I think we must have defaults here

tsorya: they can be None though

discovery-infra/tests/base_test.py
```diff
@@ -258,3 +258,57 @@ def wait_for_nodes_status_installing_or_installed(
                       consts.NodesStatus.INSTALLED],
             interval=30,
         )
+
```
tsorya, line 261: I think base test starts to be too large file, lets create cluster object and move all cluster functionality to it

discovery-infra/tests/base_test.py
```diff
@@ -258,3 +258,57 @@ def wait_for_nodes_status_installing_or_installed(
...
+    @staticmethod
+    def assert_http_error_code(api_client, api_call, status, reason, **kwargs):
+        with pytest.raises(ApiException) as response:
+            api_call(api_client=api_client, **kwargs)
+        assert response.value.status == status
+        assert response.value.reason == reason
+
+    @staticmethod
+    def get_cluster(cluster_id, api_client):
```
tsorya, line 293: not sure i understand why we add static function for every api call?

discovery-infra/tests/conftest.py
```diff
@@ -24,7 +24,9 @@
 env_variables = {"ssh_public_key": utils.get_env('SSH_PUB_KEY'),
                  "remote_service_url": utils.get_env('REMOTE_SERVICE_URL'),
                  "pull_secret": utils.get_env('PULL_SECRET'),
+                 "second_pull_secret": utils.get_env('SECOND_PULL_SECRET'),
```
tsorya, line 27: why os env? it is not something test specific?

discovery-infra/tests/test_auth.py
```diff
@@ -0,0 +1,282 @@
...
+        #user2 cannot generate ISO user1's cluster
+        self.assert_http_error_code(
+            api_client=client_user2,
+            api_call=self.generate_and_download_image,
+            status=404,
+            reason="Not Found",
+            cluster_id=user1_cluster_id
+        )
+
+        self.generate_and_download_image(cluster_id=user1_cluster_id, api_client=client_user1)
```
tsorya, line 99: i think we have this part in every test we add. please lets move all those thing to cluster object and reuse

discovery-infra/test_infra/assisted_service_api.py
nshidlin: default is here:
 https://github.com/openshift/assisted-test-infra/blob/bd4328952346c3cdc17b2ca15f3b3bd3824fb35d/discovery-infra/test_infra/assisted_service_api.py#L323

discovery-infra/tests/conftest.py
tsorya: i am not sure somebody will ever remember to add SECOND_PULL_SECRET to run it's test same with SECOND_OFFLINE_TOKEN

discovery-infra/tests/base_test.py
nshidlin: I need the functions to be called in the same way for:
https://github.com/openshift/assisted-test-infra/blob/bd4328952346c3cdc17b2ca15f3b3bd3824fb35d/discovery-infra/tests/base_test.py#L286

discovery-infra/tests/conftest.py
```diff
@@ -52,13 +52,16 @@
...
-            utils.get_env('PROFILE'), utils.get_env('NAMESPACE'), 'assisted-service', utils.get_env('DEPLOY_TARGET'))
+def get_api_client(offline_token=env_variables['offline_token'], **kwargs):
+        url = env_variables['remote_service_url']
 
-    yield assisted_service_api.create_client(url)
+        if not url:
+            url = utils.get_local_assisted_service_url(
+                utils.get_env('PROFILE'), utils.get_env('NAMESPACE'), 'assisted-service', utils.get_env('DEPLOY_TARGET'))
+        
+        return assisted_service_api.create_client(url, offline_token, **kwargs)
```
osherdp, line 64: You have two tabs of indentation here

discovery-infra/tests/test_auth.py
```diff
@@ -0,0 +1,216 @@
...
+    yield get_api_client
+
+@pytest.mark.skipif(not env_variables['offline_token'], reason="not cloud env")
+class TestAuth(BaseTest):
+    @pytest.fixture()
+    def cluster(self):
+        clusters = []
+
+        def get_cluster_func(api_client, cluster_name: Optional[str] = None, cluster_id: Optional[str] = None):
+
```
osherdp, line 26: delete blank line

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -28,6 +28,21 @@ def delete(self):
         self.api_client.delete_cluster(self.id)
 
 
+    def get_details(self):
+        self.api_client.cluster_get(self.id)
```
osherdp, line 32: It shouldn't return anything?
a bit weird for a get method

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -28,6 +28,21 @@ def delete(self):
...
+        return self.api_client.get_cluster_hosts(self.id)
+
+    def get_host_ids(self):
+        host_ids = []
+        hosts = self.get_hosts()
+
+        for host in hosts:
+            host_ids.append(host['id'])
+        
+        return host_ids
```
osherdp, line 44:
```suggestion
        return [host["id"] for host in self.get_hosts()]
```

osherdp (changes requested):

PR #224, opened 2020-11-02
add role selection tests

osherdp:
* Test letting the system do all role assignments automatically
* Test making some of the assignments and let the system fill-in for the other hosts
* TODO: test masters with not enough capabilities scenario


slaviered:
1. It might be a good idea to extract the steps to a common place because many tests used them 
2. Please don't set the default worker to 2. It really slows down the flow and most tests do not need them. If you run your test with the env variable of WORKER_COUNT=2 you get the desired result

osherdp:
@slaviered how can I specify this env variable only for my test?
I think that the relevant environment (e.g. Jenkins) sets it globally for all the tests

nshidlin:
What worries me about pytest-step, is that it makes it tricky to follow teardown and setup in fixtures. Why not just group these steps into helper functions? wdty?

osherdp:
@nshidlin for those tests I suppose it can be done, I'll work on it

But in the more complex tests it will get unreadable I think, so we should address it next
I suppose cleanup should be implemented in the base class, leaving the test entirely clean from consequences :)

Makefile
```diff
@@ -25,7 +25,7 @@ ISO := $(or $(ISO), "") # ISO should point to a file that has the '.iso' extensi
 NUM_MASTERS :=  $(or $(NUM_MASTERS),3)
 WORKER_MEMORY ?= 8892
 MASTER_MEMORY ?= 16984
-NUM_WORKERS := $(or $(NUM_WORKERS),0)
+NUM_WORKERS := $(or $(NUM_WORKERS),2)
```
tsorya, line 28: lets change this envs default only for tests, not sure though how to do it in Makefie

discovery-infra/tests/test_auto_role_selection.py
```diff
@@ -0,0 +1,119 @@
...
+    def prepare_for_installation(self, steps_data):
+        self.set_network_params(cluster_id=steps_data.cluster.id,
+                                api_client=steps_data.api_client,
+                                controller=steps_data.node_controller)
+    
+    def expect_ready_to_install(self, steps_data):
+        self.wait_until_cluster_is_ready_for_install(cluster_id=steps_data.cluster.id,
+                                                     api_client=steps_data.api_client)
+    
+    @depends_on(prepare_for_installation)
```
tsorya, line 38: nice

Makefile
```diff
@@ -327,7 +327,7 @@ _lint:
 	pre-commit run --all-files
 
 test:
-	skipper make $(SKIPPER_PARAMS) _test
+	NUM_WORKERS=2 skipper make $(SKIPPER_PARAMS) _test
```
osherdp, line 330: @slaviered is this what you meant?

Makefile
osherdp: look now, changed only for the ``test`` target

Makefile
tsorya: maybe better to move it to conftest.py? and do something like this max(2, os.get(NUM_WORKERS)) ? to be able to add more hosts?

discovery-infra/tests/base_test.py
```diff
@@ -86,10 +86,26 @@ def generate_and_download_image(
             image_path=iso_download_path
         )
 
+    @staticmethod
+    def wait_until_hosts_are_registered(
```
slaviered, line 90: per the discussion in the channel, this should be unified with wait_until_host_are_discovered()

osherdp: in another issue. lots of the tests are depending on ```wait_until__host_are_discovered()`` and I think they mean a non-insufficient discovered host

discovery-infra/tests/base_test.py
```diff
@@ -258,3 +285,56 @@ def wait_for_nodes_status_installing_or_installed(
...
+        hosts = api_client.get_cluster_hosts(cluster_id=cluster_id)
+        return {h["id"]: h["role"] for h in hosts}
+
+    def assert_assignments(self, actual_assignments, expected_assignments):
+        assert Counter(actual_assignments.values()) == expected_assignments
+    
+    def assert_manually_assigned_roles(self, actual_assignments, manually_assigned_roles):
+        assert set(tuple(a.values()) for a in manually_assigned_roles) <= set(actual_assignments.items())
+
+    def get_victims(self, hosts, host_type, count):
```
YuviGold, line 320: victims?

discovery-infra/tests/base_test.py
```diff
@@ -258,3 +285,56 @@ def wait_for_nodes_status_installing_or_installed(
...
+    
+    def start_installation(self, cluster_id, api_client):
+        self.start_cluster_install(cluster_id=cluster_id,
+                                   api_client=api_client)
+        self.wait_until_cluster_starts_installation(cluster_id=cluster_id,
+                                                    api_client=api_client)
+        hosts = api_client.get_cluster_hosts(cluster_id=cluster_id)
+        return {h["id"]: h["role"] for h in hosts}
+
+    def assert_assignments(self, actual_assignments, expected_assignments):
```
YuviGold, line 314: IMO helper functions shouldn't have asserts in order to support negative scenarios
Moreover, I don't understand the meaning of `assert_assignments`. Please have meaningful function names

YuviGold (approved):
/approve
/lgtm

discovery-infra/tests/base_test.py
osherdp: Done

discovery-infra/tests/base_test.py
osherdp: Done
In the long term I don't agree with you. It's not unit tests and it would become a nightmare to understand each and every assertion in the entire context of a more complex test. In the short-term of those initial tests you are right so I'm changing it

PR #225, opened 2020-11-02
Move CVO_PARAMS to be as part of ADDITIONAL_PARAMS

YuviGold:
In order to pass `-cvo` flag as a configuration I need to use `ADDITIONAL_PARAMS`
So we would be able to run a normal flow like
```bash
# 1. setups environment and runs everything
make

# 2. runs everything without setup
make run_full_flow_with_install 

# 3. deploy nodes and install
make deploy_nodes_with_install
```

And also the same processes passing parameters `-cvo` flag
```bash
# 1. exporting before running
export ADDITIONAL_PARAMS=-cvo && make

# 2. Passing parameters to make
make deploy_nodes_with_install ADDITIONAL_PARAMS=-cvo 
```

This way providing configuration, it is much easier to compound different workflows.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -235,13 +235,10 @@ install_cluster:
 #########
 
 _deploy_nodes:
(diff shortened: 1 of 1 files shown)


yevgeny-shnaidman:
i think you need to add a description somewhere regarding how to use ADDITIONAL params ( README) or maybe even in Makefile itself

tsorya:
like

romfreiman:
Birthday present to @yevgeny-shnaidman 
/approve

PR #226, opened 2020-11-02
Added test cases to test_discovery_ignition_override

eranco74:


carbonin:
Looks like you committed a `.swp` file

discovery-infra/tests/test_discovery_ignition_override.py
```diff
@@ -1,37 +1,159 @@
...
         # Boot nodes into ISO
         hosts = node_controller.start_all_nodes()
         # Wait until hosts are discovered and update host roles
         self.wait_until_hosts_are_discovered(cluster_id=cluster_id, api_client=api_client)
         # Verify override
+        self.validate_ignition_override(hosts, TestDiscoveryIgnition.override_path)
+
+    @pytest.mark.regression
+    def test_discovery_ignition_after_ISO_was_created(self, api_client, node_controller, cluster):
+        """ Verify that we create a new ISO upon discovery igniton override in case one already exists.
```
carbonin, line 54: I don't think we do this, but would probably be a good enhancement.

discovery-infra/tests/test_discovery_ignition_override.py
```diff
@@ -1,37 +1,159 @@
...
+    def validate_ignition_override(self, hosts, file_path, expected_content=TEST_STRING):
         for test_host in hosts.values():
-            file_content = test_host.run_command("cat {}".format(override_path))
-            assert file_content == test_string
+            logging.info("Verifying {} applied the override".format(test_host.name))
+            file_content = test_host.run_command("cat {}".format(file_path))
+            assert file_content == expected_content
 
     @pytest.mark.regression
     def test_discovery_ignition_bad_version(self, api_client, node_controller, cluster):
```
carbonin, line 89: This is covered by unit tests ... but :man_shrugging:

discovery-infra/tests/test_discovery_ignition_override.py
```diff
@@ -1,37 +1,159 @@
...
+        api_client.patch_cluster_discovery_ignition(cluster_id, ignition_override_2)
+
+        # Generate and download cluster ISO
+        self.generate_and_download_image(cluster_id=cluster_id, api_client=api_client)
+        # Boot nodes into ISO
+        hosts = node_controller.start_all_nodes()
+        # Wait until hosts are discovered and update host roles
+        self.wait_until_hosts_are_discovered(cluster_id=cluster_id, api_client=api_client)
+        # Verify override
+        self.validate_ignition_override(hosts, override_path)
```
carbonin, line 158: This will fail I think. We don't merge subsequent overrides into the original, it's a full replacement.

carbonin (commented):
We can probably remove https://github.com/openshift/assisted-test-infra/blob/3196d35a5de531eb315c4fa6d76b85d89c87256b/discovery-infra/start_discovery.py#L356-L370 as these tests will cover this behavior.

discovery-infra/tests/test_discovery_ignition_override.py
eranco74: I'll remove it

discovery-infra/tests/test_discovery_ignition_override.py
eranco74: I'll open a ticket for it and comment out the test

carbonin (approved):

discovery-infra/tests/test_discovery_ignition_override.py
```diff
@@ -1,37 +1,142 @@
...
+import copy
 import pytest
 import base64
 from tests.base_test import BaseTest
 
+TEST_STRING = "I can write tests all day long"
+
 
 class TestDiscoveryIgnition(BaseTest):
+    base64_encoded_test_string = base64.b64encode(TEST_STRING.encode("utf-8")).decode("utf8")
```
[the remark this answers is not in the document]
eranco74: Without encoding it will fail:
`
TypeError: a bytes-like object is required, not 'str'
`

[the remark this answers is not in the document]
osherdp: No, I mean omitting encoding type. Like so: ``TEST_STRING.encode()``

PR #229, opened 2020-11-02
add cancel reset test with disabled worker nodes

lalon4:


discovery-infra/tests/test_cancel_and_reset.py
```diff
@@ -135,3 +136,48 @@ def test_cancel_reset_one_node_unavailable(self, api_client, node_controller, cl
...
+        self.wait_until_hosts_are_discovered(cluster_id=cluster_id,
+                                             api_client=api_client,
+                                             nodes_count=env_variables['num_masters'])
+        self.wait_until_cluster_is_ready_for_install(cluster_id=cluster_id, api_client=api_client)
+        # Install Cluster
+        self.start_cluster_install(cluster_id=cluster_id, api_client=api_client)
+        self.wait_for_nodes_to_install(cluster_id=cluster_id,
+                                       api_client=api_client,
+                                       nodes_count=env_variables['num_masters'])
+        self.wait_for_cluster_to_install(cluster_id=cluster_id, api_client=api_client)
```
tsorya, line 183: not sure we really need to wait for cluster full installation. Not enough to verify we nodes are installed in this case?

discovery-infra/tests/test_cancel_and_reset.py
```diff
@@ -1,6 +1,7 @@
 import pytest
 
 from tests.base_test import BaseTest
+from tests.conftest import env_variables
```
tsorya, line 4: can't we use it as Base test env?

lalon4: not that im aware of, any suggestions?

discovery-infra/tests/test_cancel_and_reset.py
lalon4: well. its not a must but we should test it at least once in the cancel\reset scope. there is no single installation completion in the cancel\reset class. WDYT? i can add a TODO or leave it as is

tsorya: agree i thought we will return the previous ones from TODO but as u decide

discovery-infra/tests/test_cancel_and_reset.py
tsorya: i think better to have func in base test that will return envs

PR #232, opened 2020-11-03
Adds Cluster class to perform operations on cluster within the tests

nshidlin:
The idea is to migrate helper functions from BaseTest to Cluster. 
No breaking changes yet. Subsequent PRs will rebase existing tests to use Cluster object
test_cancel_reset_before_node_boot is rebased to use Cluster object

### DIFF
--- a/discovery-infra/test_infra/helper_classes/cluster.py
+++ b/discovery-infra/test_infra/helper_classes/cluster.py
@@ -1,0 +1,174 @@
+import logging
+
+from tests.conftest import env_variables
+from test_infra import consts, utils
+
+
+class Cluster:
+    
+    def __init__(self, api_client, cluster_name, cluster_id=None):
+        self.api_client = api_client
+
+        if cluster_id:
+            self.id = cluster_id
+        else:
+            self.id = self._create(cluster_name).id
+    
+    def _create(self, cluster_name):
+        return self.api_client.create_cluster(
+            cluster_name,
+            ssh_public_key=env_variables['ssh_public_key'],
+            openshift_version=env_variables['openshift_version'],
+            pull_secret=env_variables['pull_secret'],
+            base_dns_domain=env_variables['base_domain'],
+            vip_dhcp_allocation=env_variables['vip_dhcp_allocation']
+        )
+
+    def delete(self):
+        self.api_client.delete_cluster(self.id)
+
+
+    def generate_and_download_image(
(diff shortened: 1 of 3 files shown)


discovery-infra/test_infra/cluster.py
```diff
@@ -0,0 +1,237 @@
...
+
+class Cluster:
+    
+    def __init__(self, api_client, cluster_name):
+        res = self._create(api_client, cluster_name)
+
+        self.id = res.id
+        self.api_client = api_client
+    
+    def _create(self, api_client, cluster_name):
```
tsorya, line 15: if we have api client in init , why we need to provide it to every function?

nshidlin: I need to be able override the client for auth tests

discovery-infra/test_infra/cluster.py
```diff
@@ -0,0 +1,237 @@
+import logging
+
+from tests.conftest import env_variables
+from test_infra import consts, utils
+
+
+class Cluster:
```
tsorya, line 7: maybe lets add folder for such sort of objects? not sure what its name but will be nice to have it in folder.
help_classes? maybe

discovery-infra/test_infra/cluster.py
```diff
@@ -0,0 +1,237 @@
...
+    
+    def __init__(self, api_client, cluster_name):
+        res = self._create(api_client, cluster_name)
```
tsorya, line 10: maybe lets allow to provide cluster id and if it is provided don't create one? just run get?

nshidlin: I don't see the use case for this, if the way to create a cluster is by creating a Cluster obeject. wdty?

discovery-infra/test_infra/cluster.py
```diff
@@ -0,0 +1,237 @@
...
+        api_client.generate_and_download_image(
+            cluster_id=self.id,
+            ssh_key=ssh_key,
+            image_path=iso_download_path
+        )
+
+    def wait_until_hosts_are_discovered(
+        self, api_client=None, nodes_count=env_variables['num_nodes']
+    ):
+        if api_client is None:
```
tsorya, line 50: if u need it for some reason, lets create decorator for this one, you can add this line to all functions.

PR #234, opened 2020-11-03
Add download logs tests

slaviered:


discovery-infra/tests/test_logs.py
```diff
@@ -0,0 +1,91 @@
...
+        #download logs into a file. At this point logs are not uploaded
+        path = "/tmp/test_on-success_logs.tar"
+        self.verify_no_logs_uploaded(cluster_id, api_client, path)
+        
+        #install the cluster
+        ts_start_at = time.time()
+        self.install_cluster_and_wait(cluster_id, api_client)
+
+        #download logs into a file. At this point logs exist
+        expected_min_log_num = len(node_controller.list_nodes()) + 1
```
YuviGold, line 48: Comment why `+1`

discovery-infra/tests/test_logs.py
```diff
@@ -0,0 +1,91 @@
...
+        cluster_id = cluster().id
+        self.setup_hosts(cluster_id, api_client, node_controller)
+        self.prepare_for_installation(cluster_id, api_client, node_controller)
+        
+        #download logs into a file. At this point logs are not uploaded
+        path = "/tmp/test_on-success_logs.tar"
+        self.verify_no_logs_uploaded(cluster_id, api_client, path)
+        
+        #install the cluster
+        ts_start_at = time.time()
```
YuviGold, line 44: Please use clear variable names. I have no idea what's `ts` stands for.

discovery-infra/tests/test_logs.py
```diff
@@ -0,0 +1,91 @@
...
+        with pytest.raises(ApiException) as ex:
+            api_client.download_cluster_logs(cluster_id, path)
+
+    def verify_logs_uploaded(self, cluster_id, api_client, path, ts_start_at, expected_min_log_num):
+        api_client.download_cluster_logs(cluster_id, path)
+        tar = tarfile.open(path)
+        loglist = []
+        for tarinfo in tar:
+            loglist.append(tarinfo.name)
+            assert tarinfo.mtime > ts_start_at
```
YuviGold, line 89: You're just verifying when the tar items has been created. @tsorya isn't that going to be always up-to-date to when we download the logs? That's when the tar is being created.
So IMO this condition doesn't actually verify anything.
Maybe we can check last "updated_at" value of the last log to know when it was last updated.

discovery-infra/tests/test_logs.py
```diff
@@ -0,0 +1,91 @@
...
+
+        #download logs into a file. At this point logs exist
+        expected_min_log_num = len(node_controller.list_nodes()) + 1
+        self.verify_logs_uploaded(cluster_id, api_client, path, ts_start_at, expected_min_log_num)
+
+    
+    def test_collect_logs_on_failure(self, api_client, node_controller, cluster):
+        '''cacnel insllation after at least one host is booted and check that logs are uploaded'''
+        cluster_id = cluster().id
+        self.generate_and_download_image(cluster_id=cluster_id, api_client=api_client)
```
YuviGold, line 55: Why don't you use the previous functions you just defined? `setup_hosts` and `prepare_for_installation`

discovery-infra/tests/test_logs.py
```diff
@@ -0,0 +1,91 @@
+import pytest
+import tarfile
+import time
+
+from assisted_service_client.rest import ApiException
+from tests.base_test import BaseTest
+
+class TestDownloadLogs(BaseTest):
+    def setup_hosts(self, cluster_id, api_client, node_controller):
```
YuviGold, line 9: These are nice helper functions. They aren't only relevant for DownloadLogs scenarios. They can be very helpful as public helper functions

discovery-infra/tests/test_logs.py
```diff
@@ -0,0 +1,91 @@
...
+            api_client=api_client
+        )
+         #download logs into a file. At this point logs exist
+        path = "/tmp/test_on_cancel_logs.tar"
+        expected_min_log_num = 1 #at least the booted master has logs
+        self.verify_logs_uploaded(cluster_id, api_client, path, ts_start_at, expected_min_log_num)
+
+    def verify_no_logs_uploaded(self, cluster_id, api_client, path):
+        with pytest.raises(ApiException) as ex:
+            api_client.download_cluster_logs(cluster_id, path)
```
YuviGold, line 81: The name of this function is misleading because `ApiException` might raise even if there is no connection to the service.
The reason for the failure should be checked to actually prove there are no logs.

YuviGold (changes requested):

discovery-infra/tests/test_logs.py
lalon4: can we expect number of files?

discovery-infra/tests/test_logs.py
```diff
@@ -0,0 +1,91 @@
+import pytest
+import tarfile
+import time
+
+from assisted_service_client.rest import ApiException
+from tests.base_test import BaseTest
+
+class TestDownloadLogs(BaseTest):
```
lalon4, line 8: can we add a cleanup method for the downloaded file?

discovery-infra/tests/test_logs.py
```diff
@@ -0,0 +1,91 @@
...
+        self.set_network_params(cluster_id=cluster_id,
+                                 api_client=api_client,
+                                 controller=node_controller
+                                 )
+
+    def install_cluster_and_wait(self, cluster_id, api_client):
+        '''start the cluster and wait for host to go to installing-in-progress state'''
+        # Start cluster install
+        self.start_cluster_install(cluster_id=cluster_id, api_client=api_client)
+        # wait until all nodes are in Installed status, and the cluster moved to status installed
```
lalon4, line 30: Please add self.wait_for_nodes_to_install(cluster_id=cluster_id, api_client=api_client), so if one of the nodes fails to install, then test fails rather then wait for timeout

discovery-infra/tests/test_logs.py
lalon4: yes, agree. you can move it to the base. And in general, if your using other methods here then "test_" so add a _ prefix to it, to differ from tests methods and make it more readable

slaviered: They will go into the code line with osher's PR and then I'll refactor mine code to use them

discovery-infra/tests/test_logs.py
ybettan: Probably controller @YuviGold

discovery-infra/tests/test_logs.py
```diff
@@ -0,0 +1,91 @@
...
+        self.set_network_params(
+            cluster_id=cluster_id,
+            api_client=api_client,
+            controller=node_controller
+        )
+
+        ts_start_at = time.time()
+        self.start_cluster_install(cluster_id=cluster_id, api_client=api_client)
+        # Cancel cluster install once at least one host booted
+        self.wait_for_one_host_to_boot_during_install(cluster_id=cluster_id, api_client=api_client)
```
ybettan, line 68: Are you waiting for `boot` or `reboot` here?

discovery-infra/tests/test_logs.py
```diff
@@ -0,0 +1,91 @@
...
+        self.cancel_cluster_install(cluster_id=cluster_id, api_client=api_client)
+        assert self.is_cluster_in_cancelled_status(
+            cluster_id=cluster_id,
+            api_client=api_client
+        )
+         #download logs into a file. At this point logs exist
+        path = "/tmp/test_on_cancel_logs.tar"
+        expected_min_log_num = 1 #at least the booted master has logs
```
ybettan, line 76: Here I think we should expect `expected_min_log_num = #rebooted masters + 2` - this time `+1` refers to the `installer-gather.sh` logs generated in case of cluster installation failure/cancelation and another `+1` for bootstrap logs.

discovery-infra/tests/test_logs.py
ybettan: Agree, I think it should be `503` (conflict)

discovery-infra/tests/test_logs.py
tsorya: @YuviGold i think it will be too much to validate log line. we can though validate log_uploaded_at flag per host and controller

discovery-infra/tests/test_logs.py
slaviered: We can expect exact number of files only on the happy flow. On the cancel flow, it depends on the host status which differ from run to run. As for cleaning the file, I will add a delete but since they are in /tmp they will be also cleared automatically, eventually

discovery-infra/tests/test_logs.py
slaviered: @ybettan the api retruns 500 with a message. asserted on that

discovery-infra/tests/test_logs.py
slaviered: the result shows that the collected logs belong only to the booted host

discovery-infra/tests/test_logs.py
```diff
@@ -0,0 +1,104 @@
...
+        #verify that logs were collected from all expected sources
+        assert(os.path.exists(path))
+        try:
+            tar = tarfile.open(path)
+            logging.info(f'downloaded logs: {tar.getnames()}')
+            assert len(tar.getnames()) >= expected_min_log_num
+            tar.close()
+            #verify that logs are recent
+            for collected_at in logs_collected_at:
+                 if collected_at > time.time() - 86400000:
```
tsorya, line 98: not sure i understand what is this int?

tsorya: i think u need to verify that collected_at is bigger than starting_installation time, just to be sure those are latest logs and relevant to specific installation

slaviered: In the intermdeidate state not all host has valid collected_at so this filters the illegal timestamps out

tsorya: as i wrote in slack, the next one that will create new test that state will be different in it. for example get logs after first installation was reset. This func will not validate it right in case of bug. So why just not to take time before installation in UTC and validate on it?

PR #236, opened 2020-11-03
Revert "Terraform utils fixes"

YuviGold:
Reverts openshift/assisted-test-infra#228


YuviGold:
Fixed on https://github.com/openshift/assisted-test-infra/pull/237

PR #254, opened 2020-11-08
Igal/nodes

tsorya:


discovery-infra/test_infra/helper_classes/nodes.py
```diff
@@ -0,0 +1,121 @@
...
+    def get_random_node(self):
+        return random.choice(self.nodes)
+
+    def shutdown_all(self):
+        self.run_for_all_nodes("shutdown")
+
+    def start_all(self):
+        self.run_for_all_nodes("start")
+
+    def format_all(self):
```
nshidlin, line 58: format_all_disks would be a better name

discovery-infra/tests/base_test.py
```diff
@@ -333,10 +318,3 @@ def assign_roles(self, cluster_id, api_client, hosts, requested_roles):
             hosts_with_roles=assigned_roles)
 
         return assigned_roles
-        
-    @staticmethod
-    def assert_http_error_code(api_call, status, reason, **kwargs):
```
nshidlin, line 338: why did you remove this? I use it in test_auth

PR #258, opened 2020-11-09
Add validation E2E tests

avishayt:


discovery-infra/test_infra/utils.py
```diff
@@ -346,6 +346,91 @@ def is_cluster_in_status(client, cluster_id, statuses):
         log.exception("Failed to get cluster %s info", cluster_id)
 
 
+def wait_till_cluster_validation_is_in_status(
```
nshidlin, line 349: these functions can go in Cluster class

discovery-infra/tests/test_validations.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class TestValidations(BaseTest):
+    @pytest.mark.regression
+    def test_basic_cluster_validations(self, api_client, cluster):
+        cluster_id = cluster().id
+
+        # Check initial validations
+        self.wait_until_cluster_is_pending_for_input(cluster_id, api_client)
+        orig_cluster = api_client.cluster_get(cluster_id)
```
nshidlin, line 16: use cluster.get_details

discovery-infra/tests/test_validations.py
```diff
@@ -0,0 +1,76 @@
...
+        vcpu_node.shutdown()
+        vcpu_node.set_cpu_cores(orig_vcpu_value)
+        vcpu_node.start()
+
+        ram_node.shutdown()
+        ram_node.set_ram_kib(int(env_variables['master_memory']) * 1024)
+        ram_node.start()
+
+        self.wait_until_hosts_are_discovered(cluster_id=cluster_id, api_client=api_client, additional_statuses=[consts.NodesStatus.INSUFFICIENT])
+        self.wait_for_host_validation(api_client, cluster_id, vcpu_host_id, "hardware", "has-min-cpu-cores", ["success"])
```
osherdp, line 75: Maybe the strings part will benefit from keyword arguments, to explain what ``"hardware"``, ``"has-min-cpu-cores"`` and ``["success"]`` are exactly

discovery-infra/test_infra/controllers/node_controllers/node.py
```diff
@@ -98,3 +98,15 @@ def set_boot_order_flow(self, cd_first=False, start=True):
...
+    def get_host_id(self):
+        return self.node_controller.get_host_id(self.name)
+
+    def get_cpu_cores(self):
+        return self.node_controller.get_cpu_cores(self.name)
+
+    def set_cpu_cores(self, core_count):
+        self.node_controller.set_cpu_cores(self.name, core_count)
+
+    def set_ram_kib(self, ram_kib):
```
osherdp, line 111: kib meaning 1000 bytes? maybe add docstring so people will know definitely the units

discovery-infra/tests/test_validations.py
```diff
@@ -0,0 +1,76 @@
...
+        self.wait_for_cluster_validation(api_client, cluster_id, "network", "dns-domain-defined", ["failure"])
+        api_client.update_cluster(cluster_id, {"base_dns_domain": orig_cluster.base_dns_domain})
+        self.wait_for_cluster_validation(api_client, cluster_id, "network", "dns-domain-defined", ["success"])
+
+        # Check pull secret
+        api_client.update_cluster(cluster_id, {"pull_secret": ""})
+        self.wait_for_cluster_validation(api_client, cluster_id, "configuration", "pull-secret-set", ["failure"])
+        api_client.update_cluster(cluster_id, {"pull_secret": env_variables['pull_secret']})
+        self.wait_for_cluster_validation(api_client, cluster_id, "configuration", "pull-secret-set", ["success"])
+
```
lalon4, line 44: add regression mark

discovery-infra/tests/test_validations.py
```diff
@@ -0,0 +1,76 @@
...
+    def test_host_insufficient(self, api_client, cluster, nodes):
+        cluster_id = cluster().id
+        self.generate_and_download_image(cluster_id=cluster_id, api_client=api_client)
+
+        # Modify vCPU count on the first node to be insufficient
```
osherdp, line 49: is it possible to replace comments with logs?

discovery-infra/tests/base_test.py
```diff
@@ -216,6 +220,14 @@ def is_cluster_in_insufficient_status(cluster_id, api_client):
             statuses=[consts.ClusterStatus.INSUFFICIENT]
         )
 
+    @staticmethod
+    def wait_until_cluster_is_pending_for_input(cluster_id, api_client):
```
lalon4, line 224: move to cluster class

discovery-infra/test_infra/utils.py
```diff
@@ -346,6 +346,91 @@ def is_cluster_in_status(client, cluster_id, statuses):
...
+                validation_section=validation_section,
+                validation_id=validation_id,
+                statuses=statuses
+            ),
+            timeout_seconds=timeout,
+            sleep_seconds=interval,
+            waiting_for="Cluster validation to be in status %s" % statuses,
+        )
+    except:
+        log.error("Cluster validation status is: %s", get_cluster_validation_value(client.cluster_get(cluster_id), validation_section, validation_id))
```
lalon4, line 367: Python PEP8 recommend no more then 80 in one line, try to break the line whenever needed. top 120 is OK

discovery-infra/tests/base_test.py
```diff
@@ -106,13 +106,17 @@ def wait_until_hosts_are_registered(
     def wait_until_hosts_are_discovered(
         cluster_id,
         api_client,
-        nodes_count=env_variables['num_masters']+env_variables['num_workers']
+        nodes_count=env_variables['num_masters']+env_variables['num_workers'],
```
lalon4, line 109: base class methods moved to cluster, we are deprecating most of the methods here

discovery-infra/tests/test_validations.py
```diff
@@ -0,0 +1,76 @@
...
+        self.assert_cluster_validation(orig_cluster, "network", "ingress-vip-valid", "pending")
+        self.assert_cluster_validation(orig_cluster, "network", "dns-domain-defined", "success")
+        self.assert_cluster_validation(orig_cluster, "network", "cluster-cidr-defined", "success")
+        self.assert_cluster_validation(orig_cluster, "network", "service-cidr-defined", "success")
+        self.assert_cluster_validation(orig_cluster, "network", "no-cidrs-overlapping", "pending")
+        self.assert_cluster_validation(orig_cluster, "network", "network-prefix-valid", "success")
+        self.assert_cluster_validation(orig_cluster, "network", "ntp-server-configured", "success")
+
+        # Check base DNS domain
+        api_client.update_cluster(cluster_id, {"base_dns_domain": ""})
```
nshidlin, line 34: same here use Cluster

osherdp (changes requested):

discovery-infra/tests/test_validations.py
lalon4, line 34: update empty dns doesn't fail?

discovery-infra/tests/test_validations.py
```diff
@@ -0,0 +1,76 @@
...
+        self.wait_for_cluster_validation(api_client, cluster_id, "configuration", "pull-secret-set", ["failure"])
+        api_client.update_cluster(cluster_id, {"pull_secret": env_variables['pull_secret']})
+        self.wait_for_cluster_validation(api_client, cluster_id, "configuration", "pull-secret-set", ["success"])
+
+    def test_host_insufficient(self, api_client, cluster, nodes):
+        cluster_id = cluster().id
+        self.generate_and_download_image(cluster_id=cluster_id, api_client=api_client)
+
+        # Modify vCPU count on the first node to be insufficient
+        vcpu_node = nodes.nodes[0]
```
lalon4, line 50: use nodes.get_random_node(), then node.name

discovery-infra/tests/test_validations.py
lalon4: all the methods that being called have logs, why do we need it here?

osherdp: Even if the getter methods have logs (which I'm not sure of it), logs on the test level should explain the macro of what the test expects. Something like ``"choosing randomly host and change its CPU cores number to be non-sufficient as a master"``
The implementation logs should be more implementation-driven, to give details in a case of a bug
I also would like our test logs to be of level of info and implementation to have debug level, but it's a more major change

lalon4: in my POV it is bad practice to log too much, and to log each step in the test level sounds irrelevant as in any case there is a need to understand test purpose, by reading its code (and comments) or a written TP.

discovery-infra/test_infra/controllers/node_controllers/node.py
avishayt: No, KiB is 1024 - I put the units into the variable name so that it will be clear

discovery-infra/test_infra/utils.py
avishayt: Will change, but if you want to adhere to pep8 it should run as part of the make process

discovery-infra/tests/test_validations.py
avishayt: It's not a required parameter when creating the cluster, so the user should be allowed to unset the value

discovery-infra/tests/test_validations.py
osherdp: It's fine, we can agree to disagree :)
But I'm sure that if we could have the framework to report on the major things it would benefit us. Probably more on the more complex tests than on the simple ones

discovery-infra/tests/test_validations.py
avishayt: How do I know that I won't get the same random node below, for ram_node?

discovery-infra/tests/test_validations.py
lalon4: but we do have a validation for dns, did the test passed?

discovery-infra/tests/test_validations.py
lalon4: Agree :)

discovery-infra/tests/test_validations.py
lalon4: ho, i thought that node.nodes is also kind of a random

discovery-infra/test_infra/utils.py
lalon4: it is just that the line is too long and less convenient to read via github and on editor

discovery-infra/tests/test_validations.py
```diff
@@ -0,0 +1,76 @@
...
+        self.wait_for_host_validation(api_client, cluster_id, vcpu_host_id, "hardware", "has-min-cpu-cores", ["failure"])
+        self.wait_for_host_validation(api_client, cluster_id, ram_host_id, "hardware", "has-min-memory", ["failure"])
+
+        # Return to original settings and verify that validations pass
+        vcpu_node.shutdown()
+        vcpu_node.set_cpu_cores(orig_vcpu_value)
+        vcpu_node.start()
+
+        ram_node.shutdown()
+        ram_node.set_ram_kib(int(env_variables['master_memory']) * 1024)
```
lalon4, line 71: Im thinking now that if test will fail somewhere, then there wont be a clean up for the CPU \ RAM changes (to bring it back to correct values). 
can you please add a fixture for applying a proper cleanup for this test?

discovery-infra/tests/test_validations.py
avishayt: We have validation, if it's non-empty.  Yes, the test passes.

discovery-infra/tests/test_validations.py
avishayt: We reuse VMs between tests?

lalon4: yes, but; 
- stop all nodes
- format disk
- set correct boot order
before\after each test

lalon4: see https://github.com/openshift/assisted-test-infra/blob/master/discovery-infra/tests/base_test.py#L25

avishayt: Done

discovery-infra/tests/test_validations.py
avishayt: I think it's clear, given the method deals with validations?

avishayt:
@lalon4 @nshidlin Please take another look

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -337,3 +358,73 @@ def setup_nodes(self, nodes):
         nodes.start_all()
         self.wait_until_hosts_are_discovered(nodes_count=len(nodes))
         return nodes.create_nodes_cluster_hosts_mapping(cluster=self)
+
+    def wait_for_cluster_validation(
```
nshidlin, line 362: naming: wait_for_validation
wait_for_host_validation already differentiates 
wdty?

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -149,6 +163,13 @@ def wait_for_cluster_in_error_status(self):
             statuses=[consts.ClusterStatus.ERROR]
         )
 
+    def wait_for_cluster_in_pending_for_input_status(self):
```
nshidlin, line 166: naming: wait_for_pending_for_input_status

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -337,3 +358,73 @@ def setup_nodes(self, nodes):
...
+                waiting_for="Cluster validation to be in status %s" % statuses,
+            )
+        except:
+            logging.error("Cluster validation status is: %s",
+                utils.get_cluster_validation_value(
+                    self.api_client.cluster_get(self.id), validation_section,
+                        validation_id))
+            raise
+
+    def is_cluster_validation_in_status(
```
nshidlin, line 386: naming, same

discovery-infra/test_infra/helper_classes/cluster.py
avishayt: Isn't it better to be explicit?  If someone saw wait_for_validation they might assume it's for both?

nshidlin: okay, sure

discovery-infra/tests/base_test.py
```diff
@@ -26,6 +26,8 @@ def nodes(self, setup_node_controller):
         controller = setup_node_controller
         nodes = Nodes(controller, env_variables["private_ssh_key_path"])
         nodes.set_correct_boot_order(start_nodes=False)
+        nodes.run_for_all_nodes("reset_cpu_cores")
```
lalon4, line 29: missed that - this should be running after the yield (both this and the next line)
no?
save orig, run tests, do changes, then back to orig

discovery-infra/consts.py
```diff
@@ -10,6 +10,7 @@
 NODES_REGISTERED_TIMEOUT = 600
 CLUSTER_INSTALLATION_TIMEOUT = 60 * 60   # 60 minutes
 START_CLUSTER_INSTALLATION_TIMEOUT = 360
+VALIDATION_TIMEOUT = 3 * 60
```
tsorya, line 13: no need here. it will be removed

discovery-infra/test_infra/helper_classes/cluster.py
tsorya: @avishayt what is cluster validation?

discovery-infra/test_infra/utils.py
```diff
@@ -348,6 +348,30 @@ def is_cluster_in_status(client, cluster_id, statuses):
...
+    found_status = "validation not found"
+    validations = json.loads(cluster_info.validations_info)
+    for validation in validations[validation_section]:
+        if validation['id'] == validation_id:
+            found_status = validation['status']
+            break
+    return found_status
+
+
+def get_host_validation_value(cluster_info, host_id, validation_section, validation_id):
```
tsorya, line 361: why not to take relevant host_object?

discovery-infra/tests/base_test.py
tsorya, line 29: this is not something specific to test?
I think this must be part of specific test fixture there is not need for this for all tests

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -52,12 +53,17 @@ def generate_and_download_image(
             image_path=iso_download_path
         )
 
-    def wait_until_hosts_are_discovered(self,nodes_count=env_variables['num_nodes']):
+    def wait_until_hosts_are_discovered(self, nodes_count=env_variables['num_nodes'],
```
tsorya, line 56: @avishayt why we need additional_statuses ? why i can't do it with  wait_till_all_hosts_are_in_status?

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -337,3 +358,73 @@ def setup_nodes(self, nodes):
...
+                        validation_id))
+            raise
+
+    def is_cluster_validation_in_status(
+        self, validation_section, validation_id, statuses
+    ):
+        logging.info("Is cluster %s validation %s in status %s",
+            self.id, validation_id, statuses)
+        try:
+            return utils.get_cluster_validation_value(
```
tsorya, line 392: not sure why it is utils func? why it doesn't match cluster object?

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -337,3 +358,73 @@ def setup_nodes(self, nodes):
...
+                utils.get_host_validation_value(self.api_client.cluster_get(self.id),
+                host_id, validation_section, validation_id))
+            raise
+
+    def is_host_validation_in_status(
+            self, host_id, validation_section, validation_id, statuses
+    ):
+        logging.info("Is host %s validation %s in status %s", host_id, validation_id, statuses)
+        try:
+            return utils.get_host_validation_value(self.api_client.cluster_get(self.id),
```
tsorya, line 427: same question

discovery-infra/test_infra/utils.py
```diff
@@ -348,6 +348,30 @@ def is_cluster_in_status(client, cluster_id, statuses):
         log.exception("Failed to get cluster %s info", cluster_id)
 
 
+def get_cluster_validation_value(cluster_info, validation_section, validation_id):
```
tsorya, line 351: i don't think this is the place. It is not utils functions, lets move them to cluster object

discovery-infra/tests/base_test.py
lalon4: maybe we will leave it here anyway? its one place for cleanup, more intuitive. what WDYT?

discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
```diff
@@ -167,3 +167,38 @@ def set_boot_order(self, node_name, cd_first=False):
...
+        return int(memory_element.firstChild.nodeValue)
+
+    def set_ram_kib(self, node_name, ram_kib):
+        logging.info(f"Going to set memory to {ram_kib} for node: {node_name}")
+        xml = self._get_xml(node_name)
+        memory_element = xml.getElementsByTagName('memory')[0]
+        memory_element.firstChild.replaceWholeText(ram_kib)
+        dom = self.libvirt_connection.defineXML(xml.toprettyxml())
+        if dom is None:
+            raise Exception(f"Failed to set memory for node: {node_name}")
```
RazRegev, line 198: wouldn't it be better to use an assert instead of an exception?

PR #261, opened 2020-11-09
Refactor logs and cancel tests

lalon4:


discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -209,7 +209,7 @@ def wait_for_hosts_to_install(
             timeout=timeout,
         )
 
-    def wait_for_install(
+    def wait_for_cluster_install(
```
nshidlin, line 212: no need for cluster in the name

discovery-infra/tests/test_logs.py
```diff
@@ -10,112 +10,78 @@
...
-    def install_cluster_and_wait(self, cluster_id, api_client):
-        '''start the cluster and wait for host to go to installing-in-progress state'''
-        # Start cluster install
-        self.start_cluster_install(cluster_id=cluster_id, api_client=api_client)
-        # wait until all nodes are in Installed status, and the cluster moved to status installed
-        self.wait_for_nodes_to_install(cluster_id=cluster_id, api_client=api_client)
-        self.wait_for_cluster_to_install(cluster_id=cluster_id, api_client=api_client)
-
-    def reset_cluster_and_wait(self, cluster_obj, api_client, nodes):
+    def _reset_cluster_and_wait(self, cluster, nodes):
```
nshidlin, line 18: naming: maybe something like reset_install_and_wait_for_ready

discovery-infra/tests/test_logs.py
```diff
@@ -10,112 +10,78 @@
...
+        self._verify_logs_are_current(started_cluster_install_at, logs_collected_at)
 
-    def verify_no_logs_uploaded(self, cluster_id, api_client, path):
+    def _verify_no_logs_uploaded(self, cluster_id, api_client, path):
         with pytest.raises(ApiException) as ex:
             api_client.download_cluster_logs(cluster_id, path)
         assert "No log files" in str(ex.value)
 
-    def verify_logs_uploaded(self, cluster_id, api_client, path, expected_min_log_num):
+    def _verify_logs_uploaded(self, cluster_id, api_client, path, expected_min_log_num):
```
nshidlin, line 84: add download_logs to Cluster and use that, it will probably be useful  for other tests

discovery-infra/tests/test_logs.py
```diff
@@ -10,114 +10,80 @@
...
 class TestDownloadLogs(BaseTest):
-    def setup_hosts(self, cluster_id, api_client, nodes):
-        '''setup nodes from ISO image and wait until they are registered'''
-        # Generate and download cluster ISO
-        self.generate_and_download_image(cluster_id=cluster_id, api_client=api_client)
-        # Boot nodes into ISO
-        nodes.start_all()
-        # Wait until hosts are discovered and update host roles
-        self.wait_until_hosts_are_discovered(cluster_id=cluster_id, api_client=api_client)
+    def _install_cluster_and_wait(self, cluster):
```
tsorya, line 13: why not to have this func in cluster?
add one that does this with flags as:
def _install_cluster_and_wait(self, wait_for_hosts=True, wait_for_cluster_inst=True)

PR #263, opened 2020-11-10
refactor test_general

lalon4:


discovery-infra/tests/test_general.py
```diff
@@ -34,16 +35,17 @@ def test_cluster_unique_name(self, api_client, cluster):
...
 
+    @pytest.mark.regression
     def test_select_roles(self, api_client, cluster, nodes):
-        cluster_id = self.test_discovery(api_client, cluster, nodes)
-        self.set_host_roles(cluster_id=cluster_id, api_client=api_client)
-        hosts = api_client.get_cluster_hosts(cluster_id=cluster_id)
+        c = self.test_discovery(api_client, cluster, nodes)
+        c.set_host_roles()
+        hosts = c.get_hosts()
         for host in hosts:
```
nshidlin, line 49: you can use cluster.set_host_roles here

lalon4: but here we assert the role in between

nshidlin: right, missed that

YuviGold (approved):
/approve

PR #267, opened 2020-11-11
refactoring proxy and wrong boot order tests to use cluster

nshidlin:


discovery-infra/tests/test_wrong_boot_order.py
```diff
@@ -7,35 +7,25 @@ class TestWrongBootOrder(BaseTest):
     @pytest.mark.regression
     def test_wrong_boot_order_one_node(self, api_client, nodes, cluster):
```
lalon4, line 8: please remove api_client

nshidlin: changed

YuviGold (approved):
LGTM

PR #268, opened 2020-11-12
MGMT-2881: e2e tests: added wrong boot order tests

RazRegev:
These tests will include the new cluster status "installing-pending-user-action".

Update the existing test: "Installation succeeded after master node had incorrect boot order".
Create "Installation succeeded on incorrect boot order - timeout is ignored".
Create "Installation succeeded after all nodes have an incorrect boot order".
Create "Reset\Cancel from Incorrect boot order"

### DIFF
--- a/discovery-infra/test_infra/helper_classes/cluster.py
+++ b/discovery-infra/test_infra/helper_classes/cluster.py
@@ -235,6 +235,21 @@ class Cluster:
             fall_on_error_status=fall_on_error_status,
         )
 
+    def wait_for_hosts_to_be_in_wrong_boot_order(
+            self,
+            nodes_count=env_variables['num_nodes'],
+            timeout=consts.CLUSTER_INSTALLATION_TIMEOUT,
+            fall_on_error_status=True
+    ):
+        utils.wait_till_all_hosts_are_in_status(
+            client=self.api_client,
+            cluster_id=self.id,
+            statuses=[consts.ClusterStatus.INSTALLING_PENDING_USER_ACTION],
+            nodes_count=nodes_count,
+            timeout=timeout,
+            fall_on_error_status=fall_on_error_status
+        )
+
     def wait_for_ready_to_install(self):
         utils.wait_till_cluster_is_in_status(
             client=self.api_client,
@@ -401,7 +416,7 @@ class Cluster:
                 self.api_client.cluster_get(self.id),
                 validation_section, validation_id) in statuses
         except:
-            log.exception("Failed to get cluster %s validation info", self.id)
+            logging.exception("Failed to get cluster %s validation info", self.id)
 
     def wait_for_host_validation(
         self, host_id, validation_section, validation_id, statuses,
@@ -436,3 +451,17 @@ class Cluster:
                 host_id, validation_section, validation_id) in statuses
         except:
             logging.exception("Failed to get cluster %s validation info", self.id)
+
+    def wait_for_cluster_to_be_in_installing_pending_user_action_status(self):
+        utils.wait_till_cluster_is_in_status(
+            client=self.api_client,
+            cluster_id=self.id,
+            statuses=[consts.ClusterStatus.INSTALLING_PENDING_USER_ACTION]
+        )
+
+    def wait_for_cluster_to_be_in_installing_status(self):
+        utils.wait_till_cluster_is_in_status(
+            client=self.api_client,
+            cluster_id=self.id,
+            statuses=[consts.ClusterStatus.INSTALLING]
+        )
--- a/discovery-infra/tests/test_wrong_boot_order.py
+++ b/discovery-infra/tests/test_wrong_boot_order.py
@@ -1,8 +1,14 @@
 import pytest
+import time
+import logging
 
 from tests.base_test import BaseTest
 
 
+logger = logging.getLogger(__name__)
+
+
+@pytest.mark.wrong_boot_order
 class TestWrongBootOrder(BaseTest):
     @pytest.mark.regression
     def test_wrong_boot_order_one_node(self, nodes, cluster):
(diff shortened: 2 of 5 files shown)


discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -213,6 +213,21 @@ def wait_for_one_host_to_be_in_wrong_boot_order(self, fall_on_error_status=True)
             fall_on_error_status=fall_on_error_status,
         )
 
+    def wait_for_hosts_to_be_in_wrong_boot_order(
```
lalon4, line 216: instead of adding this method, please user this:
cluster.wait_for_one_host_to_be_in_wrong_boot_order
You can rename it to wait_for_atleast_one_host_to_be_in_wrong_boot_order, and pass node_count
it has fall_on_error_status, which is better in case of unexpected host error

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -213,6 +213,21 @@ def wait_for_one_host_to_be_in_wrong_boot_order(self, fall_on_error_status=True)
...
+            self,
+            nodes_count=env_variables['num_nodes'],
+            timeout=consts.CLUSTER_INSTALLATION_TIMEOUT,
+            fall_on_error_status=True
+    ):
+        utils.wait_till_all_hosts_are_in_status(
```
lalon4, line 222: rename to wait_till_hosts_are_in_status

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -213,6 +213,21 @@ def wait_for_one_host_to_be_in_wrong_boot_order(self, fall_on_error_status=True)
...
+            client=self.api_client,
+            cluster_id=self.id,
+            statuses=[consts.ClusterStatus.INSTALLING_PENDING_USER_ACTION],
+            nodes_count=nodes_count,
```
lalon4, line 226: nodes_count should have a default of num_nodes

discovery-infra/tests/test_wrong_boot_order.py
```diff
@@ -20,12 +25,165 @@ def test_wrong_boot_order_one_node(self, nodes, cluster):
...
         node.shutdown()
         node.set_boot_order(cd_first=False)
         node.start()
 
         # wait until all nodes are in Installed status, will fail in case one host in error
+        new_cluster.wait_for_cluster_to_be_in_installing_status()
+        new_cluster.wait_for_hosts_to_install()
+        new_cluster.wait_for_install()
+
+    @pytest.mark.regression
```
lalon4, line 40: Im a bit concern about it, cause this tests wont run on Staging (not yet in)
please remove the regressions mark for now. instead we can have another mark representing new_feature \ master?
@nshidlin @RazRegev @tsorya WDYT?

discovery-infra/tests/test_wrong_boot_order.py
```diff
@@ -20,12 +25,165 @@ def test_wrong_boot_order_one_node(self, nodes, cluster):
...
+            n.set_boot_order(cd_first=True)
+
+        # Start cluster install
+        new_cluster.prepare_for_install(nodes)
+        new_cluster.start_install()
+        new_cluster.wait_for_installing_in_progress()
+
+        # Wait until wrong boot order - all hosts except bootstrap
+        new_cluster.wait_for_hosts_to_be_in_wrong_boot_order(len(nodes)-1)
+        new_cluster.wait_for_cluster_to_be_in_installing_pending_user_action_status()
```
lalon4, line 58: What is the behavior when cluster moves to installing pending user action? 
in such state installation continues? 
cuase here you wait for all hosts to have wrong boot order, so when does the cluster status changes? after first wrong boot node?

discovery-infra/tests/test_wrong_boot_order.py
```diff
@@ -20,12 +25,165 @@ def test_wrong_boot_order_one_node(self, nodes, cluster):
...
+
+        # Wait for an hour+, in normal cases we expect to get timeout error
+        # after an hour
+        logger.info('Waiting 65 minutes before fixing wrong boot order')
+        time.sleep(65 * 60)
```
lalon4, line 163: as we spoke instead of regression mark, add wrong_boot_order mark, and assign this mark to all relevant tests (also one in cancel\reset) . that way we will be able to run wrong_boot_order desperately, and this will include the wait test

lalon4: or maybe it is better to have it as a regression?
@nshidlin WDYT?

nshidlin: the wrong_boot_order mark can be placed on the class, and it's applied to all the tests in the class

discovery-infra/tests/test_wrong_boot_order.py
nshidlin: I think just leaving the regression mark off is good. Then test runs against staging will just run regression mark

lalon4: OK sounds good

lalon4: not completely actually, in this case the wrong boot order wont run at all (cause also the existing one changes)
we must have branches in some point

lalon4:
can you please add here a step to check cluster status? 
https://github.com/openshift/assisted-test-infra/blob/master/discovery-infra/tests/test_cancel_and_reset.py#L273

discovery-infra/tests/test_wrong_boot_order.py
lalon4: @RazRegev please leave the existing wrong boot order tests as is (as master) with regression mark (for staging running).
for all new tests (4 in total), do not add any mark.

discovery-infra/test_infra/helper_classes/cluster.py
RazRegev: this function has also `fall_on_error_status`, am I missing something?
or you mainly want me to unify between these functions?

discovery-infra/test_infra/helper_classes/cluster.py
RazRegev: this function was here before, not that I have a problem to rename it but this isn't a new function and will include refactoring out of the task scope.

discovery-infra/test_infra/helper_classes/cluster.py
RazRegev: it already has..

discovery-infra/tests/test_wrong_boot_order.py
RazRegev: resolved since we discussed it in private.

discovery-infra/test_infra/helper_classes/cluster.py
lalon4: OK, sorry missed that

discovery-infra/test_infra/helper_classes/cluster.py
lalon4: missed that too. so just unify

discovery-infra/test_infra/helper_classes/cluster.py
lalon4: OK so leave it as is.

PR #269, opened 2020-11-12
cleaning up methods moved from base test to cluster

nshidlin:
/cc @tsorya @lalon4


discovery-infra/tests/base_test.py
```diff
@@ -46,24 +46,12 @@ def get_cluster_func(cluster_name: Optional[str] = None):
 
         for cluster in clusters:
             logging.info(f'--- TEARDOWN --- deleting created cluster {cluster.id}')
```
lalon4, line 48: can you please add a separate line log for TEARDOWN?
that way it is easy to see in logs when test ended. it separate the screen

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -299,7 +307,7 @@ def download_installation_logs(self, path):
         self.api_client.download_cluster_logs(self.id, path)
 
     def get_install_config(self):
-        return self.api_client.get_cluster_install_config(self.id)
+        return yaml.load(self.api_client.get_cluster_install_config(self.id), Loader=yaml.SafeLoader)
```
tsorya, line 310: not sure i understand why we need it?

discovery-infra/tests/base_test.py
tsorya: it is relevant not only to cluster

discovery-infra/test_infra/helper_classes/cluster.py
nshidlin: I see if proxy values are set in the install config

discovery-infra/tests/base_test.py
nshidlin: yeah, i can add TEARDOWN log to the other fixtures
wdty?

discovery-infra/test_infra/helper_classes/cluster.py
tsorya: but you get json? why u need to change it to yaml?

tsorya: got it

PR #273, opened 2020-11-15
Add e2e test: max cluster and hosts names length

lalon4:


osherdp (approved):

discovery-infra/tests/test_validations.py
```diff
@@ -80,3 +81,30 @@ def test_host_insufficient(self, cluster, modified_nodes):
         new_cluster.wait_until_hosts_are_discovered(allow_insufficient=True)
         new_cluster.wait_for_host_validation(vcpu_host_id, "hardware", "has-min-cpu-cores", ["success"])
         new_cluster.wait_for_host_validation(ram_host_id, "hardware", "has-min-memory", ["success"])
+
+    @pytest.mark.regression
+    @pytest.mark.skip(reason="BZ:1897916")
```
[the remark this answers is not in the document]
lalon4: That there is a Bugzila bug on it (https://bugzilla.redhat.com/show_bug.cgi?id=1897916)

discovery-infra/tests/test_validations.py
```diff
@@ -80,3 +81,30 @@ def test_host_insufficient(self, cluster, modified_nodes):
...
+        # Trying to update cluster name to have more then 54 chars
+        with pytest.raises(ApiException):
+            c.set_cluster_name(cluster_name)
+        # Set cluster name with exactly 54 chars
+        cluster_name = "this-is-exactly-54-chars-cluster-name-for-testing-abc"
+        c.set_cluster_name(cluster_name)
+        c.prepare_for_install(nodes=nodes)
+        random_master_host = c.get_hosts_by_role(role=consts.NodeRoles.MASTER)[0]
+        # Trying to set host name with more then 63 chars
+        long_hostname = "this-is-a-more-then-63chars-long-host-name-this-is-more-then-63chars-long-host"
```
gamli75, line 104: spelling mistake - than instead of then

discovery-infra/tests/test_validations.py
```diff
@@ -80,3 +81,30 @@ def test_host_insufficient(self, cluster, modified_nodes):
...
+    def test_hosts_and_cluster_max_length(self, cluster, nodes):
+        # Trying to create a cluster with more then 54 chars
+        cluster_name = "this-is-a-very-long-cluster-name-this-is-a-very-long-cluster-name-this" \
+                       "-is-a-very-long-cluster-name"
+        # TODO: assert on error code and reason
+        with pytest.raises(ApiException):
+            cluster(cluster_name)
+        c = cluster()
+        # Trying to update cluster name to have more then 54 chars
+        with pytest.raises(ApiException):
```
gamli75, line 96: Why we do not validate the error code and the error message?

tsorya: Agree , will be nice to have an option to validate specific code

lalon4: right, this is why i added a comment in line 91, with TODO to add. Currently due to the bug 1897916 i do not get any error code, once ill have it, it will be added

discovery-infra/tests/test_validations.py
```diff
@@ -80,3 +81,30 @@ def test_host_insufficient(self, cluster, modified_nodes):
         new_cluster.wait_until_hosts_are_discovered(allow_insufficient=True)
         new_cluster.wait_for_host_validation(vcpu_host_id, "hardware", "has-min-cpu-cores", ["success"])
         new_cluster.wait_for_host_validation(ram_host_id, "hardware", "has-min-memory", ["success"])
+
+    @pytest.mark.regression
+    @pytest.mark.skip(reason="BZ:1897916")
+    def test_hosts_and_cluster_max_length(self, cluster, nodes):
+        # Trying to create a cluster with more then 54 chars
```
YuviGold, line 88: Sometimes you mention 64 chars and sometimes 54 chars in the comments.
I would prefer to generate a new `cluster_name` each time with the given length (there is no validation that the current `cluster_name`s that you provided are the correct length).

discovery-infra/tests/test_validations.py
```diff
@@ -80,3 +81,30 @@ def test_host_insufficient(self, cluster, modified_nodes):
...
+            cluster(cluster_name)
+        c = cluster()
+        # Trying to update cluster name to have more then 54 chars
+        with pytest.raises(ApiException):
+            c.set_cluster_name(cluster_name)
+        # Set cluster name with exactly 54 chars
+        cluster_name = "this-is-exactly-54-chars-cluster-name-for-testing-abc"
+        c.set_cluster_name(cluster_name)
+        c.prepare_for_install(nodes=nodes)
+        random_master_host = c.get_hosts_by_role(role=consts.NodeRoles.MASTER)[0]
```
YuviGold, line 102: for random let's use `random.choice(c.get_hosts_by_role(role=consts.NodeRoles.MASTER)`

discovery-infra/tests/test_validations.py
lalon4: fixed

discovery-infra/tests/test_validations.py
lalon4: 63 is the max of host name
54 is the max for cluster name

gamli75: so maybe it's worth to write a comment at the beginning of the test what is the max length for host & cluster name

lalon4: i actually would like my string to have - (dash) in between, and i cannot see other use than this test for a method to generate a string in certain length. WDYT about adding a method that will check\assert given string length?

discovery-infra/tests/test_validations.py
tsorya, line 102: lets add it to hosts. there is already get random host. lets have get random host by role

discovery-infra/tests/test_validations.py
lalon4: added assert_string_length

PR #281, opened 2020-11-16
Add wrong boot order e2e flow

lalon4:


discovery-infra/tests/test_wrong_boot_order.py
```diff
@@ -188,3 +189,36 @@ def test_installation_succeeded_on_incorrect_boot_order_timeout_is_ignored(self,
...
+        # Wait until one host is in wrong boot order
+        new_cluster.wait_for_one_host_to_be_in_wrong_boot_order()
+        new_cluster.wait_for_cluster_to_be_in_installing_pending_user_action_status()
+        # Kill bootstrap installer to simulate cluster error
+        bootstrap = nodes.get_bootstrap_node(cluster=new_cluster)
+        bootstrap.kill_podman_container_by_name("assisted-installer")
+        # Wait for node and cluster Error
+        new_cluster.wait_for_host_status([consts.NodesStatus.ERROR])
+        new_cluster.wait_for_cluster_in_error_status()
+        assert new_cluster.is_in_error_status()
```
nshidlin, line 213: isn't this assert redundant? The test will fail new_cluster.wait_for_cluster_in_error_status() once the time out is reached

lalon4: removed

YuviGold (approved):
LGTM

discovery-infra/tests/test_cancel_and_reset.py
```diff
@@ -233,6 +233,7 @@ def test_cancel_reset_after_installation_failure(self, nodes, cluster):
         # new_cluster.start_install_and_wait_for_installed()
 
     @pytest.mark.regression
+    @pytest.mark.wrong_boot_order
```
tsorya, line 236: what is the idea of this one? it is added for all wrong_boot_order tests?

tsorya: i think i got it, u mean it is testing 2 things , cancel reset and wrong boot order?

discovery-infra/tests/test_wrong_boot_order.py
```diff
@@ -188,3 +189,35 @@ def test_installation_succeeded_on_incorrect_boot_order_timeout_is_ignored(self,
...
+        # Start cluster install
+        new_cluster.prepare_for_install(nodes=nodes)
+        new_cluster.start_install()
+        new_cluster.wait_for_installing_in_progress()
+        # Wait until one host is in wrong boot order
+        new_cluster.wait_for_one_host_to_be_in_wrong_boot_order()
+        new_cluster.wait_for_cluster_to_be_in_installing_pending_user_action_status()
+        # Kill bootstrap installer to simulate cluster error
+        bootstrap = nodes.get_bootstrap_node(cluster=new_cluster)
+        bootstrap.kill_podman_container_by_name("assisted-installer")
```
tsorya, line 209: didn't u add function for kill installer?

discovery-infra/tests/test_wrong_boot_order.py
```diff
@@ -188,3 +189,35 @@ def test_installation_succeeded_on_incorrect_boot_order_timeout_is_ignored(self,
...
+        # Wait for node and cluster Error
+        new_cluster.wait_for_host_status([consts.NodesStatus.ERROR])
```
tsorya, line 211: lets have it as cluster function? wait for cluster error?

discovery-infra/tests/test_cancel_and_reset.py
lalon4: First reason is that there ia another wrong_boot_order test that is under cancel\reset. 
the second reason is that wrong boot order has 1 test, added by Raz, that validate that there is no timeout when cluster is in this state (timeout removed) so we wait for 1 h. I understood from Raz that in the future the timeout will be changed to 24 h, so i would like us to have the ability to run wrong_boot_order seperatly

discovery-infra/tests/test_wrong_boot_order.py
```diff
@@ -188,3 +189,35 @@ def test_installation_succeeded_on_incorrect_boot_order_timeout_is_ignored(self,
...
+        new_cluster.wait_for_cluster_in_error_status()
+        # Reset cluster install
+        new_cluster.reset_install()
+        assert new_cluster.is_in_insufficient_status()
+        # Fix boot order and reboot required nodes into ISO
+        nodes.set_correct_boot_order(master_nodes, False)
+        new_cluster.reboot_required_nodes_into_iso_after_reset(nodes=nodes)
+        # Wait for hosts to be rediscovered
+        new_cluster.wait_until_hosts_are_discovered()
```
tsorya, line 220: not enough to wait for new_cluster.wait_for_ready_to_install()?

discovery-infra/tests/test_wrong_boot_order.py
lalon4: Yeah, at first i added is_cluster_in_error, then Nir mentioned that it will be redundant. Cause we wait for error so why we need the assert afterwards, if error wont occur then wait will fail

discovery-infra/tests/test_wrong_boot_order.py
lalon4: we can wait for ready to install, but i prefer to have this verification as well, it wont cost more time in running, and when fails, we more or less know in which stage.

PR #283, opened 2020-11-16
MGMT-2837 - add day2 on ocp test flow

yevgeny-shnaidman:
/assign @danielerez

### DIFF
--- a/discovery-infra/day2.py
+++ b/discovery-infra/day2.py
@@ -15,7 +15,14 @@ def set_cluster_pull_secret(client, cluster_id, pull_secret):
     client.set_pull_secret(cluster_id, pull_secret)
 
 
-def execute_day2_flow(ocp_cluster_id, args):
+def execute_day2_cloud_flow(ocp_cluster_id, args):
+    execute_day2_flow(ocp_cluster_id, args, "cloud")
+
+
+def execute_day2_ocp_flow(ocp_cluster_id, args):
+    execute_day2_flow(ocp_cluster_id, args, "ocp")
+
+def execute_day2_flow(ocp_cluster_id, args, day2_type_flag):
     utils.recreate_folder(consts.IMAGE_FOLDER, force_recreate=False)
     client = assisted_service_api.create_client(
             url=utils.get_assisted_service_url_by_args(args=args)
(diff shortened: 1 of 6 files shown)


discovery-infra/day2.py
```diff
@@ -25,13 +32,18 @@ def execute_day2_flow(ocp_cluster_id, args):
...
-    cluster_id = str(uuid.uuid4())
-    cluster = client.create_day2_cluster(
-        ocp_cluster_name + "-day2", cluster_id, **_day2_cluster_create_params(ocp_openshift_version, ocp_api_vip_dnsname)
-    )
-
-    set_cluster_pull_secret(client, cluster_id, args.pull_secret)
+    if day2_type_flag == "ocp":
+        cluster = ocp_cluster
+        terraform_cluster_dir_prefix = "test-infra-cluster-assisted-installer"
+    else:
```
danielerez, line 38: should we check explicitly for "cloud" flag? or we can handle it if adding a new type.

discovery-infra/start_discovery.py
```diff
@@ -616,8 +618,13 @@ def main():
...
     parser.add_argument(
-        "--day2-cluster",
-        help="day2 cluster",
+        "--day2-cloud-cluster",
+        help="day2 cloud cluster",
+        action="store_true",
+    )
+    parser.add_argument(
+        "--day2-ocp-cluster",
+        help="day2 ocp luster",
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

discovery-infra/day2.py
yevgeny-shnaidman: currently we have only 2 flags, so in case there will be another one ( not likely), then we will handle it

discovery-infra/day2.py
```diff
@@ -25,13 +32,18 @@ def execute_day2_flow(ocp_cluster_id, args):
...
     ocp_api_vip_ip = ocp_cluster.api_vip
-    cluster_id = str(uuid.uuid4())
-    cluster = client.create_day2_cluster(
-        ocp_cluster_name + "-day2", cluster_id, **_day2_cluster_create_params(ocp_openshift_version, ocp_api_vip_dnsname)
-    )
-
-    set_cluster_pull_secret(client, cluster_id, args.pull_secret)
+    if day2_type_flag == "ocp":
+        cluster = ocp_cluster
+        terraform_cluster_dir_prefix = "test-infra-cluster-assisted-installer"
```
tsorya, line 37: i am really not sure regarding dir name.
@RazRegev how it will work with profile?

discovery-infra/day2.py
```diff
@@ -25,13 +32,18 @@ def execute_day2_flow(ocp_cluster_id, args):
...
+    else:
+        cluster_id = str(uuid.uuid4())
+        cluster = client.create_day2_cluster(
+            ocp_cluster_name + "-day2", cluster_id, **_day2_cluster_create_params(ocp_openshift_version, ocp_api_vip_dnsname)
+        )
+        set_cluster_pull_secret(client, cluster_id, args.pull_secret)
+        terraform_cluster_dir_prefix = ocp_cluster_name
```
tsorya, line 44: i think after adding new  ocp type name as ocp_cluster_name is a little bit confusing if it is not unders if ... == ocp

yevgeny-shnaidman: not sure i understand your comment

discovery-infra/day2.py
RazRegev: Profile is relevant only if DEPLOY_TARGET is minikube. need to make sure this is not the case.

RazRegev: I don't see where is the problem, where do we use profile for a dirname?

PR #287, opened 2020-11-18
MGMT-2465: fetch ocp cluster id on deploy_controller

danielerez:
- As OCP cluster id is required for the controller, fetched the it form clusters list on deploy_controller.sh.
- Removed CLUSTER_ID from config_etc_hosts_for_ocp_cluster target (not required as build/kubeconfig is used).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -214,12 +214,6 @@ kill_all_port_forwardings:
 #########
 
 deploy_on_ocp_cluster: bring_assisted_installer
-	# controller
-	DEPLOY_TARGET=ocp NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE)) \
-		DEPLOY_TAG=$(DEPLOY_TAG) DEPLOY_MANIFEST_TAG=$(DEPLOY_MANIFEST_TAG) OCP_KUBECONFIG=$(OCP_KUBECONFIG) \
-		PROFILE=$(PROFILE) CONTROLLER_OCP=$(CONTROLLER_OCP) \
-		scripts/deploy_controller.sh
-
 	# service
 	DEPLOY_TARGET=ocp NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE)) \
 		DEPLOY_TAG=$(DEPLOY_TAG) DEPLOY_MANIFEST_TAG=$(DEPLOY_MANIFEST_TAG) OCP_KUBECONFIG=$(OCP_KUBECONFIG) \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -213,13 +213,7 @@ kill_all_port_forwardings:
...
 #########
 
-deploy_on_ocp_cluster: bring_assisted_installer
-	# controller
-	DEPLOY_TARGET=ocp NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE)) \
-		DEPLOY_TAG=$(DEPLOY_TAG) DEPLOY_MANIFEST_TAG=$(DEPLOY_MANIFEST_TAG) OCP_KUBECONFIG=$(OCP_KUBECONFIG) \
-		PROFILE=$(PROFILE) CONTROLLER_OCP=$(CONTROLLER_OCP) \
-		scripts/deploy_controller.sh
-
+deploy_on_ocp_cluster:
```
yevgeny-shnaidman, line 216: why it is now a different target? Why can't it be part of deploy_on_ocp_cluster

danielerez: The controller needs the ID of the new cluster, so separated into two stages:
1. deploy_on_ocp_cluster
2. deploy_controller_on_ocp_cluster CLUSTER_ID=123
I.e. The cluster_id is available after the cluster is registered in step 1.

Makefile
```diff
@@ -231,8 +225,15 @@ deploy_on_ocp_cluster: bring_assisted_installer
 		DEPLOY_TAG=$(DEPLOY_TAG) DEPLOY_MANIFEST_TAG=$(DEPLOY_MANIFEST_TAG) OCP_KUBECONFIG=$(OCP_KUBECONFIG) PROFILE=$(PROFILE) \
 		scripts/deploy_ui.sh
 
+deploy_controller_on_ocp_cluster: bring_assisted_installer
```
eranco74, line 228: Consider renaming it to day_2_controller

danielerez: Moved back to 'deploy_on_ocp_cluster' target, so already removed this one :)

Makefile
danielerez: Fetched the cluster id in deploy_controller.sh, so no need to manually specify it now.

scripts/deploy_controller.sh
```diff
@@ -11,5 +11,8 @@ export SERVICE_BASE_URL=${SERVICE_BASE_URL:-"http://${SERVICE_URL}:${SERVICE_POR
 mkdir -p build
 
 if [ "${DEPLOY_TARGET}" == "ocp" ]; then
-    skipper run "scripts/ocp.sh deploy_controller $OCP_KUBECONFIG $SERVICE_BASE_URL $NAMESPACE $CONTROLLER_OCP"
+    CLUSTER_ID=$(curl -s $SERVICE_BASE_URL/api/assisted-install/v1/clusters | jq -r '.[0] | .id')
```
yevgeny-shnaidman, line 14: I thought we are gonna do it at the beginning of the main function in a loop. Otherwise, it is racy: assisted-service on OCP might not have created the cluster yet when you are quiring it for cluster id

danielerez: I'm not sure it's an actual risk, as the query is called after the service and ui are ready. So it should be plenty of time for the cluster to be created in db.

tsorya:
Please open a ticket to move all inside day2 logic and remove all the deployment to python code. It is not the same as deploying service cause it is going to happened in a middle of the run and there is no point for this code independently

PR #290, opened 2020-11-19
Add initial support of IPv6

empovit:
- Set up host to support IPv6 addresses
- Deploy guests with IPv6
- Enable IPv6 connections from guests to assisted service
- Work around libvirt Terraform module not setting hostnames


discovery-infra/test_infra/tools/terraform_utils.py
```diff
@@ -14,14 +15,19 @@ def __init__(self, working_dir):
         self.init_tf()
 
     def init_tf(self):
-        self.tf.cmd("init -plugin-dir=/root/.terraform.d/plugins/", raise_on_error=True)
+        try:
+            self.tf.cmd("init -plugin-dir=/root/.terraform.d/plugins/", raise_on_error=True)
+        except:
+            logging.exception("Init")
+            time.sleep(30000)
```
YuviGold, line 22: How would sleep help?

empovit: @ori-amizur ^^

Makefile
```diff
@@ -49,6 +49,8 @@ HTTP_PROXY_URL := $(or $(HTTP_PROXY_URL), "")
 HTTPS_PROXY_URL := $(or $(HTTPS_PROXY_URL), "")
 NO_PROXY_VALUES := $(or $(NO_PROXY_VALUES), "")
 VIP_DHCP_ALLOCATION := $(or $(VIP_DHCP_ALLOCATION),yes)
+IPv4 := $(or $(IPv4),yes)
```
YuviGold, line 52: Why there are two parameters? Could it be like an IP mode? Do we want to support no IPv4?

discovery-infra/test_infra/tools/terraform_utils.py
ori-amizur: @empovit Please remove.  It was for debug only

Makefile
ori-amizur: Modes are:
IPv6 only
IPv4 only
Dual

empovit: 1) IPv4, 2) IPv6, 3) IPv4+IPv6. There's a validation that it has to be at lease one.

discovery-infra/test_infra/tools/terraform_utils.py
empovit: Done

discovery-infra/start_discovery.py
```diff
@@ -353,6 +431,23 @@ def nodes_flow(client, cluster_name, cluster, image_path):
                 utils.config_etc_hosts(cluster_info)
                 utils.wait_for_cvo_available()
 
+def _get_libvirt_nodes_from_tf_state(network_name, tf_state):
+    nodes = _extract_nodes(tf_state, network_name, consts.NodeRoles.MASTER)
+    nodes.update(_extract_nodes(tf_state, network_name, consts.NodeRoles.WORKER))
```
ori-amizur, line 436: Since the function is called for both roles, why call it twice for each role? Why the function does not return all roles?

discovery-infra/start_discovery.py
```diff
@@ -353,6 +431,23 @@ def nodes_flow(client, cluster_name, cluster, image_path):
...
+    return nodes
+
+def _extract_nodes(tf_state, network_name, role):
```
ori-amizur, line 439: Maybe call it extract_nodes_from tfstate  ?  Maybe put in util?

discovery-infra/test_infra/tools/terraform_utils.py
```diff
@@ -14,7 +15,10 @@ def __init__(self, working_dir):
         self.init_tf()
 
     def init_tf(self):
-        self.tf.cmd("init -plugin-dir=/root/.terraform.d/plugins/", raise_on_error=True)
+        try:
```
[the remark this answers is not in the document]
empovit: Remove what? The code was there, you wrapped it in try-except. Do you want me to remove the try?

discovery-infra/start_discovery.py
empovit: There are two sections in tfstate - one is called "master", the other "worker". I need to do the same sequence of steps for each of the sections. Putting this in one function would mean duplicating the code, instead of applying the same code to one section at a time.

discovery-infra/start_discovery.py
empovit: It's a workaround, I didn't want to put it in utils because it will be "public".

discovery-infra/test_infra/tools/terraform_utils.py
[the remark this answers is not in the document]
ori-amizur: Yes.  It was just for debug.

discovery-infra/start_discovery.py
ori-amizur: OK

discovery-infra/test_infra/tools/terraform_utils.py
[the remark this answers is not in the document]
empovit: Done

discovery-infra/utils.py
```diff
@@ -541,6 +541,12 @@ def get_network_leases(network_name):
 def create_ip_address_list(node_count, starting_ip_addr):
     return [str(ipaddress.ip_address(starting_ip_addr) + i) for i in range(node_count)]
 
+def create_ip_address_nested_list(node_count, starting_ip_addr):
```
tsorya, line 544: lets move all changes to discovery-infra/test_infra/utils.py
I want to deprecate this one

scripts/install_environment.sh
```diff
@@ -112,6 +113,34 @@ function config_firewalld() {
     sudo systemctl start firewalld
 }
 
+function config_squid() {
+    echo "Config squid"
+    sudo dnf install -y squid
```
tsorya, line 118: add it to install_env... script

scripts/install_environment.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 set -euo pipefail
+source scripts/utils.sh
```
tsorya, line 3: i think i miss what function from utils you are using here?

terraform_files/main.tf
```diff
@@ -77,14 +78,16 @@ resource "libvirt_domain" "master" {
   }
 
   network_interface {
-    network_name = var.libvirt_network_name
+    network_name = libvirt_network.net.name
     hostname   = "${var.cluster_name}-master-${count.index}.${var.cluster_domain}"
-    addresses  = [var.libvirt_master_ips[count.index]]
+    addresses  = var.libvirt_master_ips[count.index]
+    wait_for_lease = true
```
tsorya, line 84: and if we don't start domain? only creating it? How it will work with wait_for_lease?

discovery-infra/utils.py
empovit: Got it. I was wondering why there were two :)

terraform_files/variables-libvirt.tf
```diff
@@ -20,12 +20,32 @@ variable "cluster_domain" {
...
-  description = "Provisioning domain"
+  description = "Provisioning IPv4 domain"
+}
+
+variable provisioning_cidr_addresses {
+    type = list(string)
+    description = "Addresses for provisioning CIDR network"
+}
+
+variable "machine_cidr6" {
```
tsorya, line 41: I don't see any usage of this one?

terraform_files/variables-libvirt.tf
```diff
@@ -20,12 +20,32 @@ variable "cluster_domain" {
 
 variable "machine_cidr" {
```
tsorya, line 21: don't see usage if it anymore

scripts/install_environment.sh
empovit: Good catch, probably it's not needed. @ori-amizur it it OK to remove?

terraform_files/variables-libvirt.tf
```diff
@@ -59,22 +79,22 @@ variable "libvirt_network_mtu" {
 }
 
 variable "libvirt_master_ips" {
-  type        = list(string)
+  type        = list(list(string))
```
tsorya, line 82: those changes will break e2e tests, they use the same files.

terraform_files/main.tf
empovit: Do we do that? @ori-amizur can you answer @tsorya's question?

terraform_files/variables-libvirt.tf
empovit: https://github.com/empovit/assisted-test-infra/blob/b54bd3829f6c4e83e6f1cacee481b81b0fe338d7/discovery-infra%2Fstart_discovery.py#L303

terraform_files/variables-libvirt.tf
empovit: Some of them are populated by scripts and then applied. Terraform had been complaining until we defined them in variables-libvirt.tf.

terraform_files/variables-libvirt.tf
empovit: E2e where? Where's the code? I made sure to change all code that references this variable.

discovery-infra/utils.py
empovit: BTW they are already there - an exact copy. So when you decide to get rid of this one just change the import (it currently seems to be used in some places, not related to this PR).

terraform_files/variables-libvirt.tf
empovit: https://github.com/openshift/assisted-test-infra/blob/b54bd3829f6c4e83e6f1cacee481b81b0fe338d7/discovery-infra%2Ftest_infra%2Fcontrollers%2Fnode_controllers%2Fterraform_controller.py#L41

scripts/install_environment.sh
ori-amizur: Yes

terraform_files/main.tf
ori-amizur: wait_for_lease is not mandatory.  OK to remove

terraform_files/variables-libvirt.tf
tsorya: but i don't see any usage of machine_cidr in terraform, so why to leave

discovery-infra/assisted_service_api.py
```diff
@@ -134,11 +134,11 @@ def generate_and_download_image(self, cluster_id, ssh_key, image_path):
         self.generate_image(cluster_id=cluster_id, ssh_key=ssh_key)
         self.download_image(cluster_id=cluster_id, image_path=image_path)
 
-    def set_hosts_roles(self, cluster_id, hosts_with_roles):
+    def set_hosts_roles(self, cluster_id, hosts_with_roles, hosts_names=None):
```
tsorya, line 137: why we have host_names here?

discovery-infra/test_infra/utils.py
```diff
@@ -555,6 +555,12 @@ def create_ip_address_list(node_count, starting_ip_addr):
     return [str(ipaddress.ip_address(starting_ip_addr) + i) for i in range(node_count)]
 
 
+def create_ip_address_nested_list(node_count, starting_ip_addr):
+    return [[str(ipaddress.ip_address(starting_ip_addr) + i)] for i in range(node_count)]
+
+def create_empty_nested_list(node_count):
```
tsorya, line 561: don't see any usage of it

discovery-infra/test_infra/utils.py
```diff
@@ -555,6 +555,12 @@ def create_ip_address_list(node_count, starting_ip_addr):
     return [str(ipaddress.ip_address(starting_ip_addr) + i) for i in range(node_count)]
 
 
+def create_ip_address_nested_list(node_count, starting_ip_addr):
```
tsorya, line 558: can i have some explanation why we need nested list? cause it seems strange

discovery-infra/utils.py
```diff
@@ -126,7 +126,7 @@ def is_cvo_available():
 
 
 def get_libvirt_nodes_macs(network_name):
-    return get_libvirt_nodes_mac_role_ip_and_name(network_name).keys()
```
tsorya, line 129: lets just move usage to test_infra/utils? for those functions? and delete them from here? Will help a lot

terraform_files/variables-libvirt.tf
tsorya: but you deleted it from main.tf so u need to change all the places that are using it and change them and to delete the old variables. Right?

terraform_files/variables-libvirt.tf
tsorya: i mean in terraform main.tf? if we don't use it there why we are providing data to it? Not sure for what we have it?

terraform_files/variables-libvirt.tf
tsorya: try running e2e tests. Check that they pass

discovery-infra/assisted_service_api.py
empovit: The API call in this function can update host names along with roles. For now we need it as a workaround. In all other scenarios it will work as before, updating just the roles.

discovery-infra/test_infra/utils.py
empovit: https://github.com/empovit/assisted-test-infra/blob/93c536cfe495442ce6aff0779f8f5767e9a0a143/discovery-infra%2Fstart_discovery.py#L84
https://github.com/empovit/assisted-test-infra/blob/93c536cfe495442ce6aff0779f8f5767e9a0a143/discovery-infra%2Fstart_discovery.py#L155

discovery-infra/test_infra/utils.py
empovit: Each NIC may have two IP addresses, one for IPv4 and another for IPv6. The IPv4 addresses are static, IPv6 are allocated  by DHCP, but in order for Terraform to insert them the IP addresses of a NIC must be a list (as opposed to a scalar value). @ori-amizur please correct me if I'm wrong.

discovery-infra/utils.py
empovit: It's already there, as I mentioned. Just change the imports and it will continue to work. There are other functions in use from the wrong utils, so I couldn't change the imports in this PR without making it even larger.

https://github.com/openshift/assisted-test-infra/pull/290/files#diff-61d18d81e7d4ae0105d0475fc2abd8dc0a66b687df5e35c56c8a337fbd7a7e4eL108

terraform_files/variables-libvirt.tf
empovit: Done

discovery-infra/assisted_service_api.py
tsorya: in this case u need to give it another name. Setting role != changing hostname

discovery-infra/utils.py
tsorya, line 544: lets remove all code from this file and move test_infra.utils

ori-amizur (approved):
/approve
/lgtm

PR #292, opened 2020-11-22
MGMT-2219 Add unique vip hostname test

YuviGold:
Testing https://github.com/openshift/assisted-installer-agent/pull/92

### DIFF
--- a/discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
+++ b/discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
@@ -36,6 +36,12 @@ class LibvirtController(NodeController):
         logging.info("Found domains %s", nodes)
         return nodes
 
+    def list_networks(self):
+        return self.libvirt_connection.listAllNetworks()
+
+    def list_leases(self, network_name):
+        return self.libvirt_connection.networkLookupByName(network_name).DHCPLeases()
+
     def shutdown_node(self, node_name):
         logging.info("Going to shutdown %s", node_name)
         node = self.libvirt_connection.lookupByName(node_name)
--- a/discovery-infra/tests/conftest.py
+++ b/discovery-infra/tests/conftest.py
@@ -8,8 +8,13 @@ from test_infra import assisted_service_api, consts, utils
 
 qe_env = False
 
+
+def is_qe_env():
+    return os.environ.get('NODE_ENV') == 'QE_VM'
+
+
 # TODO changes it
-if os.environ.get('NODE_ENV') == 'QE_VM':
+if is_qe_env():
     from test_infra.controllers.node_controllers.qe_vm_controler import \
         QeVmController as nodeController
     qe_env = True
--- a/discovery-infra/tests/test_dhcp.py
+++ b/discovery-infra/tests/test_dhcp.py
@@ -1,0 +1,33 @@
+import pytest
+
+from tests.conftest import env_variables, is_qe_env
+from tests.base_test import BaseTest
+
+
+class TestDHCP(BaseTest):
+    def test_unique_vip_hostname(self, api_client, cluster, nodes):
+        vips = []
+
+        # First time
+        first_cluster = cluster()
+        first_cluster.prepare_for_install(nodes=nodes, vip_dhcp_allocation=True)
+        vips.append(self._get_api_vip(nodes.controller))
+
+        first_cluster.delete()
+        nodes.controller.destroy_all_nodes()
+
+        # Second time
+        nodes.controller.prepare_nodes()
+        second_cluster = cluster()
+        second_cluster.prepare_for_install(nodes=nodes, vip_dhcp_allocation=True)
+        vips.append(self._get_api_vip(nodes.controller))
+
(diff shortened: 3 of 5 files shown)


discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
```diff
@@ -36,6 +36,12 @@ def list_nodes_with_name_filter(self, name_filter):
         logging.info("Found domains %s", nodes)
         return nodes
 
+    def list_networks(self):
```
tsorya, line 39: don't it need to be part of interface?

tsorya: maybe even network controller?

discovery-infra/tests/test_dhcp.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+@pytest.mark.skipif(not env_variables['vip_dhcp_allocation'], reason="DHCP allocation must be enabled")
+@pytest.mark.skipif(is_qe_env(), reason="Implemented only for local env")
+class TestDHCP(BaseTest):
+    def test_unique_vip_hostname(self, api_client, cluster, nodes):
+        vips = []
+
+        # First time
+        c = cluster()
```
tsorya, line 14: c? it is not go. Please have more self explaining var name

discovery-infra/tests/test_dhcp.py
```diff
@@ -0,0 +1,35 @@
+import pytest
+
+from tests.conftest import env_variables, is_qe_env
+from tests.base_test import BaseTest
+
+
+@pytest.mark.skipif(not env_variables['vip_dhcp_allocation'], reason="DHCP allocation must be enabled")
```
lalon4, line 7: This means DHCP and none DHCP cannot run at the same run,
Need to find a way to set it up in test level and not using the vip_dhcp_allocation

discovery-infra/tests/test_dhcp.py
```diff
@@ -0,0 +1,35 @@
...
+@pytest.mark.skipif(is_qe_env(), reason="Implemented only for local env")
```
lalon4, line 8: Tried to run on QE env. Run fails on cluster preparation

discovery-infra/tests/test_dhcp.py
nshidlin: How about adding the ability, to override the cluster creation params in the fitxture:
  https://github.com/openshift/assisted-test-infra/blob/580aea5aab7c6b822d5719b2862e5f175e53062b/discovery-infra/tests/base_test.py#L37

YuviGold: This test isn't relevant on non-DHCP mode

discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
YuviGold: But this is calling libvirt functionatlity

discovery-infra/tests/test_dhcp.py
lalon4: yes, but the mode defined in the cluster level

discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
tsorya: still think it must be part of interface

discovery-infra/tests/test_dhcp.py
lalon4: currently the mode is depended on the env vip_dhcp_allocation, so each test run has true or false, meaning we wont be able to run both DHCP and none DHCP tests, and we do want that

discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
lalon4: why do we need another controller? its libvirt

discovery-infra/tests/test_dhcp.py
YuviGold: It would require a bigger change to the infrastrucutre. For now, I just made sure the test wouldn't run with inappropriate configuration.

lalon4: The test wont run unless youll explicitly run it, it wont run as part of the regression, so what is the point?
It is not a big change, and it is required

nshidlin: We can remove VIP_DHCP_ALLOCATION env variable all together and default vip_dhcp_allocation=false durring cluster creation and allow to override this

discovery-infra/tests/conftest.py
```diff
@@ -8,8 +8,13 @@
 
 qe_env = False
 
+
+def is_qe_env():
```
tsorya, line 12: why it is function? why not to set variable in env?

YuviGold: Globals in Python are global to a module, not across all modules.
So instead of depending on a variable, and instead of creating a Singleton instance, I chose to have a function :)

PR #299, opened 2020-11-25
Config etc hosts on download logs for must gather

YuviGold:


discovery-infra/download_logs.py
```diff
@@ -14,7 +14,7 @@
 from assisted_service_api import InventoryClient, create_client
 from consts import ClusterStatus
 from logger import log
-from utils import recreate_folder, run_command
+from utils import recreate_folder, run_command, config_etc_hosts
```
tsorya, line 17: lets move to utils in test_infra

PR #300, opened 2020-11-25
Set default expiration time for images in s3 to 10m when running e2e tests

RazRegev:
This in order to avoid storage overflow which causes the pods to crash and tests to fail.


RazRegev:
closed since this is the relevant pr https://github.com/openshift/assisted-test-infra/pull/304

PR #308, opened 2020-11-30
removing redundant assisted_service_api

nshidlin:


YuviGold:
LGTM

Something else should be verified before we remove the old assisted-service-api? @tsorya  @yevgeny-shnaidman @RazRegev

tsorya:
bless you!!!

tsorya:
same with utils please :)

PR #309, opened 2020-12-01
adding download logs to cluster teardown

nshidlin:
This downloads cluster logs and the end of each test

At the beginning of the test session the previous logs are cleared


discovery-infra/test_infra/consts.py
```diff
@@ -21,6 +21,7 @@
 WAIT_FOR_BM_API = 900
 NAMESPACE_POOL_SIZE = 15
 PODMAN_FLAGS = "--cgroup-manager=cgroupfs --storage-driver=vfs --events-backend=file"
+LOG_FOLDER = "/tmp/assisted_test_infra_logs"
```
YuviGold, line 24: Don't we want to pass this parameter from pytest? then it would be easy to get archived

discovery-infra/tests/base_test.py
```diff
@@ -77,3 +82,10 @@ def assert_cluster_validation(cluster_info, validation_section, validation_id, e
     def assert_string_length(string, expected_len):
         assert len(string) == expected_len, "Expected len string of: " + str(expected_len) + \
                                             " rather than: " + str(len(string)) + " String value: " + string
+
+    @staticmethod
+    def collect_test_logs(cluster, api_client, test_name):
+        log_dir_name = f'{consts.LOG_FOLDER}/{test_name}'
+        with suppress(ApiException):
```
YuviGold, line 89: I think there is no need to suppress ApiException while `download_logs` should do it

discovery-infra/tests/base_test.py
```diff
@@ -77,3 +82,10 @@ def assert_cluster_validation(cluster_info, validation_section, validation_id, e
...
+            cluster_detials = json.loads(json.dumps(cluster.get_details().to_dict(), sort_keys=True, default=str))
+            download_logs(api_client, cluster_detials, log_dir_name, True)
```
YuviGold, line 91: Be aware that downloading `must-gather` would explode your storage. I think you should consider if you want to download it

discovery-infra/tests/base_test.py
```diff
@@ -77,3 +82,10 @@ def assert_cluster_validation(cluster_info, validation_section, validation_id, e
     def assert_string_length(string, expected_len):
         assert len(string) == expected_len, "Expected len string of: " + str(expected_len) + \
                                             " rather than: " + str(len(string)) + " String value: " + string
+
+    @staticmethod
+    def collect_test_logs(cluster, api_client, test_name):
+        log_dir_name = f'{consts.LOG_FOLDER}/{test_name}'
+        with suppress(ApiException):
+            cluster_detials = json.loads(json.dumps(cluster.get_details().to_dict(), sort_keys=True, default=str))
```
YuviGold, line 90: `cluster_details`

discovery-infra/test_infra/consts.py
nshidlin: What do you mean? As a command line argument?

discovery-infra/tests/base_test.py
nshidlin: changed

discovery-infra/test_infra/consts.py
lalon4: question - its a skipper folder? or local?

discovery-infra/tests/base_test.py
nshidlin: In some tests the cluster is deleted during the test e.g https://github.com/openshift/assisted-test-infra/blob/master/discovery-infra/tests/test_general.py#L17
suppress ApiException addresses this

discovery-infra/tests/base_test.py
nshidlin: These logs are helpful for debugging. I can  download must-gather in the case failure. wdty?

discovery-infra/test_infra/consts.py
nshidlin: /tmp is mounted to the container
https://github.com/openshift/assisted-test-infra/blob/master/skipper.yaml#L11

discovery-infra/tests/base_test.py
YuviGold: I think it should be done.
You can keep it for now, and look if it's too much.

discovery-infra/test_infra/consts.py
YuviGold: > What do you mean? As a command line argument?

Yes,
So if you would want to run on another environment or something, you will be able to change the path of the log.

discovery-infra/tests/base_test.py
nshidlin: changed to only download must-gather in the case of test failure

discovery-infra/test_infra/consts.py
nshidlin: changed

YuviGold (approved):
LGTM

PR #310, opened 2020-12-02
Remove oc install from install_minikube.sh script

YuviGold:


tsorya:
don't push it please, i think it is very convenient to have oc on host that we use

PR #312, opened 2020-12-03
Adding support to run pytests in parallel in terraform env

lalon4:


Makefile
```diff
@@ -351,4 +351,16 @@ test:
 	skipper make $(SKIPPER_PARAMS) _test
 
 _test: $(REPORTS)
+	if [ -d /tmp/assisted_test_infra_logs ]; then rm -rf /tmp/assisted_test_infra_logs; fi
```
nshidlin, line 354: No need for this. Done here:
https://github.com/openshift/assisted-test-infra/blob/master/discovery-infra/tests/conftest.py#L81

discovery-infra/tests/conftest.py
```diff
@@ -69,19 +83,23 @@ def get_api_client(offline_token=env_variables['offline_token'], **kwargs):
...
+        controller = nodeController(**env_variables)
+        controller.prepare_nodes()
+        yield controller
+    yield nodeController
     logging.info(f'--- TEARDOWN --- node controller\n')
-    controller.destroy_all_nodes()
+    if is_qe_env():
+        controller.destroy_all_nodes()
 
-@pytest.fixture(scope="session", autouse=True)
```
nshidlin, line 81: Why did you remove this?

nshidlin: I see running with pytest-xdist creates a new session for each test:
https://github.com/pytest-dev/pytest-xdist/issues/271

lalon4: Not working as multi run uses different process for each test run. Than it run at the first of each test which caused a raise issues..need to make this setup in the beginning of each RUN. Therefore added to Makefile

discovery-infra/tests/conftest.py
```diff
@@ -69,19 +83,23 @@ def get_api_client(offline_token=env_variables['offline_token'], **kwargs):
     return assisted_service_api.create_client(url, offline_token, **kwargs)
 
 
+@pytest.fixture(scope="session")
+def is_dev_env():
```
YuviGold, line 87: why is it a fixture?

Makefile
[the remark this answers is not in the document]
lalon4: No, see below, needed to remove the session scope.
How can you tell whats run inside the container and what not?

Makefile
```diff
@@ -351,4 +351,16 @@ test:
...
 	python3 -m pytest $(or ${TEST},discovery-infra/tests) -k $(or ${TEST_FUNC},'') -m $(or ${TEST_MARKER},'') --verbose -s --junit-xml=$(REPORTS)/unittest.xml
+
+test_parallel:
+	if [ -d /tmp/assisted_test_infra_logs ]; then rm -rf /tmp/assisted_test_infra_logs; fi
+	mkdir /tmp/assisted_test_infra_logs
+	cp -p discovery-infra/test_infra/tools/tf_network_pool.json /tmp/tf_network_pool.json
+	skipper make $(SKIPPER_PARAMS) _test_parallel
+
+_test_parallel: $(REPORTS)
+	python3 -m pytest -n $(or ${MULTI_NUM}, '2') $(or ${TEST},discovery-infra/tests) -k $(or ${TEST_FUNC},'') -m $(or ${TEST_MARKER},'') --verbose -s --junit-xml=$(REPORTS)/unittest.xml
```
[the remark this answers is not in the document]
lalon4: The reason is to be able to run also a single test during development for example.
With -n option, no live logs until test are finished. 
Also, -n parameter cannot be ' ' so its either set, or not exist at all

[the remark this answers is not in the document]
lalon4: so unable to merge, and ill change the name to TEST_WORKERS_NUM

discovery-infra/test_infra/tools/assets.py
```diff
@@ -0,0 +1,39 @@
...
+            with open(self.assets_file) as _file:
+                all_assets = json.load(_file)
+            asset = Munch.fromDict(all_assets.pop(0))
+            with open(self.assets_file, "w") as _file:
+                json.dump(all_assets, _file)
+            self._took_assets.append(asset)
+        logging.info("Taken network asset: %s", asset)
+        return asset
+
+    def return_assets(self, assets):
```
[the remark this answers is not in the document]
lalon4: it return the asset into the pool

discovery-infra/test_infra/tools/assets.py
```diff
@@ -0,0 +1,39 @@
+import json
+import logging
+from munch import Munch
+from test_infra import utils
+from test_infra import consts
+
+
+class Assets(object):
```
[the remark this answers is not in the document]
lalon4: assest represent a networking asset

discovery-infra/tests/conftest.py
```diff
@@ -47,8 +53,15 @@ def is_qe_env():
                  "kubeconfig_path": utils.get_env('KUBECONFIG', ''),
                  "log_folder": utils.get_env('LOG_FOLDER', consts.LOG_FOLDER)}
 
-image = utils.get_env('ISO',
```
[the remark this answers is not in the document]
lalon4: cause we need unique ISO name

discovery-infra/tests/conftest.py
lalon4: using it from base_test.py

Makefile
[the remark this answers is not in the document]
lalon4: Ho i understand now. But why does it matter? tmp is mounted

discovery-infra/test_infra/tools/assets.py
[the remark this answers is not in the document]
nshidlin: so maybe rename to something like free_assests ?

Makefile
nshidlin, line 366: tf_network_pool.json only has 2 networks. what happens if MUTLI_NUM > 2? The number of networks in itf_network_pool.json should match the number of parallel tests. Maybe remove this option until this is supported?

skipper.yaml
```diff
@@ -77,6 +77,7 @@ env:
   TEST: $TEST
   TEST_FUNC: $TEST_FUNC
   TEST_MARKER: $TEST_MARKER
+  MULTI_NUM: $MULTI_NUM
```
nshidlin, line 80: Where is this env variable used?

discovery-infra/test_infra/tools/assets.py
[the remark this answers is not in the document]
YuviGold: It still isn't clear to me. But let's at least add filedoc or docstring to the class

Makefile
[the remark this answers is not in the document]
YuviGold: but isn't `-n` can't be equal to 1?
What happens when trying to run a single test in parallel? Does it crush?

discovery-infra/test_infra/tools/assets.py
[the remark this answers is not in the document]
lalon4: I agree that the name Asset is unclear. But Lock is also not represent entirely this class. 
How about NetworkAsset? 
Ill also add a docstring, no problem.

Makefile
tsorya: @lalon4 lets add logic that if we don't have free asset we just wait for it

tsorya: @YuviGold regarding merging , -n 1 != without -n. 
If u will want to debug your tests better to run without -n cause adding -n option sort of disables the interactive option.
The problem is that this plugin runs in subprocess. Better to have 2 those options or we can have one option that will validate if parallel == 1 not to add -n

discovery-infra/test_infra/tools/assets.py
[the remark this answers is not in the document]
tsorya: first of all it is not Lock for sure. I like "asset" name. Not sure what is not clear in it

discovery-infra/test_infra/tools/assets.py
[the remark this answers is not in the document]
tsorya: release(self, asset)

discovery-infra/test_infra/tools/assets.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def __init__(self, assets_file=consts.TF_NETWORK_POOL_PATH, lock_file="/tmp/network_pool.lock"):
+        self.assets_file = assets_file
+        self.lock_file = lock_file
+        self._took_assets = []
+
+    def get_asset(self):
```
tsorya, line 15: just get()

discovery-infra/tests/base_test.py
```diff
@@ -25,13 +22,27 @@ def random_name():
...
-        nodes = Nodes(controller, env_variables["private_ssh_key_path"])
-        nodes.set_correct_boot_order(start_nodes=False)
-        yield nodes
-        nodes.shutdown_all()
-        nodes.format_all_disks()
+    def nodes(self, setup_node_controller, is_dev_env):
+        if is_dev_env:
+            asset = None
+            try:
+                asset = Assets()
```
tsorya, line 29: i think this way better cause it is much less code and not try finally in IF scope

```
@pytest.fixture(scope="function")
def nodes(self, setup_node_controller):
    asset = None
    try:
        if is_dev_env:
            asset = assets.Assets()
            env_variables["asset"] = asset.get_asset()
        controller = setup_node_controller(**env_variables)
        controller.prepare_nodes()
        nodes = Nodes(controller, env_variables["private_ssh_key_path"])
        yield nodes
        logging.info(f'--- TEARDOWN --- node controller\n')
        controller.destroy_all_nodes()
    finally:
        if asset:
            asset.return_all()
```

discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
```diff
@@ -31,6 +34,9 @@ def _create_tf_folder(self):
         copy_tree(consts.TF_TEMPLATE, tf_folder)
         return tf_folder
 
+    def _get_random_name(self):
+        return ''.join(random.choice(ascii_lowercase) for i in range(8))
```
tsorya, line 38:
```suggestion
        uuid.uuid4().hex[:6].lower()
```

Makefile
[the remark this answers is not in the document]
lalon4: no it wont crush it will work, just that you wont have live logs until test is PASSED\FAILED. 
So need to be able to run tests without -n

discovery-infra/tests/conftest.py
tsorya: i think we have something relevant to this in conftest, can u check?

discovery-infra/tests/conftest.py
```diff
@@ -69,19 +83,23 @@ def get_api_client(offline_token=env_variables['offline_token'], **kwargs):
...
+    return not is_qe_env()
+
+
 @pytest.fixture(scope="session")
 def setup_node_controller():
     logging.info(f'--- SETUP --- node controller\n')
-    controller = nodeController(**env_variables)
-    controller.prepare_nodes()
-    yield controller
+    if is_qe_env():
```
tsorya, line 94: not sure i follower the logic here? I think all logic must move to nodes fixture and here just return the relevant module.
Why to have same if's in different files

logging.info(f'--- SETUP --- node controller\n')
yield nodeController

Makefile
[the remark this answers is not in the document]
YuviGold: Can you check this workaround?
https://stackoverflow.com/questions/27006884/pytest-xdist-without-capturing-output

discovery-infra/test_infra/tools/assets.py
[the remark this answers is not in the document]
YuviGold: Because asset is too generic. It isn't clear what does it do or who owns it, what is it related to.
If you would ask me out-of-context what is an "asset" I would probably say either a token/credentials security related

discovery-infra/test_infra/tools/assets.py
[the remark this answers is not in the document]
YuviGold: Release SGTM

discovery-infra/tests/conftest.py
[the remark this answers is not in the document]
YuviGold: Between qe and dev envs? I don't understand why. These two environments won't run in parallel.

discovery-infra/tests/conftest.py
YuviGold: Why not just a function?

discovery-infra/tests/conftest.py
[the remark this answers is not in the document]
lalon4: right but for qe env, we set an ENV:  than we use it in a certain folder. ISO="/var/lib/libvirt/images/cluster-discovery.iso";

discovery-infra/tests/conftest.py
lalon4: cause than ill need to import conftest into base, and it is not the intention of it

Makefile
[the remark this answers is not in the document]
lalon4: didn't help so solve :(  same output

discovery-infra/tests/conftest.py
```diff
@@ -1,11 +1,17 @@
...
+import random
+from string import ascii_lowercase
 from distutils import util
 from pathlib import Path
 
 import pytest
 from test_infra import assisted_service_api, consts, utils
 
+
+def _get_random_name():
```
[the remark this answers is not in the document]
lalon4: but than ill need to import conftest

Makefile
lalon4: @tsorya its a bit tricky cause its inside the lock, so waiting will keep the file locked.
Cant we just declare that we are supporting up to X number of tests in parallel?
for now i will add 3 or 4 (depends on the time i have to test it).

tsorya: i think better to have wait

discovery-infra/test_infra/tools/assets.py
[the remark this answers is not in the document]
tsorya: Assets meant to be generic, if u want to add specific asset type u can. Just inherit

Makefile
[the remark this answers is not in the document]
lalon4: Well not exactly the same, its now show ERROR on real time but doesn't help much. 
I noticed on the web that when using -n that you also get progress bar - but when i run multi tests - no progress bar. 
I also tried to use a plugin called sugar, but it looks like all "graphical" output are somehow filtered, maybe because we run it from within a container.

lalon4: I also think it is nice to have, and not hard to implement. but i think we should not implement such a mechanism.
is_free_asset should be done outside the Asset obj, therefore there will be X processes that will hang and wait for free asset. meanwhile, each process should query a file (load into memo) in a loop, and in some cases for hours (until free asset). 
while we have 15 VM running, containers, and many more processes.
So i think this mechanism will consume more resources. I also have no time to invest in testing such cases. 
Now we have 40 tests, next month 100... dont know the influence 
In this PR scope, added 3 assets and this is what was tested. So for start we will support 3
WDYT?

scripts/install_environment.sh
```diff
@@ -160,6 +160,10 @@ function additional_configs() {
     echo "disabling selinux by setenforce 0"
     sudo setenforce 0 || true
     sudo chmod 600 ssh_key/key
+    sudo firewall-cmd --zone=libvirt --add-port=59151/tcp
```
YuviGold, line 163: You can use range `59151-59154`

Makefile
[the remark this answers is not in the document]
YuviGold: It makes sense if all of our logical workflows would run inside of a container. Makefile shouldn't hold code logic, it's just a nice way to wrap everything up.
In addition, I'd like to avoid code duplication.

discovery-infra/tests/conftest.py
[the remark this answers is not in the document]
YuviGold: What's the problem with that?

discovery-infra/tests/conftest.py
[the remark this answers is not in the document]
YuviGold: OK. Please add a comment on why it is done

discovery-infra/tests/conftest.py
YuviGold: What's the problem with that?

Makefile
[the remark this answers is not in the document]
nshidlin: this logic can be moved to its own make target, or a shell script that gets called from a make target wdyt?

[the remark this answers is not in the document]
tsorya: First of all lets delete all the "if" logic from here. Just rm -rf and mkdir -p enough

Makefile
```diff
@@ -351,4 +351,18 @@ test:
 	skipper make $(SKIPPER_PARAMS) _test
 
 _test: $(REPORTS)
+	if [ -d /tmp/assisted_test_infra_logs ]; then rm -rf /tmp/assisted_test_infra_logs; fi
+	mkdir /tmp/assisted_test_infra_logs
+	if [ -d /tmp/test_images ]; then rm -rf /tmp/test_images; fi
+	cp -p discovery-infra/test_infra/tools/tf_network_pool.json /tmp/tf_network_pool.json
```
tsorya, line 357: why tmp? and not build?

Makefile
```diff
@@ -351,4 +351,18 @@ test:
 	skipper make $(SKIPPER_PARAMS) _test
 
 _test: $(REPORTS)
+	if [ -d /tmp/assisted_test_infra_logs ]; then rm -rf /tmp/assisted_test_infra_logs; fi
+	mkdir /tmp/assisted_test_infra_logs
+	if [ -d /tmp/test_images ]; then rm -rf /tmp/test_images; fi
```
tsorya, line 356: add all those stuff to make clean target and delete all needed folders there.

discovery-infra/tests/conftest.py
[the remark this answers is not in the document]
nshidlin: it can be moved into utils and be imported into both conftest and base test

Makefile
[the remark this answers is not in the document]
lalon4: tried with separated target, didn't work, ill try again

discovery-infra/tests/base_test.py
```diff
@@ -25,13 +22,21 @@ def random_name():
...
-        nodes.shutdown_all()
-        nodes.format_all_disks()
+    def nodes(self, setup_node_controller, is_dev_env):
+        net_asset = None
+        try:
+            if is_dev_env:
+                net_asset = NetworkAssets()
+                env_variables["net_asset"] = net_asset.get()
+            controller = setup_node_controller(**env_variables)
+            controller.prepare_nodes()
```
tsorya, line 32: lets move it to nodes.prepare_nodes()

discovery-infra/tests/base_test.py
```diff
@@ -25,13 +22,21 @@ def random_name():
...
+            nodes = Nodes(controller, env_variables["private_ssh_key_path"])
+            yield nodes
+            logging.info(f'--- TEARDOWN --- node controller\n')
+            controller.destroy_all_nodes()
```
tsorya, line 36: same

discovery-infra/utils.py
```diff
@@ -356,6 +359,10 @@ def is_cluster_in_status(client, cluster_id, statuses):
         log.exception("Failed to get cluster %s info", cluster_id)
 
 
+def get_random_name(length=8):
```
nshidlin, line 362: added to the wrong utils file

This one is no longer with us

Makefile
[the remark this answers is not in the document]
lalon4: Fixed

discovery-infra/tests/conftest.py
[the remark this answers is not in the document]
lalon4: Fixed

discovery-infra/tests/conftest.py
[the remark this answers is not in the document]
lalon4: Fixed

discovery-infra/tests/conftest.py
lalon4: Fixed

discovery-infra/tests/conftest.py
lalon4: Fixed

scripts/install_environment.sh
lalon4: Fixed

Makefile
lalon4: Fixed

Makefile
lalon4: why build?

discovery-infra/utils.py
lalon4: Fixed

Makefile
lalon4: its for tests usage

discovery-infra/tests/base_test.py
lalon4: not in this scope. it requires more tests on both envs. 
why do we want that? isnt it OK that its done from the controller?

YuviGold (approved):

discovery-infra/tests/base_test.py
lalon4: Fixed

discovery-infra/tests/base_test.py
lalon4: fixed

PR #313, opened 2020-12-06
MGMT-3175 - move day2 flow to use InstallHost instead of InstallHosts

yevgeny-shnaidman:


discovery-infra/day2.py
```diff
@@ -117,7 +117,8 @@ def day2_nodes_flow(client,
     if install_cluster_flag:
         log.info("Start installing all known nodes in the cluster %s", cluster.id)
         ocp_orig_ready_nodes = get_ocp_cluster_ready_nodes_num()
-        client.install_day2_cluster(cluster.id)
+        hosts = client.get_cluster_hosts(cluster.id)
+        [client.install_day2_host(cluster.id, host['id']) for host in hosts if host["status"] == 'known']
```
YuviGold, line 121: Why do we need to filter by hosts' status?
Aren't them all in `consts.NodesStatus.KNOWN` from the previous call of `wait_till_all_hosts_are_in_status` ?

yevgeny-shnaidman: just to make sure. In day2 user may call install_host while other hosts are not in known state.

YuviGold (approved):

PR #321, opened 2020-12-07
moving tests downstream

nshidlin:
As conftest is moving downstream, moved the environment variables to there own file


YuviGold (approved):
LGTM

PR #323, opened 2020-12-08
Set libgcrypt installation as permissive

YuviGold:


YuviGold:
Was resloved on https://github.com/openshift/assisted-test-infra/pull/324

PR #325, opened 2020-12-09
Upgrade from python3.6 to python3.8

YuviGold:


YuviGold:
Prow base images don't support it at the moment. Would have to see how it can be overcome.

PR #335, opened 2020-12-15
MGMT-3269 IPv6 fixes

ori-amizur:
- Change default subnets and prefixes
- During cluster installation, query cluster for host statuses, and do not query libvirt for DHCP leases
/cc @YuviGold 
/cc @tsorya

### DIFF
--- a/discovery-infra/start_discovery.py
+++ b/discovery-infra/start_discovery.py
@@ -39,7 +39,7 @@ class MachineNetwork(object):
         self.cidr_v4 = machine_cidr_4
         self.cidr_v6 = machine_cidr_6
         self.provisioning_cidr_v4 = _get_provisioning_cidr(machine_cidr_4, ns_index)
-        self.provisioning_cidr_v6 = _get_provisioning_cidr(machine_cidr_6, ns_index)
+        self.provisioning_cidr_v6 = _get_provisioning_cidr6(machine_cidr_6, ns_index)
 
 
 # Filling tfvars json files with terraform needed variables to spawn vms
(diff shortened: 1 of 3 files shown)


discovery-infra/start_discovery.py
```diff
@@ -330,6 +330,13 @@ def _get_provisioning_cidr(cidr, ns_index):
     provisioning_cidr += ns_index + consts.NAMESPACE_POOL_SIZE
     return str(provisioning_cidr)
 
+def _get_provisioning_cidr6(cidr):
```
YuviGold, line 333: Maybe support multi deployments with ns_index?

discovery-infra/start_discovery.py
```diff
@@ -204,22 +204,22 @@ def create_nodes_and_wait_till_registered(
     )
 
     # TODO: Check for only new nodes
-    utils.wait_till_nodes_are_ready(
-        nodes_count=nodes_count, network_name=nodes_details["libvirt_network_name"]
-    )
     if not inventory_client:
+        utils.wait_till_nodes_are_ready(
```
YuviGold, line 208: The log doesn't match the action. Why in case no inventory_url you want to wait for them to register?

discovery-infra/test_infra/utils.py
```diff
@@ -612,18 +612,26 @@ def file_lock_context(filepath='/tmp/discovery-infra.lock', timeout=300):
...
+    desc = md.parseString(net.XMLDesc())
+    try:
+        hosts = desc.getElementsByTagName("network")[0].getElementsByTagName("ip")[0].getElementsByTagName("dhcp")[0].getElementsByTagName("host")
+        return list(map(lambda host: {"mac": host.getAttribute("mac"), "ipaddr": host.getAttribute("ip"), "hostname": host.getAttribute("name")}, hosts))
+    except IndexError:
+        return []
+
+def _merge(leases, hosts):
+    lips = [l["ipaddr"] for l in leases]
+    ret = leases + [h for h in hosts if h["ipaddr"] not in lips]
```
YuviGold, line 625: Shouldn't it check by hostname?

ori-amizur: Hostname does not always appear as part of the lease

discovery-infra/start_discovery.py
ori-amizur: The opposite.  When there is no inventory_url we are not waiting for registration.
The only different here is that in case there is no inventory client we wait for node leases.  Otherwise we skip to check nodes in the cluster.

discovery-infra/start_discovery.py
ori-amizur: The cidr should be already based on ns_index.

YuviGold: iirc, @RazRegev added this so the CI can run simultaneously, every run runs with the same cidr.

discovery-infra/test_infra/utils.py
YuviGold: IP yes and hostname no? wouldn't it break other code depending on that?
https://github.com/openshift/assisted-test-infra/blob/e211a3fff26aa3d3d2f62a3ba7e758e3b1af3b82/discovery-infra/test_infra/utils.py#L96

discovery-infra/start_discovery.py
YuviGold: Please elaborate with comments on why we don't wait for registration. The code isn't clear on that workflow.

discovery-infra/start_discovery.py
ori-amizur: OK, will add

discovery-infra/test_infra/utils.py
ori-amizur: Hostname is not mandatory in lease.  In IPv6 we don't have hostnames

discovery-infra/start_discovery.py
ori-amizur: OK

YuviGold (approved):
LGTM

PR #337, opened 2020-12-15
MGMT-2755: None platform integration

RazRegev:
To test none platform integration test_infra will initialize vms using
another terraform template, where the secondary network is not part of
the original masters and workers but has its own secondary masters and
secondary workers. When NUM_SECONDARY_WORKERS or NUM_SECONDARY_MASTERS
is given the mode will be change automatically to none platform.


YuviGold:
Could there be a different number of workers/masters to a regular flow and the new flow?
So this won't be two new env vars but just a toggle for "start the none_platform mechanism"

I don't really like the implementation of duplicating everything.
Could we use the same core terraform JSON and resources and just extend them?

RazRegev:
> Could there be a different number of workers/masters to a regular flow and the new flow?
> So this won't be two new env vars but just a toggle for "start the none_platform mechanism"
> 
> I don't really like the implementation of duplicating everything.
> Could we use the same core terraform JSON and resources and just extend them?

The thing is that we added to each master multiple dns that we do not necessarily want in a regular flow, and in addition, the secondary network from the regular flow was moved from the original masters to the secondary masters.
To make things easier to understand and manage, and to avoid the risk of this will effect the regular flow, Freddy and I thought it will be better to make a separate template for this flow. Otherwise, it will be a big complex template file with some irrelevant configuration for the common use cases. What do you think?

scripts/install_environment.sh
```diff
@@ -151,6 +151,13 @@ EOF
   sudo chmod +x $fname
 }
 
+function configure_iptables() {
```
tsorya, line 154: i think it is bad place. Try to set this inside container. I think it is bad idea to have it on each node

Makefile
```diff
@@ -85,6 +85,9 @@ ifdef KEEP_ISO
     KEEP_ISO_FLAG = --keep-iso
 endif
 
+NUM_SECONDARY_MASTERS := $(or $(NUM_SECONDARY_MASTERS), 0)
```
tsorya, line 88: I must say i don't like this option. Not sure why it should be configurable at all

Makefile
```diff
@@ -85,6 +85,9 @@ ifdef KEEP_ISO
...
+NUM_SECONDARY_WORKERS := $(or $(NUM_SECONDARY_WORKERS), 0)
```
tsorya, line 89: same here. Just do it inside code. Why it matters how much workers are on another network?
And the name is really not clear

discovery-infra/start_discovery.py
```diff
@@ -370,7 +385,13 @@ def nodes_flow(client, cluster_name, cluster, image_path):
 
     tf_folder = utils.get_tf_folder(cluster_name, args.namespace)
     utils.recreate_folder(tf_folder)
-    copy_tree(consts.TF_TEMPLATE, tf_folder)
+
+    is_none_platform = args.sec_master_count or args.sec_worker_count
```
tsorya, line 389: As i said, i think better to have flag that says "none_platform" and not to looks on those flags. I think don't need them at all

discovery-infra/start_discovery.py
```diff
@@ -460,6 +481,14 @@ def _extract_nodes_from_tf_state(tf_state, network_name, role):
 
     return data
 
+
+def _enable_multiple_networks_installation(base_url, cluster_id):
```
tsorya, line 485: move it to api

terraform_files/none_platform_flow/main.tf
```diff
@@ -0,0 +1,351 @@
...
+  bridge = "virbr126"
+  mtu = var.libvirt_network_mtu
+  domain = var.cluster_domain
+  addresses = var.machine_cidr_addresses
+  autostart = true
+
+  dns {
+    dynamic "hosts" {
+      for_each = concat(
+      data.libvirt_network_dns_host_template.masters.*.rendered,
```
tsorya, line 51: do we need full path? can't we have just wildcard?

terraform_files/none_platform_flow/terraform.tfvars.json
```diff
@@ -0,0 +1,30 @@
...
+  "cluster_domain": "redhat.com",
+  "cluster_name": "test-infra-cluster",
+  "machine_cidr_addresses": ["192.168.126.0/24"],
+  "provisioning_cidr_addresses": ["192.168.140.0/24"],
+  "master_count": "3",
+  "sec_master_count": "1",
+  "worker_count": "1",
+  "image_path": "tmp",
+  "libvirt_uri": "qemu+tcp://192.168.122.1/system",
+  "libvirt_network_if": "tt0",
```
tsorya, line 11: not relevant, please go through params and clean irrelevant

Makefile
RazRegev: Why not? Why can you configure regular masters and workers and not secondary (network) masters and workers?

tsorya: i don't see any usecase for it? so why to add irrelevant envS?

RazRegev:
will reopen in another pr

PR #340, opened 2020-12-17
Set default of 3 service replicas

YuviGold:
The service is going to be default with 1 replica in a development environment. https://github.com/openshift/assisted-service/pull/807
e2e should run with 3 replicas.


nmagnezi (commented):
/test e2e-metal-assisted-onprem

nmagnezi (commented):
/retest

PR #341, opened 2020-12-17
NO-ISSUE cleanup before test execution

djzager:


carbonin:
@YuviGold what do you think about this? Without this we'll get output like this when trying to run using a different `SERVICE_BRANCH` value.
```
11:01:49  Fetching origin
11:01:49  fatal: ambiguous argument 'origin/rm-assisted-iso-create': unknown revision or path not in the working tree.
11:01:49  Use '--' to separate paths from revisions, like this:
11:01:49  'git <command> [<revision>...] -- [<file>...]'
11:01:49  make: *** [Makefile:300: bring_assisted_service] Error 128
```

Alternatively we could make `fetch_assisted_service` more clever, but it might require a script depending on how much we want to do.

YuviGold:
> fetch_assisted_service

I'd prefer to change all the way that we depend on deployment scripts.
Currently, there's both an integration with images and also another one for deployment scripts...

I elaborated on it on a task [MGMT-3312 Run e2e workflow by assisted-service](https://issues.redhat.com/browse/MGMT-3312) as part of [MGMT-3311 Run e2e-metal-assisted workflow in Prow by every component] epic.

For now, cleaning is LGTM

YuviGold (approved):
LGTM

PR #342, opened 2020-12-18
NO-ISSUE Handle auth when querying release info

carbonin:
Previously, if we were pulling an image which required authentication,
we would fail when trying to execute `oc adm release info` unless
we had logged in with docker, before running.

This uses the pull secret env var to pull the image which should work
as we would need the same auth when deploying the cluster.


YuviGold (approved):
LGTM

PR #352, opened 2020-12-22
Adding events tests

iovadia:

Add methods for events in cluster
Add conts for events
Add hosts id parameter to get_events on assisted_service_api in order to get host events


discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -515,3 +551,18 @@ def wait_for_cluster_to_be_in_installing_status(self):
             cluster_id=self.id,
             statuses=[consts.ClusterStatus.INSTALLING]
         )
+
+    def get_events(self, host_id=''):
+        return self.api_client.get_events(cluster_id=self.id, host_id=host_id)
+
+    def verify_events(self, event_to_find, params_list, host_id='', fail_on_error=False):
```
lalon4, line 558: where your using this fail_on_error?
and what is param_list? can you give an example?

lalon4: Can you please replace verify_events with verify_event or find_event?

lalon4: params_list=[]

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -515,3 +551,20 @@ def wait_for_cluster_to_be_in_installing_status(self):
...
+    def get_events(self, host_id=''):
+        return self.api_client.get_events(cluster_id=self.id, host_id=host_id)
+
+    def find_event(self, event_to_find, params_list=[], host_id='', fail_on_error=True):
+        events_list = self.get_events(host_id=host_id)
+        event_exist = False
+        logging.info(f"Searching for event: {event_to_find}")
+        for event in events_list:
+            if event_to_find in event['message']:
+                if all(param in event['message'] for param in params_list):
```
gamli75, line 564: I think you should verify that the EventTime is after the test start time or between time ranges - if there is a test that should create the same event 5 times - you will decide that the event_exist even if the event appears only once

discovery-infra/test_infra/assisted_service_api.py
```diff
@@ -248,9 +248,9 @@ def download_cluster_logs(self, cluster_id, output_file):
         with open(output_file, "wb") as _file:
             _file.write(response.data)
 
-    def get_events(self, cluster_id):
+    def get_events(self, cluster_id, host_id=''):
```
gamli75, line 251: I think that we should change it to waitForEvent because not all the events are raised synchronously - poll the event list every 5 seconds with a timeout

gamli75:
I also think that we should verify the event message, for example, we wrote hostname and not host id or there are not illegal character, such as %d
I think you should review the bugs that related to events and try to cover them in the testing

discovery-infra/test_infra/assisted_service_api.py
iovadia: Done, please review

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -545,3 +581,39 @@ def reset_cluster_and_wait_for_ready(self, cluster, nodes):
...
+                    event_to_find,
+                    reference_time,
+                    params_list,
+                    host_id
+                ),
+                timeout_seconds=30,
+                sleep_seconds=2,
+                waiting_for="Event: %s" % event_to_find,
+            )
+        except:
```
nshidlin, line 616: except on the specific exception `waiting.exceptions.TimeoutExpired`

discovery-infra/test_infra/helper_classes/cluster.py
iovadia: Done, please review

discovery-infra/test_infra/helper_classes/cluster.py
iovadia: Done, please review.
Added reference_time parameter

discovery-infra/test_infra/helper_classes/cluster.py
iovadia: Done, please review

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -545,3 +581,36 @@ def reset_cluster_and_wait_for_ready(self, cluster, nodes):
...
+        event_exist = False
+        for event in events_list:
+            if event_to_find in event['message']:
+                # Adding a 2 sec buffer to account for a small time diff between the machine and the time on staging
+                if utils.to_utc(event['event_time']) >= reference_time-2:
+                    if all(param in event['message'] for param in params_list):
+                        event_exist = True
+                        logging.info(f"Event to find: {event_to_find} exists with its params")
+                        return event_exist
+        return event_exist
```
YuviGold, line 598: Use `for-else` instead of `event_exist` variable

```python
    def _find_event(self, event_to_find, reference_time, params_list, host_id):
        events_list = self.get_events(host_id=host_id)
        for event in events_list:
            if event_to_find in event['message']:
                # Adding a 2 sec buffer to account for a small time diff between the machine and the time on staging
                if utils.to_utc(event['event_time']) >= reference_time-2:
                    if all(param in event['message'] for param in params_list):
                        logging.info(f"Event to find: {event_to_find} exists with its params")
                        return True
        else:
            return False

iovadia: Done, please review

YuviGold (approved):
LGTM

PR #355, opened 2020-12-23
Add TEMP sleep before declaring that cluster is Ready

lalon4:


YuviGold:
Why do we wait twice? To verify consistency?

lalon4:
According to the failures we had, looks like between the wait_for_ready and start_install, some hosts are moving to insufficient. 
So need to make sure they are connected before start install otherwise test fails. 
This is TEMP solution, for now we need to understand if its a defect need to be fixed in the service or real machine connectivity issue

YuviGold (approved):

PR #357, opened 2020-12-24
MGMT-3258 Smart logs collection

YuviGold:
Moving logs utils into a separate module `logs_utils.py`.
Added error messages to the `assert`s.

Using the `verify_logs_uploaded` function to verify downloaded logs, in case of invalid logs - retry after some interval sleep.

/cc @ybettan  @nshidlin @lalon4 @slaviered


discovery-infra/test_infra/logs_utils.py
```diff
@@ -0,0 +1,69 @@
...
+    if not installation_success:
+        for logs_type in ["dmesg.logs", "log-bundle"]:
+            assert any(logs_type in s for s in logs), f"{logs_type} isn't found in {logs}"
+        # test that installer-gather gathered logs from all masters
+        lb_path = [s for s in logs if "log-bundle" in s][0]
+        gz.extract(lb_path, dir_path)
+        lb = tarfile.open(os.path.join(dir_path, lb_path))
+        lb.extractall(dir_path)
+        cp_path = [s for s in lb.getnames() if "control-plane" in s][0]
+        assert len(os.listdir(os.path.join(dir_path, cp_path))) == NUMBER_OF_MASTERS - 1, f"expecting {os.listdir(os.path.join(dir_path, cp_path))} to have {NUMBER_OF_MASTERS - 1} values"
```
ybettan, line 60: Please use `env_variables["num_masters"]` instead of `NUMBER_OF_MASTERS`

YuviGold: I can't use that, while it's a code inside the tests

discovery-infra/test_infra/logs_utils.py
```diff
@@ -0,0 +1,69 @@
+import logging
+import os
+import shutil
+import tarfile
+import time
+
+from test_infra.consts import NUMBER_OF_MASTERS
+
+
+def verify_logs_uploaded_and_delete(cluster_tar_path, expected_min_log_num, installation_success):
```
ybettan, line 10: This is used in the d/s repo?

discovery-infra/download_logs.py
```diff
@@ -62,10 +69,29 @@ def download_logs(client: InventoryClient, cluster: dict, dest: str, must_gather
...
 
     with suppress(assisted_service_client.rest.ApiException):
-        client.download_cluster_logs(cluster['id'], os.path.join(output_folder, f"cluster_{cluster['id']}_logs.tar"))
+        for i in range(MAX_RETRIES):
+            cluster_logs_tar = os.path.join(output_folder, f"cluster_{cluster['id']}_logs.tar")
+
+            with suppress(FileNotFoundError):
+                os.remove(cluster_logs_tar)
+
+            client.download_cluster_logs(cluster['id'], cluster_logs_tar)
```
ybettan, line 81: What will happen if `cluster_logs_tar` already exist?

discovery-infra/test_infra/logs_utils.py
YuviGold: Yes.

discovery-infra/download_logs.py
YuviGold: On the previous line, I'm trying to remove it if it already exists.

discovery-infra/download_logs.py
```diff
@@ -62,10 +69,29 @@ def download_logs(client: InventoryClient, cluster: dict, dest: str, must_gather
...
+
+            min_number_of_logs = len(cluster['hosts']) + 1 if cluster['status'] == ClusterStatus.INSTALLED else len(cluster['hosts'])
+
+            try:
+                verify_logs_uploaded(cluster_logs_tar, min_number_of_logs, cluster['status'] == ClusterStatus.INSTALLED)
```
ybettan, line 86: If you failed on the first try you will fail on all other tries because you don't delete the create dir that holds the extracted files, therefore on each try it will iterate also on old files (that failed in previous try) and fail again.

YuviGold: Good catch. Changing implementation to use `tempfile.TemporaryDirectory`

lalon4:
Thinking about the logs_utils, do we really need a separated util for logs? cant we have them as part of utils?

discovery-infra/test_infra/logs_utils.py
```diff
@@ -0,0 +1,69 @@
...
+import os
+import shutil
+import tarfile
+import time
+from tempfile import TemporaryDirectory
+
+from test_infra.consts import NUMBER_OF_MASTERS
+
+
+def verify_logs_uploaded_and_delete(cluster_tar_path, expected_min_log_num, installation_success):
```
ybettan, line 11: The only thing that this function does and `verify_logs_uploaded` doesn't is removing `cluster_tar_path` - it is redundant IMO.

Lets just keep it simple and use only `verify_logs_uploaded` and delete `cluster_tar_path` in it as it was originally.

It is not a big deal to download the tar on each try and it is state less - so simpler and more robust.

Jenkinsfile.download_logs
```diff
@@ -5,7 +5,7 @@ pipeline {
         string(name: 'REMOTE_SERVICE_URL', defaultValue: 'https://api.openshift.com', description: 'Service URL')
     }
 
-    triggers { cron('H/10 * * * *') }
+    triggers { cron('H/30 * * * *') }
```
romfreiman, line 8: why?

discovery-infra/test_infra/logs_utils.py
romfreiman, line 11: I dont understand why u need it in this context. How does it help?

Jenkinsfile.download_logs
YuviGold: Now each run might take longer. So together with @ErezAlster we figured out we should extend the interval as well.

discovery-infra/test_infra/logs_utils.py
YuviGold: I don't need it. These functions are also used in the e2e logs tests.
Anyway, this function got deleted.

YuviGold:
> Thinking about the logs_utils, do we really need a separated util for logs? cant we have them as part of utils?

It is just cleaner.

PR #359, opened 2020-12-24
MGMT-3348 Add chrony NTP server for both IPv4 and IPv6

ori-amizur:


scripts/install_environment.sh
```diff
@@ -150,6 +150,25 @@ EOF
...
 
+function config_nginx() {
+  echo "Config nginx"
+  sudo dnf install -y nginx
+  fname=/etc/nginx/ntp.conf
+  sudo cat <<EOF > $fname
+stream {
+	server {
+		listen [::]:123 udp;
+		proxy_pass clock.redhat.com:123;
```
empovit, line 161: Do I understand correctly that you always forward NTP requests to clock.redhat.com? If so, it might not work on some machines. IMO this should be a variable. Better yet, run a local NTP server and use it as an additional NTP source.

ori-amizur: Regarding clock.redhat.com, maybe ask @YuviGold .  We do it today for IPv4.  Here we just convert to IPv6.

discovery-infra/start_discovery.py
```diff
@@ -297,6 +298,13 @@ def _get_vips_ips(machine_net):
 # Converts params from args to assisted-service cluster params
 def _cluster_create_params():
     ipv4 = args.ipv4 and args.ipv4.lower() in MachineNetwork.YES_VALUES
+    ipv6 = args.ipv6 and args.ipv6.lower() in MachineNetwork.YES_VALUES
+    ntp_source = None
```
empovit, line 302: you don't need this in Python

discovery-infra/start_discovery.py
```diff
@@ -297,6 +298,13 @@ def _get_vips_ips(machine_net):
...
+    if ipv6 and not ipv4:
+        ntp_source = str(IPNetwork(args.vm_network_cidr6).ip + 1)
```
empovit, line 304: Don't we already have the VM IP address calculated somewhere? Doing the same calculation in multiple places bothers me :/

discovery-infra/start_discovery.py
ori-amizur: Why not?

discovery-infra/start_discovery.py
ori-amizur: This is not VM IP, but the address on the host

terraform_files/main.tf
```diff
@@ -36,6 +36,13 @@ resource "libvirt_network" "net" {
       ip = var.api_vip
       hostname = "api.${var.cluster_name}.${var.cluster_domain}"
     }
+    dynamic "hosts" {
+      for_each = var.extra_dns_names
+      content {
+        hostname = hosts.key
+        ip = hosts.value
+      }
+    }
```
empovit, line 45: Is it still needed when you use the IP address as NTP source?

terraform_files/variables-libvirt.tf
```diff
@@ -144,3 +144,10 @@ variable "running" {
 variable "cluster_inventory_id" {
   type      = string
 }
+
+variable "extra_dns_names" {
+  type = map(string)
+  default = {
+  }
+}
```
empovit, line 152: Is this still needed?

discovery-infra/start_discovery.py
empovit: Doesn't matter, if we already calculate it somewhere else it should be re-used.

discovery-infra/start_discovery.py
empovit: It's redundant. ntp_source will be assigned either by the `if` or by the `else` branch. There's no point in declaring it or initializing it with `None`.

discovery-infra/start_discovery.py
YuviGold: Will it always be the IP on the host? Shouldn't we do something like `get_ip_by_interface_name`?

ori-amizur: @empovit Where else do we do this calculation?

terraform_files/main.tf
ori-amizur: No

terraform_files/variables-libvirt.tf
ori-amizur: No

YuviGold (approved):
LGTM

PR #363, opened 2020-12-28
MGMT-3408 - test-infra hosts VMs should be configured with static MAC…

yevgeny-shnaidman:
…s on network interfaces


YuviGold (approved):
LGTM

PR #364, opened 2020-12-28
MGMT-3292 - Added test infrastructure for attaching disks

omertuc:
(Ignore the branch name, it's a mistake and can't be changed in GitHub after a PR is created)

Adds ability to attach disks to hosts. This in order to allow writing tests for the "Show all disks" epic (where we show all disks, and why they aren't eligible for installation, rather than just disks that are eligible for installation).

### DIFF
--- a/discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
+++ b/discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
@@ -1,5 +1,12 @@
 import os
+import re
+import string
+import uuid
 import logging
+import tempfile
+from abc import ABC
+from typing import List
+
 import libvirt
 import waiting
 from xml.dom import minidom
@@ -10,7 +17,8 @@ from test_infra import consts
 from test_infra.controllers.node_controllers.node_controller import NodeController
 
 
-class LibvirtController(NodeController):
+class LibvirtController(NodeController, ABC):
+    TEST_DISKS_PREFIX = "ua-TestInfraDisk"
 
     def __init__(self, **kwargs):
         self.libvirt_connection = libvirt.open('qemu:///system')
@@ -20,15 +28,15 @@ class LibvirtController(NodeController):
     def __del__(self):
         with suppress(Exception):
             self.libvirt_connection.close()
-
+            
     @property
     def setup_time(self):
         return self._setup_timestamp
 
-    def list_nodes(self):
+    def list_nodes(self) -> List[libvirt.virDomain]:
         return self.list_nodes_with_name_filter(None)
 
-    def list_nodes_with_name_filter(self, name_filter):
+    def list_nodes_with_name_filter(self, name_filter) -> List[libvirt.virDomain]:
         logging.info("Listing current hosts with name filter %s", name_filter)
         nodes = []
         domains = self.libvirt_connection.listAllDomains()
@@ -84,7 +92,12 @@ class LibvirtController(NodeController):
         return nodes
 
     @staticmethod
-    def format_disk(disk_path):
+    def create_disk(disk_path, disk_size):
+        command = f'qemu-img create -f qcow2 {disk_path} {disk_size}'
+        utils.run_command(command, shell=True)
+
+    @classmethod
+    def format_disk(cls, disk_path):
         logging.info("Formatting disk %s", disk_path)
         if not os.path.exists(disk_path):
             logging.info("Path to %s disk not exists. Skipping", disk_path)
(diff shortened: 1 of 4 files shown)


discovery-infra/test_infra/controllers/node_controllers/node_controller.py
```diff
@@ -33,6 +33,24 @@ def format_node_disk(self, node_name: str) -> None:
     def format_all_node_disks(self) -> None:
         raise NotImplementedError
 
+    def attach_disk(self, node_name, disk_size):
```
YuviGold, line 36: Please add type hints

discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
```diff
@@ -91,8 +99,40 @@ def format_disk(disk_path):
...
+    def attach_disk(self, node, disk_size):
+        node = self.libvirt_connection.lookupByName(node)
+
+        # <target> dev attribute needs some unique device identifier for the disk (target_dev).
+        # This path will not be the same path used by the guest for this disk.
+        # See https://libvirt.org/formatdomain.html#elementsDisks:~:text=The%20target%20element%20controls%20the%20bus
+        # We generate a random one with a lot of digits so collision is very unlikely
+        # When detaching, we use the alias rather than target_dev,
+        # (unlike virsh attach-disk and detach-disk commands)
+        disk_id = random.randrange(int(1e12))
```
YuviGold, line 114: Use `uuid.uuid4()`

omertuc: I need decimal digits without hyphens or hex... vda1892379817239 vda1237818723 etc

YuviGold:
```
In [11]: x                                                                      
Out[11]: UUID('ceb3ed14-bc4f-45a0-9294-b3cccc78f431')

In [12]: x.hex                                                                  
Out[12]: 'ceb3ed14bc4f45a09294b3cccc78f431'

In [13]: x.int                                                                  
Out[13]: 274755196843288735486359503772499637297
```

omertuc: While checking to see if such long numbers are supported (truncating UUID's is a bad idea) I stumbled upon a potential limitation/bug that makes this entire line irrelevant, so resolving for now.

/hold

discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
```diff
@@ -91,8 +99,40 @@ def format_disk(disk_path):
...
+        target_dev = f"vdtest{disk_id}"
+        disk_alias = f"ua-testDisk{disk_id}"
+
+        tmp_disk = tempfile.mktemp()
```
nshidlin, line 118: `mktemp()` is deprecated
https://docs.python.org/3/library/tempfile.html#tempfile.mktemp

discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
```diff
@@ -96,8 +109,124 @@ def format_disk(disk_path):
         # Fix for libvirt 6.0.0
         if image_size.isdigit():
             image_size += "G"
-        command = f'qemu-img create -f qcow2 {disk_path} {image_size}'
-        utils.run_command(command, shell=True)
+
+        cls.create_disk(disk_path, image_size)
+
+    @staticmethod
```
YuviGold, line 115: What's the benefit of having all these functions as `staticmethod`s and also at the same time private functions? Isn't this contradictional?

discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
```diff
@@ -10,7 +17,8 @@
 from test_infra.controllers.node_controllers.node_controller import NodeController
 
 
-class LibvirtController(NodeController):
+class LibvirtController(NodeController, ABC):
```
YuviGold, line 20: It seems like there are lots of disk functions.. Maybe we can separate it somehow logically? inner-class?

discovery-infra/test_infra/controllers/node_controllers/node_controller.py
```diff
@@ -1,73 +1,114 @@
...
-from test_infra.controllers.node_controllers import node
+from typing import Dict, List, Any, Tuple
+from test_infra.controllers.node_controllers.node import Node
+from abc import ABC, abstractmethod
 
 
-class NodeController:
-    def list_nodes(self) -> Dict[str, node.Node]:
-        raise NotImplementedError
+class NodeController(ABC):
```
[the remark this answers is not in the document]
omertuc: ABC is better than no ABC, and this is how you work with ABC. ABC makes sure you don't even instantiate a class that doesn't have all its (and its parents) abstract methods implemented. You clearly mark methods as abstract with `@abstractmethod`. Since it won't even let you instantiate the class, it will catch missing implementation errors much more early. If you only raise when the method is called, you only catch this bug during runtime, and if that method is not called on your happy path, that bug can hide for a long time before you even figure out that it's missing

discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
```diff
@@ -96,8 +109,124 @@ def format_disk(disk_path):
...
+
+        # Prefixing the disk's target element's dev attribute with `sd` makes libvirt create an SCSI disk.
+        # We don't use `vd` virtio disks because libvirt overwrites our aliases if we do so, coming up with
+        # its own `virtio-<num>` aliases instead. Those aliases allow us to identify disks created by this
+        # function when we perform `detach_all_test_disks` for cleanup.
+        target_dev = f"sd{self._get_available_scsi_identifier(node)}"
+        disk_alias = f"{self.TEST_DISKS_PREFIX}-{target_dev}"
+
+        with tempfile.NamedTemporaryFile(delete=False) as f:
+            tmp_disk = f.name
```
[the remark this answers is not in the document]
omertuc: Come to think of it I prefer it gets deleted, so I'll just remove the delete=False

discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
omertuc: `@staticmethod` technically only makes the calling requirements less strict, so if a method *can* be static, it *should* be static. PyCharm even detects methods that don't use self and suggests you convert them to static, rightfully so in my opinion.

One argument for why you would want to make a private method static - Say you want to add a future, public method to the class, that actually *needs* to be static (for whatever reason). That method might want to make use of one of my private methods. If my methods are not static, that public static method cannot make use of them.

discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
omertuc: That thought crossed my mind, I personally think adding another abstraction / layer / class here will make things more confusing/complicated than just simply having those methods inside the class. They do logically fit here.

discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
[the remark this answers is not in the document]
omertuc: In this case I want qemu to create the file, all I ever wanted was a tmp file name with no security considerations. The only sane way to get a temp file name without creating a file is using `mktemp`, which is deprecated, so you're stuck with `NamedTemporaryFile` or `mkstemp` which force you to actually create a file you don't even need.

discovery-infra/test_infra/controllers/node_controllers/node_controller.py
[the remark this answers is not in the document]
YuviGold: FYI @tsorya @nshidlin  @lalon4

YuviGold (approved):
LGTM

PR #367, opened 2020-12-29
adding iptables fixture and controller

nshidlin:
This PR adds an iptables controller and fixture in base_test.

This allows to add and remove iptables rules from within the tests based on node ips

### DIFF
--- a/discovery-infra/test_infra/controllers/iptables.py
+++ b/discovery-infra/test_infra/controllers/iptables.py
@@ -1,0 +1,48 @@
+import logging
+
+from test_infra.utils import run_command
+
+class IptableRule:
+    def __init__(self, chain, target, protocol, dest_port='', sources=None, extra_args=''):
+        self._chain = chain
+        self._target = target
+        self._protocol = protocol
+        self._dest_port = dest_port
+        self._sources = sources if sources else []
+        self._extra_args = extra_args
+
+    def _build_command_string(self, option):
+        sources_string = ','.join(self._sources)
+        rule_template = ["iptables", f"--{option}", self._chain, "-p" ,self._protocol, "-j", self._target]
+
+        if self._sources:
+            rule_template += ["-s", sources_string]
+
+        if self._dest_port:
+            rule_template += ["--dport", self._dest_port]
+        
+        if self._extra_args:
+            rule_template += [self._extra_args]
+
+        return " ".join(rule_template)
+    
+    def _does_rule_exist(self):
+        check_rule = self._build_command_string('check')
+        _, _, exit_code = run_command(check_rule, shell=True, raise_errors=False)
+
+        return exit_code == 0
+
+    def add_sources(self, sources):
+        self._sources += sources
+
+    def insert(self):
+        if not self._does_rule_exist():
+            insert_rule = self._build_command_string('insert')
+            logging.info(f'Setting iptable rule: {insert_rule}')
+            run_command(insert_rule, shell=True)
+
+    def delete(self):
+        if self._does_rule_exist():
+            delete_rule = self._build_command_string('delete')
+            logging.info(f'Removing iptable rule: {delete_rule}')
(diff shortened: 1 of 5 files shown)


discovery-infra/test_infra/controllers/iptables.py
```diff
@@ -0,0 +1,44 @@
...
+        rule += f' -s {sources_string}' if self._sources else ''
+        rule += f' --dport {self._dest_port}' if self._dest_port else ''
+        rule += f' {self._extra_agrs}' if self._extra_agrs else ''
+        rule += f' -j {self._target}'
+
+        return rule
+    
+    def _does_rule_exist(self):
+        check_rule = self._build_command_string('check')
+        res = run_command(check_rule, shell=True, raise_errors=False)
```
lalon4, line 28: Assuming the rule doesn't exist, than you apply it. no?

nshidlin: this function just checks if the rule exists

This is checked only on delete

You're right will check on insert as well

discovery-infra/test_infra/controllers/iptables.py
```diff
@@ -0,0 +1,45 @@
...
+        self._chain = chain
+        self._target = target
+        self._protocol = protocol
+        self._dest_port = dest_port
+        self._sources = sources if sources else []
+        self._extra_agrs = extra_args
+
+    def _build_command_string(self, option):
+        sources_string = ','.join(self._sources)
+        rule = f'iptables --{option}'
```
YuviGold, line 16: This isn't so readable..  Maybe a core list with `" ".join(...)` and only add the relevant flags if needed? 
i.e.
```python
rule = ["iptables", f"--{option}", self._chain, "-p" ,self._protocol, "-j", self._target]

if self.sources:
    rule += ["-s", sources_string]

if self._dest_port:
    rule += ["--dport", self._dest_port]

if self._extra_args:
    rule += [self._extra_args]

return " ".join(rule_template)

File: discovery-infra/test_infra/controllers/iptables.py [hunk 3]
```diff
@@ -0,0 +1,45 @@
...
+from test_infra.utils import run_command
+
+class IptableRule:
+    def __init__(self, chain, target, protocol, dest_port='', sources=None, extra_args=''):
+        self._chain = chain
+        self._target = target
+        self._protocol = protocol
+        self._dest_port = dest_port
+        self._sources = sources if sources else []
+        self._extra_agrs = extra_args
```
YuviGold, line 12: typo
`self._extra_agrs` -> `self._extra_args`

discovery-infra/test_infra/controllers/iptables.py
```diff
@@ -0,0 +1,45 @@
...
+        rule += f' -s {sources_string}' if self._sources else ''
+        rule += f' --dport {self._dest_port}' if self._dest_port else ''
+        rule += f' {self._extra_agrs}' if self._extra_agrs else ''
+        rule += f' -j {self._target}'
+
+        return rule
+    
+    def _does_rule_exist(self):
+        check_rule = self._build_command_string('check')
+        res = run_command(check_rule, shell=True, raise_errors=False)
```
YuviGold, line 28:
```python
_, _, exit_code = ....

return exit_code == 0
```

discovery-infra/test_infra/controllers/iptables.py
nshidlin: done

discovery-infra/test_infra/controllers/iptables.py
nshidlin: done

discovery-infra/test_infra/controllers/iptables.py
nshidlin: done

YuviGold (approved):
LGTM

PR #368, opened 2020-12-29
Adding libvirt logs

lalon4:
In some cases the nodes fails to get IPs when booting. Not yet sure if its ISO issue or libvirt or lack of machine resources.

This PR collect more logs (mainly libvirt logs) on test failure, in order to properly understand the root cause when reproduces.

### DIFF
--- a/discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
+++ b/discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
@@ -15,10 +15,15 @@ class LibvirtController(NodeController):
     def __init__(self, **kwargs):
         self.libvirt_connection = libvirt.open('qemu:///system')
         self.private_ssh_key_path = kwargs.get("private_ssh_key_path")
+        self._setup_timestamp = utils.run_command("date +\"%Y-%m-%d %T\"")[0]
 
     def __del__(self):
         with suppress(Exception):
             self.libvirt_connection.close()
+
+    @property
+    def setup_time(self):
+        return self._setup_timestamp
 
     def list_nodes(self):
         return self.list_nodes_with_name_filter(None)
--- a/discovery-infra/tests/base_test.py
+++ b/discovery-infra/tests/base_test.py
@@ -2,9 +2,6 @@ import logging
 import pytest
 import json
 import os
-import tarfile
-import shutil
-import time
 from contextlib import suppress
 from typing import Optional
 
@@ -38,7 +35,7 @@ class BaseTest:
                 net_asset.release_all()
 
     @pytest.fixture()
-    def cluster(self, api_client, request):
+    def cluster(self, api_client, request, nodes):
         clusters = []
 
         def get_cluster_func(cluster_name: Optional[str] = None,
@@ -56,7 +53,7 @@ class BaseTest:
 
         for cluster in clusters:
             logging.info(f'--- TEARDOWN --- Collecting Logs for test: {request.node.name}\n')
-            self.collect_test_logs(cluster, api_client, request.node)
+            self.collect_test_logs(cluster, api_client, request.node, nodes)
             logging.info(f'--- TEARDOWN --- deleting created cluster {cluster.id}\n')
             if cluster.is_installing() or cluster.is_finalizing():
                 cluster.cancel_install()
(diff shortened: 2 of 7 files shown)


discovery-infra/tests/base_test.py
```diff
@@ -90,12 +87,45 @@ def assert_string_length(string, expected_len):
...
         log_dir_name = f"{env_variables['log_folder']}/{test.name}"
         with suppress(ApiException):
             cluster_details = json.loads(json.dumps(cluster.get_details().to_dict(), sort_keys=True, default=str))
             download_logs(api_client, cluster_details, log_dir_name, test.result_call.failed)
+        if test.result_call.failed:
+            self._collect_virsh_logs(nodes, log_dir_name)
+
+    @staticmethod
+    def _collect_virsh_logs(nodes, log_dir_name):
+        logging.info('---TEARDOWN --- Collecting virsh logs\n')
```
YuviGold, line 100: This is a utility, it wouldn't necessary be called on teardown. I suggest to remove it

discovery-infra/tests/base_test.py
```diff
@@ -90,12 +87,45 @@ def assert_string_length(string, expected_len):
...
+        if not os.path.exists(log_dir_name):
+            logging.info(f"Creating virsh logs directory: {log_dir_name}")
```
YuviGold, line 102: I think this log is redundant.

discovery-infra/tests/base_test.py
```diff
@@ -90,12 +87,45 @@ def assert_string_length(string, expected_len):
...
+            os.makedirs(log_dir_name, exist_ok=False)
```
YuviGold, line 103: You can keep switch to `exist_ok=True` and remove the previous `os.path.exists` check

discovery-infra/tests/base_test.py
lalon4: It is necessarily called in teardown, and in specific order (we make sure nodes are not teadwon before we capture all virsh data)
I dont see other usage than logs debugging.

discovery-infra/tests/base_test.py
lalon4: It is the directory that can be created at collect_test_logs method, but, this folder created only if cluster obj exist. Not in all failures cluster obj created, so for that cases im adding folder creation.

discovery-infra/tests/base_test.py
lalon4: exists_ok True\ false is kind of irrelevant, no? cause im already checking if this folder doesnt exist, and than run create dir

discovery-infra/tests/base_test.py
```diff
@@ -90,12 +87,45 @@ def assert_string_length(string, expected_len):
...
-    def collect_test_logs(cluster, api_client, test):
+    def collect_test_logs(self, cluster, api_client, test, nodes):
         log_dir_name = f"{env_variables['log_folder']}/{test.name}"
         with suppress(ApiException):
             cluster_details = json.loads(json.dumps(cluster.get_details().to_dict(), sort_keys=True, default=str))
             download_logs(api_client, cluster_details, log_dir_name, test.result_call.failed)
+        if test.result_call.failed:
+            self._collect_virsh_logs(nodes, log_dir_name)
+
+    @staticmethod
```
YuviGold, line 98: Why is this function both `staticmethod` and private?
I think the `staticmethod` decorator can be removed.

discovery-infra/tests/base_test.py
YuviGold: Nevertheless, this function is a utility, maybe it will become useful in the future.
IMO it shouldn't show `TEARDOWN` in the logs. The calling function already mentioned it's a teardown related
```
logging.info(f'--- TEARDOWN --- Collecting Logs for test: {request.node.name}\n')
```

discovery-infra/tests/base_test.py
YuviGold: That's right. But to make the code cleaner you can just keep it simple without any conditions.
```python
def _collect_virsh_logs(nodes, log_dir_name):
    logging.info('Collecting virsh logs\n')
    os.makedirs(log_dir_name, exist_ok=True)
    ...
```

discovery-infra/tests/base_test.py
YuviGold: I still don't understand why it is an important log for the log collection workflow.
Especially in case `exist_ok=True` would be set. Already created/newly created - it doesn't really matter.

discovery-infra/tests/base_test.py
```diff
@@ -90,12 +87,42 @@ def assert_string_length(string, expected_len):
...
+
+        qemu_libvirt_path = os.path.join(virsh_log_path, "qemu_libvirt_logs")
+        os.makedirs(qemu_libvirt_path, exist_ok=False)
+        for node in nodes:
+            infra_utils.run_command(f"cp -p /var/log/libvirt/qemu/{node.name}.log "
+                                    f"{qemu_libvirt_path}/{node.name}.log",
+                                    shell=True)
+
+        libvird_log_path = os.path.join(virsh_log_path, "libvirtd_journal")
+        infra_utils.run_command(f"journalctl -D /run/log/journal >> {libvird_log_path}", shell=True)
```
nshidlin, line 125: this doesn't filter logs only for libvirtd 
`journalctl -u libvirtd -D /run/log/journal`

discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
```diff
@@ -15,11 +15,16 @@ class LibvirtController(NodeController):
     def __init__(self, **kwargs):
         self.libvirt_connection = libvirt.open('qemu:///system')
         self.private_ssh_key_path = kwargs.get("private_ssh_key_path")
+        self.setup_timestamp = utils.run_command("date +\"%Y-%m-%d %T\"")[0]
```
nshidlin, line 18: should be `self._setup_timestamp` if you have a getter for it

YuviGold (approved):
LGTM

PR #369, opened 2020-12-29
MGMT-3393 Add openshift_version parameter to Cluster object & Install…

YuviGold:
… all OPENSHIFT_VERSIONS


discovery-infra/tests/base_test.py
```diff
@@ -42,13 +42,15 @@ def cluster(self, api_client, request):
         clusters = []
 
         def get_cluster_func(cluster_name: Optional[str] = None,
-                             additional_ntp_source: Optional[str] = consts.DEFAULT_ADDITIONAL_NTP_SOURCE):
+                             additional_ntp_source: Optional[str] = consts.DEFAULT_ADDITIONAL_NTP_SOURCE,
+                             openshift_version: Optional[str] = env_variables['openshift_version']):
```
lalon4, line 46: This will break all the other tests cause the default of env_variables['openshift_version'] is
 "openshift_version": utils.get_openshift_version()
conftest need to have default:
utils.get_env('OPENSHIFT_VERSION', utils.get_openshift_version()),

discovery-infra/tests/test_install.py
```diff
@@ -1,11 +1,22 @@
...
+import subprocess
+
 import pytest
 
 from tests.base_test import BaseTest
 
+
+def pytest_generate_tests(metafunc):
+    if "openshift_version" in metafunc.fixturenames:
+        cmd = "kubectl get configmap assisted-service-config -n assisted-installer -o jsonpath='{.data.OPENSHIFT_VERSIONS}'"
```
nshidlin, line 11: You can get the openshift versions through with an API call to assisted-service, wouldn't that be better?

discovery-infra/tests/base_test.py
YuviGold: `get_openshift_version` function have `OPENSHIFT_VERSION` reference.

lalon4: ho right, missed that

YuviGold:
@ronniel1 You want that each PR would run both tests? Or is that too much?

/hold

YuviGold:
Running in parallel
/unhold

discovery-infra/test_infra/controllers/node_controllers/qe_vm_controler.py
```diff
@@ -22,3 +24,6 @@ def prepare_nodes(self):
 
     def get_cluster_network(self):
         return "baremetal-0"
+
+    def get_machine_cidr(self):
+        return utils.get_env('NETWORK_CIDR', '192.168.126.0/24')
```
lalon4, line 29: Double checking linchpin CIDR, and its 192.168.123.0/24, can you please update?

PR #371, opened 2020-12-31
Mgmt 3108 - use latest release image for installing SNO with bootstrap-in-place

eranco74:


discovery-infra/test_infra/utils.py
```diff
@@ -58,9 +58,9 @@ def _io_buffer_to_str(buf):
     return out, err, process.returncode
 
 
-def run_command_with_output(command):
+def run_command_with_output(command, env):
```
osherdp, line 61: Backwards compatibility not needed?
If so, use empty dict / ``os.environ`` as default value

terraform_files/baremetal/main.tf
```diff
@@ -88,11 +88,13 @@ resource "libvirt_domain" "master" {
     addresses  = var.libvirt_master_ips[count.index]
   }
 
-  network_interface {
-    network_name = libvirt_network.secondary_net.name
-    addresses  = var.libvirt_secondary_master_ips[count.index]
+  dynamic "network_interface" {
+    for_each = var.bootstrap_in_place ? [] : ["secondary_net"]
+    content {
+        network_name = libvirt_network.secondary_net.name
```
osherdp, line 94: deindent two spaces

osherdp (changes requested):

PR #374, opened 2020-12-31
fix-flaky-ci-failure-due-numeric-runid

RazRegev:
Added string prefix to PROFILE and NAMESPACE variables in Jenkinesfile in order to eliminate the possibility that it will be interpreted as an integer (which can't be used as namespace or profile names).


RazRegev:
closed due duplication https://github.com/openshift/assisted-test-infra/pull/375

PR #380, opened 2021-01-03
MGMT-3490 Add flag to control log collector must-gather

oshercc:


YuviGold:
Why?

ErezAlster:
@yuv

ErezAlster:
@YuviGold, we want to analyze all installation attempt snapshots for data analysis.

YuviGold:
@ErezAlster 
We don't have enough storage to save each test-infra run with must-gather artifacts on Jenkins

ErezAlster:
@oshercc

Makefile
```diff
@@ -312,7 +312,7 @@ delete_all_virsh_resources: destroy_all_nodes delete_minikube kill_all_port_forw
 	skipper run $(SKIPPER_PARAMS) 'discovery-infra/delete_nodes.py -ns $(NAMESPACE) -a'
 
 _download_logs:
-	discovery-infra/download_logs.py $(REMOTE_SERVICE_URL) $(LOGS_DEST) --cluster-id $(CLUSTER_ID)
+	discovery-infra/download_logs.py $(REMOTE_SERVICE_URL) $(LOGS_DEST) --cluster-id $(CLUSTER_ID) --must-gather
```
YuviGold, line 315: must-gather shouldn't always be provided. Only for `download_all_logs`

Jenkinsfile.download_all_logs
```diff
@@ -0,0 +1,45 @@
+pipeline {
```
YuviGold, line 1: I don't think we should create another pipeline

YuviGold:
> @oshercc, what about controller logs, installer gather?
> I think that the paremater should be something like: "take only metadata and events log"

IMO collecting only events and metadata is the right approach. It's very lightweight and it doesn't arise storage concerns.

Jenkinsfile.download_all_logs
oshercc: there is a different pipeline, 
renamed the old one to download_failed_cluster_logs and added Jenkinsfile.download_all_logs

Makefile
oshercc: now download_logs  collects only the failed logs with must-gather
so this job is not supposed to change.

added a second pipeline that collects all logs that will not collect must-gather.

ErezAlster:
>> @oshercc, what about controller logs, installer gather?
>> I think that the paremater should be something like: "take only metadata and events log"
>
>IMO collecting only events and metadata is the right approach. It's very lightweight and it doesn't arise storage concerns.

I agree. only saying that going to the white list approach is better than what not to do.

Jenkinsfile.download_all_logs
YuviGold: Why do we need two pipelines? It can be the same one with different parameters.

Makefile
YuviGold: I would prefer a way to quickly download logs without must-gather.
What do you say about:

```makefile
_download_logs:
	discovery-infra/download_logs.py $(REMOTE_SERVICE_URL) $(LOGS_DEST) --cluster-id $(CLUSTER_ID) $(ADDITIONAL_PARAMS)

download_logs:
	skipper make $(SKIPPER_PARAMS) _download_logs $(ADDITIONAL_PARAMS)

download_all_logs:
	skipper make $(SKIPPER_PARAMS) _download_all_logs ADDITIONAL_PARAMS="'--download-all --must-gather  ${ADDITIONAL_PARAMS}'"
```

Jenkinsfile.download_logs
```diff
@@ -3,6 +3,9 @@ pipeline {
 
     parameters {
         string(name: 'REMOTE_SERVICE_URL', defaultValue: 'https://api.openshift.com', description: 'Service URL')
+        string(name: 'DOWNLOAD_ALL', defaultValue: '--download-all', description: 'Collect all logs (not only from failed clusters)')
```
YuviGold, line 6: Change both `DOWNLOAD_ALL` and `MUST_GATHER` to boolean parameters.
If this parameter was ticked - add the matching flag.

Default `DOWNLOAD_ALL` is `false`, while `MUST_GATHER` is `true`

Makefile
```diff
@@ -312,16 +312,16 @@ delete_all_virsh_resources: destroy_all_nodes delete_minikube kill_all_port_forw
...
 
 _download_logs:
-	discovery-infra/download_logs.py $(REMOTE_SERVICE_URL) $(LOGS_DEST) --cluster-id $(CLUSTER_ID)
+	discovery-infra/download_logs.py $(REMOTE_SERVICE_URL) $(LOGS_DEST) --cluster-id $(CLUSTER_ID) $(ADDITIONAL_PARAMS)
 
 download_logs:
-	skipper make $(SKIPPER_PARAMS) _download_logs
+	skipper make $(SKIPPER_PARAMS) _download_logs $(ADDITIONAL_PARAMS)
 
 _download_all_logs:
```
YuviGold, line 320: You can delete `_download_all_logs` and call from `download_all_logs` to `_download_logs` directly.

YuviGold (approved):

YuviGold (approved):

oshercc:
retest

YuviGold (approved):

PR #384, opened 2021-01-04
Change DEPLOY_TARGET podman-localhost to onprem

YuviGold:
/cc @rwsu

### DIFF
--- a/discovery-infra/test_infra/utils.py
+++ b/discovery-infra/test_infra/utils.py
@@ -517,7 +517,7 @@ def get_remote_assisted_service_url(oc, namespace, service, scheme):
 
 
 def get_local_assisted_service_url(profile, namespace, service, deploy_target):
-    if deploy_target == "podman-localhost":
+    if deploy_target in ["onprem", "podman-localhost"]:
         assisted_hostname_or_ip = os.environ["ASSISTED_SERVICE_HOST"]
         return f'http://{assisted_hostname_or_ip}:8090'
     elif deploy_target == "ocp":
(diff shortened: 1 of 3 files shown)


discovery-infra/test_infra/utils.py
```diff
@@ -517,7 +517,7 @@ def get_remote_assisted_service_url(oc, namespace, service, scheme):
 
 
 def get_local_assisted_service_url(profile, namespace, service, deploy_target):
-    if deploy_target == "podman-localhost":
+    if deploy_target in ["onprem", "podman-localhost"]:
```
rwsu, line 520: I originally named the target "podman-localhost" to be more specific as to how assisted-service is deployed. "onprem" is a bit of an overloaded term. But if we want to rename "podman-localhost" to "onprem" that is fine. If it is renamed to "onprem", do we still want to keep "podman-localhost" around?

rwsu (changes requested):

discovery-infra/test_infra/utils.py
YuviGold: I think this naming isn't clear, while we use `onprem` reference anywhere else including in the [assisted-service code](https://github.com/openshift/assisted-service/blob/36b517f589451192591948bf60e7f8537e62b1dc/Makefile#L345) as DEPLOY_TARGET.

Just in order to keep the CI works. It has to be done in two stages (support both, then remove the old one.)

rwsu (approved):
/lgtm

PR #385, opened 2021-01-05
Collecting journalctl from nodes

nshidlin:
In cases where a test fails before the nodes register to the service or in cases of connectivity issues with the machine running the tests collecting journalctl from the vms can provide helpful insight into the failure

### DIFF
--- a/discovery-infra/tests/base_test.py
+++ b/discovery-infra/tests/base_test.py
@@ -5,6 +5,7 @@ import os
 from contextlib import suppress
 from typing import Optional
 from pathlib import Path
+from paramiko import SSHException
 
 from test_infra import consts
 import test_infra.utils as infra_utils
@@ -202,15 +203,14 @@ class BaseTest:
     def _collect_journalctl(self, nodes, log_dir_name):
         logging.info('Collecting journalctl\n')
         infra_utils.recreate_folder(log_dir_name, with_chmod=False ,force_recreate=False)
-        journal_ctl_path = Path(log_dir_name)
+        journal_ctl_path = Path(log_dir_name) / 'nodes_journalctl'
         infra_utils.recreate_folder(journal_ctl_path, with_chmod=False)
         for node in nodes:
             try:
-                journal = node.run_command(f'sudo journalctl')
+                node.run_command(f'sudo journalctl >> /tmp/{node.name}-journalctl')
                 journal_path = journal_ctl_path / node.name
-                with open(journal_path, 'w') as _file:
-                    _file.write(journal)
-            except:
+                node.download_file(f'/tmp/{node.name}-journalctl', str(journal_path))
+            except (RuntimeError, TimeoutError, SSHException):
                 logging.info(f'Could not collect journalctl for {node.name}')
 
     @staticmethod
--- a/discovery-infra/tests/base_test.py
+++ b/discovery-infra/tests/base_test.py
@@ -4,6 +4,7 @@ import json
 import os
 from contextlib import suppress
 from typing import Optional
+from pathlib import Path
 
 from test_infra import consts
 import test_infra.utils as infra_utils
@@ -166,6 +167,7 @@ class BaseTest:
             download_logs(api_client, cluster_details, log_dir_name, test.result_call.failed)
         if test.result_call.failed:
             self._collect_virsh_logs(nodes, log_dir_name)
+            self._collect_journalctl(nodes, log_dir_name)
 
     def _collect_virsh_logs(self, nodes, log_dir_name):
         logging.info('Collecting virsh logs\n')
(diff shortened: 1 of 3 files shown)


discovery-infra/tests/base_test.py
```diff
@@ -180,6 +181,20 @@ def _collect_virsh_logs(self, nodes, log_dir_name):
         infra_utils.run_command(f"journalctl --since \"{nodes.setup_time}\" "
                                 f"-u libvirtd -D /run/log/journal >> {libvird_log_path}", shell=True)
 
+    def _collect_journalctl(self, nodes, log_dir_name):
+        logging.info('Collecting journalctl\n')
+        os.makedirs(log_dir_name, exist_ok=True)
+        journal_ctl_path = os.path.join(log_dir_name, "journalctl")
+        os.makedirs(journal_ctl_path, exist_ok=False)
```
YuviGold, line 188: The first directory is OK to exist but its subdirectory not?
You should expect a raise.. What is the behavior you expect? I would suggest to delete the `journal_ctl_path` dir in case it exists and you don't want it to exists.

discovery-infra/tests/base_test.py
```diff
@@ -180,6 +181,20 @@ def _collect_virsh_logs(self, nodes, log_dir_name):
...
+        for node in nodes:
+            try:
+                journal = node.run_command(f'sudo journalctl')
+                journal_path = os.path.join(journal_ctl_path, node.name)
+                with open(journal_path, 'w') as _file:
+                    _file.write(journal)
+            except:
```
YuviGold, line 195: Please be explicit. I guess in the case of `AttributeError` or `IndexError` you would want to crash in order to fix that failure.
Did you mean `paramiko.SSHException`?

discovery-infra/tests/base_test.py
```diff
@@ -180,6 +181,20 @@ def _collect_virsh_logs(self, nodes, log_dir_name):
...
 
+    def _collect_journalctl(self, nodes, log_dir_name):
+        logging.info('Collecting journalctl\n')
+        os.makedirs(log_dir_name, exist_ok=True)
+        journal_ctl_path = os.path.join(log_dir_name, "journalctl")
+        os.makedirs(journal_ctl_path, exist_ok=False)
+        for node in nodes:
+            try:
+                journal = node.run_command(f'sudo journalctl')
+                journal_path = os.path.join(journal_ctl_path, node.name)
```
YuviGold, line 192: Let's start to use a more mordern way using `pathlib`.

```python
from pathlib import Path
...
    journal_path = Path(journal_ctl_path) / node.name
    with journal_path.open('w') as _file:
        _file.write(journal)
```

discovery-infra/tests/base_test.py
nshidlin: In the case the node is inactive a generic exception is raised:
https://github.com/openshift/assisted-test-infra/blob/master/discovery-infra/test_infra/controllers/node_controllers/node.py#L60

Also in the case where the node does not respond:
https://github.com/openshift/assisted-test-infra/blob/master/discovery-infra/test_infra/controllers/node_controllers/ssh.py#L55

YuviGold: So we should change these `Exception`s as well to something more relevant. Like `RuntimeError` and `TimeoutError`

romfreiman:
Dont we prefer to do it from the system rather that from the infra?

nshidlin:
@romfreiman a new set of vms is set up and torn down for each test. Collecting these as part of the test teardown is  our only option

YuviGold (approved):
LGTM

discovery-infra/tests/base_test.py
```diff
@@ -180,6 +183,19 @@ def _collect_virsh_logs(self, nodes, log_dir_name):
...
+    def _collect_journalctl(self, nodes, log_dir_name):
+        logging.info('Collecting journalctl\n')
+        infra_utils.recreate_folder(log_dir_name, with_chmod=False ,force_recreate=False)
+        journal_ctl_path = Path(log_dir_name) / 'nodes_journalctl'
+        infra_utils.recreate_folder(journal_ctl_path, with_chmod=False)
+        for node in nodes:
+            try:
+                node.run_command(f'sudo journalctl >> /tmp/{node.name}-journalctl')
+                journal_path = journal_ctl_path / node.name
+                node.download_file(f'/tmp/{node.name}-journalctl', journal_path)
```
YuviGold, line 195: I think this should be `str(journal_path)` since `download_file` expects `str` and not a `Path` object. 
Sorry for making everything complex.. :)

nshidlin: hmm, not sure how this worked

changing

YuviGold (approved):
LGTM

PR #386, opened 2021-01-05
Fixing reboot issue

iovadia:
/cc @YuviGold @lalon4 @nshidlin 

Add wait for stages:
   one including reboot stage
   and second is passed reboot stage

### DIFF
--- a/discovery-infra/test_infra/consts.py
+++ b/discovery-infra/test_infra/consts.py
@@ -71,6 +71,12 @@ class HostsProgressStages:
     DONE = "Done"
 
 
+all_host_stages = [HostsProgressStages.START_INSTALLATION, HostsProgressStages.INSTALLING,
+                   HostsProgressStages.WRITE_IMAGE_TO_DISK, HostsProgressStages.WAIT_FOR_CONTROL_PLANE,
+                   HostsProgressStages.REBOOTING, HostsProgressStages.WAIT_FOR_IGNITION,
+                   HostsProgressStages.CONFIGURING, HostsProgressStages.JOINED, HostsProgressStages.DONE]
+
+
 class Events:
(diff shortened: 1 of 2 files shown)


discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -250,6 +250,18 @@ def wait_for_hosts_passed_reboot(self, nodes_count=env_variables['num_nodes']):
             nodes_count=nodes_count
         )
 
+    def wait_for_hosts_passed_reboot(self, nodes_count=env_variables['num_nodes']):
```
YuviGold, line 253: Let's make the previous function with another parameter `hasStagePassed: bool = False` instead of having two functions.

discovery-infra/test_infra/consts.py
```diff
@@ -27,6 +27,11 @@
 DEFAULT_ADDITIONAL_NTP_SOURCE = "clock.redhat.com"
 
 
+all_host_stages = ['START_INSTALLATION', 'INSTALLING', 'WRITE_IMAGE_TO_DISK',
```
YuviGold, line 30: You should declare it after the `HostsProgressStages` class and use its values.
i.e.
```python
HOSTS_STAGES = [HostsProgressStages.START_INSTALLATION, HostsProgressStages.INSTALLING, ...]
```

We can even do a smarter usage of the `enum` class and convert the class into an enum, then iterating upon it.

```python
    from enum import Enum
    class HostsProgressStages(Enum):
        ...

    stages_values = list(map(lambda stage: stage.value ,HostsProgressStages))
    index = stages_values.index(stage)
    ....
```

But it might break other functions use this class.

YuviGold (approved):

PR #387, opened 2021-01-05
enable a fix_uid.sh script for prow tests

osherdp:
https://github.com/openshift/release/pull/14552#discussion_r551833515

/cc @YuviGold


YuviGold:
I honestly don't think we should add it.

osherdp:
> I honestly don't think we should add it.

Maybe. But I cannot know why if you won't elaborate... :)

YuviGold:
> > I honestly don't think we should add it.
> 
> Maybe. But I cannot know why if you won't elaborate... :)

Isn't that obvious? ;-)

PR #389, opened 2021-01-05
MGMT-3504 Move proxy_server fixture to base_test

ori-amizur:
- Add port 3129 to firewall
/cc @nshidlin 
/cc @YuviGold

### DIFF
--- a/discovery-infra/tests/base_test.py
+++ b/discovery-infra/tests/base_test.py
@@ -8,6 +8,7 @@ from typing import Optional
 from test_infra import consts
 import test_infra.utils as infra_utils
 from test_infra.tools.assets import NetworkAssets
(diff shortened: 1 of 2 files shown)


discovery-infra/tests/base_test.py
```diff
@@ -116,6 +117,22 @@ def attach(node, disk_size):
         if modified_node is not None:
             modified_node.detach_all_test_disks()
 
+    @pytest.fixture()
+    def proxy_server(self):
```
tsorya, line 121: just to be sure. Commit is called "move  proxy_server fixture to base_test"  but i don't see any deleted fixture. What i am missing?

nshidlin: it's moved over from the ds repo

YuviGold (approved):

PR #390, opened 2021-01-05
Reducing tests timeout

lalon4:
In some cases tests fails due to a real bug or environment issue. Currently there is a generic TIMEOUT which causes to wait 1h in some failures. 
This PR is to avoid waiting too much time


discovery-infra/test_infra/consts.py
```diff
@@ -9,7 +9,13 @@
 NODES_REGISTERED_TIMEOUT = 600
 CLUSTER_INSTALLATION_TIMEOUT = 60 * 60   # 60 minutes
 START_CLUSTER_INSTALLATION_TIMEOUT = 360
+INSTALLING_IN_PROGRESS_TIMEOUT = 60 * 10
 VALIDATION_TIMEOUT = 6 * 60
+DISCOVERY_TIMEOUT = 60 * 15
```
nshidlin, line 14: ` DISCOVERY_TIMEOUT`  seems redundant with `NODES_REGISTERED_TIMEOUT` wdyt?

lalon4: Agree. removing `DISCOVERY_TIMEOUT` and raising `INSTALLING_IN_PROGRESS_TIMEOUT` to 15.

YuviGold (approved):

PR #391, opened 2021-01-06
MGMT-3408 - adapt test-infra to allow static ip configuration flow

yevgeny-shnaidman:


Makefile
```diff
@@ -261,10 +262,19 @@ deploy_nodes_with_install:
...
+	$(bash scripts/utils.sh local_setup_before_deployment $(PLATFORM) $(NAMESPACE) $(OC_FLAG))
+	skipper make $(SKIPPER_PARAMS) _deploy_nodes NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE) $(OC_FLAG)) NAMESPACE=$(NAMESPACE) ADDITIONAL_PARAMS="'-in -sic $(STATIC_IPS_CONFIG)'" $(SKIPPER_PARAMS) DAY1_PARAMS=--day1-cluster
+
 deploy_nodes:
 	$(bash scripts/utils.sh local_setup_before_deployment $(PLATFORM) $(NAMESPACE) $(OC_FLAG))
 	skipper make $(SKIPPER_PARAMS) _deploy_nodes NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE) $(OC_FLAG)) NAMESPACE=$(NAMESPACE) DAY1_PARAMS=--day1-cluster
 
+deploy_static_ips_nodes:
+	$(bash scripts/utils.sh local_setup_before_deployment $(PLATFORM) $(NAMESPACE) $(OC_FLAG))
+	skipper make $(SKIPPER_PARAMS) _deploy_nodes NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE) $(OC_FLAG)) NAMESPACE=$(NAMESPACE) DAY1_PARAMS=--day1-cluster ADDITIONAL_PARAMS="'-sic $(STATIC_IPS_CONFIG)'"
```
tsorya, line 276: can't we just set ADDITIONAL_PARAMS="'-sic $(STATIC_IPS_CONFIG)'" and run deploy_nodes ?

discovery-infra/start_discovery.py
```diff
@@ -446,6 +458,28 @@ def set_hosts_roles(client, cluster, nodes_details, machine_net, tf, master_coun
                        update_roles=master_count > 1)
 
 
+def parse_static_ips_config(config_file):
+    with open(config_file) as _file:
+        res = json.load(_file)
+        return res
+
+
+def set_hostames_for_static_ips_hosts(client, cluster_id):
```
tsorya, line 467: we have this logic for ipv6 , please use it

tsorya: we don't really need specific name, just need to set ip per current hostname right?

Makefile
yevgeny-shnaidman: i wanted a different target, so that user does not needs to know what additional parameters to set, more convenient

static_ips_config.json
```diff
@@ -0,0 +1,44 @@
+[
+  {
+    "mac": "52:54:00:c2:f4:4d",
```
tsorya, line 3: can we generate it in the code? just get list of ips generate some random static mac and to +1 to every vm?

discovery-infra/start_discovery.py
yevgeny-shnaidman: fixed

static_ips_config.json
yevgeny-shnaidman: fixed

Makefile
tsorya: i mean if this target can call deploy_nodes and just provide it's own ADDITIONAL_PARAMS, without duplicating same code

discovery-infra/start_discovery.py
```diff
@@ -149,6 +158,45 @@ def _secondary_tfvars(master_count, nodes_details, machine_net):
         }
 
 
+def _generate_static_ips_data(num_masters, num_workers, machine_net):
```
tsorya, line 161: can we move it to , static_ip_tools.py ?

Makefile
yevgeny-shnaidman: this is what it does, look at all the other targets.  ADDITIONAL_PARAMS, Namespace and OC_FLAG

discovery-infra/start_discovery.py
yevgeny-shnaidman: fixed

Makefile
yevgeny-shnaidman: fixed

discovery-infra/start_discovery.py
```diff
@@ -436,7 +451,11 @@ def set_hosts_roles(client, cluster, nodes_details, machine_net, tf, master_coun
     if machine_net.has_ip_v4:
         libvirt_nodes = utils.get_libvirt_nodes_mac_role_ip_and_name(networks_names[0])
         libvirt_nodes.update(utils.get_libvirt_nodes_mac_role_ip_and_name(networks_names[1]))
-        update_hostnames = False
+        if static_ip_mode:
+            log.info("Setting hostnames when running in static ips mode")
+            update_hostnames = True
+        else:
+            update_hostnames = True
```
empovit, line 458: Looks like a bug. You're setting `update_hostname = True` in both cases.

discovery-infra/start_discovery.py
```diff
@@ -618,6 +647,11 @@ def main():
     parser.add_argument(
         "-ps", "--pull-secret", help="Pull secret", type=str, default=""
     )
+    parser.add_argument(
```
empovit, line 650: Probably the documentation is worth updating with this flag and an example of setting configuring static IPs.

discovery-infra/start_discovery.py
yevgeny-shnaidman: yes, i will fix it as part of MGMT-3568

discovery-infra/test_infra/tools/static_ips.py
```diff
@@ -0,0 +1,43 @@
...
+                    (workers_macs, workers_static_ips, mask, gw_dns, num_workers),
+                    (workers_secondary_macs, workers_static_secondary_ips, mask_secondary, gw_dns_secondary, num_workers)]:
+        data = [{'mac': netdata[0][i], 'ip': netdata[1][i], 'mask': netdata[2], 'gateway': netdata[3], 'dns': netdata[3]} for i in range(netdata[4])]
+        static_ips = static_ips + data
+
+    return static_ips,(masters_macs, masters_secondary_macs, workers_macs, workers_secondary_macs)
+
+
+def _generate_macs(count):
+    return ["02:00:00:%02x:%02x:%02x" % (random.randint(0, 255), random.randint(0, 255), random.randint(0, 255)) for x in range(count)]
```
empovit, line 43: There's a small chance that the same MAC address will be generated twice. And it will be very difficult to troubleshoot. WDYS if I change one of the octets (e.g. the last) to sequential (and throw an error if count > 256).

yevgeny-shnaidman: i think that the chance of the same MAC address is very, very small : (1/255)^3, even taking into account that randint is pseudo random.  And in case it does happen it will be pretty straight forward , since VM will simply refuse to boot

empovit: Are you sure it won't boot? And it will be clear why? I suspect the VMs will boot, but we'll see weird networking issues. Maybe worth verifying.

PR #399, opened 2021-01-07
MGMT-3474 - Add e2e test to oc must-gather

slaviered:


YuviGold (changes requested):
Apparently a bad rebase.
* `openshift_version` was added.
* `cluster_machince_cidr` should be called from the `controller`.
* `set_hostnames` should be called to fix IPv6 hostnames.

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -237,6 +233,15 @@ def wait_for_non_bootstrap_masters_to_reach_configuring_state_during_install(sel
             nodes_count=env_variables['num_masters']-1
         )
 
+    def wait_for_non_bootstrap_masters_to_reach_joined_state_during_install(self):
```
lalon4, line 236: Why not making this method more generic by calling it `wait_for_hosts_to_reach_joined`, and accept `nodes_count` as a parameter?

YuviGold (approved):

PR #400, opened 2021-01-07
Adding support for single node in assisted service flow

eranco74:
Mo need to wait for insufficient in case of single node


osherdp (approved):
/lgtm

PR #401, opened 2021-01-07
[WIP] MGMT-3554: fix api-vip update to refresh state of terraform

osherdp:


osherdp:
there's a better option: #402

PR #415, opened 2021-01-11
Setting api_vip for single node and none platform.

tsorya:
It will allow to download logs and wait for cvo.

### DIFF
--- a/discovery-infra/download_logs.py
+++ b/discovery-infra/download_logs.py
@@ -18,7 +18,7 @@ from logger import log
 from test_infra.assisted_service_api import InventoryClient, create_client
 from test_infra.consts import ClusterStatus
 from test_infra.logs_utils import verify_logs_uploaded
-from test_infra.utils import config_etc_hosts, recreate_folder, run_command
+from test_infra.utils import config_etc_hosts, recreate_folder, run_command, get_api_vip_from_cluster
 
 TIME_FORMAT = '%Y-%m-%d_%H:%M:%S'
 MAX_RETRIES = 3
@@ -105,7 +105,7 @@ def download_logs(client: InventoryClient, cluster: dict, dest: str, must_gather
 
             if must_gather:
                 recreate_folder(os.path.join(output_folder, "must-gather"))
-                config_etc_hosts(cluster['name'], cluster['base_dns_domain'], cluster['api_vip'])
+                config_etc_hosts(cluster['name'], cluster['base_dns_domain'], get_api_vip_from_cluster(cluster))
                 download_must_gather(kubeconfig_path, os.path.join(output_folder, "must-gather"))
     finally:
         run_command(f"chmod -R ugo+rx '{output_folder}'")
--- a/discovery-infra/test_infra/tools/connection.py
+++ b/discovery-infra/test_infra/tools/connection.py
@@ -1,0 +1,17 @@
+import socket
+import logging
+
+
+def is_open(ip, port, timeout=5):
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    s.settimeout(timeout)
+    try:
+        s.connect((ip, int(port)))
+        s.shutdown(socket.SHUT_RDWR)
+        logging.info("Ip %s with port %s are reachable", ip, port)
+        return True
+    except:
+        logging.info("Ip %s with port %s are not reachable", ip, port)
+        return False
+    finally:
+        s.close()
--- a/discovery-infra/test_infra/utils.py
+++ b/discovery-infra/test_infra/utils.py
@@ -16,17 +16,21 @@ from pathlib import Path
 from functools import wraps
 from contextlib import contextmanager
 import xml.dom.minidom as md
+from typing import Dict
 
 import libvirt
 import waiting
 import requests
 import filelock
 from test_infra import consts
+from test_infra.tools import connection
 import oc_utils
 from logger import log
 from retry import retry
 from pprint import pformat
 from distutils.dir_util import copy_tree
+from assisted_service_client import models
+
 
 conn = libvirt.open("qemu:///system")
 
@@ -124,12 +128,13 @@ def is_cvo_available():
         conditions = json.loads(res)['items'][0]['status']['conditions']
         for condition in conditions:
             log.info(
-                f"CVO condition <{condition['type']}> status is <{condition['status']}>, because: {condition['message']}")
+                f"CVO condition <{condition['type']}> status is <{condition['status']}>, "
+                f"because: {condition.get('message')}")
 
             if condition['type'] == 'Available' and condition['status'] == 'True':
                 return True
     except:
-        log.info("exception in access the cluster api server")
+        log.exception("exception in access the cluster api server")
     return False
 
 
@@ -659,10 +664,12 @@ def create_ip_address_nested_list(node_count, starting_ip_addr):
 def create_empty_nested_list(node_count):
     return [[] for i in range(node_count)]
 
+
 def get_libvirt_nodes_from_tf_state(network_names, tf_state):
     nodes = extract_nodes_from_tf_state(tf_state, network_names, consts.NodeRoles.MASTER)
     nodes.update(extract_nodes_from_tf_state(tf_state, network_names, consts.NodeRoles.WORKER))
     return nodes
+
 
 def extract_nodes_from_tf_state(tf_state, network_names, role):
     domains = next(r["instances"] for r in tf_state.resources if r["type"] == "libvirt_domain" and r["name"] == role)
@@ -673,7 +680,7 @@ def extract_nodes_from_tf_state(tf_state, network_names, role):
             if nic["network_name"] not in network_names:
                 continue
 
-            data[nic["mac"]] =  {"ip": nic["addresses"], "name": d["attributes"]["name"], "role": role}
+            data[nic["mac"]] = {"ip": nic["addresses"], "name": d["attributes"]["name"], "role": role}
 
     return data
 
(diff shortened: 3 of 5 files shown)


discovery-infra/test_infra/utils.py
```diff
@@ -801,13 +808,57 @@ def update_hosts(client, cluster_id, libvirt_nodes, update_hostnames=False, upda
...
     log.info(f'{response.stdout}')
     return response.stdout
+
+
+def get_network_interface_ip(interface):
+    if len(interface.ipv4_addresses) > 0:
+        return interface.ipv4_addresses[0].split("/")[0]
+    if len(interface.ipv6_addresses) > 0:
+        return interface.ipv6_addresses[0].split("/")[0]
+    return " "
```
YuviGold, line 830: raise an error or return None.. Dont return a space!

discovery-infra/test_infra/utils.py
```diff
@@ -801,13 +808,57 @@ def update_hosts(client, cluster_id, libvirt_nodes, update_hostnames=False, upda
...
+    if cluster.get("api_vip"):
+        logging.info("API VIP is set: %s", cluster["api_vip"])
+        return cluster["api_vip"]
+    # workaround for MGMT-3583
+    if cluster.get("user_managed_networking") or cluster.get("user-managed-networking"):
+        logging.info("API VIP is not set, taking masters reachable ip")
+        masters = get_masters_from_cluster(cluster)
+        ips = get_inventory_host_reachable_ips(masters[0], 443)
+        logging.info("master %s first reachable ip is %s", masters[0], ips[0])
+        return ips[0]
```
YuviGold, line 864: I'm not sure this is a valid way to get the API VIP.
You must search for 'api' interface by name. Other than that, you're right it could be any master.

discovery-infra/test_infra/utils.py
```diff
@@ -801,13 +808,57 @@ def update_hosts(client, cluster_id, libvirt_nodes, update_hostnames=False, upda
...
+    ips = get_inventory_host_ips_data(host)
+    return [ip for ip in ips if connection.is_open(ip, port)]
+
+
+def get_masters_from_cluster(cluster: dict):
+    return [host for host in cluster["hosts"] if host["role"] == consts.NodeRoles.MASTER]
+
+
+def get_api_vip_from_cluster(cluster: dict):
+    if cluster.get("api_vip"):
```
YuviGold, line 855: `else` is missing

discovery-infra/test_infra/utils.py
```diff
@@ -801,13 +808,57 @@ def update_hosts(client, cluster_id, libvirt_nodes, update_hostnames=False, upda
...
     response = subprocess.run(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
     if response.returncode != 0:
         log.error(f'failed to get controller status: {response.stderr}')
         return b''
 
     log.info(f'{response.stdout}')
     return response.stdout
+
+
+def get_network_interface_ip(interface):
```
YuviGold, line 825: What type is `interface`? Especially adding this typehint would help to understand while it seems like an object with specific attributes

discovery-infra/test_infra/tools/connection.py
```diff
@@ -0,0 +1,17 @@
...
+def is_open(ip, port, timeout=5):
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    s.settimeout(timeout)
+    try:
+        s.connect((ip, int(port)))
+        s.shutdown(socket.SHUT_RDWR)
+        logging.info("Ip %s with port %s are reachable", ip, port)
+        return True
+    except:
+        logging.info("Ip %s with port %s are not reachable", ip, port)
```
YuviGold, line 14: I don't think that this utility should have logs inside. But anyway please transform logs inside the `except` clause to be of type `logging.exception` so the exception won't get lost.

discovery-infra/test_infra/tools/connection.py
```diff
@@ -0,0 +1,17 @@
...
+        return False
+    finally:
+        s.close()
```
YuviGold, line 17: Instead of having this `finally` clause you can use `with` statement for the `socket` object

discovery-infra/test_infra/tools/connection.py
```diff
@@ -0,0 +1,17 @@
...
+
+def is_open(ip, port, timeout=5):
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    s.settimeout(timeout)
+    try:
+        s.connect((ip, int(port)))
+        s.shutdown(socket.SHUT_RDWR)
+        logging.info("Ip %s with port %s are reachable", ip, port)
+        return True
+    except:
```
YuviGold, line 13: Set specific exceptions

discovery-infra/download_logs.py
```diff
@@ -105,7 +105,7 @@ def download_logs(client: InventoryClient, cluster: dict, dest: str, must_gather
 
             if must_gather:
                 recreate_folder(os.path.join(output_folder, "must-gather"))
-                config_etc_hosts(cluster['name'], cluster['base_dns_domain'], cluster['api_vip'])
+                config_etc_hosts(cluster['name'], cluster['base_dns_domain'], get_api_vip_from_cluster(cluster))
```
YuviGold, line 108: Isn't `cluster['api']` must be defined correctly when we ask to download the logs? Otherwise, the cluster itself is misconfigured.

discovery-infra/test_infra/utils.py
```diff
@@ -801,13 +808,57 @@ def update_hosts(client, cluster_id, libvirt_nodes, update_hostnames=False, upda
...
+
+def get_api_vip_from_cluster(cluster: dict):
+    if cluster.get("api_vip"):
+        logging.info("API VIP is set: %s", cluster["api_vip"])
+        return cluster["api_vip"]
+    # workaround for MGMT-3583
+    if cluster.get("user_managed_networking") or cluster.get("user-managed-networking"):
+        logging.info("API VIP is not set, taking masters reachable ip")
+        masters = get_masters_from_cluster(cluster)
+        ips = get_inventory_host_reachable_ips(masters[0], 443)
```
YuviGold, line 862: Please use constant and add a comment why the first master is being chosen. (I know all of them should have the same vip derived from the cluster id. But please add a comment describing this behavior.

YuviGold (changes requested):

discovery-infra/test_infra/utils.py
tsorya: right. copied this code and missed it

discovery-infra/test_infra/utils.py
tsorya: how u know which interface is api vip? If u have an idea will be glad to implement

discovery-infra/test_infra/tools/connection.py
tsorya: no reason for it to be logging.exception. It is on purpose info

discovery-infra/test_infra/tools/connection.py
tsorya: agree

discovery-infra/test_infra/tools/connection.py
tsorya: no need, it is fine to fall on any error, don't want to handle here any logic

discovery-infra/download_logs.py
tsorya: no, in None platform or single node. U don't have ip for vip

discovery-infra/test_infra/utils.py
tsorya: what constant u want?

discovery-infra/test_infra/utils.py
YuviGold: It has the name `api` since that's how it is being generated.
i.e.
```
4: api@ens3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP mode DEFAULT group default 
    link/ether 00:1a:4a:54:c6:3d brd ff:ff:ff:ff:ff:ff
5: ingress@ens3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP mode DEFAULT group default 
    link/ether 00:1a:4a:46:45:d8 brd ff:ff:ff:ff:ff:ff
```

discovery-infra/test_infra/tools/connection.py
YuviGold: I think it is useful when there's an exception to log that exception.  Errors should never pass silently.

discovery-infra/test_infra/tools/connection.py
YuviGold: If you provided something strange, like a wrong name `NameError` would be raised. Other examples like `RecursionError`, `RunTimeError`, `IndexError`.. - All of those you want to know about.
You should explicitly choose the accepted exceptions. I would suggest even to take the whole socket errors family `socket.error` (`socket.timeout` is a subclass of `socket.error`)

discovery-infra/test_infra/utils.py
tsorya: else it will return None

discovery-infra/test_infra/utils.py
tsorya: u don't have api vip in None platform

discovery-infra/test_infra/utils.py
YuviGold: `http.client.HTTPS_PORT`

discovery-infra/download_logs.py
YuviGold: You don't an API virtual IP or you don't have an API endpoint at all?
DHCP mode isn't running on none-platform and single-node modes?

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -637,3 +641,63 @@ def wait_for_event(self, event_to_find, reference_time, params_list=[], host_id=
...
+        return api_vip
+
+    # validate if kube-api is ready on given address
+    @staticmethod
+    def is_kubeapi_service_ready(ip_or_dns):
+        try:
+            response = requests.get(f'https://{ip_or_dns}:6443/readyz', verify=False)
+            if response.status_code == 200:
+                return True
+        except:
```
YuviGold, line 691: set explicit exceptions

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -637,3 +641,63 @@ def wait_for_event(self, event_to_find, reference_time, params_list=[], host_id=
...
+            masters = self.get_hosts_by_role(consts.NodeRoles.MASTER, hosts=cluster.to_dict()["hosts"])
+            api_vip = self.get_kube_api_ip(hosts=masters)
+
+        logging.info("api vip is %s", api_vip)
+        return api_vip
+
+    # validate if kube-api is ready on given address
+    @staticmethod
+    def is_kubeapi_service_ready(ip_or_dns):
+        try:
```
YuviGold, line 687: use `contextlib.suprress(<explicit_exceptions>)`

YuviGold (approved):

PR #417, opened 2021-01-11
Adding enable host method

lalon4:
Adding NTP validation timeout, and removing redundant PENDING_FOR_INPUT const


discovery-infra/test_infra/consts.py
```diff
@@ -89,7 +90,6 @@ class Events:
     GENERATED_IMAGE = "Generated image (SSH public key is set)"
     DOWNLOAD_IMAGE = "Started image download"
     HOST_REGISTERED_TO_CLUSTER = ": registered to cluster"
-    PENDING_FOR_INPUT = "pending-for-input"
```
YuviGold, line 92: This isn't being used anywhere?

lalon4: It is used by cluster and test events, but its redundant. So removed one of them
https://github.com/openshift/assisted-test-infra/blob/master/discovery-infra/test_infra/consts.py#L88
https://github.com/openshift/assisted-test-infra/blob/master/discovery-infra/test_infra/consts.py#L92

YuviGold (approved):

PR #418, opened 2021-01-12
MGMT-3584 Generate IPv6 static addresses

empovit:


discovery-infra/test_infra/tools/static_ips.py
```diff
@@ -12,50 +12,82 @@ def generate_macs(count):
...
 
-    mask = str(ipaddress.IPv4Network(machine_cidr).prefixlen)
-    mask_secondary = str(ipaddress.IPv4Network(provisioning_cidr).prefixlen)
-    gw_dns = str(ipaddress.ip_address(ipaddress.IPv4Network(machine_cidr).network_address) + 1)
-    gw_dns_secondary = str(ipaddress.ip_address(ipaddress.IPv4Network(provisioning_cidr).network_address) + 1)
+    if first_net.version == 6:
+        ipv6_gen = _static_conf_gen(num_nodes, first_net, address_offset)
+    else:
+        ipv6_gen = _static_conf_gen(num_nodes, ip_network(
+            cidrs[1]), address_offset) if len(cidrs) > 1 else _empty_conf_gen()
```
yevgeny-shnaidman, line 65: in case we have only ipv4 cidr machine/provisioning cidr , the  cidr[1] will throw an exception

empovit: `if len(cidrs) > 1`

discovery-infra/test_infra/tools/static_ips.py
```diff
@@ -12,50 +12,82 @@ def generate_macs(count):
...
+
+def _static_conf_gen(num_nodes, network, address_offset=0):
+    starting_ip = str(ip_address(network.network_address)
+                      + 30 + address_offset)
+    ips = utils.create_ip_address_list(num_nodes, starting_ip)
+    mask = str(network.prefixlen)
+    gw_dns = str(ip_address(network.network_address) + 1)
+
+    for i in range(num_nodes):
+        yield (ips[i], mask, gw_dns, gw_dns)
```
YuviGold, line 88: I don't like it so much.. The meaning of each returned value isn't clear, and afterward, you address them as indexes.. I would suggest using [namedtuple](https://docs.python.org/3/library/collections.html#collections.namedtuple) or even better a [dataclass](https://docs.python.org/3/library/dataclasses.html).

For example
```python
from dataclasses import dataclass
from typing import Optional

@dataclass 
class StaticIPConf: 
    ip: Optional[str] = None 
    mask: Optional[str] = None
    gateway: Optional[str] = None 
    dns: Optional[str] = None 
...

In [10]: StaticIPConf()                                                             
Out[10]: StaticIPConf(ip=None, mask=None, gateway=None, dns=None)

In [12]: StaticIPConf('asd')                                                        
Out[12]: StaticIP(ip='asd', mask=None, gateway=None, dns=None)

In [13]: StaticIPConf('asd').ip                                                     
Out[13]: 'asd'

```

Then `_empty_conf_gen` won't be needed.

discovery-infra/test_infra/tools/static_ips.py
```diff
@@ -12,50 +12,82 @@ def generate_macs(count):
...
+        num_nodes, first_net, address_offset) if first_net.version == 4 else _empty_conf_gen()
 
-    mask = str(ipaddress.IPv4Network(machine_cidr).prefixlen)
-    mask_secondary = str(ipaddress.IPv4Network(provisioning_cidr).prefixlen)
-    gw_dns = str(ipaddress.ip_address(ipaddress.IPv4Network(machine_cidr).network_address) + 1)
-    gw_dns_secondary = str(ipaddress.ip_address(ipaddress.IPv4Network(provisioning_cidr).network_address) + 1)
+    if first_net.version == 6:
+        ipv6_gen = _static_conf_gen(num_nodes, first_net, address_offset)
+    else:
+        ipv6_gen = _static_conf_gen(num_nodes, ip_network(
```
YuviGold, line 64: The new line is coming in a bad place. This isn't so clear. Should be
```python
ipv6_gen = _static_conf_gen(num_nodes, ip_network(cidrs[1]), address_offset) if len(cidrs) > 1 \
         else _empty_conf_gen()
```

discovery-infra/test_infra/tools/static_ips.py
```diff
@@ -12,50 +12,82 @@ def generate_macs(count):
...
-    workers_static_secondary_ips = utils.create_ip_address_list(num_workers, workers_static_secondary_starting_ip)
+    return static_ips
+
+
+def _genetate_ips(mac_addresses, cidrs, address_offset=0):
+
+    num_nodes = len(mac_addresses)
+
+    first_net = ip_network(cidrs[0])
+    ipv4_gen = _static_conf_gen(
```
YuviGold, line 58: The new line is coming in a bad place. It is difficult to understand what the "if" is referring to.  I gave an example of how to use it in the next comment.

discovery-infra/test_infra/tools/static_ips.py
```diff
@@ -12,50 +12,82 @@ def generate_macs(count):
...
+            'ip': ipv4_conf[0], 'mask': ipv4_conf[1], 'gateway': ipv4_conf[2], 'dns': ipv4_conf[3],
+            'ip_v6': ipv6_conf[0], 'mask_v6': ipv6_conf[1], 'gateway_v6': ipv6_conf[2], 'dns_v6': ipv6_conf[3]
+        })
 
     return static_ips
 
+
+def _static_conf_gen(num_nodes, network, address_offset=0):
+    starting_ip = str(ip_address(network.network_address)
+                      + 30 + address_offset)
```
YuviGold, line 82: set const `DEFAULT_STATIC_IP_OFFSET`

discovery-infra/test_infra/tools/static_ips.py
empovit: I agree it's not ideal, but that's what automatic formatting does. If I want to keep the file auto-formatted, I prefer not to override the formatting manually.

discovery-infra/test_infra/tools/static_ips.py
empovit: Doesn't make much sense. This value is, and will be, used exactly once.

YuviGold: It's fine that it is used only one. Just so our code won't have magic numbers.

discovery-infra/test_infra/tools/static_ips.py
```diff
@@ -7,55 +7,99 @@
...
+    return static_ips
+
+
+def _genetate_ips(mac_addresses, cidrs, address_offset):
+
+    num_nodes = len(mac_addresses)
+
+    first_net = ip_network(cidrs[0])
+    ipv4_gen = _static_conf_gen(num_nodes, first_net, address_offset) if first_net.version == 4 \
+        else _empty_conf_gen()
```
YuviGold, line 71:
```python
 else iter(lambda: StaticIPEntry(), 1)
```

discovery-infra/test_infra/tools/static_ips.py
```diff
@@ -7,55 +7,99 @@
 from test_infra import consts
 
 
+class StaticIPEntry(object):
```
YuviGold, line 10: Since Python3 all classes inherit from `object` by default.

discovery-infra/test_infra/tools/static_ips.py
```diff
@@ -7,55 +7,99 @@
...
+
+    def __init__(self, ip=None, mask=None, gateway=None, dns=None):
```
YuviGold, line 12: If you will use `dataclass` you won't need to declare a ctor

YuviGold (approved):

PR #421, opened 2021-01-12
MGMT-3600 Add makefile target to run IPv6 full flow

ori-amizur:
/cc @YuviGold


YuviGold (changes requested):
Did you neglect `VIP_DHCP_ALLOCATION` on purpose?
I think that it would be simplest if the default proxy URL and non proxy values could be inferred inside the code. Like we already define subnets, cidrs and stuff for IPv6. These should be included.
Then we don't need a new Makefile target. I just don't think I should know the URL&Proxy values while they might change in the future.
IPv6, IPv4, DHCP mode env vars are fine to be defined previously.

IMO it should be like that:
```shell
IPv6=yes IPv4=no VIP_DHCP_ALLOCATION=no PROXY=yes make run_full_flow_with_install
```

YuviGold (approved):

PR #422, opened 2021-01-12
MGMT-3065 Handle IPv6 addresses in day 2

empovit:


discovery-infra/day2.py
```diff
@@ -170,13 +170,21 @@ def get_network_nodes_from_terraform(tf_folder):
...
-    master_end_ip = tfvars[master_ip_type][-1][0]
-    workers_ip_list = tfvars[worker_ip_type]
-    if not len(workers_ip_list):
-        worker_starting_ip = ipaddress.ip_address(master_end_ip)
+
+    old_worker_ips_list = tfvars[worker_ip_type]
+
+    if tfvars[master_ip_type][-1]:
+        master_end_ip = tfvars[master_ip_type][-1][0]
+        if not len(old_worker_ips_list):
```
YuviGold, line 178:
```python
if not old_worker_ips_list:
```

discovery-infra/day2.py
```diff
@@ -170,13 +170,21 @@ def get_network_nodes_from_terraform(tf_folder):
...
 
 def set_workers_ips_by_type(tfvars, num_worker_nodes, master_ip_type, worker_ip_type):
-    master_end_ip = tfvars[master_ip_type][-1][0]
-    workers_ip_list = tfvars[worker_ip_type]
-    if not len(workers_ip_list):
-        worker_starting_ip = ipaddress.ip_address(master_end_ip)
+
+    old_worker_ips_list = tfvars[worker_ip_type]
+
+    if tfvars[master_ip_type][-1]:
```
YuviGold, line 176: What is index `-1`? Why was it chosen and not others?

YuviGold (changes requested):
Sorry, I'm unable to understand anything from this function. Please insert some comments.

discovery-infra/day2.py
empovit: Last from the end. Because that + 1 will serve as the beginning of the added IP range. Would `tfvars[master_ip_type][0]` be clearer?

YuviGold: Your explanation was good enough. Please add a comment about it (since you don't assign it to a variable, it just looks like you chose it arbitrarily for the check)

YuviGold (approved):

PR #424, opened 2021-01-13
MGMT-3417 - infra changes for static ips e2e test

yevgeny-shnaidman:


tsorya:
We need to add get ip functionality to e2e test in static ip mode, currently it will not work

yevgeny-shnaidman:
> We need to add get ip functionality to e2e test in static ip mode, currently it will not work

i will open another a Jira ticket that will implement this functionality in terraform controller based on tfvars and ping

PR #425, opened 2021-01-13
Adding delete_host method and get_workers nodes

lalon4:


discovery-infra/test_infra/assisted_service_api.py
```diff
@@ -189,6 +189,10 @@ def delete_cluster(self, cluster_id):
         log.info("Deleting cluster %s", cluster_id)
         self.client.deregister_cluster(cluster_id=cluster_id)
 
+    def delete_host(self, cluster_id, host_id):
```
nshidlin, line 192: Maybe should be `deregister_host`? wdty?

lalon4: isn't delete_host is more intuitive when:
new_cluster.delete_host(host)?
i mean, youll know to search deregister if you know that the API called deregister rather then delete
WDYT?

lalon4: ho its the api, so yeah it can be deregistered, changing

PR #426, opened 2021-01-13
Add advanced networking cluster methods

trewest:
Adding following methods to cluster.py:

`def set_advanced_networking` - Updates Cluster CIDR, Service CIDR, and Cluster Host Prefix under AI Advanced Networking
`def set_advanced_cluster_cidr` - Updates Cluster Network CIDR under AI Advanced Networking
`def set_advanced_service_cidr` - Updates Service Network CIDR under AI Advanced Networking
`def set_advanced_cluster_host_prefix` - Updates Cluster Network Host Prefix under AI Advanced Networking

### DIFF
--- a/discovery-infra/test_infra/helper_classes/cluster.py
+++ b/discovery-infra/test_infra/helper_classes/cluster.py
@@ -160,6 +160,23 @@ class Cluster:
     def set_base_dns_domain(self, base_dns_domain):
         logging.info(f"Setting base DNS domain:{base_dns_domain} for cluster: {self.id}")
         self.api_client.update_cluster(self.id, {"base_dns_domain": base_dns_domain})
+    
+    def set_advanced_networking(self, cluster_cidr, service_cidr, cluster_host_prefix):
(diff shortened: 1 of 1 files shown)


discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -160,6 +160,23 @@ def set_ssh_key(self, ssh_key):
     def set_base_dns_domain(self, base_dns_domain):
         logging.info(f"Setting base DNS domain:{base_dns_domain} for cluster: {self.id}")
         self.api_client.update_cluster(self.id, {"base_dns_domain": base_dns_domain})
+    
+    def set_advanced_networking(self, cluster_cidr, service_cidr, cluster_host_prefix):
```
nshidlin, line 164: I think just this `set_advanced_networking` is enough no? Have defaults for the params and build the dict from them. Or just except **kwargs
wdty?

trewest: The only issue is if you only want to update one of the parameters you will have to update all three with defaults.

nshidlin: /lgtm

PR #432, opened 2021-01-16
MGMT-3463 collect full journal logs on error

slaviered:
/hold

dmesg and journal logs should be collected on all nodes in case of an error

### DIFF
--- a/discovery-infra/test_infra/logs_utils.py
+++ b/discovery-infra/test_infra/logs_utils.py
@@ -23,10 +23,10 @@ def verify_logs_uploaded(cluster_tar_path, expected_min_log_num, installation_su
             tar.extractall(tempdir)
             for gz in os.listdir(tempdir):
                 if "bootstrap" in gz:
-                    _verify_node_logs_uploaded(tempdir, gz)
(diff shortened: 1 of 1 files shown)


discovery-infra/test_infra/logs_utils.py
```diff
@@ -67,7 +67,7 @@ def _verify_bootstrap_logs_uploaded(dir_path, file_path, installation_success):
     logs = gz.getnames()
     assert any("bootkube.logs" in s for s in logs), f"bootkube.logs isn't found in {logs}"
     if not installation_success:
-        for logs_type in ["dmesg.logs", "log-bundle"]:
+        for logs_type in ["dmesg.logs", "log-bundle", "journal.logs"]:
```
ybettan, line 70: `journal.logs` and `dmesg.logs` are collected on all nodes only if failure.

They both need to be added to `_verify_node_logs_uploaded` and add a `installation_success` to the function in order to test it only on failure.

YuviGold:
@slaviered Still relevant?

YuviGold (approved):
LGTM

PR #434, opened 2021-01-17
Add wait for controller method

iovadia:

Used in tests to wait for controller and cause cluster error status

### DIFF
--- a/discovery-infra/test_infra/logs_utils.py
+++ b/discovery-infra/test_infra/logs_utils.py
@@ -11,7 +11,7 @@ OC_DOWNLOAD_LOGS_INTERVAL = 10 * 60
 NUM_OF_RETRIES = 6
 
 
-def verify_logs_uploaded(cluster_tar_path, expected_min_log_num, installation_success, check_oc=False):
+def verify_logs_uploaded(cluster_tar_path, expected_min_log_num, installation_success, verify_control_plane=False, check_oc=False):
     assert os.path.exists(cluster_tar_path), f"{cluster_tar_path} doesn't exist"
 
     with TemporaryDirectory() as tempdir:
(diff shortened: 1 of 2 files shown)


YuviGold (approved):

discovery-infra/test_infra/logs_utils.py
```diff
@@ -11,7 +11,7 @@
 NUM_OF_RETRIES = 6
 
 
-def verify_logs_uploaded(cluster_tar_path, expected_min_log_num, installation_success, check_oc=False):
+def verify_logs_uploaded(cluster_tar_path, expected_min_log_num, installation_success, verify_control_plane, check_oc=False):
```
YuviGold, line 14: Please set `verify_control_plane` with a default value so the API won't get broken.
For example, this function is being called from [download_logs module](https://github.com/openshift/assisted-test-infra/blob/5c615a2b85b2c7b0a5867d75bf944463e0f08745/discovery-infra/download_logs.py#L133-L135) and also on the [d/s tests](https://gitlab.cee.redhat.com/ocp-edge-qe/kni-assisted-installer-auto/blob/9bcbbfc15586cb0e782d003e27b05533c8056cb9/api_tests/test_logs.py#L33)

YuviGold (changes requested):

discovery-infra/test_infra/logs_utils.py
iovadia: Thank you for the review, it has a default value.
https://github.com/openshift/assisted-test-infra/pull/434/files#diff-c389358d0828525582921b728933e4e8fac124f747d6bc1a955f4edf2ae3667aR73

YuviGold: Only the inner private function has a default value, while the public function is missing one.

iovadia: Got it, thanks

YuviGold (approved):

PR #435, opened 2021-01-17
pulling squid container image from quay

nshidlin:
Pulling the container image for proxy server from quay instead of dockerhub, in order to avoid pull rate limits


YuviGold (approved):
LGTM

PR #441, opened 2021-01-18
MGMT-3603 Changes for single node IPv6

ori-amizur:
- Send machine CIDR to assisted-service in user-managed-networking single with IPv6
- Set all dns names to point to the node IP address
/cc @empovit 
/cc @tsorya

### DIFF
--- a/discovery-infra/start_discovery.py
+++ b/discovery-infra/start_discovery.py
@@ -8,7 +8,7 @@ import json
 import os
 import time
 import distutils.util
-from netaddr import IPNetwork
+from netaddr import IPNetwork, IPAddress
 
 from test_infra import assisted_service_api, consts, utils
 from test_infra.helper_classes import cluster as helper_cluster
@@ -229,12 +229,12 @@ def set_cluster_vips(client, cluster_id, machine_net):
     client.update_cluster(cluster_id, cluster_info)
 
 
-def set_cluster_machine_cidr(client, cluster_id, machine_net):
+def set_cluster_machine_cidr(client, cluster_id, machine_net, set_vip_dhcp_allocation=True):
     cluster_info = client.cluster_get(cluster_id)
-    cluster_info.vip_dhcp_allocation = True
-    cluster_info.machine_network_cidr = machine_net.cidr_v4
+    if set_vip_dhcp_allocation:
+        cluster_info.vip_dhcp_allocation = True
+    cluster_info.machine_network_cidr = machine_net.cidr_v6 if machine_net.has_ip_v6 and not machine_net.has_ip_v4 else machine_net.cidr_v4
     client.update_cluster(cluster_id, cluster_info)
-
 
 def _get_vips_ips(machine_net):
     if machine_net.has_ip_v4:
@@ -288,7 +288,7 @@ def _cluster_create_params():
     ipv4 = args.ipv4 and args.ipv4.lower() in MachineNetwork.YES_VALUES
     ipv6 = args.ipv6 and args.ipv6.lower() in MachineNetwork.YES_VALUES
     ntp_source = _get_host_ip_from_cidr(args.vm_network_cidr6 if ipv6 and not ipv4 else args.vm_network_cidr)
-    user_managed_networking = is_none_platform_mode()
+    user_managed_networking = is_user_managed_networking()
     http_proxy, https_proxy, no_proxy = _get_http_proxy_params(ipv4=ipv4, ipv6=ipv6)
     params = {
         "openshift_version": utils.get_openshift_version(),
@@ -394,7 +394,7 @@ def nodes_flow(client, cluster_name, cluster):
         cluster_info = client.cluster_get(cluster.id)
         macs = utils.get_libvirt_nodes_macs(nodes_details["libvirt_network_name"])
 
-        if not (cluster_info.api_vip and cluster_info.ingress_vip) and args.master_count != 1:
+        if not (cluster_info.api_vip and cluster_info.ingress_vip):
             utils.wait_till_hosts_with_macs_are_in_status(
                 client=client,
                 cluster_id=cluster.id,
@@ -406,7 +406,10 @@ def nodes_flow(client, cluster_name, cluster):
                 ],
             )
 
-            if args.vip_dhcp_allocation:
+            if args.master_count == 1 and machine_net.has_ip_v6 and not machine_net.has_ip_v4:
+                set_cluster_machine_cidr(client, cluster.id, machine_net, set_vip_dhcp_allocation=False)
+                tf.change_variables({"libvirt_master_ips": [[helper_cluster.Cluster.get_ip_for_single_node_v6(client, cluster.id, args.vm_network_cidr6)]]})
+            elif args.vip_dhcp_allocation:
                 set_cluster_machine_cidr(client, cluster.id, machine_net)
             else:
                 set_cluster_vips(client, cluster.id, machine_net)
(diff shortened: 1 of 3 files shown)


discovery-infra/start_discovery.py
```diff
@@ -229,12 +230,25 @@ def set_cluster_vips(client, cluster_id, machine_net):
     client.update_cluster(cluster_id, cluster_info)
 
 
-def set_cluster_machine_cidr(client, cluster_id, machine_net):
+def set_cluster_machine_cidr(client, cluster_id, machine_net, set_vip_dhcp_allocation=True):
     cluster_info = client.cluster_get(cluster_id)
-    cluster_info.vip_dhcp_allocation = True
-    cluster_info.machine_network_cidr = machine_net.cidr_v4
+    if set_vip_dhcp_allocation:
```
tsorya, line 235: no need for if ,just 
cluster_info.vip_dhcp_allocation = set_vip_dhcp_allocation

discovery-infra/start_discovery.py
```diff
@@ -229,12 +230,25 @@ def set_cluster_vips(client, cluster_id, machine_net):
...
+        cluster_info.vip_dhcp_allocation = True
+    cluster_info.machine_network_cidr = machine_net.cidr_v6 if machine_net.has_ip_v6 and not machine_net.has_ip_v4 else machine_net.cidr_v4
     client.update_cluster(cluster_id, cluster_info)
 
+def _get_ip_for_single_node_v6(client, cluster_id):
```
tsorya, line 240: can we use this method ? 
https://github.com/openshift/assisted-test-infra/blob/master/discovery-infra/test_infra/helper_classes/cluster.py#L660

tsorya: i want to have only one place that get host ips

discovery-infra/start_discovery.py
```diff
@@ -288,7 +302,7 @@ def _cluster_create_params():
     ipv4 = args.ipv4 and args.ipv4.lower() in MachineNetwork.YES_VALUES
     ipv6 = args.ipv6 and args.ipv6.lower() in MachineNetwork.YES_VALUES
     ntp_source = _get_host_ip_from_cidr(args.vm_network_cidr6 if ipv6 and not ipv4 else args.vm_network_cidr)
-    user_managed_networking = is_none_platform_mode()
+    user_managed_networking = is_user_managed_networking()
```
tsorya, line 305: one of the ideas not to change this flag was to verify that there is no need to set it true in cluster create call

discovery-infra/start_discovery.py
```diff
@@ -394,7 +408,7 @@ def nodes_flow(client, cluster_name, cluster):
         cluster_info = client.cluster_get(cluster.id)
         macs = utils.get_libvirt_nodes_macs(nodes_details["libvirt_network_name"])
 
-        if not (cluster_info.api_vip and cluster_info.ingress_vip) and args.master_count != 1:
+        if not (cluster_info.api_vip and cluster_info.ingress_vip):
```
tsorya, line 411: not sure i understand this condition, maybe you know?

terraform_files/baremetal/main.tf
```diff
@@ -162,24 +162,24 @@ resource "libvirt_domain" "worker" {
 
 data "libvirt_network_dns_host_template" "api" {
   count    = 1
-  ip       = var.bootstrap_in_place ? var.libvirt_master_ips[count.index][0] : var.api_vip
+  ip       = var.bootstrap_in_place && !var.ipv6 ? var.libvirt_master_ips[count.index][0] : var.api_vip
```
tsorya, line 165: can you explain the logic please ?

discovery-infra/start_discovery.py
ori-amizur: OK

discovery-infra/start_discovery.py
ori-amizur: This one gets IP that belongs to the machine CIDR

discovery-infra/start_discovery.py
ori-amizur: It seemed to me that the definition of is_user_managed_networking here was incorrect.  It is not just for none platform but for single node as well

empovit:
IMO this PR may interfere with static IPs, e.g. in how it handles the API VIP. What do you think @yevgeny-shnaidman ?

discovery-infra/start_discovery.py
ori-amizur: Whoever did it didn't want to set vips or machine-cidr for none platform

terraform_files/baremetal/main.tf
ori-amizur: When using ipv6, we do not know the ips before and we cannot set before.  Therefore, for DNS, we set all to point to api-vip, and when the node IP is known, we set the api-vip to point to the node-ip

discovery-infra/start_discovery.py
tsorya: but most of the logic is the same, so lets use the existing one and just verify machine-cidr ?

terraform_files/baremetal/main.tf
tsorya: it was added to support adding dns for single node and single node have no api vip. so i don't understand the logic

discovery-infra/start_discovery.py
ori-amizur: Actually since python is typeless,  prefer to set it to True explicitly

discovery-infra/start_discovery.py
ori-amizur: And also I want to verify that is is IPv6.  Anyway, I will change

ori-amizur: This approach will prefer IPv4 over IPv6.  Not sure I want that

discovery-infra/start_discovery.py
```diff
@@ -229,13 +229,13 @@ def set_cluster_vips(client, cluster_id, machine_net):
...
 
 
-def set_cluster_machine_cidr(client, cluster_id, machine_net):
+def set_cluster_machine_cidr(client, cluster_id, machine_net, set_vip_dhcp_allocation=True):
     cluster_info = client.cluster_get(cluster_id)
-    cluster_info.vip_dhcp_allocation = True
-    cluster_info.machine_network_cidr = machine_net.cidr_v4
+    if set_vip_dhcp_allocation:
+        cluster_info.vip_dhcp_allocation = True
+    cluster_info.machine_network_cidr = machine_net.cidr_v6 if machine_net.has_ip_v6 and not machine_net.has_ip_v4 else machine_net.cidr_v4
```
empovit, line 236: Can be simplified as `machine_net.cird_v4 if machine_net.has_ip_v4 else machine_net.cidr_v6`

discovery-infra/start_discovery.py
```diff
@@ -406,7 +406,10 @@ def nodes_flow(client, cluster_name, cluster):
                 ],
             )
 
-            if args.vip_dhcp_allocation:
+            if args.master_count == 1 and machine_net.has_ip_v6 and not machine_net.has_ip_v4:
+                set_cluster_machine_cidr(client, cluster.id, machine_net, set_vip_dhcp_allocation=False)
+                tf.change_variables({"libvirt_master_ips": [[helper_cluster.Cluster.get_ip_for_single_node_v6(client, cluster.id, args.vm_network_cidr6)]]})
```
empovit, line 411: This line will break static IPs as it tries to update TF variables. Currently this causes an attempt to set a MAC address, which fails.

empovit: I believe `if not args.with_static_ips:` can solve it.

discovery-infra/start_discovery.py
ori-amizur: It is not the same

PR #442, opened 2021-01-18
allowing setting node variables from test

nshidlin:
This PR adds:

- request argument to nodes fixture in order to set vm creation parameters at the test level
- modifies attach_disk fixture to support multiple disks

### DIFF
--- a/discovery-infra/tests/base_test.py
+++ b/discovery-infra/tests/base_test.py
@@ -7,6 +7,8 @@ from typing import Optional
 from pathlib import Path
 from paramiko import SSHException
 import shutil
+from copy import deepcopy
+import math
 
 from test_infra import consts
 import test_infra.utils as infra_utils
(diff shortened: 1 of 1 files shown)


discovery-infra/tests/base_test.py
```diff
@@ -228,3 +239,12 @@ def update_oc_config(self, nodes, cluster):
...
                                api_vip=api_vip)
+
+    @staticmethod
+    def get_insufficeint_disk_size_for_nodes(api_client):
+        min_host_requirements = api_client.get_host_requirements()
+        min_disk = min_host_requirements.worker.disk_size_gb
+        min_disk_bytes = min_disk * pow(1000, 3)
+        insufficient_disk_bytes = math.floor(min_disk_bytes * 0.90)
+
+        return {'worker_disk': insufficient_disk_bytes, 'master_disk': insufficient_disk_bytes}
```
lalon4, line 250: why we need to return same value twice? cant we just return node_disk?

nshidlin: `worker_disk` and `master_disk` are the names of the terrafrom variables. This makes it easier to pass to override_vm_parameters

lalon4: think its better to have this method in test level. WDYT?

discovery-infra/tests/base_test.py
```diff
@@ -18,18 +20,27 @@
 from tests.conftest import env_variables, qe_env
 from download_logs import download_logs
 
-
 class BaseTest:
+    @staticmethod
+    def override_vm_parameters(**kwargs):
```
tsorya, line 25: where it is used?

nshidlin: in the tests

nshidlin: moved the calling of this into `nodes` fixture

PR #446, opened 2021-01-19
MGMT-3693 Add wait loop for must-gather logs in download_logs

slaviered:
/assign @YuviGold

### DIFF
--- a/discovery-infra/test_infra/logs_utils.py
+++ b/discovery-infra/test_infra/logs_utils.py
@@ -10,7 +10,8 @@ from test_infra.consts import NUMBER_OF_MASTERS
 OC_DOWNLOAD_LOGS_INTERVAL = 10 * 60
 NUM_OF_RETRIES = 6
 
-def verify_logs_uploaded(cluster_tar_path, expected_min_log_num, installation_success):
+
+def verify_logs_uploaded(cluster_tar_path, expected_min_log_num, installation_success, check_oc=False):
     assert os.path.exists(cluster_tar_path), f"{cluster_tar_path} doesn't exist"
 
     with TemporaryDirectory() as tempdir:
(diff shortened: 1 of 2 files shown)


YuviGold (approved):
LGTM

discovery-infra/test_infra/logs_utils.py
```diff
@@ -24,17 +25,18 @@ def verify_logs_uploaded(cluster_tar_path, expected_min_log_num, installation_su
                     _verify_bootstrap_logs_uploaded(tempdir, gz, installation_success)
                 elif "master" in gz or "worker" in gz:
                     _verify_node_logs_uploaded(tempdir, gz)
+                elif "controller" in gz:
+                    if check_oc:
+                        _verify_oc_logs_uploaded(os.path.join(tempdir, gz))
```
ybettan, line 30: Why not to have a function `_verify_oc_logs_uploaded` that get a parameter `installation_success` and check logs on success and on failure?

We are currently missing checking controller logs on success, I would suggest something like:
```
elif "controller" in gz:
    _verify_controller_logs_uploaded(cluster_tar_path, installation_success)
 
 def _verify_controller_logs_uploaded(cluster_tar_path, installation_success):
      # check that the controller logs exist (controller.logs)
      if !installation_success:
          # check that all the subdirectories of must-gather exists (the same way it is done for installer-gather)
```

discovery-infra/test_infra/logs_utils.py
```diff
@@ -10,7 +10,8 @@
 OC_DOWNLOAD_LOGS_INTERVAL = 10 * 60
 NUM_OF_RETRIES = 6
 
-def verify_logs_uploaded(cluster_tar_path, expected_min_log_num, installation_success):
+
+def verify_logs_uploaded(cluster_tar_path, expected_min_log_num, installation_success, check_oc=False):
```
ybettan, line 14: `check_oc` parameter is redundant, you can get the same logic by using `installation_success`

discovery-infra/test_infra/logs_utils.py
slaviered: we do check controller logs on success through the check of min_num_of_logs. In addition there is no point to check within the must-gather dirctory since they are too many directories there and with temporary names. if the outer zip exists it means that must-gather tool ran successfully. Otherwise the zip is not created (unlike other log cases)

discovery-infra/test_infra/logs_utils.py
slaviered: these parameters are not identical. installation_success can be false from numerous reasons. only certain failures, or more precisely failure stages indicates that we might have must-gather logs

PR #447, opened 2021-01-20
MGMT-3702 Fix assisted test infra to work with static ips with IPv6 single node

ori-amizur:
/cc @empovit 
/cc @tsorya


terraform_files/baremetal/variables-libvirt.tf
```diff
@@ -174,3 +174,9 @@ variable "static_macs" {
   description = "If true, static macs are configured for the network interfaces"
   type        = bool
 }
+
+variable "single_node_ip" {
+  description = "IP address of single node.  Used for DNS"
+  type = string
+  default = "192.168.126.10"
```
tsorya, line 181: lets remove default and set it in the code, i want this part to fail in case nobody sets ip

ori-amizur: OK

PR #449, opened 2021-01-20
MGMT-3607: configure OS to enable IPv6 via sysctl

osherdp:
/cc @ori-amizur @empovit

### DIFF
--- a/scripts/install_environment.sh
+++ b/scripts/install_environment.sh
@@ -183,6 +183,10 @@ function additional_configs() {
     sudo setenforce 0 || true
     sudo chmod 600 ssh_key/key
     sudo firewall-cmd --zone=libvirt --add-port=59151-59154/tcp
+
+    echo "enabling ipv6"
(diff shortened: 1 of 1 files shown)


scripts/install_environment.sh
```diff
@@ -183,6 +183,10 @@ function additional_configs() {
     sudo setenforce 0 || true
     sudo chmod 600 ssh_key/key
     sudo firewall-cmd --zone=libvirt --add-port=59151-59154/tcp
+
+    echo "enabling ipv6"
+    sudo sed -ir 's/net.ipv6.conf.all.disable_ipv6 = 1/net.ipv6.conf.all.disable_ipv6 = 0/g' /etc/sysctl.conf
```
empovit, line 188: Although having spaces around the equal sign is the default, the format without spaces is also accepted. I'd recommend handling it as well, i.e. `net.ipv6.conf.all.disable_ipv6=1`. Just in case.

osherdp: Done

PR #452, opened 2021-01-23
Add bootable option for newly attached disks

trewest:
Modifies `attach_test_disk` node method and dependencies to allow for attaching bootable disks. This is useful for test cases that require a disk with the `(bootable)` flag to be set

### DIFF
--- a/discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
+++ b/discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
@@ -99,11 +99,11 @@ class LibvirtController(NodeController, ABC):
         utils.run_command(command, shell=True)
 
     @staticmethod
-    def create_bootable_disk(disk_path, disk_size):
-        create_disk_image = f'qemu-img create -f qcow2 {disk_path} {disk_size}'
-        utils.run_command(create_disk_image, shell=True)
-        create_mbr_partition = f'virt-format -a {disk_path} --partition=mbr'
-        utils.run_command(create_mbr_partition, shell=True)
+    # LIBGUESTFS_BACKEND set to mitigate errors with running libvirt as root
+    # https://libguestfs.org/guestfs-faq.1.html#permission-denied-when-running-libguestfs-as-root
+    def add_disk_bootflag(disk_path):
+        command = f'virt-format -a {disk_path} --partition=mbr'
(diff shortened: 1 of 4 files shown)


discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
```diff
@@ -205,6 +210,10 @@ def attach_test_disk(self, node_name, disk_size):
 
         self.create_disk(tmp_disk, disk_size)
 
+        if bootable:
+            os.environ["LIBGUESTFS_BACKEND"] = "direct"
```
omertuc, line 214: What is the purpose of this environment variable?

trewest:
@omertuc Check "Permission denied" when running libguestfs as root section:
https://libguestfs.org/guestfs-faq.1.html#permission-denied-when-running-libguestfs-as-root

I have seen a "No such file or directory" error message but it seems that this method fixes the issue.

omertuc:
> @omertuc Check "Permission denied" when running libguestfs as root section:
> [libguestfs.org/guestfs-faq.1.html#permission-denied-when-running-libguestfs-as-root](https://libguestfs.org/guestfs-faq.1.html#permission-denied-when-running-libguestfs-as-root)
> 
> I have seen a "No such file or directory" error message but it seems that this method fixes the issue.

I see. Is it relevant only for the invocation of `virt-format`? If so, I think it's best if you only set that environment for that specific invocation, rather than globally for the entire process, so it doesn't affect anything that might not otherwise need it.

Might want to modify https://github.com/openshift/assisted-test-infra/blob/b6d2b0099b1c5d133d69ffc9f919182c22feb7d4/discovery-infra/test_infra/utils.py#L35 to allow passing extra environment variables to allow that

trewest:
@omertuc made recommended updates. I also tested and it is passing

discovery-infra/test_infra/utils.py
```diff
@@ -32,13 +32,14 @@
...
 
-def run_command(command, shell=False, raise_errors=True):
+def run_command(command, shell=False, raise_errors=True, env=None):
     command = command if shell else shlex.split(command)
     process = subprocess.run(
         command,
         shell=shell,
         stdout=subprocess.PIPE,
         stderr=subprocess.PIPE,
+        env=env,
```
omertuc, line 42: IIRC env overwrites the entire env, so it's better to append to it rather than overwrite it completely.

Here is an example of what I mean:
image

PR #453, opened 2021-01-24
Collecting log only on failure to reduce execution time

lalon4:
In collect logs there is a retry and sleep of 5 minutes, which is unnecessary when test passes. Therefore collecting logs only on failures

### DIFF
--- a/discovery-infra/tests/base_test.py
+++ b/discovery-infra/tests/base_test.py
@@ -72,8 +72,9 @@ class BaseTest:
         yield get_cluster_func
 
         for cluster in clusters:
-            logging.info(f'--- TEARDOWN --- Collecting Logs for test: {request.node.name}\n')
-            self.collect_test_logs(cluster, api_client, request.node, nodes)
+            if request.node.result_call.failed:
+                logging.info(f'--- TEARDOWN --- Collecting Logs for test: {request.node.name}\n')
+                self.collect_test_logs(cluster, api_client, request.node, nodes)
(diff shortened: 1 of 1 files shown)


discovery-infra/tests/base_test.py
```diff
@@ -178,7 +178,8 @@ def collect_test_logs(self, cluster, api_client, test, nodes):
         log_dir_name = f"{env_variables['log_folder']}/{test.name}"
         with suppress(ApiException):
             cluster_details = json.loads(json.dumps(cluster.get_details().to_dict(), sort_keys=True, default=str))
-            download_logs(api_client, cluster_details, log_dir_name, test.result_call.failed)
+            if test.result_call.failed:
```
YuviGold, line 181: I think that this condition should be checked on the [cluster teardown](https://github.com/openshift/assisted-test-infra/blob/7ffdc1bf69463a9df01f3f765a1444175e1b0ba0/discovery-infra/tests/base_test.py#L76). no?

lalon4: right. this method is being called in cluster teardown
https://github.com/openshift/assisted-test-infra/blob/master/discovery-infra/tests/base_test.py#L76

YuviGold: This method seems more like a utility. I think it would be much proper to make that check in the calling teardown function

lalon4: Fixed

YuviGold (approved):
LGTM

PR #456, opened 2021-01-25
MGMT-3607: add some targets for testing Openshift 4.8 milestone one

osherdp:
/hold
/cc @ronniel1 @empovit

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -270,6 +270,14 @@ deploy_nodes_with_install:
 	bash scripts/utils.sh local_setup_before_deployment $(PLATFORM) $(NAMESPACE) $(OC_FLAG)
 	skipper make $(SKIPPER_PARAMS) _deploy_nodes NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE) $(OC_FLAG)) NAMESPACE=$(NAMESPACE) ADDITIONAL_PARAMS="'-in ${ADDITIONAL_PARAMS}'" $(SKIPPER_PARAMS) DAY1_PARAMS=--day1-cluster
 
+deploy_single_node_ipv6_with_install:
+	make deploy_nodes_with_install NUM_MASTERS=1 IPv6=yes
+
+deploy_single_node_static_ip_with_install:
+	make deploy_nodes_with_install NUM_MASTERS=1 ADDITIONAL_PARAMS="'--with-static-ips'"
+
+openshift_4.8_milestone_one:
+	make deploy_nodes_with_install NUM_MASTERS=1 IPv6=yes ADDITIONAL_PARAMS="'--with-static-ips' --wait-for-cvo"
 
 deploy_static_ips_nodes_with_install:
 	make deploy_nodes_with_install ADDITIONAL_PARAMS="'--with-static-ips'"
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -270,6 +270,14 @@ deploy_nodes_with_install:
 	bash scripts/utils.sh local_setup_before_deployment $(PLATFORM) $(NAMESPACE) $(OC_FLAG)
 	skipper make $(SKIPPER_PARAMS) _deploy_nodes NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE) $(OC_FLAG)) NAMESPACE=$(NAMESPACE) ADDITIONAL_PARAMS="'-in ${ADDITIONAL_PARAMS}'" $(SKIPPER_PARAMS) DAY1_PARAMS=--day1-cluster
 
+deploy_single_node_ipv6_with_install:
+	make deploy_nodes_with_install NUM_MASTERS=1 IPv6=yes
```
empovit, line 274: `IPv4=no`, otherwise many things may work not because they work fine with IPv6, but because IPv4 is there. Also, how does it work without a proxy? Do we support IPv6 without a proxy already?

Makefile
```diff
@@ -270,6 +270,14 @@ deploy_nodes_with_install:
...
+
+deploy_single_node_static_ip_with_install:
+	make deploy_nodes_with_install NUM_MASTERS=1 ADDITIONAL_PARAMS="'--with-static-ips'"
+
+openshift_4.8_milestone_one:
+	make deploy_nodes_with_install NUM_MASTERS=1 IPv6=yes ADDITIONAL_PARAMS="'--with-static-ips' --wait-for-cvo"
```
empovit, line 280: Again, you aren't disabling IPv4.

Makefile
osherdp: About IPv4, it's a nice catch. I'll change this target. Although for the larger milestone I'm not sure if it's exactly what the user will do (whether it's IPv6 only or both)
About the proxy, let me get back to you

YuviGold:
I don't like adding more and more Makefile targets. IMO We should use [environment files](https://3musketeers.io/docs/environment-variables.html)

Makefile
empovit: I guess you should be adding `PROXY=true` and it will be configured automatically. It's a recent change, see the `run_full_flow_with_ipv6` target.

empovit:
> I don't like adding more and more Makefile targets. IMO We should use [environment files](https://3musketeers.io/docs/environment-variables.html)

I agree, it becomes unmanageable. We should probably have a wider discussion about migrating to environment files. Or at least find a way to split the Makefile into multiple files, e.g. core and by feature.

Makefile
osherdp: @ronniel1 our flow talks about IPv4+IPv6 or on a pure IPv6 environment?

YuviGold:
> @YuviGold I of course agree, but please please please let's not block this PR because of that. We have no periodic builds right now for making those kind of flows and it gets unmanageable for me in a different scope... :)

The makefile shouldn't block you from creating periodics. You can look at https://github.com/openshift/release/pull/14578 how I defined the workflow, IMO it's much clearer instead of maintaining complex targets
```yaml
- as: e2e-metal-assisted-ipv6
  cron: 0 0 * * 1
  steps:
    cluster_profile: packet
    env:
      ASSISTED_CONFIG: |
        IPv6=yes
        IPv4=no
        VIP_DHCP_ALLOCATION=no
        PROXY=yes
    workflow: baremetalds-assisted
```

Then when you would want to create another new deployment combination, for example (IPv6, single node, with DHCP) you won't need to make another Makefile target.

PR #458, opened 2021-01-26
MGMT-3790: fix deploy-ibip DNS entries

osherdp:
/cc @tsorya @YuviGold 
A better solution will come in the future

### DIFF
--- a/discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
+++ b/discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
@@ -65,7 +65,8 @@ class TerraformController(LibvirtController):
                   # TODO change to namespace index
                   "libvirt_secondary_network_if": self.network_conf.libvirt_secondary_network_if,
                   "provisioning_cidr": self.network_conf.provisioning_cidr,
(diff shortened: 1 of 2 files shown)


discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
```diff
@@ -65,7 +65,8 @@ def _terraform_params(self, **kwargs):
                   # TODO change to namespace index
                   "libvirt_secondary_network_if": self.network_conf.libvirt_secondary_network_if,
                   "provisioning_cidr": self.network_conf.provisioning_cidr,
-                  "running": True
+                  "running": True,
+                  "single_node_ip": kwargs.get('single_node_ip', ''),
```
eranco74, line 69: The terrraform get this param?

osherdp: yes
so far it's got changed only on runtime, that's why I needed to add it here

YuviGold (approved):

eranco74 (commented):
/lgtm

PR #460, opened 2021-01-26
MGMT-3336 consts: added event for minimal-iso

danielerez:
Added an event const for minimal-iso and modified the current to full-iso.

### DIFF
--- a/discovery-infra/test_infra/consts.py
+++ b/discovery-infra/test_infra/consts.py
@@ -95,6 +95,8 @@ class Events:
     SUCCESSFULLY_REGISTERED_CLUSTER = "Successfully registered cluster"
     PENDING_FOR_INPUT = "to pending-for-input"
     GENERATED_IMAGE = "Generated image (SSH public key is set)"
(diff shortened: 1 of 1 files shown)


danielerez:
@lalon4 @YuviGold

discovery-infra/test_infra/consts.py
```diff
@@ -94,7 +94,8 @@ class Events:
     REGISTERED_CLUSTER = "Registered cluster"
     SUCCESSFULLY_REGISTERED_CLUSTER = "Successfully registered cluster"
     PENDING_FOR_INPUT = "to pending-for-input"
-    GENERATED_IMAGE = "Generated image (SSH public key is set)"
```
lalon4, line 97: missed that
Is the message is going to totally replaced? 
cause this will break the current regression run. 
So until this feature get to staging lets leave it, with a remark?

danielerez: Sure, so simply left both consts. When ready, we could just remove the stale one.

YuviGold (approved):

PR #462, opened 2021-01-26
MGMT-3801 Fix empty MAC addresses for day2 nodes

empovit:


discovery-infra/day2.py
```diff
@@ -169,7 +170,7 @@ def get_network_nodes_from_terraform(tf_folder):
     return tfvars['libvirt_network_name'], tfvars['master_count'] + tfvars['worker_count']
 
 
-def set_workers_ips_by_type(tfvars, num_worker_nodes, master_ip_type, worker_ip_type):
+def set_workers_ips_by_type(tfvars, num_worker_nodes, master_ip_type, worker_ip_type, worker_mac_type):
```
yevgeny-shnaidman, line 173: let's just rename this function to reflect that it also sets macs

PR #464, opened 2021-01-27
WIP: MGMT-3323 Build python client in assisted-test-infra

lranjbar:
@YuviGold @djzager Thoughts on building python client in assisted-test-infra?

### DIFF
--- a/Dockerfile.test-infra
+++ b/Dockerfile.test-infra
@@ -1,8 +1,7 @@
-FROM quay.io/ocpmetal/assisted-service:latest AS service
-
 FROM quay.io/centos/centos:8.3.2011
 
 RUN yum -y install make gcc unzip wget curl git podman httpd-tools jq nss_wrapper \
+  java \
   python3 python3-devel \
   libvirt-client libvirt-devel libguestfs-tools && \
   yum clean all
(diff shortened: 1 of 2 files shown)


romfreiman:
I think we had some issues with that

YuviGold:
As I explained in the ticket, we cannot set the python client as part of the image.
For example, with your approach, we aren't able to update for a specific client version - which means we will always use the master.
Transferring an argument as the git commit won't work as well since Prow doesn't support arguments.

djzager:
> Transferring an argument as the git commit won't work as well since Prow doesn't support arguments.

I think for prow we can have a specific container image that is built as part of the assisted-service config that contains the python client. This way, when testing the assisted-service we use the code from the PR for the client, when testing everything else we use the latest tested version of the assisted-service.

lranjbar:
So for me the disconnect I have with this request is what the repository branches should resolve to depending on the repository under test. In my experience the best way to avoid any integration deadlocks is to only pull in PR commit changes into the tests in the repositories they originate from. A separate testing repository like assisted-test-infra should only pull the latest tested (master) branches of all other repositories under test.

PR #465, opened 2021-01-27
MGMT-3065 Set hostnames via API in IPv6 setups

empovit:
Setting hostnames via Terraform doesn't work with IPv6. Work that around by setting the hostnames of a cluster via assisted service API. Day 2 scenario similar to how it's done in day 1.

### DIFF
--- a/discovery-infra/day2.py
+++ b/discovery-infra/day2.py
@@ -7,7 +7,7 @@ import waiting
 import uuid
 
 from test_infra import assisted_service_api, utils, consts
-from test_infra.tools import static_ips
+from test_infra.tools import static_ips, terraform_utils
 from logger import log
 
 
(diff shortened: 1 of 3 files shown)


discovery-infra/day2.py
```diff
@@ -140,6 +144,11 @@ def day2_nodes_flow(client,
         )
         log.info("%d worker nodes were successfully added to OCP cluster", num_worker_nodes)
 
+def set_hostnames_from_tf(client, cluster_id, tf_folder, network_name):
+    tf = terraform_utils.TerraformUtils(working_dir=tf_folder)
+    libvirt_nodes = utils.extract_nodes_from_tf_state(tf.get_state(), (network_name), consts.NodeRoles.WORKER)
+    utils.update_hosts(client, cluster_id, libvirt_nodes, update_roles=False, update_hostnames=True)
```
yevgeny-shnaidman, line 150: i think you should add a comment here, that even if 1st dat cluster had worker nodes installed, the update_hosts function will ignore them , since they are not present in the inventory of day2 cluster

PR #466, opened 2021-01-27
MGMT-3684 Do not set machine CIDR in none-platform

rollandf:
- Do not set machine CIDR in none-platform
- Enable second network on SNO


rollandf:
@tsorya

YuviGold (approved):

eranco74 (commented):
/lgtm

eranco74:
Worth mentioning that this is no longer relevant since the assisted-service will get the machine CIDR from the bootstrap node
https://github.com/openshift/assisted-service/pull/967

PR #474, opened 2021-01-30
MGMT-3841: Inject images onprem

djzager:
Make it possible to override the installer, controller, and agent images when running test-infra for onprem.


djzager:
cc @YuviGold

YuviGold:
The run is failing on
```
scripts/deploy_assisted_service.sh: line 24: INSTALLER_IMAGE: unbound variable
```
Since your script it defined with `set -u`
```
              -u      Treat unset variables and parameters other than the special parameters "@" and "*" as an error when performing parameter expansion.  If expansion is attempted on an unset variable or parameter, the shell prints
                      an error message, and, if not interactive, exits with a non-zero status.
```

IMO that's a good validation for shell scripts. You should change the script into this:
```bash
if [ -n "${INSTALLER_IMAGE:-}" ]; then
```

YuviGold (approved):
LGTM

YuviGold:
I'm able to see the images in the logs :+1:

PR #475, opened 2021-01-31
MGMT-3838: limit libvirt's DHCP allocation range on IPv6

osherdp:
/hold
/cc @ronniel1 @ori-amizur 
Not pretty, but can be easily replaced when someone implement dmacvicar/terraform-provider-libvirt#794


osherdp:
Already noticed I need to fix IPs allocation mode as well
just a sec

PR #476, opened 2021-01-31
MGMT-3571 Changes for single node IPv6

ori-amizur:
/cc @tsorya 
/cc @nshidlin

### DIFF
--- a/discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
+++ b/discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
@@ -19,7 +19,7 @@ class TerraformController(LibvirtController):
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
         self.cluster_suffix = self._get_random_name()
-        self.cluster_name = kwargs.get('cluster_name', f'{consts.CLUSTER_PREFIX}') + "-" + self.cluster_suffix
+        self.cluster_name = kwargs.get('cluster_name', f'{consts.CLUSTER_PREFIX}' + "-" + self.cluster_suffix)
         self.network_name = kwargs.get('network_name', consts.TEST_NETWORK) + self.cluster_suffix
         self.network_conf = kwargs.get('net_asset')
         self.cluster_domain = kwargs.get('base_domain', "redhat.com")
--- a/discovery-infra/test_infra/helper_classes/cluster.py
+++ b/discovery-infra/test_infra/helper_classes/cluster.py
@@ -18,16 +18,24 @@ from test_infra.tools import static_ips
 class Cluster:
 
     def __init__(self, api_client, cluster_name=None, additional_ntp_source=None,
-                 openshift_version="4.6", cluster_id=None):
+                 openshift_version="4.6", cluster_id=None, user_managed_networking=False, high_availability_mode=consts.HighAvailabilityMode.FULL):
         self.api_client = api_client
 
+        self._high_availability_mode = high_availability_mode
         if cluster_id:
             self.id = cluster_id
         else:
-            self.id = self._create(cluster_name or "test-infra-cluster", additional_ntp_source, openshift_version).id
+            cluster_name = cluster_name or env_variables.get('cluster_name', "test-infra-cluster")
+            self.id = self._create(cluster_name, additional_ntp_source, openshift_version,
+                                   user_managed_networking=user_managed_networking, high_availability_mode=high_availability_mode).id
             self.name = cluster_name
 
-    def _create(self, cluster_name, additional_ntp_source, openshift_version):
+    def _create(self,
+                cluster_name,
+                additional_ntp_source,
+                openshift_version,
+                user_managed_networking,
+                high_availability_mode):
         return self.api_client.create_cluster(
             cluster_name,
             ssh_public_key=env_variables['ssh_public_key'],
@@ -35,7 +43,9 @@ class Cluster:
             pull_secret=env_variables['pull_secret'],
             base_dns_domain=env_variables['base_domain'],
             vip_dhcp_allocation=env_variables['vip_dhcp_allocation'],
-            additional_ntp_source=additional_ntp_source
+            additional_ntp_source=additional_ntp_source,
+            user_managed_networking=user_managed_networking,
+            high_availability_mode=high_availability_mode
         )
 
     def delete(self):
@@ -149,8 +159,7 @@ class Cluster:
             "cluster_network_cidr": env_variables['cluster_cidr'],
             "cluster_network_host_prefix": env_variables['host_prefix'],
         })
-
-        if vip_dhcp_allocation:
+        if vip_dhcp_allocation or self._high_availability_mode == consts.HighAvailabilityMode.NONE:
             self.set_machine_cidr(controller.get_machine_cidr())
         else:
             self.set_ingress_and_api_vips(controller.get_ingress_and_api_vips())
@@ -486,9 +495,12 @@ class Cluster:
                 static_ips=static_ips_config
             )
         nodes.start_all()
-        self.wait_until_hosts_are_discovered(nodes_count=nodes_count)
+        self.wait_until_hosts_are_discovered(nodes_count=nodes_count, allow_insufficient=True)
         nodes.set_hostnames(self)
-        self.set_host_roles()
+        if self._high_availability_mode != consts.HighAvailabilityMode.NONE:
+            self.set_host_roles()
+        else:
+            nodes.set_single_node_ip(self)
         self.set_network_params(
             controller=nodes.controller,
             vip_dhcp_allocation=vip_dhcp_allocation,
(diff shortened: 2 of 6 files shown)


discovery-infra/tests/base_test.py
```diff
@@ -59,14 +59,18 @@ def cluster(self, api_client, request, nodes):
 
         def get_cluster_func(cluster_name: Optional[str] = None,
                              additional_ntp_source: Optional[str] = consts.DEFAULT_ADDITIONAL_NTP_SOURCE,
-                             openshift_version: Optional[str] = env_variables['openshift_version']):
+                             openshift_version: Optional[str] = env_variables['openshift_version'],
+                             user_managed_networking=False,
+                             high_availability_mode="Full"):
```
nshidlin, line 64: should `high_availability_mode="Full"` be the default for all non-sno deployements?

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -149,8 +152,8 @@ def set_network_params(
             "cluster_network_cidr": env_variables['cluster_cidr'],
             "cluster_network_host_prefix": env_variables['host_prefix'],
         })
-
-        if vip_dhcp_allocation:
+        master_count = env_variables.get('master_count' , 3)
```
nshidlin, line 155: Not sure why you default to 3 here and not just use what's defined in `env_variables`

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -486,9 +489,12 @@ def prepare_for_install(
                 static_ips=static_ips_config
             )
         nodes.start_all()
-        self.wait_until_hosts_are_discovered(nodes_count=nodes_count)
+        self.wait_until_hosts_are_discovered(nodes_count=nodes_count, allow_insufficient=True)
```
nshidlin, line 492: What's the purpose of adding `allow_insufficient=True` ? Could this be moved to be a parameter to `prepare_for_install` ?

discovery-infra/tests/base_test.py
ori-amizur: Yes

discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
```diff
@@ -19,7 +19,7 @@ class TerraformController(LibvirtController):
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
         self.cluster_suffix = self._get_random_name()
-        self.cluster_name = kwargs.get('cluster_name', f'{consts.CLUSTER_PREFIX}') + "-" + self.cluster_suffix
+        self.cluster_name = kwargs.get('cluster_name', f'{consts.CLUSTER_PREFIX}' + "-" + self.cluster_suffix)
```
tsorya, line 22: sure? i think cluster_name was always set till now, so suffix was added in any case. It was on purpose

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -18,24 +18,27 @@
 class Cluster:
 
     def __init__(self, api_client, cluster_name=None, additional_ntp_source=None,
-                 openshift_version="4.6", cluster_id=None):
+                 openshift_version="4.6", cluster_id=None, user_managed_networking=False, high_availability_mode="Full"):
```
tsorya, line 21: lets have high_availability_mode as class in consts with 2 values full and none

discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
tsorya: here it is made to be sure we have unique vm-name prefix

discovery-infra/test_infra/helper_classes/cluster.py
tsorya: no option to set vip_dhcp_allocation in the code where this function is called from? and not to have "master_count == 1"

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -486,9 +489,12 @@ def prepare_for_install(
                 static_ips=static_ips_config
             )
         nodes.start_all()
-        self.wait_until_hosts_are_discovered(nodes_count=nodes_count)
+        self.wait_until_hosts_are_discovered(nodes_count=nodes_count, allow_insufficient=True)
         nodes.set_hostnames(self)
-        self.set_host_roles()
+        if nodes_count != 1:
```
tsorya, line 494: and if it is 0 ?

discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
ori-amizur: cluster_name is what the user wanted.  Why change it?
Anyway for our purpose, we need to verify that the cluster_name here is the same as known by Assisted Service.  Otherwise, we will have problem with DNS names. For example, api-int.<cluster-name>.<base-domain>

discovery-infra/test_infra/helper_classes/cluster.py
ori-amizur: OK

discovery-infra/test_infra/helper_classes/cluster.py
tsorya: can we have parameter in class that will say, single_node ? and then just if self.single_node?

discovery-infra/test_infra/helper_classes/cluster.py
ori-amizur: OK

discovery-infra/test_infra/helper_classes/cluster.py
ori-amizur: I think we should always allow it.  I don't know why it is not always wait for insufficient as well.
I don't think the user of prepare_for_install should bother with this implementation detail

discovery-infra/test_infra/helper_classes/cluster.py
ori-amizur: OK

discovery-infra/test_infra/helper_classes/cluster.py
ori-amizur: @tsorya look at the code, not just the comment

discovery-infra/test_infra/helper_classes/cluster.py
nshidlin: In the majority of cases `wait_until_hosts_are_discovered` assumes hosts are either known or pending-for-input
an insufficient host is the exception

ori-amizur: Waiting for insufficient is more correct.

nshidlin: Sure, it doesn't break any flow as prepare_for_install still waits for cluster to be ready

discovery-infra/test_infra/helper_classes/cluster.py
ori-amizur: Is 0 a valid value ?
Do we need an additional parameter, besides the 6 we already set:
master_count
nodes_count
bootstrap_in_place
high_availability_mode
user_managed_networking ?

discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
nshidlin: the suffix is added so that the vm-names are unique when running in parallel

ori-amizur: So it should be the user responsibility.  We need to verify that the cluster name known by assisted-service is the same as known by TF.  Otherwise, the DNS names will not be set correctly.

discovery-infra/test_infra/helper_classes/cluster.py
ori-amizur: I don't think it should break anything

discovery-infra/test_infra/helper_classes/cluster.py
tsorya: i mean why not to have one and not 5
self.single_node -> will set all others

ori-amizur: It is another dependency.  Where will it set all others? And if they are not set for some reason?  Just adds more complexity

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -18,24 +18,34 @@
...
+                                   user_managed_networking=user_managed_networking, high_availability_mode=high_availability_mode).id
             self.name = cluster_name
 
-    def _create(self, cluster_name, additional_ntp_source, openshift_version):
+    def _create(self,
+                cluster_name,
+                additional_ntp_source,
+                openshift_version,
+                user_managed_networking=False,
+                high_availability_mode=consts.HighAvailabilityMode.FULL):
```
tsorya, line 38: const in in init, why to have those?

ori-amizur: Assuming we always need these values, then there must be default values for them.   What's wrong with it?

PR #481, opened 2021-02-01
NO-ISSUE: retry on every error cURL has, even non-transient ones

osherdp:
/cc @YuviGold


YuviGold (approved):
LGTM

YuviGold (approved):

osherdp:
@YuviGold maybe we indeed need to run some subset of the tests, they are way too flaky

PR #484, opened 2021-02-02
MGMT-3885: enable using reverse-proxy to access image registries

osherdp:
This PR adds an HAproxy component that enables accessing image registries (or actually any other service) that isn't resolved into a IPv6 address.
We also define the service's URL using DNS, to make it accessable via both IPv4 and IPv6.
This should be more generalized in the future with NAT64+DNS64, to allow on-the-fly translation of any service that doesn't have IPv6 address yet.


terraform_files/baremetal/main.tf
```diff
@@ -45,6 +45,22 @@ resource "libvirt_network" "net" {
         ip       = hosts.value.ip
       }
     }
+
+    dynamic "hosts" {
+      for_each = var.reverse_proxy_hosts
+      content {
+        hostname  = hosts.value
```
YuviGold, line 52: Shouldn't it be `hosts.value.hostname`?

terraform_files/baremetal/main.tf
```diff
@@ -45,6 +45,22 @@ resource "libvirt_network" "net" {
...
+        ip        = var.reverse_proxy_ip
+      }
+    }
+
+    dynamic "hosts" {
+      for_each = var.assisted_service_ips
+      content {
+        hostname  = var.assisted_service_domain
+        ip        = hosts.value
```
YuviGold, line 61: Shouldn't it be `hosts.value.ip`?

YuviGold (commented):
Please add a description for this PR describing what you did.

discovery-infra/start_discovery.py
```diff
@@ -22,6 +23,15 @@
 import bootstrap_in_place as ibip
 
 
+# Any change here should be reflected in scripts/haproxy.cfg as well
+IMAGE_REGISTRIES = [
+    "quay.io",
+    "cdn02.quay.io",
+    "mirror.openshift.com",
+    "registry.ci.openshift.org",
+    "ci-dv2np-image-registry-us-east-1-aunteqmixxpqypvdqwbmjbiloeix.s3.dualstack.us-east-1.amazonaws.com",
```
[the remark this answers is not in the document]
osherdp: Not really. It's every registry we're pulling images from
It's all a temporary solution until we'll provide DNS64 that can detect any service which doesn't have IPv6 address, so I wouldn't sweat it

discovery-infra/start_discovery.py
```diff
@@ -22,6 +23,18 @@
...
+    "registry.build02.ci.openshift.org",
+]
+
+REVERSE_PROXY_IP = "1001:db8::1"
```
ori-amizur, line 36: Can we calculate according to the network we use? Not just hard coded network?

discovery-infra/start_discovery.py
```diff
@@ -22,6 +23,18 @@
 import bootstrap_in_place as ibip
 
 
+# Any change here should be reflected in scripts/haproxy.cfg as well
+IMAGE_REGISTRIES = [
+    "quay.io",
```
ori-amizur, line 28: Since we are hard coding these, maybe we can have a single place that they are defined?

scripts/haproxy.cfg
```diff
@@ -0,0 +1,37 @@
+# Any change in this file should be reflected in discovery-infra/start_discovery.py as well
+defaults
+  timeout client 30s
+  timeout server 30s
+  timeout connect 5s
+
+frontend https
+    bind :::443 v4v6
```
ori-amizur, line 8: Do we need IPv4 binding?

scripts/install_environment.sh
```diff
@@ -160,6 +160,11 @@ function config_chronyd() {
   sudo firewall-cmd --zone=libvirt --add-port=123/udp
 }
 
+function deploy_reverse_proxy() {
```
tsorya, line 163: i don't think we need to setup ha proxy on create environment, i think it must be deployed only if needed

terraform_files/baremetal/terraform.tfvars.json
```diff
@@ -29,5 +29,9 @@
   "libvirt_network_mtu": "1500",
   "running": true,
   "bootstrap_in_place": false,
-  "static_macs": false
+  "static_macs": false,
+  "reverse_proxy_hosts": ["quay.io", "cdn02.quay.io", "mirror.openshift.com"],
```
tsorya, line 33: we will set reverse proxy on every run?

terraform_files/baremetal/terraform.tfvars.json
```diff
@@ -29,5 +29,9 @@
...
+  "reverse_proxy_ip": "1001:db8::1",
+  "assisted_service_ips": ["1001:db8::1", "10.19.176.137"],
```
tsorya, line 35: what is this default?

scripts/install_environment.sh
tsorya: from my point of view, run it from test infra when you need it

discovery-infra/start_discovery.py
```diff
@@ -99,21 +119,28 @@ def fill_tfvars(
 
     machine_cidr_addresses = []
     provisioning_cidr_addresses = []
+    assisted_service_ips = []
```
tsorya, line 122: why it is needed? I don't understand this change. Why changing something for ipv6 must require changes for ipv4?

discovery-infra/start_discovery.py
tsorya: and will be nice to have it in consts and not here

discovery-infra/start_discovery.py
tsorya: same

tsorya: on the way class MachineNetwork too :)

terraform_files/baremetal/main.tf
```diff
@@ -45,6 +45,22 @@ resource "libvirt_network" "net" {
         ip       = hosts.value.ip
       }
     }
+
+    dynamic "hosts" {
```
tsorya, line 49: why it is needed?

terraform_files/baremetal/main.tf
```diff
@@ -45,6 +45,22 @@ resource "libvirt_network" "net" {
...
+      for_each = var.reverse_proxy_hosts
+      content {
+        hostname  = hosts.value
+        ip        = var.reverse_proxy_ip
+      }
+    }
+
+    dynamic "hosts" {
+      for_each = var.assisted_service_ips
```
tsorya, line 58: why it is needed?

scripts/haproxy.cfg
tsorya: will it work with cloud or staging dpeloyments?

discovery-infra/start_discovery.py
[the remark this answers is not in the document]
YuviGold: When do we use `ci-dv2np-image-registry-us-east-1-aunteqmixxpqypvdqwbmjbiloeix...`? Seems like a temporary DNS

scripts/install_environment.sh
```diff
@@ -160,6 +160,11 @@ function config_chronyd() {
...
+    echo "Starting HAproxy for accessing image registries via IPv6"
```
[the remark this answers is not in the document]
YuviGold: How is that not relevant? If you set `SERVICE_URL` as `hostname` always, then the nodes always need to resolve (regardless of IPv4v6) the IP of the service

discovery-infra/start_discovery.py
osherdp: I hope you appreciate the lua code you made me write

terraform_files/baremetal/terraform.tfvars.json
osherdp: Only if this list is non empty

terraform_files/baremetal/terraform.tfvars.json
osherdp: changed

discovery-infra/start_discovery.py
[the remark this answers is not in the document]
YuviGold: Seems like something that's going to get broken soon.
There is no way to support redirection?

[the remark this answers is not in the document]
osherdp: Not in the scope of this issue
Sorry

scripts/choose_backend.lua
```diff
@@ -0,0 +1,19 @@
+core.Debug("Loading lua code for routing to the right backend")
```
YuviGold, line 1: Why another programming language? Although it's pretty readable, I don't think we need to support more.

osherdp: I didn't choose it, that's what HAproxy works with...

PR #492, opened 2021-02-04
Add download_all_logs_minimal target to makefile

oshercc:


YuviGold:
Changed download_logs targets on https://github.com/openshift/assisted-test-infra/pull/523.

PR #494, opened 2021-02-07
Add new interface to VM

lalon4:
This PR implement new methods to create new Network, create new Interface, and attach it to a single node.

### DIFF
--- a/discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
+++ b/discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
@@ -1,7 +1,6 @@
 import os
 import re
 import string
-import uuid
 import logging
 import tempfile
 from abc import ABC
@@ -28,7 +27,7 @@ class LibvirtController(NodeController, ABC):
     def __del__(self):
         with suppress(Exception):
             self.libvirt_connection.close()
-            
+
     @property
     def setup_time(self):
         return self._setup_timestamp
@@ -239,6 +238,80 @@ class LibvirtController(NodeController, ABC):
             assert source_file.startswith(
                 tempfile.gettempdir()), "File unexpectedly not in tmp, avoiding deletion to be on the safe side"
             os.remove(source_file)
+
+    def attach_interface(self, node_name, network_xml, target_interface=consts.TEST_TARGET_INTERFACE):
+        """
+        Create network and Interface. New interface will be attached to a given node.
+        """
+        network = self.create_network(network_xml)
+        interface_mac = self.add_interface(node_name, network.bridgeName(), target_interface)
+        return network, interface_mac
+
+    def create_network(self, network_xml):
+        """
+        Create a network from a given xml and return libvirt.virNetwork object
+        """
+        logging.info(f"Creating new network: {network_xml}")
+        network = self.libvirt_connection.networkCreateXML(network_xml)
+        if network is None:
+            raise Exception(f"Failed to create network: {network_xml}")
+        active = network.isActive()
+        if active != 1:
+            self.destroy_network(network)
+            raise Exception(f"Failed to activate network: {network_xml}")
+        logging.info(f"Successfully created and activated network. name: {network.name()}")
+        return network
+
+    def get_network_by_name(self, network_name):
+        """
+        Get network name and return libvirt.virNetwork object
+        """
+        return self.libvirt_connection.networkLookupByName(network_name)
+
+    def destroy_network(self, network):
+        """
+        Destroy network of a given libvirt.virNetwork object
+        """
+        logging.info(f"Destroy network: {network.name()}")
+        network.destroy()
+
+    def add_interface(self, node_name, network_name, target_interface):
+        """
+        Create an interface using given network name, return created interface's mac address.
+        Note: Do not use the same network for different tests
+        """
+        logging.info(f"Creating new interface attached to network: {network_name}, for node: {node_name}")
+        net_leases = self.list_leases(network_name)
+        mac_addresses = []
+        for lease in net_leases:
+            mac_addresses.append(lease['mac'])
+        command = f"virsh attach-interface {node_name} network {network_name} --target {target_interface} --persistent"
+        utils.run_command(command)
+        try:
+            waiting.wait(
+                    lambda: len(self.list_leases(network_name)) > len(mac_addresses),
+                    timeout_seconds=30,
+                    sleep_seconds=2,
+                    waiting_for="Wait for network lease"
+            )
+        except waiting.exceptions.TimeoutExpired:
+            logging.error("Network lease wasnt found for added interface")
+            raise
+
+        new_net_leases = self.list_leases(network_name)
+        for lease in new_net_leases:
+            if not lease['mac'] in mac_addresses:
+                mac_address = lease['mac']
+                break
+        logging.info(f"Successfully attached interface, network: {network_name}, mac: {mac_address}, for node:"
+                     f" {node_name}")
+        return mac_address
+
+    def undefine_interface(self, node_name, mac):
+        logging.info(f"Undefining an interface mac: {mac}, for node: {node_name}")
+        command = f"virsh detach-interface {node_name} --type network --mac {mac}"
+        utils.run_command(command, True)
+        logging.info(f"Successfully removed interface.")
 
     def restart_node(self, node_name):
         logging.info("Restarting %s", node_name)
--- a/discovery-infra/test_infra/controllers/node_controllers/node.py
+++ b/discovery-infra/test_infra/controllers/node_controllers/node.py
@@ -145,3 +145,21 @@ class Node:
 
     def detach_all_test_disks(self):
         self.node_controller.detach_all_test_disks(self.name)
+
+    def attach_interface(self, network_xml, target_interface=consts.TEST_TARGET_INTERFACE):
+        return self.node_controller.attach_interface(self.name, network_xml, target_interface)
+
+    def add_interface(self, network_name, target_interface=consts.TEST_TARGET_INTERFACE):
+        return self.node_controller.add_interface(self.name, network_name, target_interface)
+
+    def create_network(self, network_xml):
+        return self.node_controller.create_network(network_xml)
+
+    def get_network_by_name(self, network_name):
+        return self.node_controller.get_network_by_name(network_name)
+
+    def destroy_network(self, network):
+        self.node_controller.destroy_network(network)
+
+    def undefine_interface(self, mac):
+        self.node_controller.undefine_interface(self.name, mac)
--- a/discovery-infra/test_infra/controllers/node_controllers/node_controller.py
+++ b/discovery-infra/test_infra/controllers/node_controllers/node_controller.py
@@ -1,3 +1,4 @@
+import libvirt
 from typing import Dict, List, Any, Tuple
 from test_infra.controllers.node_controllers.node import Node
 from abc import ABC, abstractmethod
@@ -115,4 +116,28 @@ class NodeController(ABC):
 
     @abstractmethod
     def get_machine_cidr(self) -> str:
+        pass
+
+    @abstractmethod
+    def attach_interface(self, node_name, network_xml: str) -> Tuple[libvirt.virNetwork, str]:
+        pass
+
+    @abstractmethod
+    def add_interface(self, node_name, network_name, target_interface: str) -> str:
+        pass
+
+    @abstractmethod
+    def undefine_interface(self, node_name: str, mac: str):
+        pass
+
+    @abstractmethod
+    def create_network(self, network_xml: str) -> libvirt.virNetwork:
+        pass
+
+    @abstractmethod
+    def get_network_by_name(self, network_name: str) -> libvirt.virNetwork:
+        pass
+
+    @abstractmethod
+    def destroy_network(self, network: libvirt.virNetwork):
         pass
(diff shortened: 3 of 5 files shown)


discovery-infra/tests/base_test.py
```diff
@@ -138,6 +138,21 @@ def attach(node, disk_size, bootable=False):
...
+        def add(node, network_xml):
+            network, interface_mac = node.attach_interface(node.name, network_xml)
+            added_networks.append({"node": node, "network": network, "mac": interface_mac})
+
+        yield add
+        for added_network in added_networks:
+            logging.info(f'Deleting custom networks:{added_networks}')
+            node_obj = added_network.get("node")
+            node_obj.undefine_interface(node_obj.name, added_network.get("mac"))
+            node_obj.destroy_network(added_network.get("network"))
```
nshidlin, line 154: What happens if more than one node has an interface on the same network?

discovery-infra/test_infra/controllers/node_controllers/node.py
```diff
@@ -145,3 +145,12 @@ def attach_test_disk(self, disk_size, bootable=False):
 
     def detach_all_test_disks(self):
         self.node_controller.detach_all_test_disks(self.name)
+
+    def attach_interface(self, node_name, network_xml, target_interface="vnet3"):
+        return self.node_controller.attach_interface(node_name, network_xml, target_interface)
+
+    def destroy_network(self, network):
```
nshidlin, line 152: I don't think destroy network should be at `node` level
How about moving into `nodes`?

discovery-infra/test_infra/controllers/node_controllers/node.py
```diff
@@ -145,3 +145,12 @@ def attach_test_disk(self, disk_size, bootable=False):
 
     def detach_all_test_disks(self):
         self.node_controller.detach_all_test_disks(self.name)
+
+    def attach_interface(self, node_name, network_xml, target_interface="vnet3"):
```
nshidlin, line 149: Same, I'm not sure if the network creation should be at the `node` level

discovery-infra/test_infra/controllers/node_controllers/node.py
```diff
@@ -145,3 +145,12 @@ def attach_test_disk(self, disk_size, bootable=False):
...
+        return self.node_controller.attach_interface(node_name, network_xml, target_interface)
+
+    def destroy_network(self, network):
+        self.node_controller.destroy_network(network)
+
+    def undefine_interface(self, node_name, mac):
```
nshidlin, line 155: why not `self.node` instead of passing `node_name` to the method?

discovery-infra/tests/base_test.py
lalon4: `destroy_network` run a command so i think nothing will happen

discovery-infra/test_infra/controllers/node_controllers/node.py
lalon4: Well i agree that network is more nodes related and i dont mind to change it. But isnt it more intuitive to `node.attach_interface`? i mean, what reason do we have to create a network w/o attaching it to a specific node interface? 
this also saves the need to pass and save both nodes and node (on test level, and on base).

discovery-infra/test_infra/controllers/node_controllers/node.py
lalon4: fixed

discovery-infra/tests/base_test.py
nshidlin: Can you destroy a network with an interface attached?

discovery-infra/test_infra/controllers/node_controllers/node.py
nshidlin: `node.attach_interface` can get network name of an already created interface that you can attach another interface on the same network

discovery-infra/tests/base_test.py
lalon4: havnt tried but i am destroying the interface and than the network

discovery-infra/test_infra/controllers/node_controllers/node.py
lalon4: Fixed the code by allowing attach_interface fixture to support network and interface creation OR interface creation by given existing network name. 

I think network need to be under node, its more intuitive for me.
Also if changing to nodes, than attach_interface method should be located under nodes, which is also not accurate. 
LMK if you disagree

discovery-infra/test_infra/controllers/node_controllers/node_controller.py
```diff
@@ -116,3 +116,27 @@ def set_ram_kib(self, node_name: str, ram_kib: int) -> None:
...
+    @abstractmethod
+    def undefine_interface(self, node_name: str, mac: str):
+        pass
+
+    @abstractmethod
+    def create_network(self, network_xml: str) -> str:
+        pass
+
+    @abstractmethod
+    def get_network_by_name(self, network_name: str) -> str:
```
YuviGold, line 137: So what's `get_cluster_network`? I find it a little bit confusing.

discovery-infra/test_infra/controllers/node_controllers/node_controller.py
```diff
@@ -116,3 +116,27 @@ def set_ram_kib(self, node_name: str, ram_kib: int) -> None:
...
+        pass
+
+    @abstractmethod
+    def destroy_network(self, network: Any):
```
YuviGold, line 141: `Any`? What types do you expect over here?

discovery-infra/test_infra/controllers/node_controllers/node_controller.py
YuviGold: Moreover, from the signature of the function it sound like it's implementation should be:
```python
def get_network_by_name(self, network_name: str) -> str:
    return network_name
```
Because it gets a network name and it returns the network. Or am I missing something?

discovery-infra/tests/base_test.py
```diff
@@ -138,6 +138,26 @@ def attach(node, disk_size, bootable=False):
...
+                network, interface_mac = node.attach_interface(network_xml)
+            elif network_name:
+                interface_mac = node.add_interface(network_name)
+                network = node.get_network_by_name(network_name)
+            added_networks.append({"node": node, "network": network, "mac": interface_mac})
+
+        yield add
+        for added_network in added_networks:
+            logging.info(f'Deleting custom networks:{added_networks}')
+            with suppress(Exception):
```
YuviGold, line 156: Please use specific exceptions.

discovery-infra/test_infra/controllers/node_controllers/node.py
```diff
@@ -145,3 +145,21 @@ def attach_test_disk(self, disk_size, bootable=False):
 
     def detach_all_test_disks(self):
         self.node_controller.detach_all_test_disks(self.name)
+
+    def attach_interface(self, network_xml, target_interface="vnet3"):
```
YuviGold, line 149: I am not sure the default target interface should be defined on the node controller level.

discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
```diff
@@ -240,6 +239,79 @@ def detach_all_test_disks(self, node_name):
...
+        logging.info(f"Creating new interface attached to network: {network_name}, for node: {node_name}")
+        net_leases = self.list_leases(network_name)
+        mac_addresses = []
+        for lease in net_leases:
+            mac_addresses.append(lease['mac'])
+        command = f"virsh attach-interface {node_name} network {network_name} --target {target_interface} --persistent"
+        utils.run_command(command)
+        try:
+            waiting.wait(
+                    lambda: len(self.list_leases(network_name)) > len(mac_addresses),
```
YuviGold, line 291: In case of running the `add_interface` function in parallel, this would cause a race-condition.
You should verify the `target_interface` was created.

discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
```diff
@@ -240,6 +239,79 @@ def detach_all_test_disks(self, node_name):
...
+    def create_network(self, network_xml):
+        """
+        Create a network from a given xml and return libvirt.virNetwork object
+        """
+        logging.info(f"Creating new network: {network_xml}")
+        network = self.libvirt_connection.networkCreateXML(network_xml)
+        if network is None:
+            raise Exception(f"Failed to create network: {network_xml}")
+        active = network.isActive()
+        if active is not 1:
```
YuviGold, line 259: You shouldn't use `is` to compare integers. It would work for numbers <=256, but it's bad practice.
Explanation - https://stackoverflow.com/questions/306313/is-operator-behaves-unexpectedly-with-integers

discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
```diff
@@ -240,6 +239,79 @@ def detach_all_test_disks(self, node_name):
                 tempfile.gettempdir()), "File unexpectedly not in tmp, avoiding deletion to be on the safe side"
             os.remove(source_file)
 
+    def attach_interface(self, node_name, network_xml, target_interface="vnet3"):
```
YuviGold, line 242: Define `vnet3` as a constant variable.

discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
lalon4: The issue is that the taken target is not the actual. the machine set different target, so it make it difficult to understand what interface was added

discovery-infra/test_infra/controllers/node_controllers/node.py
lalon4: its not in the node controller, its in node :)

discovery-infra/test_infra/controllers/node_controllers/node_controller.py
lalon4: Your right. will fix

discovery-infra/test_infra/controllers/node_controllers/node_controller.py
lalon4: i expect to get libvirt network object

discovery-infra/tests/base_test.py
lalon4: but there could be few

discovery-infra/test_infra/controllers/node_controllers/node_controller.py
lalon4: we first get to node:
```
def get_network_by_name(self, network_name):
        return self.node_controller.get_network_by_name(network_name)
```
which call to node_controller, which pass it to libvirt controller. 
So the node will return the value

discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
lalon4: Test will be flaky only if they will use the same network (two different tests use same network). which should not happen.

discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
lalon4: fixed

discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
lalon4: fixed

discovery-infra/tests/base_test.py
YuviGold: So address all of them

discovery-infra/test_infra/controllers/node_controllers/node_controller.py
YuviGold: So an integer or a string wouldn't work. Is there a class for that object? is that a dictionary?

discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
YuviGold: So that's no biggy. Maybe just add a comment about that scenario so it won't be a surprise.

discovery-infra/test_infra/controllers/node_controllers/node_controller.py
YuviGold: In case it returns a network libvirt object, then you should change the returned type from `str`.

discovery-infra/test_infra/controllers/node_controllers/node_controller.py
lalon4: libvirt.virNetwork, ill import libvirt and use that

discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
lalon4: added

discovery-infra/test_infra/controllers/node_controllers/node_controller.py
lalon4: fixed

discovery-infra/test_infra/controllers/node_controllers/node_controller.py
lalon4: fixed

discovery-infra/tests/base_test.py
lalon4: In this case we need it to be general Exception, as undefine_interface uses run_command while destroy_network uses libvirt object

YuviGold (approved):

PR #497, opened 2021-02-07
NO-ISSUE Pin cryptography Python dependency to 3.3.2

empovit:
Pin the cryptography dependency to the latest version before 3.4 to fix
ModuleNotFoundError: No module named 'setuptools_rust'

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
-
 boto3==1.16.63
 dnspython==2.1.0
 ipython==7.11.1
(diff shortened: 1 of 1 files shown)


empovit:
See https://github.com/pyca/cryptography/issues/5753. However, upgrading pip as suggested causes ```ERROR: Cannot uninstall 'PyYAML'. It is a distutils installed project and thus we cannot accurately determine which files belong to it which would lead to only a partial uninstall.```. Downgrading cryptography until a better solution is found.

YuviGold (approved):
LGTM

PR #498, opened 2021-02-07
MGMT-4043: Add a post install check for metal3 enablement

asalkeld:


discovery-infra/oc_utils.py
```diff
@@ -197,3 +197,37 @@ def get_operators_status(kubeconfig):
             statuses[name] = False
 
     return statuses
+
+
+def check_metal3(kubeconfig):
+    def oc(*args):
+        command = ["/usr/local/bin/oc",
```
tsorya, line 204: can we have this command as function for all ?

PR #502, opened 2021-02-08
Disable test teardown

lalon4:
Adding the ability to set test_teardown true/false as environment variable. Its mainly use for development, when we need the cluster and nodes to stay up.


YuviGold (approved):
LGTM

PR #504, opened 2021-02-08
NO-ISSUE Fix setuptools_rust module not found

empovit:
Upgrade pip to solve `No module named 'setuptools_rust'`, 
and force-ignore installed PyYAML to solve 
```
ERROR: Cannot uninstall 'PyYAML'. It is a distutils installed project and 
thus we cannot accurately determine which files belong to it which would 
lead to only a partial uninstall.
```


empovit:
This PR is supposed to replace https://github.com/openshift/assisted-test-infra/pull/497

YuviGold (approved):

PR #505, opened 2021-02-08
MGMT-3370: Allow running in kube api mode

RazRegev:
By passing ENABLE_KUBE_API env.
Use the skipper image of assisted-service which has all
the relevant dependencies required to generate and deploy
relevant definitions and deploy the service with the kube
controller enabled.

### DIFF
--- a/scripts/deploy_assisted_service.sh
+++ b/scripts/deploy_assisted_service.sh
@@ -14,7 +14,10 @@ export PROFILE=${PROFILE:-assisted-installer}
 export OCP_SERVICE_PORT=$(( 7000 + $NAMESPACE_INDEX ))
 export OPENSHIFT_INSTALL_RELEASE_IMAGE=${OPENSHIFT_INSTALL_RELEASE_IMAGE:-}
 export PUBLIC_CONTAINER_REGISTRIES=${PUBLIC_CONTAINER_REGISTRIES:-}
-
+export ENABLE_KUBE_API_CMD=""
+if [ ! -z ${ENABLE_KUBE_API:-} ]; then
+ENABLE_KUBE_API_CMD="ENABLE_KUBE_API=true"
+fi
 mkdir -p build
 
 if [ "${DEPLOY_TARGET}" == "onprem" ]; then
@@ -50,7 +53,7 @@ elif [ "${DEPLOY_TARGET}" == "ocp" ]; then
 else
     print_log "Updating assisted_service params"
     skipper run discovery-infra/update_assisted_service_cm.py ENABLE_AUTH=${ENABLE_AUTH}
-    skipper run "make -C assisted-service/ deploy-all" ${SKIPPER_PARAMS} DEPLOY_TAG=${DEPLOY_TAG} DEPLOY_MANIFEST_PATH=${DEPLOY_MANIFEST_PATH} DEPLOY_MANIFEST_TAG=${DEPLOY_MANIFEST_TAG} NAMESPACE=${NAMESPACE} ENABLE_AUTH=${ENABLE_AUTH} PROFILE=${PROFILE}
+    cd assisted-service/ && skipper run "make deploy-all" ${SKIPPER_PARAMS} $ENABLE_KUBE_API_CMD DEPLOY_TAG=${DEPLOY_TAG} DEPLOY_MANIFEST_PATH=${DEPLOY_MANIFEST_PATH} DEPLOY_MANIFEST_TAG=${DEPLOY_MANIFEST_TAG} NAMESPACE=${NAMESPACE} ENABLE_AUTH=${ENABLE_AUTH} PROFILE=${PROFILE}
 
     print_log "Wait till ${SERVICE_NAME} api is ready"
     wait_for_url_and_run "$(minikube service ${SERVICE_NAME} --url -p $PROFILE -n ${NAMESPACE})" "echo \"waiting for ${SERVICE_NAME}\""


scripts/deploy_assisted_service.sh
```diff
@@ -50,7 +51,7 @@ elif [ "${DEPLOY_TARGET}" == "ocp" ]; then
 else
     print_log "Updating assisted_service params"
     skipper run discovery-infra/update_assisted_service_cm.py ENABLE_AUTH=${ENABLE_AUTH}
-    skipper run "make -C assisted-service/ deploy-all" ${SKIPPER_PARAMS} DEPLOY_TAG=${DEPLOY_TAG} DEPLOY_MANIFEST_PATH=${DEPLOY_MANIFEST_PATH} DEPLOY_MANIFEST_TAG=${DEPLOY_MANIFEST_TAG} NAMESPACE=${NAMESPACE} ENABLE_AUTH=${ENABLE_AUTH} PROFILE=${PROFILE}
+    cd assisted-service/ && skipper run "make deploy-all" ${SKIPPER_PARAMS} ENABLE_KUBE_API=${ENABLE_KUBE_API} DEPLOY_TAG=${DEPLOY_TAG} DEPLOY_MANIFEST_PATH=${DEPLOY_MANIFEST_PATH} DEPLOY_MANIFEST_TAG=${DEPLOY_MANIFEST_TAG} NAMESPACE=${NAMESPACE} ENABLE_AUTH=${ENABLE_AUTH} PROFILE=${PROFILE}
```
filanov, line 54: why do you cd to the directory and not running `skipper run "make -C assisted-service/ deploy-all"` ?

scripts/deploy_assisted_service.sh
```diff
@@ -14,6 +14,7 @@ export PROFILE=${PROFILE:-assisted-installer}
 export OCP_SERVICE_PORT=$(( 7000 + $NAMESPACE_INDEX ))
 export OPENSHIFT_INSTALL_RELEASE_IMAGE=${OPENSHIFT_INSTALL_RELEASE_IMAGE:-}
 export PUBLIC_CONTAINER_REGISTRIES=${PUBLIC_CONTAINER_REGISTRIES:-}
+export ENABLE_KUBE_API=${ENABLE_KUBE_API:-}
```
filanov, line 17: i would expect it to be a boolean and not a full command that user need to set.

tsorya: sounds good

scripts/deploy_assisted_service.sh
RazRegev: The reason I did it because it currently runs with a test-infra skipper image.
If we want to have the dependencies to generate manifests we need to use an assisted-service build image.
I think regardless of this specific need that this would be more appropriate to use an assisted-service build image to deploy assisted-service.

filanov: Is it because of the controller-gen? 
@tsorya is there a specific reason why assisted build image wasn't used to begin with?

RazRegev: yes, the controller gen is the main reason

tsorya: sounds logical i must say but it will require to create 2 build images in every deployment and will take very long time

filanov: Then to make it work with single build container we need to install some things from this commit https://github.com/openshift/assisted-service/commit/697a281f3a06b8314d7ed74e9a8985d6e142b8f3
Do you think it will better?

scripts/deploy_assisted_service.sh
```diff
@@ -14,7 +14,10 @@ export PROFILE=${PROFILE:-assisted-installer}
 export OCP_SERVICE_PORT=$(( 7000 + $NAMESPACE_INDEX ))
 export OPENSHIFT_INSTALL_RELEASE_IMAGE=${OPENSHIFT_INSTALL_RELEASE_IMAGE:-}
 export PUBLIC_CONTAINER_REGISTRIES=${PUBLIC_CONTAINER_REGISTRIES:-}
-
+export ENABLE_KUBE_API_CMD=""
+if [ ! -z ${ENABLE_KUBE_API:-} ]; then
+ENABLE_KUBE_API_CMD="ENABLE_KUBE_API=true"
```
osherdp, line 19: add indentation

scripts/deploy_assisted_service.sh
```diff
@@ -50,7 +53,7 @@ elif [ "${DEPLOY_TARGET}" == "ocp" ]; then
 else
     print_log "Updating assisted_service params"
     skipper run discovery-infra/update_assisted_service_cm.py ENABLE_AUTH=${ENABLE_AUTH}
-    skipper run "make -C assisted-service/ deploy-all" ${SKIPPER_PARAMS} DEPLOY_TAG=${DEPLOY_TAG} DEPLOY_MANIFEST_PATH=${DEPLOY_MANIFEST_PATH} DEPLOY_MANIFEST_TAG=${DEPLOY_MANIFEST_TAG} NAMESPACE=${NAMESPACE} ENABLE_AUTH=${ENABLE_AUTH} PROFILE=${PROFILE}
+    cd assisted-service/ && skipper run "make deploy-all" ${SKIPPER_PARAMS} $ENABLE_KUBE_API_CMD DEPLOY_TAG=${DEPLOY_TAG} DEPLOY_MANIFEST_PATH=${DEPLOY_MANIFEST_PATH} DEPLOY_MANIFEST_TAG=${DEPLOY_MANIFEST_TAG} NAMESPACE=${NAMESPACE} ENABLE_AUTH=${ENABLE_AUTH} PROFILE=${PROFILE}
```
osherdp, line 56: maybe wrap it all with parenthesis so that the ``cd`` is relevant to this line alone

osherdp (changes requested):

scripts/deploy_assisted_service.sh
filanov: Instead of doing this, maybe lets install the dependencies https://github.com/openshift/assisted-service/pull/829/files and install them in test-infra.
Then we will not need two build containers.

RazRegev:
closing, I added the dependencies to the test-infra skipper after discussing with @filanov and understood these changes aren't heavy as we thought. @tsorya 
https://github.com/openshift/assisted-test-infra/pull/521

PR #506, opened 2021-02-08
Add get_host_disks cluster method

trewest:
Adds `get_host_disks` cluster method for retrieving host disk information. Includes an optional filter that can be a regular function or a lambda function. Returns a list of disks matching the filter criteria, or if no filter is included, returns all disks.

### DIFF
--- a/discovery-infra/test_infra/helper_classes/cluster.py
+++ b/discovery-infra/test_infra/helper_classes/cluster.py
@@ -721,6 +721,15 @@ class Cluster:
                 return ip
         raise Exception("IP for single node IPv6 not found")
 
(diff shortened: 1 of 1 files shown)


discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -721,6 +721,14 @@ def get_ip_for_single_node(client, cluster_id, machine_cidr, ipv4_first=True):
                 return ip
         raise Exception("IP for single node IPv6 not found")
 
+    def get_host_disks(self, node, nodes, filter=None):
```
lalon4, line 724: This need to get host

PR #509, opened 2021-02-10
MGMT-3611 Pass proxy configuration for day 2 cluster

empovit:
Allow passing a proxy configuration to day 2 clusters.
If not present, it will be copied from the original day 1 cluster.


empovit:
This PR depends on https://github.com/openshift/assisted-service/pull/1046

PR #514, opened 2021-02-14
MGMT-4096: add canary-openshift-ingress DNS entry on terraform to enable ingress operator full operation

osherdp:
/hold
/cc @eranco74 @romfreiman

### DIFF
--- a/terraform_files/baremetal/main.tf
+++ b/terraform_files/baremetal/main.tf
@@ -39,6 +39,7 @@ resource "libvirt_network" "net" {
         data.libvirt_network_dns_host_template.api-int.*.rendered,
         data.libvirt_network_dns_host_template.oauth.*.rendered,
         data.libvirt_network_dns_host_template.console.*.rendered,
+        data.libvirt_network_dns_host_template.canary.*.rendered,
       )
       content {
         hostname = hosts.value.hostname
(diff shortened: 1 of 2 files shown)


terraform_files/baremetal/main.tf
```diff
@@ -191,3 +192,9 @@ data "libvirt_network_dns_host_template" "console" {
   ip       = var.single_node_ip
   hostname = "console-openshift-console.apps.${var.cluster_name}.${var.cluster_domain}"
 }
+
+data "libvirt_network_dns_host_template" "canary" {
+  count    = var.bootstrap_in_place && var.single_node_ip != "" ? 1 : 0
+  ip       = var.single_node_ip
+  hostname = "canary-openshift-ingress-canary.apps.${var.cluster_name}.${var.cluster_domain}"
```
romfreiman, line 199: maybe always? Why do u think it's specific to single node? I would say it's specific for none platform

osherdp: we should do some cleanup, because single-node uses the baremetal template and the other none-platform flow uses a separate template (yes, it's pretty ugly)
I guess we should first decide if we're building single-node above none platform or baremetal, and then do some cleanup in our templates

Either way I cannot add it to none template without it being covered on prow/jenkins

osherdp: anyway added to the none platform template

RazRegev (approved):

PR #520, opened 2021-02-15
MGMT-3370: Added KubeAPI helpers for e2e test infra

RazRegev:
Provides infra to create and destroy a ClusterDeployment CRD
to be used by e2e tests.

There's a PR for e2e test using this infra as well:
https://gitlab.cee.redhat.com/ocp-edge-qe/kni-assisted-installer-auto/-/merge_requests/112

### DIFF
--- a/discovery-infra/test_infra/helper_classes/kube_helpers.py
+++ b/discovery-infra/test_infra/helper_classes/kube_helpers.py
@@ -1,0 +1,531 @@
+"""
+kube_helpers.py provides infra to deploy, manage and install cluster using
+CRDs instead of restful API calls.
+
+Simplest use of this infra is performed by calling cluster_deployment_context
+fixture. With this context manager you will be able to manage a cluster
+without need to handle registration and deregistration.
+
+Example of usage:
+
+with cluster_deployment_context(kube_api_client) as cluster_deployment:
+    print(cluster_deployment.status())
+
+When a ClusterDeployment has sufficient data, installation will be started
+automatically.
+"""
+
+import logging
+import os
+import abc
+import contextlib
+import json
+
+import waiting
+
+from pprint import pformat
+from typing import Optional, Union, Dict, Tuple, ContextManager
+
+from kubernetes.config import load_kube_config
+from kubernetes.config.kube_config import Configuration
+from kubernetes.client import ApiClient, CoreV1Api, CustomObjectsApi
+from kubernetes.client.rest import ApiException
+
+from tests.conftest import env_variables
+from test_infra.utils import get_random_name
+
+
+# silence kubernetes debug messages.
+logging.getLogger('kubernetes').setLevel(logging.INFO)
+
+logger = logging.getLogger(__name__)
+
+
+def create_kube_api_client(kubeconfig_path: Optional[str] = None) -> ApiClient:
+    logger.info('creating kube client with config file: %s', kubeconfig_path)
+
+    conf = Configuration()
+    load_kube_config(config_file=kubeconfig_path, client_configuration=conf)
+    return ApiClient(configuration=conf)
+
+
+def _does_string_contain_value(s: Union[str, None]) -> bool:
+    return s and s != '""'
+
+
+class ObjectReference:
+    """
+    A class that contains the information required to to let you locate a
+    referenced Kube API resource.
+    """
+    def __init__(self, name: str, namespace: str):
+        self.name = name
+        self.namespace = namespace
+
+    def __repr__(self) -> str:
+        return str(self.as_dict())
+
+    def as_dict(self) -> dict:
+        return {'name': self.name, 'namespace': self.namespace}
+
+
+class Secret:
+    """
+    A Kube API secret resource that consists of the pull secret data, used
+    by a ClusterDeployment CRD.
+    """
+    def __init__(
+            self,
+            kube_api_client: ApiClient,
+            name: str,
+            namespace: str = env_variables['namespace']
+    ):
+        self.v1_api = CoreV1Api(kube_api_client)
+        self._reference = ObjectReference(name=name, namespace=namespace)
+
+    @property
+    def ref(self) -> ObjectReference:
+        return self._reference
+
+    def create(self, pull_secret: str = env_variables['pull_secret']) -> None:
+        self.v1_api.create_namespaced_secret(
+            body={
+                'apiVersion': 'v1',
+                'kind': 'Secret',
+                'metadata': self.ref.as_dict(),
+                'stringData': {'pullSecret': pull_secret}
+            },
+            namespace=self.ref.namespace
+        )
+
+        logger.info('created secret %s', self.ref)
+
+    def delete(self) -> None:
+        self.v1_api.delete_namespaced_secret(
+            name=self.ref.name,
+            namespace=self.ref.namespace
+        )
+
+        logger.info('deleted secret %s', self.ref)
+
+
+class BaseCustomResource(abc.ABC):
+    """
+    Base class for all CRDs, enforces basic methods that every resource must
+    have e.g create, path, get, delete and status.
+    """
+    def __init__(self,  name: str, namespace: str):
+        self._reference = ObjectReference(name=name, namespace=namespace)
+
+    @property
+    def ref(self) -> ObjectReference:
+        return self._reference
+
+    @abc.abstractmethod
+    def create(self, **kwargs) -> None:
+        pass
+
+    @abc.abstractmethod
+    def patch(self, **kwargs) -> None:
+        pass
+
+    @abc.abstractmethod
+    def get(self) -> dict:
+        pass
+
+    @abc.abstractmethod
+    def delete(self) -> None:
+        pass
+
+    @abc.abstractmethod
+    def status(self, timeout: Union[int, float]) -> dict:
+        pass
+
+
+DEFAULT_API_VIP = env_variables.get('api_vip', '')
+DEFAULT_API_VIP_DNS_NAME = env_variables.get('api_vip_dns_name', '')
+DEFAULT_INGRESS_VIP = env_variables.get('ingress_vip', '')
+
+
+class Platform:
+    """
+    A class that represents the configuration for the specific platform upon
+    which to perform the installation.
+    """
+    def __init__(
+            self,
+            api_vip: str = DEFAULT_API_VIP,
+            api_vip_dns_name: str = DEFAULT_API_VIP_DNS_NAME,
+            ingress_vip: str = DEFAULT_INGRESS_VIP,
+            vip_dhcp_allocation: bool = env_variables['vip_dhcp_allocation']
+    ):
+        self.api_vip = api_vip
+        self.api_vip_dns_name = api_vip_dns_name
+        self.ingress_vip = ingress_vip
+        self.vip_dhcp_allocation = vip_dhcp_allocation
+
+    def __repr__(self) -> str:
+        return str(self.as_dict())
+
+    def as_dict(self) -> dict:
+        vip_dhcp_allocation = 'Enabled' if self.vip_dhcp_allocation else ''
+        data = {
+            'agentBareMetal': {
+                'apiVIP': self.api_vip,
+                'ingressVIP': self.ingress_vip,
+                'VIPDHCPAllocation': vip_dhcp_allocation
+            }
+        }
+
+        if self.api_vip_dns_name:
+            data['agentBareMetal']['apiVIPDNSName'] = self.api_vip_dns_name
+
+        return data
+
+
+DEFAULT_MACHINE_CIDR = env_variables.get('machine_cidr', '')
+DEFAULT_CLUSTER_CIDR = env_variables.get('cluster_cidr', '172.30.0.0/16')
+DEFAULT_SERVICE_CIDR = env_variables.get('service_cidr', '10.128.0.0/14')
+
+
+class InstallStrategy:
+    """
+    A class that provides platform agnostic configuration for the use of
+    alternate install strategies.
+    """
+    def __init__(
+            self,
+            host_prefix: int = env_variables['host_prefix'],
+            machine_cidr: str = DEFAULT_MACHINE_CIDR,
+            cluster_cidr: str = DEFAULT_CLUSTER_CIDR,
+            service_cidr: str = DEFAULT_SERVICE_CIDR,
+            ssh_public_key: str = env_variables['ssh_public_key'],
+            control_plane_agents: int = 1,
+            worker_agents: int = 0,
+            label_selector: Optional[Dict[str, str]] = None
+    ):
+        self.host_prefix = host_prefix
+        self.machine_cidr = machine_cidr
+        self.cluster_cidr = cluster_cidr
+        self.service_cidr = service_cidr
+        self.control_plane_agents = control_plane_agents
+        self.worker_agents = worker_agents
+        self.label_selector = label_selector
+        self.ssh_public_key = ssh_public_key
+
+    def __repr__(self) -> str:
+        return str(self.as_dict())
+
+    def as_dict(self) -> dict:
+        data = {
+            'agent': {
+                'networking': {
+                    'clusterNetwork': [{
+                        'cidr': self.cluster_cidr,
+                        'hostPrefix': self.host_prefix
+                    }],
+                    'serviceNetwork': [self.service_cidr]
+                },
+                'provisionRequirements': {
+                    'controlPlaneAgents': self.control_plane_agents,
+                    'workerAgents': self.worker_agents,
+                },
+                'agentSelector': {'matchLabels': self.label_selector or {}}
+            }
+        }
+
+        if _does_string_contain_value(self.ssh_public_key):
+            data['agent']['sshPublicKey'] = self.ssh_public_key
+
+        if _does_string_contain_value(self.machine_cidr):
+            data['agent']['networking']['machineNetwork'] = [{
+                'cidr': self.machine_cidr
+            }]
+
+        return data
+
+
+DEFAULT_WAIT_FOR_CRD_STATUS_TIMEOUT = 60
+DEFAULT_WAIT_FOR_CRD_STATE_TIMEOUT = 300
+
+
+class ClusterDeployment(BaseCustomResource):
+    """
+    A CRD that represents cluster in assisted-service.
+    On creation the cluster will be registered to the service.
+    On deletion it will be unregistered from the service.
+    When has sufficient data installation will start automatically.
+    """
+
+    _hive_api_group = 'hive.openshift.io'
+    _plural = 'clusterdeployments'
+
+    def __init__(
+            self,
+            kube_api_client: ApiClient,
+            name: str,
+            namespace: str = env_variables['namespace']
+    ):
+        BaseCustomResource.__init__(self, name, namespace)
+        self.crd_api = CustomObjectsApi(kube_api_client)
+        self._assigned_secret = None
+
+    @property
+    def secret(self):
+        return self._assigned_secret
+
+    def create(
+            self,
+            platform: Platform,
+            install_strategy: InstallStrategy,
+            secret: Secret,
+            base_domain: str = env_variables['base_domain'],
+            **kwargs
+    ) -> None:
+        body = {
+            'apiVersion': f'{self._hive_api_group}/v1',
+            'kind': 'ClusterDeployment',
+            'metadata': self.ref.as_dict(),
+            'spec': {
+                'clusterName': self.ref.name,
+                'baseDomain': base_domain,
+                'platform': platform.as_dict(),
+                'provisioning': {'installStrategy': install_strategy.as_dict()},
+                'pullSecretRef': secret.ref.as_dict(),
+            }
+        }
+        body['spec'].update(kwargs)
+        self.crd_api.create_namespaced_custom_object(
+            group=self._hive_api_group,
+            version='v1',
+            plural=self._plural,
+            body=body,
+            namespace=self.ref.namespace
+        )
+        self._assigned_secret = secret
+
+        logger.info(
+            'created cluster deployment %s: %s', self.ref, pformat(body)
+        )
+
+    def patch(
+            self,
+            platform: Optional[Platform] = None,
+            install_strategy: Optional[InstallStrategy] = None,
+            secret: Optional[Secret] = None,
+            **kwargs
+    ) -> None:
+        body = {'spec': kwargs}
+
+        spec = body['spec']
+        if platform:
+            spec['platform'] = platform.as_dict()
+
+        if install_strategy:
+            spec['provisioning']['installStrategy'] = install_strategy.as_dict()
+
+        if secret:
+            spec['pullSecretRef'] = secret.ref.as_dict()
+
+        self.crd_api.patch_namespaced_custom_object(
+            group=self._hive_api_group,
+            version='v1',
+            plural=self._plural,
+            name=self.ref.name,
+            namespace=self.ref.namespace,
+            body=body
+        )
+
+        logger.info(
+            'patching cluster deployment %s: %s', self.ref, pformat(body)
+        )
+
+    def get(self) -> dict:
+        return self.crd_api.get_namespaced_custom_object(
+            group=self._hive_api_group,
+            version='v1',
+            plural=self._plural,
+            name=self.ref.name,
+            namespace=self.ref.namespace
+        )
+
+    def delete(self) -> None:
+        self.crd_api.delete_namespaced_custom_object(
+            group=self._hive_api_group,
+            version='v1',
+            plural=self._plural,
+            name=self.ref.name,
+            namespace=self.ref.namespace
+        )
+
+        logger.info('deleted cluster deployment %s', self.ref)
+
+    def status(
+            self,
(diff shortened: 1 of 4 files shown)


discovery-infra/test_infra/helper_classes/kube_helpers.py
```diff
@@ -0,0 +1,376 @@
...
+                'VIPDHCPAllocation': 'Enabled' if vip_dhcp_allocation else ''
+            }
+        )
+
+        if _does_string_contain_value(api_vip_dns_name):
+            self['agentBareMetal']['apiVIPDNSName'] = api_vip_dns_name,
+
+
+DEFAULT_MACHINE_CIDR = env_variables.get('machine_cidr', '')
+DEFAULT_CLUSTER_CIDR = env_variables.get('cluster_cidr', '172.30.0.0/16')
```
filanov, line 118: you have a new endpoint to the service that can get those values https://github.com/openshift/assisted-service/blob/master/swagger.yaml#L617-L641

filanov (commented):
Just to make things easy to edit maybe you should support deploy from file?

filanov:
@tsorya @YuviGold can you please help with the review?

RazRegev:
@cc @tsorya

discovery-infra/test_infra/helper_classes/kube_helpers.py
```diff
@@ -0,0 +1,376 @@
...
+logging.getLogger('kubernetes').setLevel(logging.INFO)
+
+logger = logging.getLogger(__name__)
+
+
+def _does_string_contain_value(s: Union[str, None]) -> bool:
+    return s and s != '""'
+
+
+class ObjectReference(dict):
```
filanov, line 28: Is it used anywhere?

discovery-infra/test_infra/helper_classes/kube_helpers.py
```diff
@@ -0,0 +1,376 @@
...
+                'platform': platform,
+                'provisioning': {'installStrategy': install_strategy},
+                'pullSecretRef': secret.ref,
+            }
+        }
+        body['spec'].update(kwargs)
+        self.crd_api.create_namespaced_custom_object(
+            group=self.hive_api_group,
+            version='v1',
+            plural='clusterdeployments',
```
filanov, line 203: Probably should be a const? I see it used in multiple places

filanov (commented):
Can you please add a comment or readme about how to run it?

discovery-infra/test_infra/helper_classes/kube_helpers.py
RazRegev: yes, in the test itself, and here internally (ref property)

discovery-infra/test_infra/helper_classes/kube_helpers.py
```diff
@@ -0,0 +1,376 @@
...
+
+    return cluster_deployment
+
+
+def delete_cluster_deployment(
+        cluster_deployment: ClusterDeployment,
+        ignore_not_found: bool = True
+) -> None:
+
+    def _try_to_delete_resource(o):
```
YuviGold, line 350: What's `o`? Please have a meaningful name and set the expected type

discovery-infra/test_infra/helper_classes/kube_helpers.py
```diff
@@ -0,0 +1,376 @@
...
+        try:
+            o.delete()
+        except ApiException as e:
+            if e.reason != 'Not Found' or not ignore_not_found:
```
YuviGold, line 354: I think `if not (e.reason == 'Not Found' and ignore_not_found)` is more readable.

discovery-infra/test_infra/helper_classes/kube_helpers.py
```diff
@@ -0,0 +1,376 @@
...
+        ignore_conflict: bool = True,
+        base_domain: str = env_variables['base_domain'],
+        platform_params: Optional[dict] = None,
+        install_strategy_params: Optional[dict] = None,
+        secret: Optional[Secret] = None,
+        **kwargs
+) -> ClusterDeployment:
+
+    platform = Platform(**platform_params or {})
+    install_strategy = InstallStrategy(**install_strategy_params or {})
```
YuviGold, line 320: Use `dict()` as a default var to avoid unnecessary `None`
```python
        platform_params: dict = dict(),
        install_strategy_params: dict = dict(),
        secret: Optional[Secret] = None,
        **kwargs
) -> ClusterDeployment:

    platform = Platform(**platform_params)
    install_strategy = InstallStrategy(**install_strategy_params)
```

discovery-infra/test_infra/helper_classes/kube_helpers.py
```diff
@@ -0,0 +1,376 @@
...
+
+        if _does_string_contain_value(machine_cidr):
+            self['agent']['machineNetwork'] = [{'cidr': machine_cidr}],
+
+
+DEFAULT_WAIT_FOR_CRD_STATUS_TIMEOUT = 60
+DEFAULT_WAIT_FOR_CRD_STATE_TIMEOUT = 300
+
+
+class ClusterDeployment(_BaseKubeAPIResource):
```
YuviGold, line 164: It seems like this API could be divided into a more generic one for CRDs.
`getCRDStatus`, `waitForCRDState`, etc.

discovery-infra/test_infra/helper_classes/kube_helpers.py
```diff
@@ -0,0 +1,376 @@
...
+    def wait_for_state(
+            self,
+            required_state: str,
+            timeout: Union[int, float] = DEFAULT_WAIT_FOR_CRD_STATE_TIMEOUT
+    ) -> None:
+
+        required_state = required_state.lower()
+
+        def _has_required_state() -> bool:
+            state, _ = self.state(0.5)
```
YuviGold, line 281: Shouldn't it use `timeout` parameter?

discovery-infra/test_infra/helper_classes/kube_helpers.py
```diff
@@ -0,0 +1,376 @@
...
+            control_plane_agents: int = 1,
+            worker_agents: int = 0,
+            label_selector: Optional[Dict[str, str]] = None,
+            host_prefix: int = env_variables['host_prefix'],
+            machine_cidr: str = DEFAULT_MACHINE_CIDR,
+            cluster_cidr: str = DEFAULT_CLUSTER_CIDR,
+            service_cidr: str = DEFAULT_SERVICE_CIDR,
+            ssh_public_key: str = env_variables['ssh_public_key'],
+    ):
+        dict.__init__(
```
YuviGold, line 135: `super().__init__(....)`

discovery-infra/tests/base_test.py
```diff
@@ -280,3 +281,7 @@ def check_status():
             sleep_seconds=90,
             waiting_for="controller to be running",
         )
+
+    @pytest.fixture()
+    def cluster_deployment_context(self):
```
YuviGold, line 286: I think `cluster_deployment` is good for a fixture

discovery-infra/test_infra/helper_classes/kube_helpers.py
```diff
@@ -0,0 +1,376 @@
...
+
+    def create(
+            self,
+            platform: Platform,
+            install_strategy: InstallStrategy,
+            secret: Secret,
+            base_domain: str = env_variables['base_domain'],
+            **kwargs
+    ) -> None:
+
```
YuviGold, line 186: Please avoid a new blank line at the beginning of functions.

discovery-infra/test_infra/helper_classes/kube_helpers.py
```diff
@@ -0,0 +1,376 @@
...
+    def ref(self) -> ObjectReference:
+        return self.__ref
+
+    @property
+    def api_client(self) -> ApiClient:
+        return self._shared_api_client[0]
+
+    @classmethod
+    def _set_api_client_if_needed(cls) -> None:
+        # reuse same client for all subclasses.
```
YuviGold, line 52: I'm not sure it would work as a Singleton for parallel multi-thread tests. You should look for Borg pattern.

discovery-infra/test_infra/helper_classes/kube_helpers.py
```diff
@@ -0,0 +1,376 @@
...
+    return s and s != '""'
+
+
+class ObjectReference(dict):
+
+    def __init__(self, name: str, namespace: str):
+        dict.__init__(self, name=name, namespace=namespace)
+
+
+class _BaseKubeAPIResource(object):
```
YuviGold, line 34: You don't have to explicitly inherit from `object` since all the classes inherit from `object` by default for Python3

discovery-infra/test_infra/helper_classes/kube_helpers.py
```diff
@@ -0,0 +1,376 @@
...
+    def api_client(self) -> ApiClient:
+        return self._shared_api_client[0]
+
+    @classmethod
+    def _set_api_client_if_needed(cls) -> None:
+        # reuse same client for all subclasses.
+        if cls._shared_api_client:
+            return
+
+        conf = Configuration()
```
YuviGold, line 56: Where this configuration is coming from? How does it know what cluster to talk to in case I have multiple clusters/kubeconfigs on my machine?

discovery-infra/test_infra/helper_classes/kube_helpers.py
```diff
@@ -0,0 +1,376 @@
...
+        pull_secret: Optional[str] = None
+) -> Secret:
+
+    if pull_secret is None:
+        pull_secret = os.environ.get('PULL_SECRET', '')
+    secret = Secret(name)
+    try:
+        secret.create(pull_secret)
+    except ApiException as e:
+        if e.reason != 'Conflict' or not ignore_conflict:
```
YuviGold, line 304: I think `if not (e.reason == 'Conflict' and ignore_conflict)` is more readable.

discovery-infra/test_infra/helper_classes/kube_helpers.py
```diff
@@ -0,0 +1,376 @@
...
+            name=self.ref['name'],
+            namespace=self.ref['namespace']
+        )
+
+        logger.info('deleted cluster deployment %s', self.ref)
+
+    def status(
+            self,
+            timeout: Union[int, float] = DEFAULT_WAIT_FOR_CRD_STATUS_TIMEOUT
+    ) -> dict:
```
YuviGold, line 234: Why `status` returns a `dict`?  I find this name misleading.

discovery-infra/test_infra/helper_classes/kube_helpers.py
```diff
@@ -0,0 +1,376 @@
...
+        )
+
+    def state(
+            self,
+            timeout: Union[int, float] = DEFAULT_WAIT_FOR_CRD_STATE_TIMEOUT
+    ) -> Tuple[str, str]:
+
+        state, state_info = None, None
+        for condition in self.status(timeout).get('conditions', []):
+            reason = condition.get('reason')
```
YuviGold, line 257: Does a `condition` always have a `reason` field?

RazRegev: nope, that's why I using get

discovery-infra/test_infra/helper_classes/kube_helpers.py
RazRegev: nope, the timeout is for the whole waiting process and not a single status call

discovery-infra/test_infra/helper_classes/kube_helpers.py
RazRegev: why?

discovery-infra/test_infra/helper_classes/kube_helpers.py
RazRegev: why?

discovery-infra/test_infra/helper_classes/kube_helpers.py
RazRegev: technically it should have but I'm not counting on it

discovery-infra/test_infra/helper_classes/kube_helpers.py
RazRegev: it reflects a Status key in the crd that contains a dict

discovery-infra/test_infra/helper_classes/kube_helpers.py
RazRegev: yes, that would be a consideration when we'll have more CRDs, 
right now it seems to me like adding an unnecessary abstraction for the current needs.

discovery-infra/test_infra/helper_classes/kube_helpers.py
RazRegev: I mean why do you insist on it?

discovery-infra/test_infra/helper_classes/kube_helpers.py
RazRegev: I'll make it to try to get the config path from env if exits

discovery-infra/test_infra/helper_classes/kube_helpers.py
RazRegev: sure, just like to be explicit, is it critical?

discovery-infra/test_infra/helper_classes/kube_helpers.py
```diff
@@ -0,0 +1,376 @@
...
+        dict.__init__(
+            self,
+            agentBareMetal={
+                'apiVIP': api_vip,
+                'ingressVIP': ingress_vip,
+                'VIPDHCPAllocation': 'Enabled' if vip_dhcp_allocation else ''
+            }
+        )
+
+        if _does_string_contain_value(api_vip_dns_name):
```
osherdp, line 113: why not just ``if api_vip_dns_name:``?

discovery-infra/test_infra/helper_classes/kube_helpers.py
```diff
@@ -0,0 +1,376 @@
...
+        self.crd_api.create_namespaced_custom_object(
+            group=self.hive_api_group,
+            version='v1',
+            plural='clusterdeployments',
+            body=body,
+            namespace=self.ref['namespace']
+        )
+        self.__secret = secret
+
+        logger.info('created cluster deployment %s: %s', self.ref, body)
```
osherdp, line 209: is the body a large block of text?
consider using ``pprint`` if so, otherwise it's ok

discovery-infra/test_infra/helper_classes/kube_helpers.py
RazRegev: I went for an abbreviation for 'object' which appears in many libraries where the scope of this variable is in a small helper function such as this (I wouldn't use that name for any other case). I will change it anyway.

discovery-infra/test_infra/helper_classes/kube_helpers.py
```diff
@@ -0,0 +1,376 @@
...
+    ) -> Tuple[str, str]:
+
+        state, state_info = None, None
+        for condition in self.status(timeout).get('conditions', []):
+            reason = condition.get('reason')
+
+            if reason == 'AgentPlatformState':
+                state = condition.get('message')
+
+            # todo: fix typo after it will be fixed in assisted-service
```
osherdp, line 262: seems like a constant's value change, no?
can we do it now?

discovery-infra/test_infra/helper_classes/kube_helpers.py
osherdp, line 281:
```suggestion
            state, _ = self.state(timeout=0.5)
```

osherdp (changes requested):

discovery-infra/test_infra/helper_classes/kube_helpers.py
YuviGold: OK. At least make it a constant variable.

discovery-infra/test_infra/helper_classes/kube_helpers.py
RazRegev: No, it's a bad practice to use mutable objects as default values. 
Take for example this class:
```
class Test(object):

    def __init__(self, key, value, dct = dict()):
        self.dct = dct
        self.dct[key] = value


t1 = Test('a', 'b')
t2 = Test('c', 'd')

print(t1.dct)
print(t2.dct)
```
You would expect that t1.dct will be `{'a': 'b'} `and t2.dct will be `{'b': 'c'}`, right?
But since I used a mutable object as default, I will get:
```
{'a': 'b', 'c': 'd'}
{'a': 'b', 'c': 'd'}
```
That is why I need to use None.

discovery-infra/test_infra/helper_classes/kube_helpers.py
YuviGold: IMO it seems more clean and Pythonic. But after searching a little bit on the web, it seems that there are people that might disagree https://fuhm.net/super-harmful/
You can keep. I will keep searching about it :)

discovery-infra/test_infra/helper_classes/kube_helpers.py
YuviGold: Although there are no details regarding this behavior on [PEP8](https://www.python.org/dev/peps/pep-0008/#blank-lines) conventions, it just makes the module unnecessary longer.

discovery-infra/test_infra/helper_classes/kube_helpers.py
YuviGold: Not critical. Just unnecessary.

discovery-infra/test_infra/helper_classes/kube_helpers.py
YuviGold: Can you add a docstring?

discovery-infra/test_infra/helper_classes/kube_helpers.py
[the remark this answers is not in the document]
RazRegev: it's nice but I'm not sure that it is necessary here, since I teat this class as dict intentionally

discovery-infra/test_infra/helper_classes/kube_helpers.py
RazRegev: I removed the empty lines although I think it makes it look less convenient to read

discovery-infra/test_infra/helper_classes/kube_helpers.py
RazRegev: thanks :)

discovery-infra/test_infra/helper_classes/kube_helpers.py
RazRegev: I change it to get the client from outside (which makes this base class a little bit unnecessary, so I removed it)

discovery-infra/test_infra/helper_classes/kube_helpers.py
RazRegev: removed

discovery-infra/test_infra/helper_classes/kube_helpers.py
RazRegev: good idea, 10x

discovery-infra/tests/base_test.py
RazRegev: I agree, but on the other hand, I think it is more indicative to tell that this fixture is context so you'll know to use it as context manager, what do you think?

discovery-infra/test_infra/helper_classes/kube_helpers.py
RazRegev: https://github.com/openshift/assisted-service/pull/1099

discovery-infra/test_infra/helper_classes/kube_helpers.py
RazRegev: done

discovery-infra/test_infra/helper_classes/kube_helpers.py
RazRegev: fixed

discovery-infra/test_infra/helper_classes/kube_helpers.py
RazRegev: I think we should add it in a seperate PR (if I get this right it should be helpful for most of thee tests)
Right now the convention is to get the defaults from env, before involving api calls to the service

discovery-infra/test_infra/helper_classes/kube_helpers.py
```diff
@@ -0,0 +1,456 @@
...
+
+from kubernetes.config import load_kube_config
+from kubernetes.config.kube_config import Configuration
+from kubernetes.client import ApiClient, CoreV1Api, CustomObjectsApi
+from kubernetes.client.rest import ApiException
+
+from tests.conftest import env_variables
+from test_infra.utils import get_random_name
+
+"""
```
osherdp, line 19: module docstring should be at the top

discovery-infra/test_infra/helper_classes/kube_helpers.py
```diff
@@ -0,0 +1,456 @@
...
+    def create(self, **kwargs) -> None:
+        raise NotADirectoryError()
+
+    @abc.abstractmethod
+    def get(self) -> dict:
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def delete(self) -> None:
+        raise NotADirectoryError()
```
osherdp, line 128: why using ``NotADirectoryError``?
btw I think a simple ``pass`` is enough. The ``abc.abstractmethod`` is enough to make subclasses uninstantable without providing implementation

osherdp (changes requested):

discovery-infra/test_infra/helper_classes/kube_helpers.py
RazRegev: ops, thanks for catching that

discovery-infra/test_infra/helper_classes/kube_helpers.py
RazRegev: But I've done it anyway (since we are going to add more CRDs for sure)

discovery-infra/tests/base_test.py
YuviGold: Not critical. :)

discovery-infra/test_infra/helper_classes/kube_helpers.py
```diff
@@ -0,0 +1,495 @@
+# kube_helpers.py provides infra to deploy, manage and install cluster using
```
osherdp, line 1: replace with triple quotes, e.g.:
```
"""Bla bla bla.

Bla bla.
"""
```

discovery-infra/test_infra/helper_classes/kube_helpers.py
```diff
@@ -0,0 +1,495 @@
...
+            kube_api_client: ApiClient,
+            name: str,
+            namespace: str = env_variables['namespace']
+    ):
+        self.v1_api = CoreV1Api(kube_api_client)
+        self.__ref = ObjectReference(name=name, namespace=namespace)
+
+    @property
+    def ref(self) -> ObjectReference:
+        return self.__ref
```
osherdp, line 84: there's no convention in Python for two underscores in start of a variable name
can we change to one underscore?

discovery-infra/test_infra/helper_classes/kube_helpers.py
```diff
@@ -0,0 +1,495 @@
...
+    def delete(self) -> None:
+        self.v1_api.delete_namespaced_secret(
+            name=self.ref.name,
+            namespace=self.ref.namespace
+        )
+
+        logger.info('deleted secret %s', self.ref)
+
+
+class _BaseCustomResource(abc.ABC):
```
osherdp, line 108: underscore in the start of a class name is also not a convention I know
if you want to decide which class is public, use ``__all__``

discovery-infra/test_infra/helper_classes/kube_helpers.py
```diff
@@ -0,0 +1,495 @@
...
+
+
+def deploy_default_secret(
+        kube_api_client: ApiClient,
+        name: str,
+        ignore_conflict: bool = True,
+        pull_secret: Optional[str] = None
+) -> Secret:
+    if pull_secret is None:
+        pull_secret = os.environ.get('PULL_SECRET', '')
```
osherdp, line 402: learning from my mistakes, we should validate this thing is a valid JSON as soon as possible
Can you do a ``json.loads`` for this validation?

osherdp (changes requested):

discovery-infra/test_infra/helper_classes/kube_helpers.py
RazRegev: Both ways are acceptable, as I've seen in different places, 
If I'll add `__all__` it will be another (long) list to maintain, I think I prefer to leave that as it

osherdp: ``__all__`` doesn't make objects unreachable, it just helps autocompletion in a non-binding way
reconsider

discovery-infra/test_infra/helper_classes/kube_helpers.py
RazRegev: the interpreter translates "dunderscore" as a class private variable, but I will change it to avoid confusion.

osherdp (approved):
/lgtm

PR #523, opened 2021-02-15
MGMT-4100 Move download logs logic to download_logs.sh script

YuviGold:
Support onprem and k8s logs with getting `ADDITIONAL_PARAMS` environment variable. i.e. `--must-gather`, `--download-all`, `--update-by-events`.

/cc @osherdp

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -24,14 +24,17 @@ pipeline {
 
     environment {
         SKIPPER_PARAMS = " "
-        PULL_SECRET = credentials('assisted-test-infra-pull-secret')
-        OCPMETAL_CREDS = credentials('docker_ocpmetal_cred')
-        SLACK_TOKEN = credentials('slack-token')
         BASE_DNS_DOMAINS = credentials('route53_dns_domain')
         ROUTE53_SECRET = credentials('route53_secret')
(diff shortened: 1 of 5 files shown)


discovery-infra/logger.py
```diff
@@ -17,6 +18,16 @@ def format(self, record):
         return self._filter(original)
 
 
+class suppressLog(suppress):
```
osherdp, line 21: it's a misleading name, I thought it just suppresses logging inside the block
Maybe ``suppressAndLog`` or ``suppressButLog``? Or ``suppressWithNotification``/``WithLogging``

osherdp (changes requested):
just the class name sounds weird

YuviGold:
I divided this PR into another PR https://github.com/openshift/assisted-test-infra/pull/524

osherdp (approved):
/lgtm

osherdp:
@YuviGold all latest failures:
```
HTTP response body: {"code":"500","href":"","id":500,"kind":"Error","reason":"coreos-installer failed: /data/coreos-installer: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory\n: exit status 127"}
```

Not relevant at all to your changes, which makes me think it's somehow got broken

PR #525, opened 2021-02-16
NO-ISSUE - if no events file look for updates

oshercc:


YuviGold (approved):
LGTM

discovery-infra/download_logs.py
```diff
@@ -79,21 +79,32 @@ def is_update_needed(output_folder: str, update_on_events_update: bool, client:
...
+        log.warn(f"no event file {destination_event_file_path}")
+        shutil.rmtree(output_folder)
+        return True
+
     with tempfile.NamedTemporaryFile() as latest_event_tp:
         with suppressAndLog(assisted_service_client.rest.ApiException):
             client.download_cluster_events(cluster['id'], latest_event_tp.name)
 
-        if filecmp.cmp(destination_event_file_path, latest_event_tp.name):
+        if new_events_available(destination_event_file_path, latest_event_tp.name):
```
eranco74, line 91: The if statement seems wrong.
```
if  new_events_available...:
    log.info("no new events found for {}".format(destination_event_file_path))


```

discovery-infra/download_logs.py
```diff
@@ -79,21 +79,32 @@ def is_update_needed(output_folder: str, update_on_events_update: bool, client:
...
             need_update =  False
         else:
             log.info("update needed, new events found, deleting {} ".format(destination_event_file_path))
-            os.remove(destination_event_file_path)
+            shutil.rmtree(output_folder)
             latest_event_tp.close()
             need_update = True
     return need_update
 
+def new_events_available(available_event_file: str, latest_event_file: str):
```
YuviGold, line 102: if you just check the length, you should use `os.path.getsize` instead of opening a file. The size can be found on the inode of the file and opening it is unnecessary.

PR #526, opened 2021-02-16
MGMT-4100 Set local service url only if REMOTE_SERVICE_URL wasn't provided

YuviGold:
…vided

Fixes getting remote cluster logs without trying to use local minikube, that might not exist.

/cc @osherdp


osherdp (approved):
/lgtm

PR #532, opened 2021-02-17
fix(install_minikube.sh): usr/local/bin is not in default 'root' $PAT…

iranzo:
usr/local/bin is not in default 'root' $PATH, only for users, but '/usr/local/sbin' is


YuviGold:
I think `oc` should be in the same place as `minikube` and `kubectl`. WDYT?

iranzo:
> I think `oc` should be in the same place as `minikube` and `kubectl`. WDYT?

Want me to also update the other paths too?

eranco74 (commented):
/lgtm

YuviGold (approved):
LGTM

YuviGold:
/hold
```
/bin/sh: oc: command not found
```

YuviGold:
@iranzo what do you think about using [arkade](https://github.com/alexellis/arkade)?

PR #533, opened 2021-02-17
prettify the files

iranzo:
- Removes executable bit of html file
- Adds encoding to python files
- Removes trailing spaces
- Applies 'black' formatter to python files
- Fixes newline at end of files


YuviGold:
There's an ongoing effort by @eliorerz to make all our scripts PEP8 compatible. We will also insert a linter as part of our CI.

iranzo:
> There's an ongoing effort by @eliorerz to make all our scripts PEP8 compatible. We will also insert a linter as part of our CI.

Better to use 'black' which not only complains but also formats them!!!

PR #536, opened 2021-02-18
Use latest minikube and kubectl releases

iranzo:
Use latest minikube and kubectl according to their website

### DIFF
--- a/scripts/install_minikube.sh
+++ b/scripts/install_minikube.sh
@@ -9,7 +9,7 @@ function install_minikube() {
 
     if ! [ -x "$(command -v minikube)" ]; then
         echo "Installing minikube..."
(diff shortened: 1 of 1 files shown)


YuviGold:
My fear with "latest" is that API could get broken from time to time. Happened to us with [terraform](https://github.com/openshift/assisted-test-infra/pull/83).

iranzo:
> My fear with "latest" is that API could get broken from time to time. Happened to us with [terraform](https://github.com/openshift/assisted-test-infra/pull/83).

For kubelet is the one labelled as stable, but for minikube didn't  found it :/

YuviGold:
Using [arkade](https://github.com/alexellis/arkade) for `minikube` and `kubectl` installation since https://github.com/openshift/assisted-test-infra/pull/660

PR #538, opened 2021-02-18
MGMT-3420 Make test infra work with none platform and IPv6

ori-amizur:
Add load balancer
The load balancer runs in its own container.  The setup of the container is in install_environment.sh.
It runs nginx as load balancer.
The nginx in the container is refreshed every 60 seconds.  It looks at directory /etc/nginx/stream.d/ , where load balancing definitions can be added or removed.
The load balancing definitions are managed by terraform (using local_file resource) which is driver by python code.  They are generated currently only for none platform.

/cc @tsorya 
/cc @empovit

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -114,7 +114,7 @@ PLATFORM := $(or ${PLATFORM},baremetal)
 all: create_full_environment run_full_flow_with_install
 
 
-destroy: destroy_nodes delete_minikube_profile kill_port_forwardings delete_podman_localhost
+destroy: destroy_nodes delete_minikube_profile kill_port_forwardings delete_podman_localhost stop_load_balancer
 
 ###############
 # Environment #
@@ -156,6 +156,17 @@ delete_minikube_profile:
 
 delete_podman_localhost:
 	make -C assisted-service/ clean-onprem || true
+
+####################
+# Load balancer    #
+####################
+
+start_load_balancer:
+	# Start load balancer if it does not already exist.  Map the directory $(HOME)/.test-infra/etc/nginx/stream.d to be /etc/nginx/stream.d so it will be used by the python code to fill up load balancing definitions
+	if [ "$(PLATFORM)" = "none" ] ; then id=`podman ps --quiet --filter "name=load_balancer"` ; ( test -z "$$id" && podman run -d --rm --net=host --name=load_balancer -v $(HOME)/.test-infra/etc/nginx/stream.d:/etc/nginx/stream.d load_balancer:latest ) || ! test -z "$$id" ; fi
+
+stop_load_balancer:
+	id=`podman ps --quiet --filter "name=load_balancer"`; test ! -z "$$id"  && podman rm -f load_balancer ; rm -f  $(HOME)/.test-infra/etc/nginx/stream.d/*.conf >& /dev/null || /bin/true
 
 #############
 # Terraform #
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -274,15 +284,15 @@ install_cluster:
 _deploy_nodes:
 	discovery-infra/start_discovery.py -i $(ISO) -n $(NUM_MASTERS) -p $(STORAGE_POOL_PATH) -k '$(SSH_PUB_KEY)' -md $(MASTER_DISK) -wd $(WORKER_DISK) -mm $(MASTER_MEMORY) -wm $(WORKER_MEMORY) -nw $(NUM_WORKERS) -ps '$(PULL_SECRET)' -bd $(BASE_DOMAIN) -cN $(CLUSTER_NAME) -vN $(NETWORK_CIDR) -nM $(NETWORK_MTU) -iU $(REMOTE_SERVICE_URL) -id $(CLUSTER_ID) -mD $(BASE_DNS_DOMAINS) -ns $(NAMESPACE) -pX $(HTTP_PROXY_URL) -sX $(HTTPS_PROXY_URL) -nX $(NO_PROXY_VALUES) --service-name $(SERVICE_NAME) --vip-dhcp-allocation $(VIP_DHCP_ALLOCATION) --profile $(PROFILE) --ns-index $(NAMESPACE_INDEX) --deploy-target $(DEPLOY_TARGET) $(DAY1_PARAMS) $(OC_PARAMS) $(KEEP_ISO_FLAG) $(ADDITIONAL_PARAMS) $(DAY2_PARAMS) -ndw $(NUM_DAY2_WORKERS) --ipv4 $(IPv4) --ipv6 $(IPv6) --platform $(PLATFORM) --proxy $(PROXY) --iso-image-type $(ISO_IMAGE_TYPE)
 
-deploy_nodes_with_install:
+deploy_nodes_with_install: start_load_balancer
```
tsorya, line 287: we need load balancer only in ipv6 none platform, why to start it every run?

discovery-infra/test_infra/utils.py
```diff
@@ -829,3 +830,32 @@ def get_assisted_controller_status(kubeconfig):
 
     log.info(f'{response.stdout}')
     return response.stdout
+
+
+def _render_socket_endpoint(ip, port):
```
tsorya, line 835: move load balancer code to load_balancer_controller maybe?

scripts/install_environment.sh
```diff
@@ -160,6 +160,22 @@ function config_chronyd() {
   sudo firewall-cmd --zone=libvirt --add-port=123/udp
 }
 
+function config_nginx() {
+  echo "Config nginx"
+
+  cat <<EOF | sudo podman build --tag load_balancer:latest -
```
tsorya, line 166: will be glad for some explanation

scripts/install_environment.sh
```diff
@@ -160,6 +160,22 @@ function config_chronyd() {
...
+FROM quay.io/centos/centos:8.3.2011
+RUN dnf install -y nginx
+RUN sed -i -e '/^http {/,\$d'  /etc/nginx/nginx.conf
+RUN sed -i -e '\$a stream {\ninclude /etc/nginx/stream.d/*.conf;\n}' -e '/^stream {/,\$d' /etc/nginx/nginx.conf
+CMD ["bash", "-c", "while /bin/true ; do (ps -ef | grep -v grep | grep -q nginx && nginx -s reload) || nginx ; sleep 60 ; done"]
+EOF
+  sudo podman rm -f load_balancer || /bin/true
+  sudo mkdir -p $HOME/.test-infra/etc/nginx/stream.d
```
tsorya, line 174: why $HOME/.test-infra? why not inside test-infra? and git ignore it?

Makefile
ori-amizur: We work with it also with IPv4

scripts/install_environment.sh
ori-amizur: Seemed OK to me.  Since it is a machine wide configuration, it should be stored on machine level

scripts/install_environment.sh
ori-amizur: Will add

discovery-infra/test_infra/utils.py
ori-amizur: Possible

Makefile
ori-amizur: Also, we need to run it outside skipper.  So this is the place since any podman we will run in skipper will be terminated when skipper terminates.  So this is the place to run it.  At this point, we don't know if we will need it.

tsorya: not sure i understand why we need it in regular flow? what it gives for us?

tsorya: so it is relevant only for None platform, lets spawn it there.

tsorya: if container is started as daemon and with nsenter? still it is killed when skipper exits?

discovery-infra/start_discovery.py
```diff
@@ -390,6 +391,10 @@ def nodes_flow(client, cluster_name, cluster):
         tf=tf
     )
 
+    is_ip4 =  machine_net.has_ip_v4 or not machine_net.has_ip_v6
```
tsorya, line 394: nit : is_ipv4

discovery-infra/start_discovery.py
```diff
@@ -419,6 +424,12 @@ def nodes_flow(client, cluster_name, cluster):
 
         set_hosts_roles(client, cluster, nodes_details, machine_net, tf, args.master_count, args.with_static_ips)
 
+        if is_none_platform_mode() and args.master_count > 1:
```
tsorya, line 427: maybe we can have function as is_none_platform_mode(haMode=true) ?

Makefile
```diff
@@ -157,6 +157,17 @@ delete_minikube_profile:
 delete_podman_localhost:
 	make -C assisted-service/ clean-onprem || true
 
+####################
+# Load balancer    #
+####################
+
+start_load_balancer:
+	# Start load balancer if it does not already exist.  Map the directory $(HOME)/.test-infra/etc/nginx/stream.d to be /etc/nginx/stream.d so it will be used by the python code to fill up load balancing definitions
+	id=`podman ps --quiet --filter "name=load_balancer"` ; ( test -z "$$id" && podman run -d --rm --net=host --name=load_balancer -v $(HOME)/.test-infra/etc/nginx/stream.d:/etc/nginx/stream.d load_balancer:latest ) || ! test -z "$$id"
```
tsorya, line 166: what is registry? docker io?

tsorya: we will need this image inside http://quay.io/

ori-amizur: Local.  We build it locally

PR #539, opened 2021-02-18
NO-ISSUE: allow operators status check errors be transient

osherdp:
/cc @osherdp 
/hold


osherdp:
/retitle NO-ISSUE: allow operators status check errors be transient

PR #541, opened 2021-02-21
MGMT-3370: Added kubeAPI infra

RazRegev:
Provides infra to deploy assisted-service with KubeAPI
controllers and relevant CRDs.
Change skipper image to deploy assisted-service using
assisted-service build image.


RazRegev:
depands on 
https://github.com/openshift/assisted-service/pull/1131

skipper.env
```diff
@@ -0,0 +1,73 @@
+PULL_SECRET
```
eranco74, line 1: Perfect.
While it may change the behavior I think it's better than the current behavior of setting empty env vars in case it's not set in the user env.

eranco74 (commented):
/lgtm

skipper.env
RazRegev: +1

PR #542, opened 2021-02-21
MGMT-3370: Allow deploying ClusterDeployment from file

RazRegev:
By given filepath (optional) deploy ClusterDeployment from
its data instead of environment variables.


osherdp (approved):
/lgtm

PR #551, opened 2021-02-22
NO-ISSUE - Perform installer-gather on single node in case of failure

omertuc:


discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
```diff
@@ -66,7 +66,7 @@ def shutdown_all_nodes(self):
         for node in nodes:
             self.shutdown_node(node.name())
 
-    def start_node(self, node_name, check_ips):
+    def start_node(self, node_name, check_ips=True):
```
omertuc, line 69: Not sure about this change... @yevgeny-shnaidman please verify this - it looks like you introduced the parameter. Does it make sense for it to be true by default (looks like the previous behavior, before the parameter, was if it was "True")?

omertuc: PR that introduced this change: #424

Not sure how this didn't come up until now... This is the line that makes use of the old signature: https://github.com/openshift/assisted-test-infra/blob/d3bf8bed79cc68f1fed0af52350fb86a0750ff67/discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py#L91

osherdp:
@omertuc unhold when build is passing

osherdp (approved):

omertuc:
Seems to have worked https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_assisted-test-infra/551/pull-ci-openshift-assisted-test-infra-master-e2e-metal-single-node-live-iso/1363969050647990272/artifacts/e2e-metal-single-node-live-iso/baremetalds-sno-gather/artifacts/build/ibip/installer-gather/

omertuc:
> @omertuc unhold when build is passing

Looks like it's failing on must-gather. Is it related to this PR? or does it always do that?

osherdp:
no, I don't think it has anything to do with the PR
I just was under the impression that it should work here

PR #555, opened 2021-02-23
MGMT-4182 - Use new pull secret without the svc registry

omertuc:
Using this new pull secret: http://assisted-jenkins.usersys.redhat.com/credentials/store/system/domain/_/credential/assisted-test-infra-pull-secret-no-svc/ This is an alternative pull secret to the previous test infra pull secret "assisted-test-infra-pull-secret" that doesn't contain an entry for the "registry.svc.ci.openshift.org" registry. The previous secret seems to have an invalid entry for that registry and it was causing CI issues.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -32,7 +32,7 @@ pipeline {
         LOGS_DEST = "${WORKSPACE}/cluster_logs"
 
         // Credentials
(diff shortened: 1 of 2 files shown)


omertuc:
obsolete

PR #557, opened 2021-02-23
NO-ISSUE - don't give up during sno log collection

omertuc:


discovery-infra/bootstrap_in_place.py
```diff
@@ -151,6 +153,37 @@ def gather_sosreport_data(node):
     node.download_file("/tmp/sosreport.tar.bz2", IBIP_DIR)
 
 
+# noinspection PyBroadException
+def log_collection(controller, vm_ip):
+    etype, _value, _tb = sys.exc_info()
+
+    logging.info(f"Collecting logs after a {('failed', 'successful')[etype is None]} installation")
```
eranco74, line 160: Didn't know this syntax

eranco74 (commented):
/lgtm

discovery-infra/bootstrap_in_place.py
omertuc: i'm actually abusing the fact that bool gets converted to 0/1 and using that as an index into the tuple.... it's very ugly but very readable

PR #559, opened 2021-02-23
NO-ISSUE: use PULL_SECRET env-var when doing 'oc adm release extract'

osherdp:
/cc @YuviGold @tsorya 
It's a bit of a more generalized solution than before.
/hold

### DIFF
--- a/discovery-infra/test_infra/utils.py
+++ b/discovery-infra/test_infra/utils.py
@@ -747,16 +747,11 @@ def get_openshift_version():
     release_image = os.getenv('OPENSHIFT_INSTALL_RELEASE_IMAGE')
 
     if release_image:
-        f = tempfile.NamedTemporaryFile(mode='w', delete=False)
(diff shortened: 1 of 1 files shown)


discovery-infra/test_infra/utils.py
```diff
@@ -771,9 +766,33 @@ def copy_template_tree(dst, none_platform_mode=False):
...
+    try:
+        with tempfile.NamedTemporaryFile(mode='w', delete=False) as f:
+            f.write(pull_secret)
+            pull_secret_path = f.name
+
+        yield pull_secret_path
+
+    finally:
+        if pull_secret_path and os.path.exists(pull_secret_path):
+            os.unlink(pull_secret_path)
```
YuviGold, line 789: can't the yield be inside the `with` and then the `delete=True` would be called as expected? Then, you don't need to explicitly call to unlink.

YuviGold (approved):
LGTM

PR #562, opened 2021-02-23
firewalld: services when available and permanent

celebdor:
Sometimes ports change between version and version, it is usually a best
practice to open services instead of ports when opening access to
services provided in the distribution.

Also, in case the environment reboots, it is good to put the fierwalld
changes as permanent.


celebdor:
I only tested the chrony part of the change

YuviGold:
@celebdor I think this is a great improvement. Could you please rebase?

PR #565, opened 2021-02-28
NOISSUE - kube_helpers.py: fixed kubeapi ClusterDeployment patch func…

RazRegev:
…tion

'provisioning' should be defined as a dictionary before adding
installStrategy


osherdp (approved):
/lgtm

PR #574, opened 2021-03-01
MGMT-4152 log collection script

oshercc:


discovery-infra/log_scrap.py
```diff
@@ -0,0 +1,216 @@
...
+        with open(metadata_dest, "w") as f:
+            json.dump(metadata_json, f, indent=4)
+
+    def log_doc(self, doc, id_):
+        try:
+            res = self.es.create(index=self.index, body=doc, id=id_)
+        except elasticsearch.exceptions.ConflictError:
+            log.debug("Hit logged event")
+            return None
+        return res
```
flaper87, line 158: Depending on the throughoput, I would recommend to either use [`bulk_inserts` ](https://elasticsearch-py.readthedocs.io/en/7.10.0/helpers.html#bulk-helpers) or put Logstash in front and post messages to it directly.

discovery-infra/log_scrap.py
```diff
@@ -0,0 +1,216 @@
...
+            json.dump(event_list, f, indent=4)
+
+        metadata_dest = os.path.join(cluster_backup_directory_path, "metadata.json")
+        with open(metadata_dest, "w") as f:
+            json.dump(metadata_json, f, indent=4)
+
+    def log_doc(self, doc, id_):
+        try:
+            res = self.es.create(index=self.index, body=doc, id=id_)
+        except elasticsearch.exceptions.ConflictError:
```
flaper87, line 155: What are the chances of conflicts? Note that if you use `index` instead of `create` Elasticsearch will index a "new" version of the document, instead of raising an error. This may be something to consider.

discovery-infra/log_scrap.py
```diff
@@ -0,0 +1,216 @@
...
+        except elasticsearch.exceptions.NotFoundError:
+            return None
+        if not res["hits"]["hits"]:
+            return None
+        return res["hits"]["hits"][0]["_source"]["event_time"]
+
+    def get_last_cluster_event(self, cluster_id):
+        res = self.es.search(index=self.index, body={
+            "size": 1,
+            "sort": {"event_time": "desc"},
```
flaper87, line 80: For cases where you only want the `event_time` you can pass `_source: ['event_time']` which will only return that field.

discovery-infra/monitoring/docker-compose.yml
```diff
@@ -0,0 +1,36 @@
+version: '2.2'
+services:
+  elasticsearch:
+    image: docker.elastic.co/elasticsearch/elasticsearch:7.10.1
```
flaper87, line 4: ELK 7.11 is out, you may want to go straight to that one.

flaper87 (changes requested):
As far as the cluster deployment goes, I think there are a few things to consider:

- What are the durability requirements for this data?

If losing data is a no-no, then I'd recommend using Logstash as doing the bulk inserts directly would require implementing retry logic, backpressure management, queue persistency, etc.

I'd also consider using more than 1 node (3 minimum) to have a replicated environment. This will not only help improve the availability of the data, but it will also improve the durability of it as we will be able to replicate the indexes to tolerate node failures.

- Consider using [index templates:](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-templates.html)

Using index templates will allow defining the index settings, minimum replicas, etc. Generally, it will help guarantee that all indexes for this data will look the same.

- Consider using [Datastreams](https://www.elastic.co/guide/en/elasticsearch/reference/master/data-streams.html)

This is a timeseries data, we will be collecting logs/events from the clusters so we won't need to update, delete records specifically. Datastreams are thought for this kind of data, so I would recommend switching to using them.

This can be set in the index template directly.

- Enable [ILM policies](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-lifecycle-management.html)

This index is an ever growing index. Using ILM policies will help us segment the index into hot, warm, cold categories so that the data we need *now* is always "ready". In addition to this, by using ILM policies, we'll be able to specify the retention policy we want. For example, do we want to keep this data forever? Or is it safe to delete it after 1 month? maybe 7 days is enough?

The ILM policy to use for these indexes can be set in the index template directly.

Hope this helps :)

oshercc:
updates will be added with the next update 
see https://issues.redhat.com/browse/MGMT-4574

discovery-infra/monitoring/process.py
```diff
@@ -0,0 +1,95 @@
...
+REMOVED_FIELDS = [
+    "cluster.image_info_ssh_public_key",
+    "cluster.ssh_public_key",
+    "cluster.image_info.ssh_public_key"
+    "cluster.connectivity_majority_groups",
+    "cluster.controller_logs_collected_at",
+    
+    "cluster.hosts.connectivity",
+    "cluster.hosts.images_status",
+    # "cluster.hosts.inventory",
```
gamli75, line 15: please remove it

discovery-infra/monitoring/process.py
```diff
@@ -0,0 +1,95 @@
...
+                host["validations_info"] = convert_field_to_json(host["validations_info"])
+
+    def remove_fields_if_exists(self):
+        for remove_field in REMOVED_FIELDS:
+            try:
+                self.pop_fields(self.metadata_json, remove_field)
+            except KeyError:
+                pass
+
+    # Delete a fieald in string path joind by "."
```
gamli75, line 76: field instead of fieald

gamli75 (approved):

discovery-infra/monitoring/process.py
```diff
@@ -0,0 +1,94 @@
...
+        if pop_list[0] not in p_json:
+            return
+        self.pop_fields(p_json[pop_list[0]], pop_list[1])
+        return
+
+def is_event_skippable(event):
+    for skippable_text in SKIPPABLE_EVENTS:
+        if skippable_text in event["message"]:
+            return True
+    return False
```
tsorya, line 94: new line

discovery-infra/monitoring/process.py
```diff
@@ -0,0 +1,94 @@
...
+            inventory = json.loads(host["inventory"])
+            vendor = inventory.get("system_vendor", None)
+            if vendor:
+                host["vendor"] = vendor
+
+    def get_processed_json(self):
+        self.remove_fields_if_exists()
+        return self.metadata_json
+
+    def convert_strings_to_dict(self):
```
tsorya, line 60: private?

tsorya: can we make all private functions private?

discovery-infra/monitoring/process.py
```diff
@@ -0,0 +1,94 @@
...
+logging.basicConfig(level=logging.WARN, format='%(levelname)-10s %(message)s')
+logger = logging.getLogger(__name__)
+logging.getLogger("__main__").setLevel(logging.INFO)
+
+
+def convert_field_to_json(converted_field):
+    try:
+        if type(converted_field) == str:
+            return json.loads(converted_field)
+    except KeyError:
```
tsorya, line 36: sure it fails only on KeyError ?

discovery-infra/monitoring/process.py
```diff
@@ -0,0 +1,94 @@
...
+SKIPPABLE_EVENTS = [
+    "reached installation stage Writing image to disk"
+]
+
+logging.basicConfig(level=logging.WARN, format='%(levelname)-10s %(message)s')
+logger = logging.getLogger(__name__)
+logging.getLogger("__main__").setLevel(logging.INFO)
+
+
+def convert_field_to_json(converted_field):
```
tsorya, line 32: this is used only here no?

tsorya, line 32: maybe part of the class as all others?

discovery-infra/log_scrap.py
```diff
@@ -0,0 +1,191 @@
+#!/usr/bin/env python3
+
```
tsorya, line 2: lets move all this inside test_infra maybe?

oshercc: not sure what you mean

discovery-infra/monitoring/process.py
oshercc: if it fails on something else we want to get notified about it

discovery-infra/monitoring/process.py
oshercc: yes, but it is not part of the class functionality so its not really an attribute of the class

PR #577, opened 2021-03-04
MGMT-3370: Jenkinsfile.test - enabled kube-api

RazRegev:
In order to make the kube-api e2e test to work we need to enable
kube-api when deploying the environment.


YuviGold:
This Jenkinsfile is deprecated. You should talk with @nshidlin in order to update the QE e2e env.

PR #580, opened 2021-03-04
MGMT-4274 Use AUTH_TYPE flag over ENABLE_AUTH

carbonin:
This will allow us to test new auth types
Removes the flag from the line to update_assisted_service_cm.py
as ENABLE_AUTH wasn't in the list of relevant env vars.

cc @YuviGold is there anywhere else that needs to change for this?
Do we have other automation somewhere that enables auth?


carbonin:
/hold

Depends on https://github.com/openshift/assisted-service/pull/1201

YuviGold (approved):
LGTM

YuviGold:
Prow and Jenkins aren't using this var. they both use the defaults.

carbonin:
openshift/assisted-service#1201 was merged so un-holding this.

/hold cancel

PR #591, opened 2021-03-08
NO-ISSUE increase wait for must-gather in download_logs

slaviered:


discovery-infra/download_logs.py
```diff
@@ -130,7 +131,7 @@ def download_logs(client: InventoryClient, cluster: dict, dest: str, must_gather
                 cluster['hosts'], [HostsProgressStages.CONFIGURING], 2)
             are_masters_in_join_state = are_host_progress_in_stage(
                 cluster['hosts'], [HostsProgressStages.JOINED], 2)
-            max_retries = 2 * MAX_RETRIES if are_masters_in_join_state else MAX_RETRIES
+            max_retries = MUST_GATHER_MAX_RETRIES if are_masters_in_join_state else MAX_RETRIES
             is_controller_expected = cluster['status'] == ClusterStatus.INSTALLED or are_masters_in_configuring_state
             min_number_of_logs = min_number_of_log_files(cluster, is_controller_expected)
```
eranco74, line 137: Please add a log for each attempt.
Also, why do you think that if we keep trying at some point it will work?

slaviered: In Production we see that many times must-gather execution time is close (or over) an hour. Then the script terminates and the logs are uploaded. This logic download the logs until it finds the must gather tar. The wait loop is another way to wait for an hour or so. We still need some logical interval because this code runs also in the CI where the performance is better and times are shorter and it also shared by host logs download and there the intervals are much shorter

slaviered: + logs are done internally in the validation so we can view the progress in the logs ( in the existing code)

YuviGold (approved):
LGTM

PR #595, opened 2021-03-09
Remove hourly cron trigger from Jenkinsfile

YuviGold:
No need to keep running the Jenkins job hourly against the master branch
since we already moved our periodics to Prow on OpenShift CI.


osherdp (approved):
/lgtm

PR #598, opened 2021-03-09
OCPBUGSM-25057 add image type to event const

danielerez:
Added STARTED_DOWNLOAD_IMAGE to consts - contains image type
as updated in: https://github.com/openshift/assisted-service/pull/1258


danielerez:
@lalon4 @iovadia

PR #599, opened 2021-03-09
Add support for writing Day2 API tests

mchernik:
- Reuse existing TF configuration
  - Always return fresh TF state
  - Clear nodes cache
  - Fetch reuested number of Day2 nodes from environment


discovery-infra/test_infra/tools/terraform_utils.py
```diff
@@ -36,6 +36,7 @@ def change_variables(self, variables, refresh=True):
         self.apply(refresh=refresh)
 
     def get_state(self):
+        self.tf.read_state_file("terraform.tfstate")
```
omertuc, line 39: Add `STATE_FILE` constant to class scope and reuse it both here and in `__init__`

omertuc: (just like `VAR_FILE`)

discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
```diff
@@ -15,14 +15,15 @@ class TerraformController(LibvirtController):
...
+        self.cluster_suffix = kwargs.get('cluster_suffix', self._get_random_name())
         self.cluster_name = kwargs.get('cluster_name', f'{consts.CLUSTER_PREFIX}' + "-" + self.cluster_suffix)
         self.network_name = kwargs.get('network_name', consts.TEST_NETWORK) + self.cluster_suffix
         self.network_conf = kwargs.get('net_asset')
         self.cluster_domain = kwargs.get('base_domain', "redhat.com")
         self.ipv6 = kwargs.get('ipv6')
         self.params = self._terraform_params(**kwargs)
-        self.tf_folder = self._create_tf_folder()
+        tf_folder = kwargs.get('tf_folder')
+        self.tf_folder = tf_folder if tf_folder else self._create_tf_folder()
```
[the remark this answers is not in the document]
mchernik: Yes. The default parameter to `get`  ( `_create_tf_folder()` method) is evaluated before calling `get`, so the TF folder is always recreated with this approach.

PR #604, opened 2021-03-10
MGMT-3550 Enable IPv6 in test environments

empovit:
Following the assisted service PR https://github.com/openshift/assisted-service/pull/1250,
enable IPv6 by default in test infra environments and allow overriding it.


YuviGold:
Don't we have already `IPv6=no`?

osherdp:
@YuviGold ``DISABLE_IPV6`` seems to affect the service, not the provisioned hosts

osherdp:
@empovit Maybe we need a better name for the env-var

empovit:
@YuviGold `IPv6=yes/no` whether to deploy IPv6 hosts and use IPv6 addresses, `DISABLE_IPV6` whether a given service deployments supports IPv6. E.g. the clouds and setups without a mirror registry can't support IPv6, so any attempt to use IPv6 for assisted installation will be rejected.

osherdp:
``DISABLE_IPV6_SUPPORT`` is better than ``DISABLE_IPV6``, and ``IPV6_SUPPORT`` is even better (I don't like creating double-negatives :) )
We can change ``IPv6`` and ``IPv4`` to ``HOSTS_IP_STACK=v4/v6/v4v6`` similarly to dev-scripts, or maybe ``HOSTS_IP_STACK=ipv4/ipv6/dual``

ori-amizur (commented):
How is the parameter passed to assisted service?

empovit:
> How is the parameter passed to assisted service?

I'm not sure I understand the question. It's an environment variable declared in `skipper.env`. All such env variables are passed to a build https://github.com/openshift/assisted-test-infra/blob/c8ab5db4d499a4956b2f75529536793a4db34b13/scripts/deploy_assisted_service.sh#L70, so it eventually makes it to the service (PR https://github.com/openshift/assisted-service/pull/1250).

YuviGold (approved):
LGTM

PR #606, opened 2021-03-11
MGMT-4373 Persistent attach of disks

priyanka19-98:


osherdp:
@priyanka19-98 I guess downstream tests will be updated accordingly?

priyanka19-98:
> @priyanka19-98 I guess downstream tests will be updated accordingly?

After the patch is merged, the downstream repo can use this new fix. It won't break anything.

PR #607, opened 2021-03-11
MGMT-3335 - Add None Platform support 

eliorerz:
…load_balancer target with test target
TerraformController now take the PLATFORM environment variable and copy the none terraform files accordingly 

https://issues.redhat.com/browse/MGMT-3335

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -398,7 +398,7 @@ lint:
 _lint:
 	pre-commit run --all-files
 
-test:
+test: start_load_balancer
 	skipper make $(SKIPPER_PARAMS) _test
 
 _test: $(REPORTS) _test_setup
--- a/discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
+++ b/discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
@@ -19,6 +19,7 @@ class TerraformController(LibvirtController):
         self.cluster_name = kwargs.get('cluster_name', f'{consts.CLUSTER_PREFIX}' + "-" + self.cluster_suffix)
         self.network_name = kwargs.get('network_name', consts.TEST_NETWORK) + self.cluster_suffix
         self.network_conf = kwargs.get('net_asset')
+        self.platform = kwargs.get('platform', consts.Platforms.BARE_METAL)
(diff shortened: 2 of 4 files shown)


discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
```diff
@@ -32,7 +33,7 @@ def _create_tf_folder(self):
         tf_folder = utils.get_tf_folder(self.cluster_name)
         logging.info("Creating %s as terraform folder", tf_folder)
         utils.recreate_folder(tf_folder)
-        utils.copy_template_tree(tf_folder)
+        utils.copy_template_tree(tf_folder, self.platform == consts.Platforms.NONE)
```
tsorya, line 36: lets add none_platform_mode=self.platform == consts.Platforms.NONE

eliorerz: Done

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -532,8 +534,12 @@ def prepare_for_install(
             ssh_key=env_variables['ssh_public_key'],
             nodes_count=env_variables['num_nodes'],
             vip_dhcp_allocation=env_variables['vip_dhcp_allocation'],
-            download_image=True
+            download_image=True,
+            platform=env_variables['platform']
     ):
+        if platform == consts.Platforms.NONE:
```
ori-amizur, line 540: IPv4 only

Makefile
```diff
@@ -398,7 +398,7 @@ lint:
 _lint:
 	pre-commit run --all-files
 
-test:
+test: start_load_balancer
```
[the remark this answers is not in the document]
ori-amizur: It has to be invoked not from a container

discovery-infra/test_infra/helper_classes/cluster.py
eliorerz: Done

Makefile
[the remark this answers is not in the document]
tsorya: It has to be invoked not from container only for the case when we to leave it after test , no?

PR #608, opened 2021-03-14
MGMT-4360 - update day2 workflow to allow static network configuration

yevgeny-shnaidman:


Makefile
```diff
@@ -308,6 +308,12 @@ deploy_day2_cloud_nodes_with_install:
 deploy_day2_ocp_nodes_with_install:
 	skipper make $(SKIPPER_PARAMS) _deploy_nodes NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE) $(OC_FLAG)) NAMESPACE=$(NAMESPACE) $(SKIPPER_PARAMS) ADDITIONAL_PARAMS="'-in --day2-ocp-cluster'" DEPLOY_TARGET=ocp
 
+deploy_static_network_config_day2_nodes:
+	skipper make $(SKIPPER_PARAMS) _deploy_nodes NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE) $(OC_FLAG)) NAMESPACE=$(NAMESPACE) $(SKIPPER_PARAMS) ADDITIONAL_PARAMS="'--day2-cloud-cluster --with-static-network-config'"
+
+deploy_static_network_config_day2_nodes_with_install:
+	skipper make $(SKIPPER_PARAMS) _deploy_nodes NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE) $(OC_FLAG)) NAMESPACE=$(NAMESPACE) $(SKIPPER_PARAMS) ADDITIONAL_PARAMS="'-in --day2-cloud-cluster --with-static-network-config'"
+
```
empovit, line 316: Can we use a common target with varying `ADDITIONAL_PARAMS`? It becomes too long, and any new parameter will require changing all the individual targets.

yevgeny-shnaidman: i thought about it, but that means that someone new will have to know which parameters to use and one. This way Makefile may be long, but any user can find out how to run a scenario without trying to understand all the parameters need to be passed

discovery-infra/day2.py
```diff
@@ -290,3 +304,20 @@ def set_cluster_proxy(client, cluster_id, copy_proxy_from_cluster, args):
...
     client.set_cluster_proxy(cluster_id, http_proxy, https_proxy, no_proxy)
+
+def set_nodes_hostnames_if_needed(client,
+                                  terraform_cluster_dir_prefix,
+                                  namespace,
+                                  with_static_network_config,
+                                  has_ipv_4,
+                                  network_name,
+                                  cluster_id):
+    tf_folder = utils.get_tf_folder(terraform_cluster_dir_prefix, namespace)
```
empovit, line 315: This part keeps reappearing in the code. IMO if TF cluster dir prefix and namespace are used only for calculating TF folder, you can make it a parameter and pass around. This will reduce the number of parameters as a bonus.

discovery-infra/day2.py
```diff
@@ -290,3 +304,20 @@ def set_cluster_proxy(client, cluster_id, copy_proxy_from_cluster, args):
...
+    tf = terraform_utils.TerraformUtils(working_dir=tf_folder)
+    libvirt_nodes = utils.extract_nodes_from_tf_state(tf.get_state(), network_name, consts.NodeRoles.WORKER)
```
empovit, line 317: You probably don't need all this when the condition is `false`. `extract_nodes_from_tf_state` seems an expensive operation, perform it only when needed.

discovery-infra/day2.py
```diff
@@ -290,3 +304,20 @@ def set_cluster_proxy(client, cluster_id, copy_proxy_from_cluster, args):
...
+    if not has_ipv_4 or with_static_network_config:
```
empovit, line 318: Why do you need to set hostnames with static network config even when you have IPv4?

discovery-infra/day2.py
```diff
@@ -290,3 +304,20 @@ def set_cluster_proxy(client, cluster_id, copy_proxy_from_cluster, args):
     https_proxy = args.https_proxy if args.https_proxy else copy_proxy_from_cluster.https_proxy
     no_proxy = args.no_proxy if args.no_proxy else copy_proxy_from_cluster.no_proxy
     client.set_cluster_proxy(cluster_id, http_proxy, https_proxy, no_proxy)
+
+def set_nodes_hostnames_if_needed(client,
```
empovit, line 308: I don't like the `if_needed` part :( This name exposes the internal logic, which might change. Bad practice IMO.

discovery-infra/test_infra/tools/static_network.py
```diff
@@ -15,6 +15,26 @@ def generate_macs(count):
             for _ in range(count)]
 
 
+def generate_day2_static_network_data_from_tf(tf_folder, num_workers):
```
empovit, line 18: `num_workers` are additional workers, aren't they? Meaning not all workers, but only day 2 ones. I think it's worth reflecting in the name.

discovery-infra/day2.py
yevgeny-shnaidman: with static ips set, hostname is not set, since those ips are not configured in the terraform. This is also what will probably happen in production

discovery-infra/day2.py
yevgeny-shnaidman: i am open to suggestions. but it explains to the reader that the hostnames are set only in case it is needed

discovery-infra/test_infra/tools/static_network.py
yevgeny-shnaidman: fixed

discovery-infra/day2.py
yevgeny-shnaidman: fixed

discovery-infra/day2.py
yevgeny-shnaidman: fixed

PR #615, opened 2021-03-18
NO-ISSUE: Added Agent CRD

RazRegev:
An Agent CRD will be created for each registered host. In order to start the
installation all agents must be approved.


RazRegev:
will open another one with the ticket name

PR #616, opened 2021-03-18
MGMT-4624: Added Agent CRD

RazRegev:
An Agent CRD will be created for each registered host. In order to start the
installation all agents must be approved.


YuviGold (approved):
LGTM

PR #620, opened 2021-03-19
NO-ISSUE: run CI in kube-api mode for e2e tests

RazRegev:


osherdp (approved):
/lgtm

YuviGold:
I'm not sure what is the outcome of this operation. Would we want always to run in kubeapi mode?

RazRegev:
> I'm not sure what is the outcome of this operation. Would we want always to run in kubeapi mode?

yes, like we do in aws ci

PR #621, opened 2021-03-21
NO-ISSUE: split kube helpers file

RazRegev:
Since it is a large file that will only get bigger,
it will be better to split it to a sub package.

### DIFF
--- a/discovery-infra/test_infra/helper_classes/kube_helpers/__init__.py
+++ b/discovery-infra/test_infra/helper_classes/kube_helpers/__init__.py
@@ -1,0 +1,44 @@
+"""
+kube_helpers package provides infra to deploy, manage and install cluster using
+CRDs instead of restful API calls.
+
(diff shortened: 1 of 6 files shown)


discovery-infra/test_infra/helper_classes/kube_helpers/__init__.py
```diff
@@ -0,0 +1,44 @@
+"""
+kube_helpers.py provides infra to deploy, manage and install cluster using
```
osherdp, line 2: module not existing :)

osherdp (changes requested):

discovery-infra/test_infra/helper_classes/kube_helpers/__init__.py
RazRegev: ops

osherdp (approved):
/lgtm

PR #623, opened 2021-03-21
Added call to create_testgrid_tickets.py to the download_logs job

ronniel1:
This script create triaging tickets for failed jobs, as marked in testgrid


omertuc:
/hold 
`deploymentRepo` has moved:
https://github.com/openshift-assisted/assisted-installer-deployment/

PR #627, opened 2021-03-22
MGMT-4342: Added InstallEnv CRD infra

RazRegev:


osherdp:
It's a bit duplicated with #635 
Maybe we want to merge some changes to the above PR?

RazRegev:
can we push this PR and then you can rebase on the top of it?
the installenv crd is a separate task, and it was here before :)

discovery-infra/test_infra/helper_classes/kube_helpers/globals.py
```diff
@@ -0,0 +1,14 @@
+from tests.conftest import env_variables as _env
```
osherdp, line 1: what's the purpose of this rename?

discovery-infra/test_infra/helper_classes/kube_helpers/cluster_deployment.py
```diff
@@ -7,18 +7,13 @@
 from kubernetes.client.rest import ApiException
 from tests.conftest import env_variables
 
-
+from .globals import *
```
osherdp, line 10: I don't really like doing it, can we just import those constants directly?

discovery-infra/test_infra/helper_classes/kube_helpers/installenv.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+class InstallEnv(BaseCustomResource):
+    """
+    InstallEnv is used to generate cluster iso.
+    Image is automatically generated on CRD deployment, after InstallEnv is
+    reconciled. Image download url will be exposed in the status.
+    """
+    _api_group = 'adi.io.my.domain'
+    _api_version = 'v1alpha1'
```
osherdp, line 25: let's be consistent with ``_version`` as in ``ClusterDeployment``

discovery-infra/test_infra/helper_classes/kube_helpers/installenv.py
```diff
@@ -0,0 +1,232 @@
...
+from .common import logger
+
+
+class InstallEnv(BaseCustomResource):
+    """
+    InstallEnv is used to generate cluster iso.
+    Image is automatically generated on CRD deployment, after InstallEnv is
+    reconciled. Image download url will be exposed in the status.
+    """
+    _api_group = 'adi.io.my.domain'
```
osherdp, line 24: Let's be consistent with ``_api_group`` as in ``ClusterDeployment``

osherdp (changes requested):

discovery-infra/test_infra/helper_classes/kube_helpers/installenv.py
RazRegev: cluster deployment CRD is part of hive definitions where it is v1
installenv CRD made by us, where we still don't have v1 for our api

discovery-infra/test_infra/helper_classes/kube_helpers/installenv.py
RazRegev: cluster deployment CRD is part of hive api group because this CRD is belong to hive
installenv is our CRD so it won't work if we deploy it to hive api group

discovery-infra/test_infra/helper_classes/kube_helpers/globals.py
RazRegev: not relevant reason, removed

discovery-infra/test_infra/helper_classes/kube_helpers/installenv.py
osherdp: I mean the variable name, not the value

discovery-infra/test_infra/helper_classes/kube_helpers/installenv.py
osherdp: Same as above ^

osherdp (changes requested):

PR #633, opened 2021-03-24
MGMT-4604 Fix scality ephemeral storage error

empovit:
Increase minikube disk allocation so that the VM
has enough space for minikube components and ephemeral storage.

### DIFF
--- a/scripts/run_minikube.sh
+++ b/scripts/run_minikube.sh
@@ -20,7 +20,7 @@ function init_minikube() {
         fi
     done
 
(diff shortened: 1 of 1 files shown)


osherdp:
@empovit you think this will help?
The error message says something about getting to some number of kilobytes, whereas it's limited by 0

empovit:
> @empovit you think this will help?
> The error message says something about getting to some number of kilobytes, whereas it's limited by 0

Yes I do. "0" means not limited, so it gets to the limit (or rather threshold) of total of ephemeral storage available on the node.

osherdp:
/lgtm
OK, great

PR #636, opened 2021-03-25
MGMT-3334 Wait for operators to finish

YuviGold:
As part of waiting for installation to complete - we look for hosts and cluster to be "installed". Now we watch for operators to be "Available" or "Failed" as well.
    
Remove "-cvo" flag to inspect cvo status since it's being done by the service.

```
2021-03-25 14:10:23,067 INFO       - 139877912267648 - Asked operators to be in one of the statuses from ['available', 'failed'] and currently operators statuses are [('console', 'available', 'All is well')]    (/home/assisted-test-infra/discovery-infra/test_infra/utils.py:232)
```

### DIFF
--- a/discovery-infra/test_infra/utils.py
+++ b/discovery-infra/test_infra/utils.py
@@ -18,7 +18,7 @@ from functools import wraps
 from pathlib import Path
 from pprint import pformat
 from string import ascii_lowercase
-from typing import List
+from typing import List, Dict
 
 import filelock
 import libvirt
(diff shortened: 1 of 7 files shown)


discovery-infra/test_infra/utils.py
```diff
@@ -225,6 +200,22 @@ def are_hosts_in_status(
...
+        "Asked operators to be in one of the statuses from %s and currently operators statuses are %s",
+        statuses,
+        [(operator["name"], operator.get("status"), operator.get("status_info")) for operator in operators],
+    )
+
+    if len([operator for operator in operators if operator.get("status") in statuses]) >= operators_count:
+        return True
+
+    if fall_on_error_status and len([operator for operator in operators if operator.get("status") == consts.OperatorStatus.FAILED]) > 0:
+        raise ValueError("All the operators must be in valid status, but got some in error")
```
pkliczewski, line 214: What do you think about using regular loop instead of list comprehension so we could give information about which operator failed. Now we got `got some in error` and it is vague

pkliczewski (approved):

pkliczewski (approved):

PR #637, opened 2021-03-25
MGMT-4761 Clear all nat iptables rules when tearing down tests using user-managed-networking

ori-amizur:
/cc @tsorya 
/cc @YuviGold

### DIFF
--- a/discovery-infra/test_infra/helper_classes/cluster.py
+++ b/discovery-infra/test_infra/helper_classes/cluster.py
@@ -3,6 +3,7 @@ import ipaddress
 import json
 import logging
 import random
+import re
 import time
 from collections import Counter
 from typing import List
@@ -539,7 +540,7 @@ class Cluster:
             platform=env_variables['platform']
     ):
         if platform == consts.Platforms.NONE and not nodes.controller.ipv6:
-            self._configure_nat(nodes.controller)
+            self.configure_nat(nodes.controller)
 
         if download_image:
             if env_variables.get('static_network_config'):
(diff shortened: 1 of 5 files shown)


discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -736,14 +737,27 @@ def _configure_load_balancer(self, controller):
...
         lb_controller.set_load_balancing_config(load_balancer_ip, master_ips, worker_ips)
 
+    @classmethod
+    def _get_namespace_index(cls, libvirt_network_if):
+        # Hack to retrieve namespace index - does not exist in tests
+        matcher = re.match(r'^tt(\d+)$', libvirt_network_if)
+        return int(matcher.groups()[0]) if matcher is not None else 0
+
     @classmethod
     def _configure_nat(cls, controller):
```
eliorerz, line 747: Maybe set it as public so we could create fixture on BaseTest

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -736,14 +737,27 @@ def _configure_load_balancer(self, controller):
...
         libvirt_network_if = controller.network_conf.libvirt_network_if
         libvirt_secondary_network_if = controller.network_conf.libvirt_secondary_network_if
         input_interfaces = [libvirt_network_if, libvirt_secondary_network_if]
+        nat_controller = NatController()
+        nat_controller.add_nat_rules(input_interfaces, cls._get_namespace_index(libvirt_network_if))
 
+    @classmethod
+    def _unconfigure_nat(cls, controller):
```
eliorerz, line 755: Same here.. Maybe set it as public so we could create fixture on BaseTest

discovery-infra/test_infra/helper_classes/cluster.py
ori-amizur: OK

discovery-infra/test_infra/helper_classes/cluster.py
ori-amizur: OK

PR #643, opened 2021-03-29
Run minikube tunnel when starting minikube

YuviGold:
In order for k8s services to have external IPs accessible from kubectl
 - minikube have to be started with a LoadBalancer.
By default minikube isn't starting with a LoadBalancer and for that we
have to run `minikube tunnel` on a separate process, since it is running
in the foreground.


flaper87 (approved):
/lgtm

PR #651, opened 2021-03-31
Increase installation timeout

ukalifon:
Increasing to 2 hours, to account for all the additional logic (for
exmaple of the operators) which increase the installation length to
well over 1 hour.


discovery-infra/test_infra/consts.py
```diff
@@ -7,7 +7,7 @@
 STORAGE_PATH = "/var/lib/libvirt/openshift-images"
 SSH_KEY = "ssh_key/key.pub"
 NODES_REGISTERED_TIMEOUT = 60 * 20
-CLUSTER_INSTALLATION_TIMEOUT = 60 * 60  # 60 minutes
+CLUSTER_INSTALLATION_TIMEOUT = 2 * 60 * 60  # 2 hours
```
lalon4, line 10: Cannot understand why it is needed, start_install_and_wait_for_installed uses 3 different method in order to declare ready:
wait_for_hosts_to_install
wait_for_operators_to_finish
wait_for_install
each method has 1 hour timeout, dont you think its enough?

YuviGold: Maybe we can even add `wait_for_finalizing` just for even shorter timeouts?

lalon4: but finalizing is where the operators are installed, so we do need a long timeout

PR #653, opened 2021-03-31
MGMT-4813: If not provided, use INSTALLER_KUBECONFIG env

RazRegev:
in order to determine to which cluster we deploy CRDs to,
in needs to be a separate env than KUBECONFIG which uses
to represents the installed cluster kubeconfig path.


discovery-infra/tests/conftest.py
```diff
@@ -54,6 +54,7 @@ def _get_cluster_name():
                  "cluster_name": _get_cluster_name(),
                  "private_ssh_key_path": utils.get_env('PRIVATE_KEY_PATH', private_ssh_key_path_default),
                  "kubeconfig_path": utils.get_env('KUBECONFIG', ''),
+                 "installer_kubeconfig_path": utils.get_env('INSTALLER_KUBECONFIG', None),
```
osherdp, line 57: we need to add ``INSTALLER_KUBECONFIG`` to ``skipper.env``, no?
also giving it a default value on the ``Makefile``

osherdp (changes requested):

discovery-infra/tests/conftest.py
RazRegev: added to skipper.env 
not sure we need to add it to Makefile, if not set it will go to the default path

PR #654, opened 2021-03-31
NO-ISSUE implement a 'kubectl apply' like interface

osherdp:
Current code is just too long, and the behavior we want is a simple 'kubectl apply'.
So I replaced deletion and creation with the new methods.
Kubernetes python client doesn't support apply right now, so it will be a simple substitution for now (until we'll mimic the 3-way patch of apply).

/hold
/cc @RazRegev

### DIFF
--- a/discovery-infra/start_discovery.py
+++ b/discovery-infra/start_discovery.py
@@ -3,7 +3,6 @@
 
 import pathlib
 import argparse
-import contextlib
 import distutils.util
 import ipaddress
 import json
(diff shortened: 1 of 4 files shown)


RazRegev:
so shouldn't this apply function needs to be added to the resource base class?

RazRegev (approved):

osherdp:
> so shouldn't this apply function needs to be added to the resource base class?

Done
I guess maybe we can extract more logic to the abstract class in the next tasks

osherdp:
kube-api job failed because of conformance tests running on presubmits
so continuing and I'll fix it

RazRegev (approved):

RazRegev (approved):

PR #655, opened 2021-04-01
MGMT-3335 teardown iptables in nodes

eliorerz:
Put `configure_nat` and `unconfigure_nat` methods in nodes.py and tear them down if needed in the nodes fixture. Also, make NatController a static class, no `self` is needed.

### DIFF
--- a/discovery-infra/test_infra/controllers/nat_controller.py
+++ b/discovery-infra/test_infra/controllers/nat_controller.py
@@ -1,83 +1,98 @@
 import logging
 from test_infra.utils import run_command
 
-#
-# Create NAT rules networks that are nat using libvirt "nat" forwarding - which is currently none platform
-# The logic behind it is to mark packets that are coming from libvirt bridges (i.e input_interfaces), and
-# reference this mark in order to perform NAT operation on these packets.
-#
+
 class NatController:
-    # Build iptables mark
-    def _build_mark(self, ns_index):
+    """
+    Create NAT rules networks that are nat using libvirt "nat" forwarding - which is currently none platform
+    The logic behind it is to mark packets that are coming from libvirt bridges (i.e input_interfaces), and
+    reference this mark in order to perform NAT operation on these packets.
+    """
+
+    @staticmethod
+    def _build_mark(ns_index):
+        """ Build iptables mark """
         return 555 + int(ns_index)
 
-    # Find all interfaces that have default route on them.  Usually it is a single interface.
-    def _get_default_interfaces(self):
+    @staticmethod
+    def _get_default_interfaces():
+        """ Find all interfaces that have default route on them.  Usually it is a single interface. """
         interfaces, _, _ = run_command(r"ip -4 route | egrep '^default ' | awk '{print $5}'", shell=True)
         return set(interfaces.strip().split())
 
-    # Mark all packets coming from the input_interface with "555".  Marking is needed because input interface
-    # query is not available in POSTROUTING chain
-    def _build_mark_string(self, input_interface, mark):
+    @staticmethod
+    def _build_mark_string(input_interface, mark):
+        """ Mark all packets coming from the input_interface with "555".  Marking is needed because input interface
+        query is not available in POSTROUTING chain """
         rule_template = ["PREROUTING", "-i", input_interface, "-j", "MARK", "--set-mark", f"{mark}"]
 
         return " ".join(rule_template)
 
-    # Perform MASQUERADE nat operation  on all marked packets with "555" and their output interface is "output_interface"
-    def _build_nat_string(self, output_interface, mark):
+    @staticmethod
+    def _build_nat_string(output_interface, mark):
+        """ Perform MASQUERADE nat operation  on all marked packets with "555" and their output interface
+        is 'output_interface' """
         rule_template = ["POSTROUTING", "-m", "mark", "--mark", f"{mark}", "-o", output_interface, "-j", "MASQUERADE"]
 
         return " ".join(rule_template)
 
-    # Build iptables command
-    def _build_rule_string(self, option, rule_suffix):
-        rule_template = ["iptables", "-t", "nat",  f"--{option}", rule_suffix]
+    @staticmethod
+    def _build_rule_string(option, rule_suffix):
+        """ Build iptables command """
+        rule_template = ["iptables", "-t", "nat", f"--{option}", rule_suffix]
 
         return " ".join(rule_template)
 
-    # Check if rule exists
-    def _does_rule_exist(self, rule_suffix):
-        check_rule = self._build_rule_string('check', rule_suffix)
+    @classmethod
+    def _does_rule_exist(cls, rule_suffix):
+        """ Check if rule exists """
+        check_rule = cls._build_rule_string('check', rule_suffix)
         _, _, exit_code = run_command(check_rule, shell=True, raise_errors=False)
 
         return exit_code == 0
 
-    # Insert a new rule
-    def _insert_rule(self, rule_suffix):
-        insert_rule = self._build_rule_string('insert', rule_suffix)
+    @classmethod
+    def _insert_rule(cls, rule_suffix):
+        """ Insert a new rule """
+        insert_rule = cls._build_rule_string('insert', rule_suffix)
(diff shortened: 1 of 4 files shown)


eliorerz:
Please take a look @ori-amizur

discovery-infra/test_infra/controllers/nat_controller.py
```diff
@@ -1,83 +1,98 @@
...
+    def _build_mark_string(cls, input_interface, mark):
+        """ Mark all packets coming from the input_interface with "555".  Marking is needed because input interface
+        query is not available in POSTROUTING chain """
         rule_template = ["PREROUTING", "-i", input_interface, "-j", "MARK", "--set-mark", f"{mark}"]
 
         return " ".join(rule_template)
 
-    # Perform MASQUERADE nat operation  on all marked packets with "555" and their output interface is "output_interface"
-    def _build_nat_string(self, output_interface, mark):
+    @classmethod
```
omertuc, line 31: @staticmethod

discovery-infra/test_infra/controllers/nat_controller.py
```diff
@@ -1,83 +1,98 @@
...
+    def _get_default_interfaces(cls):
+        """ Find all interfaces that have default route on them.  Usually it is a single interface. """
         interfaces, _, _ = run_command(r"ip -4 route | egrep '^default ' | awk '{print $5}'", shell=True)
         return set(interfaces.strip().split())
 
-    # Mark all packets coming from the input_interface with "555".  Marking is needed because input interface
-    # query is not available in POSTROUTING chain
-    def _build_mark_string(self, input_interface, mark):
+    @classmethod
```
omertuc, line 23: `@staticmethod`

discovery-infra/test_infra/controllers/nat_controller.py
```diff
@@ -1,83 +1,98 @@
...
+    """
+
+    @classmethod
+    def _build_mark(cls, ns_index):
+        """ Build iptables mark """
         return 555 + int(ns_index)
 
-    # Find all interfaces that have default route on them.  Usually it is a single interface.
-    def _get_default_interfaces(self):
+    @classmethod
```
omertuc, line 17: `@staticmethod`

discovery-infra/test_infra/controllers/nat_controller.py
```diff
@@ -1,83 +1,98 @@
...
 class NatController:
-    # Build iptables mark
-    def _build_mark(self, ns_index):
+    """
+    Create NAT rules networks that are nat using libvirt "nat" forwarding - which is currently none platform
+    The logic behind it is to mark packets that are coming from libvirt bridges (i.e input_interfaces), and
+    reference this mark in order to perform NAT operation on these packets.
+    """
+
+    @classmethod
```
omertuc, line 12: `@staticmethod`

discovery-infra/test_infra/controllers/nat_controller.py
```diff
@@ -1,83 +1,98 @@
...
+        """ Perform MASQUERADE nat operation  on all marked packets with "555" and their output interface
+        is 'output_interface' """
         rule_template = ["POSTROUTING", "-m", "mark", "--mark", f"{mark}", "-o", output_interface, "-j", "MASQUERADE"]
 
         return " ".join(rule_template)
 
-    # Build iptables command
-    def _build_rule_string(self, option, rule_suffix):
-        rule_template = ["iptables", "-t", "nat",  f"--{option}", rule_suffix]
+    @classmethod
```
omertuc, line 39: `@staticmethod`

discovery-infra/tests/base_test.py
```diff
@@ -40,17 +40,22 @@ def nodes(self, setup_node_controller, request):
         else:
             node_vars = env_variables
         net_asset = None
+        needs_nat = "platform" in node_vars and node_vars["platform"] == consts.Platforms.NONE
```
omertuc, line 43: `needs_nat = node_vars.get("platform") == consts.Platforms.NONE

omertuc, line 43: `needs_nat = node_vars.get("platform") == consts.Platforms.NONE`

discovery-infra/test_infra/helper_classes/nodes.py
```diff
@@ -41,6 +44,24 @@ def __iter__(self) -> Iterator[Node]:
         for n in self.nodes:
             yield n
 
+    @classmethod
```
omertuc, line 47: `@staticmethod`

discovery-infra/test_infra/helper_classes/nodes.py
```diff
@@ -41,6 +44,24 @@ def __iter__(self) -> Iterator[Node]:
...
+    def _get_namespace_index(cls, libvirt_network_if):
+        # Hack to retrieve namespace index - does not exist in tests
+        matcher = re.match(r'^tt(\d+)$', libvirt_network_if)
+        return int(matcher.groups()[0]) if matcher is not None else 0
+
+    def configure_nat(self):
+        libvirt_network_if = self.controller.network_conf.libvirt_network_if
```
omertuc, line 54: Maybe create

```
@property
def nat_interfaces(self):
	return (self.controller.network_conf.libvirt_network_if, self.controller.network_conf.libvirt_secondary_network_if)
```

Then the configure functions become simpler:
```
def configure_nat(self):
        self._nat_controller.add_nat_rules(self.nat_interfaces, self._get_namespace_index(libvirt_network_if))

    def unconfigure_nat(self):
        self._nat_controller.remove_nat_rules(self.nat_interfaces, self._get_namespace_index(libvirt_network_if))
```

omertuc: Oh sorry, didn't notice you just copy pasted that code. It would still be nice to change that if you're already changing this code

eliorerz:
All fixed @omertuc

discovery-infra/tests/base_test.py
```diff
@@ -40,17 +40,22 @@ def nodes(self, setup_node_controller, request):
...
+        needs_nat = node_vars.get("platform") == consts.Platforms.NONE
         try:
             if not qe_env:
                 net_asset = NetworkAssets()
                 node_vars["net_asset"] = net_asset.get()
             controller = setup_node_controller(**node_vars)
             nodes = Nodes(controller, node_vars["private_ssh_key_path"])
             nodes.prepare_nodes()
+            if needs_nat:
+                nodes.configure_nat()
```
ori-amizur, line 52: Why not put in its own fixture?

discovery-infra/test_infra/helper_classes/nodes.py
```diff
@@ -22,6 +24,7 @@ class Nodes:
     def __init__(self, node_controller: NodeController, private_ssh_key_path):
         self.controller = node_controller
         self.private_ssh_key_path = private_ssh_key_path
+        self._nat_controller = NatController()
```
ori-amizur, line 27: If the class is static why is it necessary to create this instance?

ori-amizur (commented):
Please make sure that the makefile flow (not tests), is still working after changing NatController to be static.

discovery-infra/tests/base_test.py
eliorerz: Because it races with nodes

eliorerz:
@ori-amizur  _Please make sure that the makefile flow (not tests), is still working after changing NatController to be static._
There is still no change in the API or logic. We can fix start_discovery to make it a static class but that is not supposed (In my opinion) to be in this PR.

discovery-infra/tests/base_test.py
ori-amizur: So isn't it just the order of the fixtures which is the order of parameters?

eliorerz: Both, If NAT was a standalone fixture it must have been called after nodes fixture, in this case I used it inside nodes fixture right after initializing the Nodes instance. The other option is to use it in the test itself with try finally block.
I see this implementation the best approach of the three

ori-amizur: OK

PR #661, opened 2021-04-04
NOISSUE: deploy_default_cluster_deployment with objects instead of dict

RazRegev:
and small cosmetics to cluster_deployment.py


eranco74 (commented):
/lgtm

PR #662, opened 2021-04-04
MGMT-4706 - add api to the assisted-service api to support mirror reg…

yevgeny-shnaidman:
…istry and CA configuration


YuviGold (approved):
LGTM

PR #668, opened 2021-04-05
NO-ISSUE ignore .env file to enable customization

osherdp:
/cc @YuviGold

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ minikube
 *.log
 **/RCS/**
 *~
+.env


YuviGold:
Actually, I think we should use `.env` files. For example, `dev.env`, `ci.env`, `prod.env`, instead of maintaining variables logic inside the Makefile.

osherdp:
yes, but it's a convention that ``.env`` is for the local environment
you can still create those files and they will not be ignored

YuviGold (approved):

PR #671, opened 2021-04-06
NO-ISSUE do not attemp to install from REST if kubeapi cluster is ready

osherdp:
What sometimes happens is the following:
* We define a cluster deployment
* Cluster installation doesn't start yet, and we get a ready state from it
* We attempt to install cluster from REST instead of kubeapi

So adding a condition not not do the REST installation.
/cc @YuviGold 
/hold


YuviGold (approved):
LGTM so it won't block
Although, we might want to split functions since it seems most of `run_install_flow` flow is REST oriented

PR #673, opened 2021-04-06
NO-ISSUE pull latest code on origin remote

osherdp:
Seems like we don't pull latest code on ``assisted-service`` and ``assisted-installer``.
Added now reference to the ``origin`` remote and not the saved ``master`` branch.

/hold
/cc @YuviGold @filanov

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,13 +12,13 @@ REPORTS = $(ROOT_DIR)/reports
 SKIPPER_PARAMS ?= -i
 
 # assisted-service
-SERVICE_BRANCH := $(or $(SERVICE_BRANCH), "master")
+SERVICE_BRANCH := $(or $(SERVICE_BRANCH), "origin/master")
 SERVICE_REPO := $(or $(SERVICE_REPO), "https://github.com/openshift/assisted-service")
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -12,13 +12,13 @@ REPORTS = $(ROOT_DIR)/reports
...
 # assisted-service
-SERVICE_BRANCH := $(or $(SERVICE_BRANCH), "master")
+SERVICE_BRANCH := $(or $(SERVICE_BRANCH), "origin/master")
 SERVICE_REPO := $(or $(SERVICE_REPO), "https://github.com/openshift/assisted-service")
 SERVICE := $(or $(SERVICE), quay.io/ocpmetal/assisted-service:latest)
 SERVICE_NAME := $(or $(SERVICE_NAME),assisted-service)
 
 # assisted-installer
-INSTALLER_BRANCH := $(or $(INSTALLER_BRANCH), "master")
+INSTALLER_BRANCH := $(or $(INSTALLER_BRANCH), "origin/master")
```
tsorya, line 21: if there is example in the doc, we should change it to have origin in it

osherdp: right

PR #674, opened 2021-04-06
NO-ISSUE disable kube-api by default

tsorya:
Set service replica count to 1 if kube-api is enabled

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,8 +76,6 @@ IMAGE_TAG := latest
 DEPLOY_TAG := $(or $(DEPLOY_TAG), "")
 DEPLOY_MANIFEST_PATH := $(or $(DEPLOY_MANIFEST_PATH), "")
 DEPLOY_MANIFEST_TAG := $(or $(DEPLOY_MANIFEST_TAG), "")
-SERVICE_REPLICAS_COUNT := 3
-
 IMAGE_NAME=test-infra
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -107,9 +105,13 @@ CONTROLLER_OCP := $(or ${CONTROLLER_OCP},quay.io/ocpmetal/assisted-installer-con
...
 PLATFORM := $(or ${PLATFORM},baremetal)
 IPV6_SUPPORT := $(or ${IPV6_SUPPORT},true)
-ENABLE_KUBE_API := $(or ${ENABLE_KUBE_API},true)
-
+SERVICE_REPLICAS_COUNT := 3
+ENABLE_KUBE_API := $(or ${ENABLE_KUBE_API},no)
 KUBE_API := $(or ${KUBE_API},no)
+ifeq ($(ENABLE_KUBE_API),true)
+	SERVICE_REPLICAS_COUNT=1
+	KUBE_API=yes
```
osherdp, line 113: Not sure about that
This will lead to confusion between ``ENABLE_KUBE_API`` and ``KUBE_API`` which is pretty confusing as of know anyway

osherdp (changes requested):

Makefile
tsorya: yep, i thought that if you already set ENABLE_KUBE_API it means that you want to run KUBE_API flow.

PR #676, opened 2021-04-06
MGMT-4093 - Set cluster name and dns on terraform when deploying nodes to specific cluster

eliorerz:
[MGMT-4093](https://issues.redhat.com/browse/MGMT-4093) 

When creating a cluster manually and trying to deploy hosts to it test-infra grab cluster name and base DNS from the cluster.


eliorerz:
Duplicate - https://github.com/openshift/assisted-test-infra/pull/685

PR #678, opened 2021-04-06
NOISSUE: remove any rest api call from kube api flow

RazRegev:
replace any rest api call to edit resources in assisted service
with matching kube-api call instead.

### DIFF
--- a/discovery-infra/install_cluster.py
+++ b/discovery-infra/install_cluster.py
@@ -85,60 +85,131 @@ def wait_till_installed(client, cluster, timeout=60 * 60 * 2):
 # 2. Running install cluster api
 # 3. Waiting till all nodes are in installing status
 # 4. Downloads kubeconfig for future usage
-def run_install_flow(client, cluster_id, kubeconfig_path, pull_secret,
-                     tf=None, kube_client=None, cluster_deployment=None):
+def run_install_flow(
+        client,
+        cluster_id,
+        kubeconfig_path,
+        pull_secret,
+        tf,
+        cluster_deployment=None,
+        nodes_number=None,
+):
+    if cluster_deployment is None:
+        run_installation_flow_rest_api(
+            client=client,
+            cluster_id=cluster_id,
+            pull_secret=pull_secret,
+            kubeconfig_path=kubeconfig_path,
+            tf=tf,
+        )
+    else:
+        run_installation_flow_kube_api(
+            cluster_deployment=cluster_deployment,
+            nodes_number=nodes_number,
+            kubeconfig_path=kubeconfig_path,
+        )
+
+
+def run_installation_flow_rest_api(
+        client,
+        cluster_id,
+        pull_secret,
+        kubeconfig_path,
+        tf
+):
     log.info("Verifying cluster exists")
     cluster = client.cluster_get(cluster_id)
+
+    start_cluster_installation_rest_api(
+        cluster=cluster,
+        client=client,
+        cluster_id=cluster_id,
+        pull_secret=pull_secret,
+        kubeconfig_path=kubeconfig_path,
+    )
+
+    log.info("Waiting until cluster finishes installation")
+    wait_till_installed(client=client, cluster=cluster)
+
+    log.info("Downloading kubeconfig")
+    waiting.wait(
+        lambda: client.download_kubeconfig(
+            cluster_id=cluster_id, kubeconfig_path=kubeconfig_path
+        ) is None,
+        timeout_seconds=240,
+        sleep_seconds=20,
+        expected_exceptions=Exception,
+        waiting_for="Kubeconfig",
+    )
+
+    if tf is not None:
+        update_vip_from_tf(client, cluster_id, tf)
+
+
+def start_cluster_installation_rest_api(
+        client,
+        cluster,
+        cluster_id,
+        pull_secret,
+        kubeconfig_path
+):
     log.info("Verifying pull secret")
-    verify_pull_secret(client=client, cluster=cluster, pull_secret=pull_secret)
+    verify_pull_secret(
+        client=client,
+        cluster=cluster,
+        pull_secret=pull_secret,
+    )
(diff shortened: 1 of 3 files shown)


discovery-infra/start_discovery.py
```diff
@@ -500,6 +540,42 @@ def set_hosts_roles(client, cluster, nodes_details, machine_net, tf, master_coun
...
+    )
+    return utils.get_libvirt_nodes_from_tf_state(
+        network_names=networks_names,
+        tf_state=tf.get_state(),
+    )
+
+
+def set_agent_hostname(agent, nodes_details, is_ipv4):
+    if is_ipv4 and not args.with_static_network_config:
+        return
```
osherdp, line 558: append blank line

osherdp:
@RazRegev maybe retitle the PR?

discovery-infra/start_discovery.py
```diff
@@ -409,28 +413,51 @@ def nodes_flow(client, cluster_name, cluster, machine_net, kube_client=None, clu
...
                     client, cluster.id, main_cidr, ipv4_first=is_ipv4)})
-                set_cluster_machine_cidr(client, cluster.id, machine_net, set_vip_dhcp_allocation=False)
+                if args.kube_api:
+                    cluster_deployment.patch(
+                        install_strategy=InstallStrategy(
+                            host_prefix=args.host_prefix if is_ipv4 else args.host_prefix6,
+                            machine_cidr=get_machine_cidr_from_machine_net(machine_net),
+                            cluster_cidr=args.cluster_network if is_ipv4 else args.cluster_network6,
+                            service_cidr=args.service_network if is_ipv4 else args.service_network6,
+                            ssh_public_key=args.ssh_key,
```
osherdp, line 449: I think @nmagnezi moved this field to ``InstallEnv``, no?

discovery-infra/start_discovery.py
```diff
@@ -409,28 +413,51 @@ def nodes_flow(client, cluster_name, cluster, machine_net, kube_client=None, clu
...
+                            control_plane_agents=args.master_count,
+                            worker_agents=args.number_of_workers,
+                        ),
+                    )
```
osherdp, line 453: Do we even need this patch?
I thought we already set all those fields with the right value
Alternatively, we must re-specify all the fields here?

discovery-infra/start_discovery.py
```diff
@@ -500,6 +540,42 @@ def set_hosts_roles(client, cluster, nodes_details, machine_net, tf, master_coun
...
+        tf_vars['libvirt_network_name'],
+        tf_vars['libvirt_secondary_network_name'],
+    )
+    return utils.get_libvirt_nodes_from_tf_state(
+        network_names=networks_names,
+        tf_state=tf.get_state(),
+    )
+
+
+def set_agent_hostname(agent, nodes_details, is_ipv4):
```
osherdp, line 556: maybe some of the functions should be in some utility module for kubeapi functions

osherdp (changes requested):

discovery-infra/start_discovery.py
RazRegev: you're right, this is unnecessary

discovery-infra/test_infra/kubeapi_utils.py
```diff
@@ -0,0 +1,91 @@
...
+        functools.partial(get_hostname_for_agent, nodes_details, agent),
+        timeout_seconds=60,
+        sleep_seconds=1,
+        waiting_for=f'agent={agent.ref} to find a hostname',
+    )
+    logger.info('patching agent hostname=%s', hostname)
+    agent.patch(hostname=hostname)
+
+
+def get_hostname_for_agent(nodes_details, agent):
```
osherdp, line 84: It bugs me that the order of the parameters is not the same as above function

discovery-infra/test_infra/kubeapi_utils.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def get_bmc_address():
+        interfaces = agent.status().get('inventory', {}).get('interfaces')
+        if not interfaces:
+            return
+        ip_addresses = interfaces[0].get(
+            'ipV4Addresses' if is_ipv4 else 'ipV6Addresses'
+        )
+        if not ip_addresses:
+            return
```
osherdp, line 27: append blank line

discovery-infra/test_infra/kubeapi_utils.py
```diff
@@ -0,0 +1,91 @@
...
+        interface_class = IPv4Interface if is_ipv4 else IPv6Interface
+        return str(interface_class(ip_addresses[0]).ip)
```
osherdp, line 29: not sure it's much Pythonic all this area
can we change it in some way?

discovery-infra/install_cluster.py
```diff
@@ -85,60 +85,131 @@ def wait_till_installed(client, cluster, timeout=60 * 60 * 2):
 # 2. Running install cluster api
 # 3. Waiting till all nodes are in installing status
 # 4. Downloads kubeconfig for future usage
-def run_install_flow(client, cluster_id, kubeconfig_path, pull_secret,
-                     tf=None, kube_client=None, cluster_deployment=None):
+def run_install_flow(
```
osherdp, line 88: yes, that's a much better approach :)
Maybe in the future we'll have two classes that implements the same interface, one for REST and one for kube-api

osherdp (changes requested):

discovery-infra/test_infra/kubeapi_utils.py
RazRegev: bless you

discovery-infra/install_cluster.py
RazRegev: Requires a huge refactoring but I couldn't agree more

osherdp (approved):

PR #682, opened 2021-04-07
Adding cluster methods for e2e

lalon4:
Adding wait_for_cluster_to_be_in_finalizing_status and is_operator_in_status methods for e2e that simulate operator failure

### DIFF
--- a/discovery-infra/test_infra/helper_classes/cluster.py
+++ b/discovery-infra/test_infra/helper_classes/cluster.py
@@ -536,6 +536,11 @@ class Cluster:
             fall_on_error_status=False,
         )
 
+    def is_operator_in_status(self, operator_name, status):
+        return utils.is_operator_in_status(operators=self.get_operators(),
+                                           operator_name=operator_name,
+                                           status=status)
+
     def wait_for_install(
             self,
             timeout=consts.CLUSTER_INSTALLATION_TIMEOUT
@@ -723,6 +728,15 @@ class Cluster:
             cluster_id=self.id,
             statuses=[consts.ClusterStatus.INSTALLING],
             timeout=consts.START_CLUSTER_INSTALLATION_TIMEOUT
+        )
+
+    def wait_for_cluster_to_be_in_finalizing_status(self):
+        utils.wait_till_cluster_is_in_status(
+            client=self.api_client,
+            cluster_id=self.id,
+            statuses=[consts.ClusterStatus.FINALIZING, consts.ClusterStatus.INSTALLED],
+            timeout=consts.CLUSTER_INSTALLATION_TIMEOUT,
+            break_statuses=[consts.ClusterStatus.ERROR]
         )
 
     @classmethod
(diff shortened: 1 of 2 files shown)


discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -725,6 +730,15 @@ def wait_for_cluster_to_be_in_installing_status(self):
             timeout=consts.START_CLUSTER_INSTALLATION_TIMEOUT
         )
 
+    def wait_for_cluster_to_be_in_finalizing_status(self):
+        utils.wait_till_cluster_is_in_status(
+            client=self.api_client,
+            cluster_id=self.id,
+            statuses=[consts.ClusterStatus.FINALIZING],
```
omertuc, line 737: Anyway to immediately "timeout" here in case the cluster is in a state from which it can never be finalizing? e.g. error states and "installation complete" state?

discovery-infra/test_infra/utils.py
```diff
@@ -220,6 +220,19 @@ def are_operators_in_status(operators: List[MonitoredOperator], operators_count:
...
 
 
+def is_operator_in_status(operators: List[MonitoredOperator], operator_name: str, status: str) -> bool:
+    log.info(
+        f"Asked operator %s to be in status: %s, and currently operators statuses are %s",
+        operator_name,
+        status,
+        [(operator.name, operator.status, operator.status_info) for operator in operators],
+    )
+    for operator in operators:
```
omertuc, line 230: Can be simplified to:

```python
return any(operator.name == operator_name and operator.status == status for operator in operators)
```

discovery-infra/test_infra/helper_classes/cluster.py
omertuc: Oh missed that, cool. Then can you just add `consts.ClusterStatus.INSTALLED` to `statuses=`?

lalon4: I need to get into Finalizing stage and not installed. 
In my e2e case, i wait until Finalizing, because then i can mess up the nodes in order to simulate operator failure & OCP success. 
Then i wait for operators to complete, and then waiting for cluster installed
https://gitlab.cee.redhat.com/ocp-edge-qe/kni-assisted-installer-auto/-/merge_requests/170/diffs

omertuc: > I need to get into Finalizing stage and not installed.
> In my e2e case, i wait until Finalizing, because then i can mess up the nodes in order to simulate operator failure & OCP success.
> Then i wait for operators to complete, and then waiting for cluster installed
> [gitlab.cee.redhat.com/ocp-edge-qe/kni-assisted-installer-auto/-/merge_requests/170/diffs](https://gitlab.cee.redhat.com/ocp-edge-qe/kni-assisted-installer-auto/-/merge_requests/170/diffs)

But usually when someone "Waits for finalizing", then they should also be okay with "INSTALLED". So I suggest you add it to `statuses=` and then in your test, make sure you haven't reached INSTALLED after the call to this function

PR #684, opened 2021-04-08
NOISSUE centralized kube api group and version vars

RazRegev:
To reduce future work in case of any change, use
global variables for api group and version instead
of duplicated values in every object.


nmagnezi (approved):
/lgtm

PR #686, opened 2021-04-08
NO-ISSUE Fix minikube upgrades on older systems

empovit:
1. Use awk instead of jq because older `minikube version` commands
doen't have the `-o` modifier.
2. Replace the minikube binary at its original location instead of
`/usr/local/bin/`.


YuviGold (approved):
LGTM

YuviGold (approved):

PR #688, opened 2021-04-08
MGMT-4918 Support configurations for test-infra pytest

eliorerz:
Test SNO and None platform given required environment variables using `make test` target

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ Check the [Install Guide](GUIDE.md) for installation instructions.
 | WORKER_MEMORY               | memory for worker VM, default: 8892MB                                                                                                       |
 | PUBLIC_CONTAINER_REGISTRIES | comma-separated list of registries that do not require authentication for pulling assisted installer images                                 |
 | CHECK_CLUSTER_VERSION       | If "True", the controller will wait for CVO to finish                                                                                       |
-| ENABLE_KUBE_API             | If set, deploy assisted-service with Kube API controllers (minikube only)                                                                                     |
+| ENABLE_KUBE_API             | If set, deploy assisted-service with Kube API controllers (minikube only)                                                                   |
 
 ## Instructions
 
--- a/discovery-infra/tests/conftest.py
+++ b/discovery-infra/tests/conftest.py
@@ -3,6 +3,7 @@ import os
 import uuid
 from distutils import util
 from pathlib import Path
+from typing import List
 
 import pytest
 import test_infra.utils as infra_utils
@@ -42,7 +43,7 @@ env_variables = {"ssh_public_key": utils.get_env('SSH_PUB_KEY'),
                  "openshift_version": utils.get_openshift_version(),
                  "base_domain": utils.get_env('BASE_DOMAIN', "redhat.com"),
                  "num_masters": int(utils.get_env('NUM_MASTERS', consts.NUMBER_OF_MASTERS)),
-                 "num_workers": max(2, int(utils.get_env('NUM_WORKERS', 0))),
+                 "num_workers": int(utils.get_env('NUM_WORKERS', 2)),
                  "num_day2_workers": int(utils.get_env('NUM_DAY2_WORKERS', 0)),
                  "vip_dhcp_allocation": bool(util.strtobool(utils.get_env('VIP_DHCP_ALLOCATION'))),
                  "worker_memory": int(utils.get_env('WORKER_MEMORY', '8892')),
@@ -65,7 +66,9 @@ env_variables = {"ssh_public_key": utils.get_env('SSH_PUB_KEY'),
                  "test_teardown": bool(util.strtobool(utils.get_env('TEST_TEARDOWN', 'true'))),
                  "namespace": utils.get_env('NAMESPACE', consts.DEFAULT_NAMESPACE),
                  "olm_operators": list(map(lambda name: {'name': name}, utils.get_env('OLM_OPERATORS', '').lower().split())),
-                 "platform": utils.get_env("PLATFORM", consts.Platforms.BARE_METAL)
+                 "platform": utils.get_env("PLATFORM", consts.Platforms.BARE_METAL),
+                 "user_managed_networking": False,
+                 "high_availability_mode": consts.HighAvailabilityMode.FULL,
                  }
 cluster_mid_name = infra_utils.get_random_name()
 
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -32,7 +32,7 @@ ISO := $(or $(ISO), "") # ISO should point to a file that has the '.iso' extensi
 NUM_MASTERS :=  $(or $(NUM_MASTERS),3)
 WORKER_MEMORY ?= 8892
 MASTER_MEMORY ?= 16984
-NUM_WORKERS := $(or $(NUM_WORKERS),0)
+NUM_WORKERS := $(or $(NUM_WORKERS),2)
```
osherdp, line 35: I wonder if we want to always test with two workers
it will add coverage, but will be harder to do on the local environment

README.md
```diff
@@ -111,7 +111,9 @@ Check the [Install Guide](GUIDE.md) for installation instructions.
...
 | PUBLIC_CONTAINER_REGISTRIES | comma-separated list of registries that do not require authentication for pulling assisted installer images                                 |
 | CHECK_CLUSTER_VERSION       | If "True", the controller will wait for CVO to finish                                                                                       |
-| ENABLE_KUBE_API             | If set, deploy assisted-service with Kube API controllers (minikube only)                                                                                     |
+| ENABLE_KUBE_API             | If set, deploy assisted-service with Kube API controllers (minikube only)                                                                   |
+| USER_MANAGED_NETWORKING     | If true set cluster user_managed_networking to True, default: False                                                                         |
+| HIGH_AVAILABILITY_MODE      | Set cluster high_availability_mode. Available values: Full or None, default: Full                                                           |
```
osherdp, line 116: Isn't it an implementation detail to how we do our tests?
I think that because it's an implementation detail for the UI (and user doesn't set it by himself) then maybe we shouldn't support changing it in our testing infrastructure

discovery-infra/tests/conftest.py
```diff
@@ -84,6 +87,18 @@ def _get_cluster_name():
 env_variables["num_nodes"] = env_variables["num_workers"] + env_variables["num_masters"]
 
 
+if env_variables["num_masters"] == 1:
+    env_variables["high_availability_mode"] = consts.HighAvailabilityMode.NONE
+    env_variables["user_managed_networking"] = True
```
osherdp, line 92: We should have similar functionality somewhere else that can be deleted, no?

discovery-infra/tests/conftest.py
```diff
@@ -100,6 +115,15 @@ def get_api_client(offline_token=env_variables['offline_token'], **kwargs):
     return assisted_service_api.create_client(url, offline_token, **kwargs)
 
 
+def get_available_openshift_versions() -> List[str]:
+    available_versions = list(get_api_client().get_openshift_versions().keys())
+    specific_version = env_variables["openshift_version"]
+    if specific_version and specific_version in available_versions:
```
osherdp, line 121: Maybe if specific version doesn't much what available we should raise
Doesn't seem right to keep going with what we have

osherdp (changes requested):

Makefile
eliorerz: It's the default value moved from conftest.py.
@RazRegev wanted to make sure that there is a minimum of two workers for the e2e test but that doesn't work with SNO.

eliorerz:
@osherdp I don't think that the tests are related to this issue.
No environment variables have been set [here](https://github.com/openshift/release/blob/389bbeff2d0dbd6dd848f7fd4b429fa7f9f9d2ab/ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml#L177-L185) yet. Also 'e2e-metal-assisted-kube-api' is not relevant to this issue.
Tests needed are: e2e-metal-assisted-single-node, e2e-metal-assisted-none

Makefile
tsorya: it is bad idea. We don't want by default to run with 2 workers. It is on purpose set to 0 here

README.md
tsorya: Maybe better to call it SINGLE_NODE? Or why not to use same logic as now NUM_MASTER=1, why we should have another field for it

README.md
```diff
@@ -111,7 +111,9 @@ Check the [Install Guide](GUIDE.md) for installation instructions.
 | WORKER_MEMORY               | memory for worker VM, default: 8892MB                                                                                                       |
 | PUBLIC_CONTAINER_REGISTRIES | comma-separated list of registries that do not require authentication for pulling assisted installer images                                 |
 | CHECK_CLUSTER_VERSION       | If "True", the controller will wait for CVO to finish                                                                                       |
-| ENABLE_KUBE_API             | If set, deploy assisted-service with Kube API controllers (minikube only)                                                                                     |
+| ENABLE_KUBE_API             | If set, deploy assisted-service with Kube API controllers (minikube only)                                                                   |
+| USER_MANAGED_NETWORKING     | If true set cluster user_managed_networking to True, default: False                                                                         |
```
tsorya, line 115: We have Platform env, why to use USER_MANAGED_NETWORKING ?

Makefile
eliorerz: It was minimum 2 in the code.
Maybe @slaviered or @RazRegev will can explain why it set to be minimum 2 (before SNO)

YuviGold: Let's keep it a default of 0 for local deployment and 2 in the [CI](https://github.com/openshift/release/blob/5070635c0545d5bf786d5b35dca20b511a9e6eb1/ci-operator/step-registry/baremetalds/assisted/setup/baremetalds-assisted-setup-commands.sh#L92)

discovery-infra/tests/conftest.py
```diff
@@ -100,6 +115,17 @@ def get_api_client(offline_token=env_variables['offline_token'], **kwargs):
     return assisted_service_api.create_client(url, offline_token, **kwargs)
 
 
+def get_available_openshift_versions() -> List[str]:
+    available_versions = list(get_api_client().get_openshift_versions().keys())
+    specific_version = env_variables["openshift_version"]
+    if specific_version:
```
YuviGold, line 121: I think this one is always true

Makefile
RazRegev: I think there is a misunderstanding, I don't recall wanting 2 workers, don't see any reason to do that

discovery-infra/tests/conftest.py
eliorerz: utils.get_openshift_version(default=None) returns None as default value

YuviGold:
LGTM. Please just organize the commits of this PR.
/approve

PR #689, opened 2021-04-09
MGMT-4640 Using HW_VALIDATOR_REQUIREMENTS

jakub-dzon:
This PR adds `HW_VALIDATOR_REQUIREMENTS` env variable handling in update_assisted_service_cm.py

The purpose of this change is to make sure assisted-service is always configured with expected default host requirements after legacy, separate env variable values (`HW_VALIDATOR_MIN_*` at https://github.com/openshift/assisted-service/blob/0ec56929f077cd512e4ef91e3a242ad20d3a7bd4/internal/hardware/validator.go#L41-L46)  are removed from the assisted-service and `HW_VALIDATOR_REQUIREMENTS` takes the full responsibility for setting the host requirements.

### DIFF
--- a/discovery-infra/update_assisted_service_cm.py
+++ b/discovery-infra/update_assisted_service_cm.py
@@ -11,9 +11,8 @@
 #   MY_VAR: ""
 #
 # Hence, in order to support unset env vars, avoid override in Makefile.
-
+import json
 import os
-
 import yaml
 
 CM_PATH = "assisted-service/deploy/assisted-service-configmap.yaml"
@@ -33,8 +32,26 @@ ENVS = [
     ("IMAGE_BUILDER", ""),
     ("OCM_BASE_URL", ""),
     ("PUBLIC_CONTAINER_REGISTRIES", ""),
-    ("CHECK_CLUSTER_VERSION", "")
+    ("CHECK_CLUSTER_VERSION", ""),
+    ("HW_VALIDATOR_REQUIREMENTS", "")
 ]
+DEFAULT_MASTER_REQUIREMENTS = {
+    "cpu_cores": 4,
+    "ram_mib": 8192,
+    "disk_size_gb": 10,
+    "installation_disk_speed_threshold_ms": 10
+}
+DEFAULT_WORKER_REQUIREMENTS = {
+    "cpu_cores": 2,
+    "ram_mib": 3072,
+    "disk_size_gb": 10,
+    "installation_disk_speed_threshold_ms": 10
+}
+DEFAULT_REQUIREMENTS = [{
+    "version": "default",
+    "master": DEFAULT_MASTER_REQUIREMENTS,
+    "worker": DEFAULT_WORKER_REQUIREMENTS
+}]
 
 
 def _read_yaml():
@@ -56,11 +73,26 @@ def _get_relevant_envs():
     return data
 
 
+def update_requirements(requirements_json):
+    if requirements_json == "" or requirements_json == "REPLACE_HW_VALIDATOR_REQUIREMENTS":
+        return json.dumps(DEFAULT_REQUIREMENTS)
+    requirements = json.loads(requirements_json)
+    for version_requirements in requirements:
+        if version_requirements["version"] == "default":
+            version_requirements["master"] = DEFAULT_MASTER_REQUIREMENTS
+            version_requirements["worker"] = DEFAULT_WORKER_REQUIREMENTS
+
+    return json.dumps(requirements)
+
+
 def set_envs_to_service_cm():
     cm_data = _read_yaml()
     if not cm_data:
         raise Exception("%s must exists before setting envs to it" % CM_PATH)
     cm_data["data"].update(_get_relevant_envs())
+    existing_requirements = cm_data["data"].get("HW_VALIDATOR_REQUIREMENTS", "")
+    requirements = update_requirements(existing_requirements)
+    cm_data["data"].update({"HW_VALIDATOR_REQUIREMENTS": requirements})
     with open(CM_PATH, "w") as cm_file:
         yaml.dump(cm_data, cm_file)
 


discovery-infra/update_assisted_service_cm.py
```diff
@@ -56,11 +75,29 @@ def _get_relevant_envs():
...
 
+def update_requirements(requirements_json):
+    try:
+        requirements = json.loads(requirements_json)
+        for version_requirements in requirements:
+            if version_requirements["version"] == "default":
+                version_requirements["master"] = MASTER_REQUIREMENTS
+                version_requirements["worker"] = WORKER_REQUIREMENTS
+    except json.JSONDecodeError:
+        requirements = REQUIREMENTS
```
omertuc, line 86: I would prefer an explicit failure (`raise`) if the JSON doesn't decode, rather than secretly switching it up.

jakub-dzon: Do you know if the script is ran always against already pre-processed `assisted-service-configmap.yaml` file? If not, then the variable value that is saved in the repository won't be parsed (`HW_VALIDATOR_REQUIREMENTS: REPLACE_HW_VALIDATOR_REQUIREMENTS`).

Also - changes on the assisted-service side that process that the configmap definition file (https://github.com/openshift/assisted-service/pull/1458) are not merged yet so using defaults instead of failing provides backward compatibility.

omertuc: Not sure about the answer to your question, if this is too much trouble, then maybe a warning + a `# TODO: Consider raising here in the future to avoid implicitly ignoring user errors` would be enough for now.

If you know of specific, non-json values that this var might have then maybe add a special case for them (`if .. == known_value: use_defaults`)

ronniel1:
@jakub-dzon can you provide some more info in the description?

YuviGold (approved):
LGTM

discovery-infra/update_assisted_service_cm.py
```diff
@@ -56,11 +75,31 @@ def _get_relevant_envs():
...
+    if requirements_json == "" or requirements_json == "REPLACE_HW_VALIDATOR_REQUIREMENTS":
+        return json.dumps(REQUIREMENTS)
+    try:
+        requirements = json.loads(requirements_json)
+        for version_requirements in requirements:
+            if version_requirements["version"] == "default":
+                version_requirements["master"] = MASTER_REQUIREMENTS
+                version_requirements["worker"] = WORKER_REQUIREMENTS
+    except json.JSONDecodeError:
+        raise
```
osherdp, line 88: It's a no-op
Meaning you can remove the complete wrapping of try-except and it will do the same

discovery-infra/update_assisted_service_cm.py
```diff
@@ -33,8 +34,26 @@
     ("IMAGE_BUILDER", ""),
     ("OCM_BASE_URL", ""),
     ("PUBLIC_CONTAINER_REGISTRIES", ""),
-    ("CHECK_CLUSTER_VERSION", "")
+    ("CHECK_CLUSTER_VERSION", ""),
+    ("HW_VALIDATOR_REQUIREMENTS", "")
 ]
+MASTER_REQUIREMENTS = {
```
osherdp, line 40: Rename all of those to ``DEFAULT_<name>``

discovery-infra/update_assisted_service_cm.py
```diff
@@ -56,11 +75,31 @@ def _get_relevant_envs():
...
+
+
 def set_envs_to_service_cm():
     cm_data = _read_yaml()
     if not cm_data:
         raise Exception("%s must exists before setting envs to it" % CM_PATH)
     cm_data["data"].update(_get_relevant_envs())
+    existing_requirements = ""
+    if VALIDATOR_REQUIREMENTS in cm_data["data"]:
+        existing_requirements = cm_data["data"][VALIDATOR_REQUIREMENTS]
```
osherdp, line 100:
```suggestion
    existing_requirements =cm_data["data"].get(VALIDATOR_REQUIREMENTS, "")
```

discovery-infra/update_assisted_service_cm.py
```diff
@@ -11,11 +11,12 @@
 #   MY_VAR: ""
 #
 # Hence, in order to support unset env vars, avoid override in Makefile.
-
+import json
 import os
-
 import yaml
 
+VALIDATOR_REQUIREMENTS = "HW_VALIDATOR_REQUIREMENTS"
```
osherdp, line 18: I think you can just inline this variable
This file is duplicating dictionary keys either way all over the place, and it's not that bad

discovery-infra/update_assisted_service_cm.py
```diff
@@ -33,8 +34,26 @@
...
 ]
+MASTER_REQUIREMENTS = {
+    "cpu_cores": 4,
+    "ram_mib": 8192,
+    "disk_size_gb": 10,
+    "installation_disk_speed_threshold_ms": 10
+}
+WORKER_REQUIREMENTS = {
+    "cpu_cores": 2,
+    "ram_mib": 3072,
```
osherdp, line 48: From the service code it looks like we require 4GB of RAM, no?

osherdp (changes requested):

discovery-infra/update_assisted_service_cm.py
jakub-dzon: According to this line https://github.com/openshift/assisted-test-infra/blob/1ec704a3516cb67b8abfcae92ee498deda409e11/discovery-infra/update_assisted_service_cm.py#L25 it should be 3 GiB here.

jakub-dzon:
@ronniel1 - description updated.

osherdp (approved):

omertuc:
> @ronniel1 - description updated.

Please also include the description in the commit message

jakub-dzon:
@omertuc - done.

PR #690, opened 2021-04-11
Adding 1.5h timeout for OCS installation

lalon4:
Total timeout when ocs operator fails is 1.5 hours. Also changed wait_till_all_operators_are_in_status to use fall_on_error_status parameter.


discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -538,7 +538,7 @@ def wait_for_operators_to_finish(
             operators_count=len(self.get_operators()),
             statuses=[consts.OperatorStatus.AVAILABLE, consts.OperatorStatus.FAILED],
             timeout=timeout,
-            fall_on_error_status=False,
+            fall_on_error_status=fall_on_error_status,
```
omertuc, line 541: This changes the existing behavior for tests that don't pass `fall_on_error_status` - up until now, it was `False` for them. Now it will automatically be `True` for them because that's the default value of this parameter.

This is no necessarily a bad thing, but something worth noting - please make sure no tests are negatively affected by this change.

lalon4: Thanks Omer. 
Double checked - no test expect to be impacted by that.

PR #696, opened 2021-04-12
NOISSUE: delete_nodes.py: delete kube api resources if KUBE_API=true

RazRegev:


discovery-infra/test_infra/kubeapi_utils.py
```diff
@@ -91,3 +96,24 @@ def get_hostname_for_agent(agent, nodes_details):
             if interface['macAddress'].lower() == mac_address:
                 return node_metadata['name']
 
+
+def delete_kube_api_resources_for_namespace(kube_api_client, name, namespace):
+    # any resource that created within a KubeAPIContext will be deleted on
+    # context exit.
```
osherdp, line 102: replace with docstring

discovery-infra/test_infra/kubeapi_utils.py
```diff
@@ -91,3 +96,24 @@ def get_hostname_for_agent(agent, nodes_details):
...
+        Secret(
+            kube_api_client=kube_api_client,
+            name=name,
+            namespace=namespace,
+        )
+        InstallEnv(
+            kube_api_client=kube_api_client,
+            name=f'{name}-install-env',
+            namespace=namespace,
+        )
```
osherdp, line 119: Maybe it's best to make it look like deletion explicitly

discovery-infra/test_infra/kubeapi_utils.py
```diff
@@ -91,3 +96,24 @@ def get_hostname_for_agent(agent, nodes_details):
             if interface['macAddress'].lower() == mac_address:
                 return node_metadata['name']
 
+
+def delete_kube_api_resources_for_namespace(kube_api_client, name, namespace):
```
osherdp, line 100: we need to delete nmstate config as well, no?

discovery-infra/test_infra/kubeapi_utils.py
```diff
@@ -91,3 +96,24 @@ def get_hostname_for_agent(agent, nodes_details):
...
+    # any resource that created within a KubeAPIContext will be deleted on
+    # context exit.
+    with KubeAPIContext(kube_api_client):
+        cluster_deployment = ClusterDeployment(
+            kube_api_client=kube_api_client,
+            name=name,
+            namespace=namespace,
+        )
+        cluster_deployment.list_agents()
```
osherdp, line 109: this will delete agents as well?

osherdp (changes requested):

discovery-infra/test_infra/kubeapi_utils.py
RazRegev: why? this is not a description of the function just a side comment

osherdp: ok, so https://github.com/openshift/assisted-test-infra/pull/696#discussion_r611374577 is the important part

discovery-infra/test_infra/kubeapi_utils.py
```diff
@@ -91,3 +96,24 @@ def get_hostname_for_agent(agent, nodes_details):
...
+        Secret(
+            kube_api_client=kube_api_client,
+            name=name,
```
nshidlin, line 112: Is it safe to assume the sercet and clusterdeployment have the same name?

discovery-infra/test_infra/kubeapi_utils.py
RazRegev: yes

discovery-infra/test_infra/kubeapi_utils.py
RazRegev: nope, not safe to assume that, thanks

discovery-infra/test_infra/kubeapi_utils.py
```diff
@@ -91,3 +98,45 @@ def get_hostname_for_agent(agent, nodes_details):
             if interface['macAddress'].lower() == mac_address:
                 return node_metadata['name']
 
+
+def delete_kube_api_resources_for_namespace(
+        kube_api_client,
+        name,
+        namespace,
+        **kwargs
```
osherdp, line 106: if not using it explicitly, then just replace with ``secret_name=None, installenv_name=None,...``

osherdp (changes requested):

osherdp (approved):

discovery-infra/delete_nodes.py
```diff
@@ -115,12 +121,35 @@ def delete_cluster(cluster_name, namespace):
     delete_nodes(cluster_name, namespace, tf_folder, tfvars)
 
 
+@utils.on_exception(message='Failed to delete kube api resources', silent=True)
+def delete_kube_api_resources_from_namespaces(namespace):
```
tsorya, line 125: maybe this function should be part of kube-api utils too?

discovery-infra/test_infra/kubeapi_utils.py
```diff
@@ -91,3 +98,48 @@ def get_hostname_for_agent(agent, nodes_details):
...
+        name=name,
+        namespace=namespace,
+    )
+
+    for agent in cluster_deployment.list_agents():
+        agent.delete()
+
+    cluster_deployment.delete()
+
+    Secret(
```
tsorya, line 129: don't we want to delete maximum we can? i think we should try except every delete operation or there is no way that it will fail?

RazRegev: this is basically what I've done here, the only acceptable exception is NotFound, otherwise, this is probably an error we want to investigate no?

```
    CoreV1Api.delete_namespaced_secret = suppress_not_found_error(
        fn=CoreV1Api.delete_namespaced_secret,
    )
    CustomObjectsApi.delete_namespaced_custom_object = suppress_not_found_error(
        fn=CustomObjectsApi.delete_namespaced_custom_object
    )
```

PR #697, opened 2021-04-12
NOISSUE: install_cluster.py: install predefined cluster using kube-api flow

RazRegev:


discovery-infra/install_cluster.py
```diff
@@ -227,29 +234,58 @@ def download_logs_from_all_hosts(client, cluster_id, output_folder):
...
         tf_folder = utils.get_tf_folder(cluster_name, args.namespace)
         args.cluster_id = utils.get_tfvars(tf_folder).get('cluster_inventory_id')
         tf = terraform_utils.TerraformUtils(working_dir=tf_folder)
 
-    client = assisted_service_api.create_client(
-        url=utils.get_assisted_service_url_by_args(
-            args=args,
-            wait=False
+    client, cluster_deployment, nodes_number = None, None, None
+    if args.kube_api:
```
tsorya, line 262: can we move all kube_api flow to the new file? I don't see why we don't have kube_api.py with all the functions relevant to it?

discovery-infra/install_cluster.py
```diff
@@ -227,29 +234,58 @@ def download_logs_from_all_hosts(client, cluster_id, output_folder):
         )
 
 
+def get_nodes_number_from_cluster_deployment(cluster_deployment):
```
nshidlin, line 237: might be worth to have this as a method in ClusterDeyployment

discovery-infra/install_cluster.py
RazRegev: moved installation function to kubeapi_utils.py

eliorerz:
@osherdp Basically, most of the work here will deprecated soon (install_cluster.py and start_discovery.py).
I think it would be better to invest the effort in the [ticket](https://issues.redhat.com/browse/MGMT-5294) for inserting KubeAPI logic into the new workflow.

YuviGold:
Obsolete

PR #706, opened 2021-04-14
MGMT-5189 - Add deprecation warning on pending deprecation scripts.

eliorerz:
Warn users while using discovery-infra scripts for the upcoming deprecation


YuviGold (approved):
LGTM

PR #707, opened 2021-04-15
MGMT-4792 Dynamically change vm resources based on enabled operators

pkliczewski:
Based on `OLM_OPERATORS` environment variable we need to add enough resources to nodes vms so validations for those operators would pass. Before running terraform we add resources as specified by `consts.OPERATOR_PARAMS` for corresponding operators.

### DIFF
--- a/discovery-infra/start_discovery.py
+++ b/discovery-infra/start_discovery.py
@@ -345,16 +345,20 @@ def _cluster_create_params():
         "vip_dhcp_allocation": bool(args.vip_dhcp_allocation) and not user_managed_networking,
         "additional_ntp_source": ntp_source,
         "user_managed_networking": user_managed_networking,
-        "high_availability_mode": "None" if args.master_count == 1 else "Full"
+        "high_availability_mode": consts.HighAvailabilityMode.NONE if args.master_count == 1 else consts.HighAvailabilityMode.FULL,
+        "olm_operators": [{'name': name} for name in utils.parse_olm_operators_from_env()]
     }
     return params
 
 
 # convert params from args to terraform tfvars
 def _create_node_details(cluster_name):
+    operators = utils.parse_olm_operators_from_env()
     return {
-        "libvirt_worker_memory": args.worker_memory,
-        "libvirt_master_memory": args.master_memory if not args.master_count == 1 else args.master_memory * 2,
+        "libvirt_worker_memory": utils.resource_param(args.worker_memory, "worker_memory", operators),
+        "libvirt_master_memory": utils.resource_param(args.master_memory if not args.master_count == 1 else args.master_memory * 2, "master_memory", operators),
+        "libvirt_worker_vcpu": utils.resource_param(args.worker_cpu, "worker_vcpu", operators),
+        "libvirt_master_vcpu": utils.resource_param(args.master_cpu, "master_vcpu", operators),
         "worker_count": args.number_of_workers,
         "cluster_name": cluster_name,
         "cluster_domain": args.base_dns_domain,
@@ -818,6 +822,20 @@ if __name__ == "__main__":
         help="Worker memory (ram) in mb",
         type=int,
         default=8192,
+    )
+    parser.add_argument(
+        "-mc",
+        "--master-cpu",
+        help="Master cpu count",
+        type=int,
+        default=consts.MASTER_CPU,
+    )
+    parser.add_argument(
+        "-wc",
+        "--worker-cpu",
+        help="Worker cpu count",
+        type=int,
+        default=consts.WORKER_CPU,
     )
     parser.add_argument(
         "-nw", "--number-of-workers", help="Workers count to spawn", type=int, default=0
--- a/discovery-infra/test_infra/consts.py
+++ b/discovery-infra/test_infra/consts.py
@@ -41,6 +41,23 @@ TEST_TARGET_INTERFACE = "vnet3"
 DEFAULT_NAMESPACE = 'assisted-installer'
 
 
+# operator resource requirements as coded in assisted service
+OPERATOR_PARAMS = {
+    "cnv": {
+        "worker_memory": 360,
+        "master_memory": 150,
+        "worker_vcpu": 2,
+        "master_vcpu": 4,
+    },
+    "ocs": {
+        "worker_memory": 14000,
+        "master_memory": 24000,
+        "worker_vcpu": 6,
+        "master_vcpu": 8,
+    }
+}
+
+
 class ImageType:
     FULL_ISO = "full-iso"
     MINIMAL_ISO = "minimal-iso"
--- a/discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
+++ b/discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
@@ -43,10 +43,11 @@ class TerraformController(LibvirtController):
 
     # TODO move all those to conftest and pass it as kwargs
     def _terraform_params(self, **kwargs):
-        params = {"libvirt_worker_memory": kwargs.get('worker_memory'),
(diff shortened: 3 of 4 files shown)


discovery-infra/start_discovery.py
```diff
@@ -345,16 +345,19 @@ def _cluster_create_params():
         "vip_dhcp_allocation": bool(args.vip_dhcp_allocation) and not user_managed_networking,
         "additional_ntp_source": ntp_source,
         "user_managed_networking": user_managed_networking,
-        "high_availability_mode": "None" if args.master_count == 1 else "Full"
+        "high_availability_mode": "None" if args.master_count == 1 else "Full",
+        "olm_operators": list(map(lambda name: {'name': name}, utils.get_env('OLM_OPERATORS', '').lower().split()))
```
omertuc, line 349: nit: `list(map(lambda name: ` can be rephrased with list comprehension more concisely:

`"olm_operators": list(map(lambda name: {'name': name}, utils.get_env('OLM_OPERATORS', '').lower().split()))`

to

`"olm_operators": [{'name': name} for name in utils.get_env('OLM_OPERATORS', '').lower().split()]`

discovery-infra/start_discovery.py
```diff
@@ -345,16 +345,19 @@ def _cluster_create_params():
         "vip_dhcp_allocation": bool(args.vip_dhcp_allocation) and not user_managed_networking,
         "additional_ntp_source": ntp_source,
         "user_managed_networking": user_managed_networking,
-        "high_availability_mode": "None" if args.master_count == 1 else "Full"
+        "high_availability_mode": "None" if args.master_count == 1 else "Full",
```
omertuc, line 348: Use https://github.com/openshift/assisted-test-infra/blob/ea04ad57551b67bdf1e057bfe808b8c11514fefd/discovery-infra/test_infra/consts.py#L151-L153

omertuc:
`utils.get_env("OLM_OPERATORS", "").lower().split()` is repeated twice in this PR, maybe we can put it inside a function (`def parse_olm_operators_from_env` or something similiar)

discovery-infra/start_discovery.py
```diff
@@ -369,6 +372,17 @@ def _create_node_details(cluster_name):
     }
 
 
+def _resource_param(value, param_name):
+    for operator in utils.get_env("OLM_OPERATORS", "").lower().split():
```
omertuc, line 376: `utils.get_env("OLM_OPERATORS", "").lower().split()` is repeated twice in this PR, maybe we can put it inside a function (`def parse_olm_operators_from_env` or something similiar) so this parsing logic is not repeated

discovery-infra/start_discovery.py
omertuc: Also see https://github.com/openshift/assisted-test-infra/pull/707/files#r614728967

discovery-infra/test_infra/consts.py
```diff
@@ -41,6 +41,21 @@
 DEFAULT_NAMESPACE = 'assisted-installer'
 
 
+CNV_PARAMS = {
```
omertuc, line 44: Instead of each operator having its own constant, this can be changed to a nested dictionary:

```python
OPERATOR_PARAMS = {
	  "cnv: {
	      "worker_memory": 360,
	      "master_memory": 150,
	      "worker_vcpu": 2,
	      "master_vcpu": 4,
	  },
	  "ocs: {
	      "worker_memory": 14000,
	      "master_memory": 24000,
	      "worker_vcpu": 6,
	      "master_vcpu": 8,
	  }
}
```

Then the `_resource_param` methods can be simplified from a `for` + switch-like `if-else` chain to just:

```python
def _resource_param(value, param_name):
	try:
		return sum(OPERATOR_PARAMS[operator][param_name] 
				   for operator in utils.get_env("OLM_OPERATORS", "").lower().split())
	except KeyError as e:
		raise ValueError(f"Unknown operator name {e.args[0]}. Available operators are {', '.join(operator_params)}")
```

omertuc: Also please document where all those 360, 150, 2, 4 14000, 24000, 6, 8 constants are coming from - is there some official documentation they were extracted from? Trial and error?

discovery-infra/start_discovery.py
pkliczewski: I only added `,` at the end. I do not want to change it

discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
```diff
@@ -43,10 +43,10 @@ def _get_random_name(cls):
 
     # TODO move all those to conftest and pass it as kwargs
     def _terraform_params(self, **kwargs):
-        params = {"libvirt_worker_memory": kwargs.get('worker_memory'),
-                  "libvirt_master_memory": kwargs.get('master_memory', 16984),
-                  "libvirt_worker_vcpu": kwargs.get("worker_vcpu", 4),
-                  "libvirt_master_vcpu": kwargs.get("master_vcpu", 4),
+        params = {"libvirt_worker_memory": self._resource_param(self, 'worker_memory', kwargs),
```
omertuc, line 46: Please reuse the non-kwargs version `self._resource_param` from the `start_discovery`. The summing logic should only be implemented once. Any differences specialized for simpler terraform `kwargs` access can be done in a thin wrapper.

You might want to make the `start_discovery` version of `self._resource_param` receive the tables as a parameter, making it more flexible for use by the terraform wrapper.

discovery-infra/start_discovery.py
omertuc: > I only added `,` at the end. I do not want to change it

Right, missed that.

pkliczewski:
@YuviGold please take a look

omertuc:
This currently will not work, you're using old variable names:
image


please address [my comment](https://github.com/openshift/assisted-test-infra/pull/707#discussion_r614743731)

pkliczewski:
> This currently will not work, you're still using old variable names:
> image
> 
> please address [my comment](https://github.com/openshift/assisted-test-infra/pull/707#discussion_r614743731)

You are right, I only tested one flow. Will fix.

discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
```diff
@@ -75,6 +75,14 @@ def _terraform_params(self, **kwargs):
                 params[key] = value
         return Munch.fromDict(params)
 
+    def _resource_param(self, param_name, default=None, **kwargs):
```
omertuc, line 78: Please reuse the other implementation of `_resource_param`.

See [my previous comment](https://github.com/openshift/assisted-test-infra/pull/707#discussion_r614743731)

pkliczewski: We can't reuse since we obtain operators differently and values are in different format.

omertuc: The function doesn't have to obtain the operators on its own, it can receive them as a parameter (I showed how in my previous comment). Then each caller can provide those parameters differently. 

Whatever conversions need to be done for the formats to match should be done in a wrapper function - the summing logic should not be repeated

pkliczewski: I disagree with you. We are trading small reuse for clarity but you are the maintainer so I updated.

omertuc: It's fair to disagree, didn't mean to force you to make the change - just wanted to make sure the change I was looking for is understood and considered.

And I agree that reuse has a clarity cost, I personally tend to be on the reuse side. Thank you for making the change regardless.

PR #712, opened 2021-04-19
NO-ISSUE fix cluster deployment when there is no state yet

tsorya:


osherdp (changes requested):
Maybe ``else None``?

tsorya:
> Maybe `else None`?

wanted something very obvious , doesn't really matter. Prefer None?

osherdp:
@tsorya yes I think so

osherdp (approved):

PR #717, opened 2021-04-20
MGMT-4482 Use local auth with kube-api by default

carbonin:
Couldn't test this locally as the service wouldn't deploy with these errors:
```
E0420 13:59:27.974475       1 leaderelection.go:325] error retrieving resource lock assisted-installer/77190dcb.agent-install.openshift.io: Get "https://10.96.0.1:443/api/v1/namespaces/assisted-installer/configmaps/77190dcb.agent-install.openshift.io": context deadline exceeded
I0420 13:59:27.974768       1 leaderelection.go:278] failed to renew lease assisted-installer/77190dcb.agent-install.openshift.io: timed out waiting for the condition
time="2021-04-20T13:59:27Z" level=fatal msg="failed to run manager" func=main.main.func1 file="/go/src/github.com/openshift/origin/cmd/main.go:167" error="leader election lost"
```

Hopefully CI has better luck :sweat_smile:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,12 +109,14 @@ CONTROLLER_OCP := $(or ${CONTROLLER_OCP},quay.io/ocpmetal/assisted-installer-con
 PLATFORM := $(or ${PLATFORM},baremetal)
 IPV6_SUPPORT := $(or ${IPV6_SUPPORT},true)
 SERVICE_REPLICAS_COUNT := 3
+AUTH_TYPE := $(or ${AUTH_TYPE},none)
 ENABLE_KUBE_API := $(or ${ENABLE_KUBE_API},no)
 KUBE_API := $(or ${KUBE_API},no)
 ifeq ($(ENABLE_KUBE_API),true)
 	SERVICE_REPLICAS_COUNT=1
 	KUBE_API=yes
 	VIP_DHCP_ALLOCATION=no
+	AUTH_TYPE=local
 endif
 
 .EXPORT_ALL_VARIABLES:
@@ -360,7 +362,7 @@ redeploy_nodes_with_install: destroy_nodes deploy_nodes_with_install
 
 deploy_assisted_service: start_minikube bring_assisted_service
 	mkdir -p assisted-service/build
-	DEPLOY_TAG=$(DEPLOY_TAG) NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE) $(OC_FLAG)) DEPLOY_MANIFEST_PATH=$(DEPLOY_MANIFEST_PATH) DEPLOY_MANIFEST_TAG=$(DEPLOY_MANIFEST_TAG) scripts/deploy_assisted_service.sh
+	DEPLOY_TAG=$(DEPLOY_TAG) NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE) $(OC_FLAG)) DEPLOY_MANIFEST_PATH=$(DEPLOY_MANIFEST_PATH) DEPLOY_MANIFEST_TAG=$(DEPLOY_MANIFEST_TAG) AUTH_TYPE=$(AUTH_TYPE) scripts/deploy_assisted_service.sh
 
 bring_assisted_service:
 	@if ! cd assisted-service >/dev/null 2>&1; then \


scripts/deploy_assisted_service.sh
```diff
@@ -16,6 +16,13 @@ export OPENSHIFT_INSTALL_RELEASE_IMAGE=${OPENSHIFT_INSTALL_RELEASE_IMAGE:-}
 export ENABLE_KUBE_API=${ENABLE_KUBE_API:-false}
 export ENABLE_KUBE_API_CMD="ENABLE_KUBE_API=${ENABLE_KUBE_API}"
 
+# If auth type isn't specified and we're using kube-api, use local
+if [ "$ENABLE_KUBE_API" == "true" -a "$AUTH_TYPE" == "" ]; then
```
osherdp, line 20: we want to always use local auth on kube-api?
it that's true, I think we have a place where we set all kind of things related to kube-api settings. So put it there

osherdp (changes requested):

scripts/deploy_assisted_service.sh
carbonin: I did it this way so that it was still configurable if someone wanted to set it back to none. But I might be able to do the same elsewhere.

carbonin: > I think we have a place where we set all kind of things related to kube-api settings

Maybe this? https://github.com/openshift/assisted-test-infra/blob/c9beeddc7597138d54c89c8e5608305ec7cd81e8/Makefile#L113-L117

osherdp: yes

discovery-infra/test_infra/assisted_service_api.py
```diff
@@ -419,11 +409,8 @@ def create_client(
         url,
         offline_token=utils.get_env('OFFLINE_TOKEN'),
         pull_secret="",
-        wait_for_api=True,
```
RazRegev, line 422: why not just set it to `False` by default, (except the discovery entry point)?

discovery-infra/test_infra/assisted_service_api.py
```diff
@@ -419,11 +409,8 @@ def create_client(
...
         timeout=consts.WAIT_FOR_BM_API
 ):
     log.info('Creating assisted-service client for url: %s', url)
     c = InventoryClient(url, offline_token, pull_secret)
-    if wait_for_api:
-        c.wait_for_api_readiness(timeout)
     return c
```
osherdp, line 416: if so, just return ``InventoryClient`` object directly

discovery-infra/test_infra/assisted_service_api.py
osherdp: what Raz said. On REST flow we want to fail fast if there's a problem

discovery-infra/test_infra/assisted_service_api.py
```diff
@@ -419,11 +409,8 @@ def create_client(
         url,
         offline_token=utils.get_env('OFFLINE_TOKEN'),
         pull_secret="",
-        wait_for_api=True,
         timeout=consts.WAIT_FOR_BM_API
 ):
     log.info('Creating assisted-service client for url: %s', url)
     c = InventoryClient(url, offline_token, pull_secret)
```
osherdp, line 415: And on kubeapi flow, why getting into client creation in the first place?

osherdp (changes requested):

discovery-infra/test_infra/assisted_service_api.py
RazRegev: it won't, that is something Igal changed recently, seems like it was merged today

discovery-infra/test_infra/assisted_service_api.py
RazRegev: and we might want to keep this logic also for the e2e tests

carbonin: My point here (in the commit message) was that we wait for the service to be ready in the deploy make targets in assisted service, so this isn't really necessary here in any flow.

ref: https://github.com/openshift/assisted-service/blob/89e2797e57d0b77d4fafb151d40cf8fdb6aedd06/Makefile#L227-L231

Additionally this uses the ready endpoint rather than querying for clusters so it wouldn't run into the same auth problem I had here.

In any case it's no longer relevant to this PR so I took the change out.

carbonin:
Rebased and tests look good now. @osherdp @RazRegev can you take another look?

osherdp (approved):

PR #719, opened 2021-04-20
MGMT-4577 kube_helpers: Added ClusterImageSet class

danielerez:
* Added ClusterImageSet and ClusterImageSetReference
* Created a ClusterImageSet as part of execute_day1_flow.
* Defined OPENSHIFT_VERSIONS env var in deploy_assisted_service.

### DIFF
--- a/discovery-infra/test_infra/helper_classes/kube_helpers/cluster_image_set.py
+++ b/discovery-infra/test_infra/helper_classes/kube_helpers/cluster_image_set.py
@@ -1,0 +1,104 @@
+import logging
+
+from pprint import pformat
+
+import yaml
+from kubernetes.client import ApiClient, CustomObjectsApi
+from tests.conftest import env_variables
+
+from .base_resource import BaseResource
+
+logger = logging.getLogger(__name__)
+
+
+class ClusterImageSetReference:
+    """
+    A CRD that represents a reference for a ClusterImageSet resource.
+    The reference contains only the name of the ClusterImageSet,
+    so the assisted-service could fetch the resource accordingly.
+    """
+    def __init__(
+            self,
+            name: str,
+    ):
+        self.name = name
+
+    def __repr__(self) -> str:
+        return str(self.as_dict())
+
+    def as_dict(self) -> dict:
+        return {'name': self.name}
+
+
+class ClusterImageSet(BaseResource):
+    """
+    A CRD that represents a ClusterImageSet resource that contains the release image URI.
+    Upon creating a cluster deployment, the release image is fetched by the assisted-service
+    from the image set.
+    """
+
+    _api_group = 'hive.openshift.io'
+    _api_version = 'v1'
+    _plural = 'clusterimagesets'
+
+    def __init__(
+            self,
+            kube_api_client: ApiClient,
+            name: str,
+            namespace: str = env_variables['namespace'],
+    ):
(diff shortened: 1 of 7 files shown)


danielerez:
verified with versions 4.7 and 4.8.
@tsorya @osherdp @YuviGold please take a look.

scripts/deploy_assisted_service.sh
```diff
@@ -15,6 +15,19 @@ export OCP_SERVICE_PORT=$(( 7000 + $NAMESPACE_INDEX ))
 export OPENSHIFT_INSTALL_RELEASE_IMAGE=${OPENSHIFT_INSTALL_RELEASE_IMAGE:-}
 export ENABLE_KUBE_API=${ENABLE_KUBE_API:-false}
 export ENABLE_KUBE_API_CMD="ENABLE_KUBE_API=${ENABLE_KUBE_API}"
+export OPENSHIFT_VERSIONS=${OPENSHIFT_VERSIONS:-}
+export OPENSHIFT_VERSIONS_CMD=""
+
+if [[ "${ENABLE_KUBE_API}" == "true" && -z "${OPENSHIFT_VERSIONS}" ]]; then
```
osherdp, line 21: Maybe we should put it inside https://github.com/openshift/assisted-test-infra/blob/f47f6d9e3706cb786ce171a9300c6e39d231ca85/Makefile#L111
That way all things related to KUBE_API are in one place

scripts/deploy_assisted_service.sh
```diff
@@ -15,6 +15,19 @@ export OCP_SERVICE_PORT=$(( 7000 + $NAMESPACE_INDEX ))
...
+    # Supporting version 4.8 for kube-api
+    supported_version=$(cat assisted-service/default_ocp_versions.json |
+        jq -rc 'with_entries(.key = "4.8")')
+        # ToDo: include only 'rhcos_image' and 'rhcos_version' when custom
```
osherdp, line 25: ToDo -> TODO

discovery-infra/test_infra/helper_classes/kube_helpers/cluster_image_set.py
```diff
@@ -0,0 +1,107 @@
...
+            self,
+            name: str,
+    ):
+        self.name = name
+
+    def __repr__(self) -> str:
+        return str(self.as_dict())
+
+    def as_dict(self) -> dict:
+        data = {
```
osherdp, line 30: inline variable

osherdp (changes requested):

scripts/deploy_assisted_service.sh
danielerez: done

discovery-infra/test_infra/helper_classes/kube_helpers/cluster_image_set.py
danielerez: done

scripts/deploy_assisted_service.sh
danielerez: You mean moving the json parse logic to the Makefile? Not sure it'll be cleaner as it would clutter the Makefile:/
Also, this logic is relevant only for the deploy service flow, so I guess it's a proper place for having context.

osherdp (approved):

PR #721, opened 2021-04-20
MGMT-5289: Remove redundant REST calls in Kube-API flow

RazRegev:
There are a few places that are still using REST API calls in the Kube-API flow.
They will cause failure in case the REST is not available.
In this PR I replace these calls with Kube-API calls instead.


RazRegev:
@tsorya refactor separated kube-api flow from the rest flow, this PR is irrelevant

PR #723, opened 2021-04-21
MGMT-4986 remove all environment variables from the interior test infra package

eliorerz:
Remove all environment variables from cluster.py and nodes.py.
Create alternative for passing environment variables to the package
https://issues.redhat.com/browse/MGMT-4986

====
This PR remove all environment variables from cluster.py and nodes.py while changing some of the classes signatures.
You can see in the first commit how it was looked without config objects 

Note: This is just one step of changing test-infra environment variables mechanism 

Changes:

-  Create temporary wrap for `env_variables` named `EnvConfig` that allow only get from `env_variables`
-  Remove all environment variables from cluster.py and nodes.py (controllers still uses `env_variables`)
-  Join cluster configuration to single object - Interface named `BaseClusterConfig` - implementation named `ClusterConfig`
-  Create placeholder for nodes configuration
- Configurations created as private modules and can access to configs objects only via its package


eliorerz:
FYI @lalon4

discovery-infra/download_logs.py
```diff
@@ -167,7 +167,7 @@ def download_logs(client: InventoryClient, cluster: dict, dest: str, must_gather
             if must_gather:
                 recreate_folder(os.path.join(output_folder, "must-gather"))
                 config_etc_hosts(cluster['name'], cluster['base_dns_domain'],
-                                 helper_cluster.get_api_vip_from_cluster(client, cluster))
+                                 helper_cluster.get_api_vip_from_cluster(client, cluster, ""))
```
osherdp, line 170: It's not understandable what's this ``""``
Can you use keyword arguments? and extract api vip into a variable

discovery-infra/test_infra/config/__init__.py
```diff
@@ -0,0 +1,9 @@
+from ._config import EnvConfig
+from ._cluster_config import DefaultClusterConfig, DefaultInstallConfig
```
osherdp, line 2: although possible, it's not very convenient setting private modules
so I would suggest rename ``_<module>`` to just ``<module>``

discovery-infra/test_infra/config/_cluster_config.py
```diff
@@ -0,0 +1,55 @@
+from dataclasses import dataclass
```
osherdp, line 1: to the best of my knowledge, we use Python 3.6 that doesn't have it
it has been changed?

discovery-infra/test_infra/config/_cluster_config.py
```diff
@@ -0,0 +1,55 @@
...
+from test_infra.config import EnvConfig
+from test_infra.helper_classes.cluster_config import ClusterConfig, ClusterInstallConfig
+
+
+@dataclass
+class DefaultInstallConfig(ClusterInstallConfig):
+
+    def __post_init__(self):
+        self.iso_download_path = self.iso_download_path \
+            if self.iso_download_path is not None else EnvConfig.get("iso_download_path")
```
osherdp, line 12: maybe ``self.iso_download_path = self.iso_download_path or EnvConfig.get("iso_download_path")``?

discovery-infra/test_infra/config/_cluster_config.py
```diff
@@ -0,0 +1,55 @@
...
+            if self.cluster_network_host_prefix is not None else EnvConfig.get("host_prefix")
+
+
+@dataclass
+class DefaultClusterConfig(ClusterConfig):
+
+    def __post_init__(self):
+        """
+        If not set get from environment variables
+        :return:
```
osherdp, line 37: delete

discovery-infra/test_infra/config/_cluster_config.py
```diff
@@ -0,0 +1,55 @@
...
+        self.cluster_network_host_prefix = self.cluster_network_host_prefix \
+            if self.cluster_network_host_prefix is not None else EnvConfig.get("host_prefix")
+
+
+@dataclass
+class DefaultClusterConfig(ClusterConfig):
+
+    def __post_init__(self):
+        """
+        If not set get from environment variables
```
osherdp, line 36: weird sentence
maybe ``Apply default configuration values from environment variables``
and do it a oneline docstring

discovery-infra/test_infra/config/_cluster_config.py
```diff
@@ -0,0 +1,55 @@
...
+            if self.openshift_version is not None else EnvConfig.get("openshift_version")
+        self.cluster_id = self.cluster_id if self.cluster_id is not None else EnvConfig.get("cluster_id")
+        self.user_managed_networking = self.user_managed_networking \
+            if self.user_managed_networking is not None else EnvConfig.get("user_managed_networking")
+        self.high_availability_mode = self.high_availability_mode \
+            if self.high_availability_mode is not None else EnvConfig.get("high_availability_mode")
+        self.olm_operators = self.olm_operators \
+            if self.olm_operators is not None else EnvConfig.get("olm_operators")
+        self.base_dns_domain = self.base_dns_domain \
+            if self.base_dns_domain is not None else EnvConfig.get("base_domain")
```
osherdp, line 53: any reason to call it ``base_dns_domain`` in one place and ``base_domain`` in the other?

osherdp (changes requested):
partial review. will continue later

osherdp:
Does it work with downstream tests?

discovery-infra/test_infra/config/__init__.py
eliorerz: I think it's a good practice to control the each module accessibility and not give access to all module ant it's objects

discovery-infra/test_infra/config/_cluster_config.py
eliorerz: there is a backport for this  https://pypi.org/project/dataclasses/

discovery-infra/test_infra/config/_cluster_config.py
eliorerz: For boolean values it might not work

discovery-infra/test_infra/config/_cluster_config.py
eliorerz: Forgot to change it :)

discovery-infra/test_infra/config/_cluster_config.py
eliorerz: We need to change it to follow up the api_client (InventoryClient) arguments naming.
The environment variable key we might change later on

eliorerz:
> Does it work with downstream tests?
Nope. @lalon4 will start working on this soon

discovery-infra/tests/base_test.py
```diff
@@ -68,21 +70,10 @@ def cluster(self, api_client, request, nodes):
         clusters = []
 
         @JunitTestCase()
-        def get_cluster_func(cluster_name: Optional[str] = None,
-                             additional_ntp_source: Optional[str] = consts.DEFAULT_ADDITIONAL_NTP_SOURCE,
```
lalon4, line 72: We still need it as default

discovery-infra/download_logs.py
```diff
@@ -45,7 +45,7 @@ def main():
     if args.cluster_id:
         cluster = client.cluster_get(args.cluster_id)
         download_logs(client, json.loads(json.dumps(cluster.to_dict(), sort_keys=True, default=str)), args.dest,
-                      args.must_gather, args.update_by_events)
+                      args.must_gather, args.update_by_events, args.pull_secret)
```
tsorya, line 48: i think this commit should not be part of this PR. Lets have a little bit smaller scope.

discovery-infra/test_infra/config/__init__.py
tsorya: @eliorerz i don't understand why it is needed

discovery-infra/test_infra/config/_cluster_config.py
```diff
@@ -0,0 +1,55 @@
+from dataclasses import dataclass
+
+from test_infra.config import EnvConfig
+from test_infra.helper_classes.cluster_config import ClusterConfig, ClusterInstallConfig
+
+
+@dataclass
+class DefaultInstallConfig(ClusterInstallConfig):
```
tsorya, line 8: Not sure it is very clear for me why we need DefaultInstallConfig and ClusterInstallConfig ? 
I think that we will always will use default and will change only some values in it in case test needs it.
Why not to have same functionality in ClusterInstallConfig and delete this class?

discovery-infra/tests/test_install.py
```diff
@@ -9,7 +14,8 @@ class TestInstall(BaseTest):
 
     @JunitTestSuite()
     @pytest.mark.parametrize("openshift_version", get_available_openshift_versions())
-    def test_install(self, nodes, cluster, openshift_version):
-        new_cluster = cluster(openshift_version=openshift_version)
-        new_cluster.prepare_for_install(nodes=nodes)
-        new_cluster.start_install_and_wait_for_installed()
+    def test_install(self, nodes: Nodes, cluster, openshift_version):
+        config = DefaultClusterConfig(openshift_version=openshift_version)
```
tsorya, line 18: I think that usage of DefaultClusterConfig is a little bit strange

tsorya: I think that if we already set all the cluster params we should not need to provide same parameters twice.

tsorya: maybe we can allow to override them but not to provide full config for sure

discovery-infra/tests/base_test.py
```diff
@@ -68,21 +70,10 @@ def cluster(self, api_client, request, nodes):
         clusters = []
 
         @JunitTestCase()
```
tsorya, line 72: why it is test case?

discovery-infra/tests/base_test.py
```diff
@@ -68,21 +70,10 @@ def cluster(self, api_client, request, nodes):
...
-            if not cluster_name:
-                cluster_name = env_variables.get('cluster_name', infra_utils.get_random_name(length=10))
-            res = Cluster(api_client=api_client,
-                          cluster_name=cluster_name,
-                          additional_ntp_source=additional_ntp_source,
-                          openshift_version=openshift_version,
-                          user_managed_networking=user_managed_networking,
-                          high_availability_mode=high_availability_mode,
-                          olm_operators=olm_operators)
+        def get_cluster_func(cluster_config: DefaultClusterConfig):
```
tsorya, line 73: DefaultClusterConfig -> ClusterConfig

discovery-infra/tests/test_install.py
eliorerz: I wanted to preserve the current flow and not pass all argument on Cluster initialization.
Parameters that were passed on Cluster initialization stays there and parameters that were passed on installation passed there.

discovery-infra/tests/base_test.py
eliorerz: ClusterConfig is the base cluster config (it will be abstract soon). 
DefaultClusterConfig is another abject with default values inside

discovery-infra/tests/base_test.py
eliorerz: You missed that PR? :)

discovery-infra/tests/test_install.py
eliorerz: >I think that if we already set all the cluster params we should not need to provide same parameters twice.
This is the point where it passes the cluster parameters.

discovery-infra/tests/base_test.py
eliorerz: Will be fixed on next commit

discovery-infra/download_logs.py
eliorerz: The API of `cluster.get_api_vip_from_cluster` has changed.
This function wasn't supposed to be inside cluster.py. On one of the next tasks it will be deprecated and than we will be able to remove it.

discovery-infra/tests/base_test.py
eliorerz: We can remove it if you think it shouldn't be there, But this is for another PR.

discovery-infra/tests/test_install.py
tsorya: not sure i understand why we need it in case we will have config class. What it will give us? It will require from user to create 2 classes that describes same cluster. Previously maybe there was a point for it (arguable though) but in case we deal with configuration class i think there is no point to preserve same logic. Still i think it will be nice to have an option to override those params.

discovery-infra/tests/base_test.py
tsorya: but you should expect base class and not default one

discovery-infra/tests/test_install.py
eliorerz: If there is no any case that we want to pass different parameters on installation (maybe we don't know all of theme at cluster initialization).
That mean that `prepare_for_install` will receive only nodes and will not be able to receive any other arguments (due to single config object that passes to Cluster on initialization)

discovery-infra/test_infra/config/_cluster_config.py
```diff
@@ -0,0 +1,55 @@
...
+        self.iso_download_path = self.iso_download_path \
+            if self.iso_download_path is not None else EnvConfig.get("iso_download_path")
+        self.iso_image_type = self.iso_image_type \
+            if self.iso_image_type is not None else EnvConfig.get("iso_image_type")
+        self.nodes_count = self.nodes_count \
+            if self.nodes_count is not None else EnvConfig.get("num_nodes")
+        self.vip_dhcp_allocation = self.vip_dhcp_allocation \
+            if self.vip_dhcp_allocation is not None else EnvConfig.get("vip_dhcp_allocation")
+        self.download_image = self.download_image \
+            if self.download_image is not None else EnvConfig.get("download_image")
```
nshidlin, line 20: wdty, about removing download_image from here and leaving it as a parameter to `prepare_for_installation`?
This is not a test wide parameter, it is only relevant to `prepare_for_installation`

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -341,49 +346,48 @@ def wait_for_pending_for_input_status(self):
...
         utils.wait_till_at_least_one_host_is_in_stage(
             client=self.api_client,
             cluster_id=self.id,
             stages=[consts.HostsProgressStages.CONFIGURING],
-            nodes_count=env_variables['num_masters'] - 1
+            nodes_count=num_masters - 1
         )
 
-    def wait_for_non_bootstrap_masters_to_reach_joined_state_during_install(self):
+    def wait_for_non_bootstrap_masters_to_reach_joined_state_during_install(self, num_masters):
```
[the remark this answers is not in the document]
eliorerz: @lalon4 There is no num_masters in config

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -16,47 +16,50 @@
...
+from test_infra.helper_classes.config import BaseClusterConfig
+from test_infra.helper_classes.nodes import Nodes
 from test_infra.tools import static_network, terraform_utils
-from tests.conftest import env_variables
 
 
 class Cluster:
 
-    def __init__(self, api_client, cluster_name=None, additional_ntp_source=None,
-                 openshift_version="4.7", cluster_id=None, user_managed_networking=False,
```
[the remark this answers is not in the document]
eliorerz: Just pass `ClusterConfig(cluster_id=<your_cluster_id>)` for day2

discovery-infra/test_infra/helper_classes/cluster.py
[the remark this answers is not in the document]
lalon4: why not add it?

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -16,25 +16,38 @@
...
-    def __init__(self, api_client, config: ClusterConfig):
+    def __init__(self, api_client, config: BaseClusterConfig):
         self._config = config
         self.api_client = api_client
 
         self._high_availability_mode = config.high_availability_mode
         if config.cluster_id:
             self.id = config.cluster_id
-            self.name = config.cluster_name or api_client.cluster_get(config.cluster_id).name
+            cluster = api_client.cluster_get(config.cluster_id)
```
nshidlin, line 33: Why not just use the passed ClusterConfig?
In the case of the auth tests an api_client configured to another user is passed and this API call will fail

discovery-infra/test_infra/helper_classes/cluster.py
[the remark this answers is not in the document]
tsorya: maybe we should explicitly say day2 clusterId?

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -16,47 +16,50 @@
...
-from tests.conftest import env_variables
 
 
 class Cluster:
 
-    def __init__(self, api_client, cluster_name=None, additional_ntp_source=None,
-                 openshift_version="4.7", cluster_id=None, user_managed_networking=False,
-                 high_availability_mode=consts.HighAvailabilityMode.FULL, olm_operators=None):
+    def __init__(self, api_client, config: BaseClusterConfig):
+        self._config = config
```
tsorya, line 27: make it public please, there is no need for it as private.  It should be easy to change any param from config

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -341,49 +346,50 @@ def wait_for_pending_for_input_status(self):
...
         utils.wait_till_at_least_one_host_is_in_stage(
             client=self.api_client,
             cluster_id=self.id,
             stages=[consts.HostsProgressStages.REBOOTING],
-            nodes_count=nodes_count
+            nodes_count=self._config.nodes_count
         )
 
-    def wait_for_non_bootstrap_masters_to_reach_configuring_state_during_install(self):
+    def wait_for_non_bootstrap_masters_to_reach_configuring_state_during_install(self, num_masters: int = None):
```
tsorya, line 357: most of the functions in this class doesn't set expected type for env, so can we have standart here? or we change all or lets leave the way it is now.

tsorya: it can be done with another PR

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -563,53 +569,46 @@ def wait_for_install(
...
-            iso_image_type=env_variables['iso_image_type'],
-            ssh_key=env_variables['ssh_public_key'],
-            nodes_count=env_variables['num_nodes'],
-            vip_dhcp_allocation=env_variables['vip_dhcp_allocation'],
-            download_image=True,
-            platform=env_variables['platform']
-    ):
-        if download_image:
-            if env_variables.get('static_network_config'):
+    def prepare_for_installation(self, nodes: Nodes, static_network_config=None, **kwargs):
```
tsorya, line 572: add comment regarding kwargs please

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -16,47 +16,50 @@
...
-                cluster_name,
-                additional_ntp_source,
-                openshift_version,
-                user_managed_networking,
-                high_availability_mode,
-                olm_operators):
+            self.id = self._create().id
+            self.name = config.cluster_name
+
+    def _update_config(self, source_cluster: models.cluster.Cluster):
```
tsorya, line 41: please lets have a name that describes that it is day2 flow

discovery-infra/test_infra/helper_classes/cluster.py
tsorya: Lets move it to function update_config that will get kwargs

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -934,8 +931,17 @@ def wait_and_kill_installer(self, nodes, host):
         selected_node.kill_installer()
 
 
-def get_api_vip_from_cluster(api_client, cluster_info: models.cluster.Cluster):
+def get_api_vip_from_cluster(api_client, cluster_info: Union[dict, models.cluster.Cluster], pull_secret):
+    import warnings
```
tsorya, line 935: why we have internal imports?

tsorya: Ticket is too large and very difficult to understand what you mean

discovery-infra/test_infra/helper_classes/cluster.py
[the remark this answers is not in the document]
nshidlin: Well, it's not only for the day2 tests, also for the authentication tests

discovery-infra/tests/config/__init__.py
```diff
@@ -0,0 +1,9 @@
+from ._global_configs import ClusterConfig, NodesConfig
```
tsorya, line 1: I really don't like this way. I don't think it is somehow required.
@osherdp  @YuviGold what do you think?

discovery-infra/test_infra/helper_classes/config/__init__.py
```diff
@@ -0,0 +1,8 @@
+from ._cluster_config import BaseClusterConfig
```
tsorya, line 1: same, i don't think this is needed

discovery-infra/test_infra/helper_classes/cluster.py
eliorerz: @nshidlin Just wanted to preserve the current logic. If we wish to change it, it can be done later on

nshidlin: @eliorerz I can work around it

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -84,45 +87,43 @@ def get_host_assigned_roles(self):
...
 
     @JunitTestCase()
     def generate_and_download_image(
             self,
-            iso_download_path=env_variables['iso_download_path'],
-            ssh_key=env_variables['ssh_public_key'],
+            iso_download_path=None,
             static_network_config=None,
-            iso_image_type=env_variables['iso_image_type']
+            iso_image_type=None
```
lalon4, line 98: `iso_image_type or self._config.iso_image_type`

discovery-infra/test_infra/helper_classes/cluster.py
eliorerz: Don't think the whole config needs to be public. If there is something we want to change or get we can define a setter/getter accordingly

discovery-infra/test_infra/helper_classes/cluster.py
eliorerz: @tsorya Already pushed the fix

discovery-infra/test_infra/helper_classes/cluster.py
eliorerz: >why we have internal imports? Cause we are going to deprecate it?

Yes, The internal import is just because it will be deprecated soon.


>i think you should explain what we should use in this case

The function and solution for that case have not been determined yet. I might be on another module of as a classmethod within Cluster class.

discovery-infra/tests/config/__init__.py
eliorerz: @tsorya What way?

eliorerz: If you are talking about the private modules:

Because the library and usage of the library are on the same repo, we want to allow only specific object to be exported, for the same reason we need to use private/protected properties in our classes.
All public can cause us a lot of issues that will cost us time and effort.

osherdp: @tsorya please be more specific next time
If talking about the private modules, then I agree. @eliorerz, no one does that on Python and it has different rules and conventions than golan/Java/C# etc.

tsorya: i meant that i don't like adding these __init__ file and using private modules too.

eliorerz: @osherdp @tsorya  Using protected modules is not that common but it indeed in use (e.g. pytest, flake8)
It is more needed in python (and not on other languages) because of the "freedom" that the developer has, more boundaries are needed.
Anyway, I think it will prevent us for making more mistakes as we did before (especially while the infra and its usage are on the same place).
I think if you don't have a major objection to this we can leave it and refactor it back is pretty easy, else I can just remove it (but again I think it's wrong).

osherdp: Sorry, I do object this
As it's a simple change, let's change this now to stay coherent to our repo's conventions

osherdp: About the init usage, I don't have any preferences

eliorerz: @osherdp It just the first step (and first task), but a lot of this repo's conventions are going to change (during MGMT-4975).

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -563,53 +574,43 @@ def wait_for_install(
...
-    def download_kubeconfig_no_ingress(
-            self, kubeconfig_path=env_variables['kubeconfig_path']
-    ):
+    def download_kubeconfig_no_ingress(self, kubeconfig_path):
         self.api_client.download_kubeconfig_no_ingress(self.id, kubeconfig_path)
 
-    def download_kubeconfig(
-            self, kubeconfig_path=env_variables['kubeconfig_path']
-    ):
+    def download_kubeconfig(self, kubeconfig_path):
```
nshidlin, line 613: This should have a default value for kubeconfig_path, coming from self._config

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -563,53 +574,43 @@ def wait_for_install(
...
         self.wait_for_ready_to_install()
 
-        if platform == consts.Platforms.NONE:
+        if self._config.platform == consts.Platforms.NONE:
             self._configure_load_balancer(nodes.controller)
 
-    def download_kubeconfig_no_ingress(
-            self, kubeconfig_path=env_variables['kubeconfig_path']
-    ):
+    def download_kubeconfig_no_ingress(self, kubeconfig_path):
```
nshidlin, line 610: This should have a default value for kubeconfig_path, coming from self._config

discovery-infra/tests/config/__init__.py
osherdp: Sorry, I just can't approve this specific convention change.
No hurt feelings, though :yellow_heart:

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -84,45 +101,43 @@ def get_host_assigned_roles(self):
     def get_operators(self):
         return self.api_client.get_cluster_operators(self.id)
 
-    def generate_image(self, ssh_key=env_variables['ssh_public_key']):
-        self.api_client.generate_image(cluster_id=self.id, ssh_key=ssh_key)
+    def generate_image(self):
+        self.api_client.generate_image(cluster_id=self.id, ssh_key=self._config.ssh_public_key)
 
     @JunitTestCase()
     def generate_and_download_image(
```
nshidlin, line 108: removal of ssh_key breaks `set_iptables_rules_for_nodes`

ssh_key should be removed from there as well

discovery-infra/test_infra/helper_classes/cluster.py
eliorerz: Why can't you just `cluster.download_kubeconfig(kubeconfig_path=env_variables["kubeconfig_path"])`?

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -341,49 +360,50 @@ def wait_for_pending_for_input_status(self):
...
         )
 
-    def wait_for_at_least_one_host_to_boot_during_install(self, nodes_count=1):
+    def wait_for_at_least_one_host_to_boot_during_install(self):
         utils.wait_till_at_least_one_host_is_in_stage(
             client=self.api_client,
             cluster_id=self.id,
             stages=[consts.HostsProgressStages.REBOOTING],
-            nodes_count=nodes_count
+            nodes_count=self._config.nodes_count
```
lalon4, line 368: please set default 1 instead of self._config.nodes_count

eliorerz: Why? This is a different logic..

discovery-infra/test_infra/helper_classes/cluster.py
nshidlin: I can take it from env_variables["kubeconfig_path"], but i thought we are trying to move away from using env_variables dict from within the tests

discovery-infra/test_infra/helper_classes/cluster.py
lalon4: why it is different? it was nodes_count=1 before the change
method name also indicate that at least 1 node is required :) 
wait_for_**at_least_one_host_to**_boot_during_install

discovery-infra/test_infra/helper_classes/cluster.py
eliorerz: You can use `cluster.download_kubeconfig_no_ingress(kubeconfig_path=env_variables["kubeconfig_path"])`

discovery-infra/test_infra/helper_classes/cluster.py
lalon4: also there are numerous of tests that uses this, its better to always have default in this method, the cases where it is overwritten is very low

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -174,9 +190,9 @@ def set_olm_operator(self, operator_name, properties=None):
 
         self.api_client.update_cluster(self.id, {'olm_operators': olm_operators})
 
-    def set_host_roles(self, requested_roles=None):
+    def set_host_roles(self, num_masters, num_workers, requested_roles=None):
```
lalon4, line 193: default for num_masters and num_workers should be taken from _config

discovery-infra/test_infra/helper_classes/nodes.py
```diff
@@ -93,8 +92,7 @@ def get_random_node(self):
     def shutdown_all(self):
         self.run_for_all_nodes("shutdown")
 
-    def start_all(self):
-        static_ips_config = env_variables.get('static_ips_config')
+    def start_all(self, static_ips_config):
```
lalon4, line 95: static_ips_configs should have a default

nshidlin: agreed should default to False

discovery-infra/test_infra/helper_classes/cluster.py
eliorerz: num_workers is not part of cluster configuration

discovery-infra/test_infra/helper_classes/nodes.py
eliorerz: 👍🏻

discovery-infra/test_infra/helper_classes/cluster.py
eliorerz: I'm adding num_masters temporarily to Config and probably will be moved to Nodes in MGMT-5370

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -84,45 +102,43 @@ def get_host_assigned_roles(self):
...
-    def generate_image(self, ssh_key=env_variables['ssh_public_key']):
-        self.api_client.generate_image(cluster_id=self.id, ssh_key=ssh_key)
+    def generate_image(self):
+        self.api_client.generate_image(cluster_id=self.id, ssh_key=self._config.ssh_public_key)
 
     @JunitTestCase()
     def generate_and_download_image(
             self,
-            iso_download_path=env_variables['iso_download_path'],
-            ssh_key=env_variables['ssh_public_key'],
```
lalon4, line 94: Need to be able to overwrite ssh_key for negative tests

eliorerz: Why not override config value?

lalon4: i can do that, but not sure its what need to be done
see this test verifying ssh key validation: 
https://gitlab.cee.redhat.com/ocp-edge-qe/kni-assisted-installer-auto/-/blob/master/api_tests/test_validations.py#L711
WDYT?

eliorerz: Added option to override

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -341,49 +385,50 @@ def wait_for_pending_for_input_status(self):
...
             client=self.api_client,
             cluster_id=self.id,
             stages=[consts.HostsProgressStages.CONFIGURING],
-            nodes_count=env_variables['num_masters'] - 1
+            nodes_count=num_masters - 1
         )
 
-    def wait_for_non_bootstrap_masters_to_reach_joined_state_during_install(self):
+    def wait_for_non_bootstrap_masters_to_reach_joined_state_during_install(self, num_masters: int = None):
+        num_masters = num_masters if num_masters is None else self._config.masters_count
```
lalon4, line 406: is not None

PR #730, opened 2021-04-23
MGMT-4792 LSO support

pkliczewski:
In this PR we add ability to configure `LSO` operator as well use when `OCS` operator is set we make sure we have at least 3 worker nodes with 5GB disk attached.


pkliczewski:
@YuviGold I do not have an env to test this change. There is 5GB requirement per disk and I am not sure whether I need to add it or it would work correctly. Do you have any idea where I could check this?

discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
```diff
@@ -48,7 +48,7 @@ def _terraform_params(self, **kwargs):
                   "libvirt_master_memory": utils.resource_param(kwargs.get("master_memory", 16984), "master_memory", operators),
                   "libvirt_worker_vcpu": utils.resource_param(kwargs.get("worker_vcpu", 4), "worker_vcpu", operators),
                   "libvirt_master_vcpu": utils.resource_param(kwargs.get("master_vcpu", 4), "master_vcpu", operators),
-                  "worker_count": kwargs.get('num_workers', 0),
+                  "worker_count": utils.ocs_resource_param(kwargs.get('num_workers', 0), "num_workers", operators),
```
YuviGold, line 51: I don't think it should specifically mention `ocs`. Why not use the same `resource_param`?

discovery-infra/test_infra/utils.py
```diff
@@ -953,4 +953,10 @@ def resource_param(value, param_name, operators):
     try:
         return sum((consts.OPERATOR_PARAMS[operator][param_name] for operator in operators), value)
     except KeyError as e:
-        raise ValueError(f"Unknown operator name {e.args[0]}")
\ No newline at end of file
+        raise ValueError(f"Unknown operator name {e.args[0]}")
+
+def ocs_resource_param(value, param_name, operators):
+    if "ocs" in operators and value > consts.OPERATOR_PARAMS["ocs"][param_name]:
```
YuviGold, line 959: How would it work with multiple OLMs? Do we need sum of all the requirements or just getting max?
We can use `resource_param` to receive a function

```
def resource_param(value, param_name, operators, func):
    try:
        return func((consts.OPERATOR_PARAMS[operator][param_name] for operator in operators), value)
    except KeyError as e:
        raise ValueError(f"Unknown operator name {e.args[0]}")
```

YuviGold:
> @YuviGold I do not have an env to test this change. There is 5GB requirement per disk and I am not sure whether I need to add it or it would work correctly. Do you have any idea where I could check this?

@lalon4 maybe you know?

discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
pkliczewski: The logic is a bit different we could add `OCS` specific requirements to other operator with 0 value. Do you suggest to do this instead?

discovery-infra/test_infra/utils.py
pkliczewski: We need to sum up all the requirements

discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
YuviGold: What is the specific requirement? additional disks? If so, then yes.
I don't want to have "OCS" reference in the terraform generic conf.

pkliczewski: Additional disks and number of workers. We need to make sure we have at least 3 when deploying `OCS`

discovery-infra/test_infra/utils.py
YuviGold: So I think `resource_param` already takes care of that logic. no?

discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
YuviGold: Okay. It's fine to add zeros to others

discovery-infra/test_infra/utils.py
pkliczewski: Assuming we add worker_count and worker_disk to other operator we can reuse.

lalon4:
In order to install OCS there is a need for 1 additional disks. LSO creating pv on the extra disk, and OCS uses those. (@priyanka19-98 :) )

YuviGold (approved):
LGTM

YuviGold:
LSO isn't reported
```
INFO     root:utils.py:209 Asked operators to be in one of the statuses from ['available'] and currently operators statuses are [('cnv', 'available', 'install strategy completed with no errors'), ('console', 'available', 'All is well'), ('cvo', 'available', 'Done applying 4.7.8'), ('lso', None, None)]
```

controller log
```
time="2021-04-27T15:53:48Z" level=warning msg="Failed to get lso" error="resource name may not be empty"
```

pkliczewski:
> LSO isn't reported

Thanks, will investigate.

YuviGold:
> > LSO isn't reported
> 
> Thanks, will investigate.

@danielerez 
LSO workflow got broken due to https://github.com/openshift/assisted-service/pull/1319 .
```
E0428 04:37:14.092160       1 queueinformer_operator.go:290] sync "openshift-local-storage" failed: constraints not satisfiable: subscription local-storage-operator exists, no operators found matching the criteria of subscription local-storage-operator
```

```
  spec:
    channel: 4.7.8
    installPlanApproval: Automatic
    name: local-storage-operator
    source: redhat-operators
    sourceNamespace: openshift-marketplace
```

I find it a little bit strange that LSO should follow `major.minor` channel, CNV follows `stable` and OCS doesn't follow a channel at all. We should align them.

pkliczewski:
> I find it a little bit strange that LSO should follow `major.minor` channel, CNV follows `stable` and OCS doesn't follow a channel at all. We should align them.

I do not think we have a way to enforce alignment here. Each team defines how they want to be consumed. I am not sure whether there are any best practices here which would help us to raise this point.

YuviGold:
https://github.com/openshift/assisted-service/pull/1597 was merged.
/retest

YuviGold:
Looks good
```
      {
        "cluster_id": "6defa3ab-a422-4aab-8ffa-9d2236ff0cee",
        "name": "cvo",
        "operator_type": "builtin",
        "status": "available",
        "status_info": "Done applying 4.7.8",
        "status_updated_at": "2021-04-28T18:13:08.624Z",
        "timeout_seconds": 3600
      },
      {
        "cluster_id": "6defa3ab-a422-4aab-8ffa-9d2236ff0cee",
        "name": "lso",
        "namespace": "openshift-local-storage",
        "operator_type": "olm",
        "status": "available",
        "status_info": "install strategy completed with no errors",
        "status_updated_at": "2021-04-28T18:15:45.653Z",
        "subscription_name": "local-storage-operator",
        "timeout_seconds": 4200
      }
```

Please rebase after @eliorerz  PR https://github.com/openshift/assisted-test-infra/pull/743

YuviGold:
Fixes MGMT-5296

PR #734, opened 2021-04-25
MGMT-5360 - remove support for mirror registry configuration from the…

yevgeny-shnaidman:
… image create flow

the mirror registry configuration will be setinto Assisted Service during deployment,
so we no longer need this option in the create image API


eliorerz (approved):
/lgtm

PR #740, opened 2021-04-26
MGMT-4577 kube_api: use 4.8 release image

danielerez:
As only 4.8 is supported when KUBE_API is enabled (see scripts/deploy_assisted_service.sh),
the relevant release image should be fetched accordingly.


danielerez:
@YuviGold @osherdp

YuviGold (approved):

PR #747, opened 2021-04-28
OCPBUGSM-24075 fix events using English spelling canceled

eliorerz:
Waiting for https://github.com/openshift/assisted-service/pull/1610


discovery-infra/test_infra/consts/consts.py
```diff
@@ -124,8 +124,8 @@ class Events:
     IGNITION = "reached installation stage Waiting for ignition"
     CONFIGURING = "reached installation stage Configuring"
     DONE = "reached installation stage Done"
-    CANCELED_CLUSTER_INSTALLATION = "Canceled cluster installation"
-    CANCELED_FOR_HOST = "Installation canceled for host"
+    CANCELED_CLUSTER_INSTALLATION = "Cancelled cluster installation"
+    CANCELED_FOR_HOST = "Installation cancelled for host"
```
YuviGold, line 128: We should update the variables names as well

eliorerz: Don't think so, because we want to change it all later on (here and on the service) to canceled (as it is partially on the service also).

YuviGold (approved):
LGTM

PR #749, opened 2021-04-29
MGMT-5381 gather sosreport from nodes

osherdp:
Adding a CLI option for gathering sosreport-like data via SSH.
/cc @eranco74 @tsorya

### DIFF
--- a/discovery-infra/bootstrap_in_place.py
+++ b/discovery-infra/bootstrap_in_place.py
@@ -12,7 +12,7 @@ from test_infra.tools.assets import NetworkAssets
 from test_infra.helper_classes.nodes import Nodes
 from test_infra.controllers.node_controllers.terraform_controller import TerraformController
 
-from download_logs import download_must_gather
+from download_logs import download_must_gather, gather_sosreport_from_node
 from oc_utils import get_operators_status
 
 warn_deprecate()
@@ -30,7 +30,6 @@ MUST_GATHER_DIR = os.path.join(IBIP_DIR, "must-gather")
 INSTALLER_GATHER_DIR = os.path.join(IBIP_DIR, "installer-gather")
 INSTALLER_GATHER_DEBUG_STDOUT = os.path.join(INSTALLER_GATHER_DIR, "gather.stdout.log")
 INSTALLER_GATHER_DEBUG_STDERR = os.path.join(INSTALLER_GATHER_DIR, "gather.stderr.log")
-SOSREPORT_SCRIPT = os.path.join(RESOURCES_DIR, "man_sosreport.sh")
 SSH_KEY = os.path.join("ssh_key", "key")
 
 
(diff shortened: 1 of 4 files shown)


eranco74 (commented):
/lgtm

osherdp:
/hold
to get more reviews and to see live-iso build status

discovery-infra/test_infra/controllers/node_controllers/node_controller.py
```diff
@@ -45,10 +45,6 @@ def start_all_nodes(self) -> List[Node]:
     def restart_node(self, node_name: str) -> None:
         pass
 
-    @abstractmethod
-    def format_node_disk(self, node_name: str) -> None:
```
tsorya, line 49: didn't we talked to leave this one?

osherdp: the implementation is dependent on cluster_id unfortunately

discovery-infra/download_logs.py
```diff
@@ -224,6 +236,25 @@ def download_must_gather(kubeconfig: str, dest_dir: str):
         log.warning(f"Failed to run must gather: {ex}")
 
 
+def gather_sosreport_data(output_dir: str):
+    sosreport_output = os.path.join(output_dir, "sosreport")
+    recreate_folder(sosreport_output)
+    controller = LibvirtController(private_ssh_key_path=private_ssh_key_path_default)
+    for node in controller.list_nodes():
```
tsorya, line 243: maybe run it in parallel? with run_concurrently ?

discovery-infra/download_logs.py
```diff
@@ -224,6 +236,25 @@ def download_must_gather(kubeconfig: str, dest_dir: str):
...
+        try:
+            gather_sosreport_from_node(node, sosreport_output)
+        except (TimeoutError, IndexError, SSHException) as e:
```
tsorya, line 246: why not to catch any exception too? i think it is better not to fail and bring all we can then to write nice exception code

discovery-infra/test_infra/controllers/node_controllers/node_controller.py
tsorya: i think better to add empty function to libvirt controller then to remove this one from interface

osherdp: Maybe let's meet in the middle and raise ``NotImplementedError``?
Instead of just passing, which is a silent error

It will still allow instantiating the class, but not using this method

discovery-infra/download_logs.py
osherdp: if we have a real error?

discovery-infra/download_logs.py
osherdp: done

eliorerz (approved):

discovery-infra/bootstrap_in_place.py
```diff
@@ -173,7 +165,7 @@ def log_collection(controller, vm_ip):
     try:
         logging.info("Gathering sosreport data from host...")
         node = Nodes(controller, private_ssh_key_path=SSH_KEY)[0]
-        gather_sosreport_data(node)
+        gather_sosreport_from_node(node)
     except Exception:
```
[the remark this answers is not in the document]
osherdp: no, guys, no
``KeyboardInterrupt`` will be caught as well when catching ``BaseException``
catching ``Exception`` is bad enough as is :)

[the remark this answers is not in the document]
eliorerz: You are right if we want catch only python built-in exceptions, but in case of using an external libraries, how can you tell if some exceptions inherits directly from `BaseException`?
On large projects I believe it is better to use `BaseException` if we want to catch all and not trace who inherit from `Exception` and who from   `BaseException` 🤷🏻‍♂️

PR #751, opened 2021-04-29
Add test test_olm_operator to test all supported OLMs

YuviGold:


discovery-infra/tests/test_install.py
```diff
@@ -16,3 +16,12 @@ def test_install(self, nodes: Nodes, cluster, openshift_version):
         new_cluster = cluster(cluster_config=ClusterConfig(openshift_version=openshift_version))
         new_cluster.prepare_for_installation(nodes)
         new_cluster.start_install_and_wait_for_installed()
+
+    @JunitTestSuite()
+    # TODO: Fix OCS
+    # @pytest.mark.parametrize("olm_operator", get_api_client().get_supported_operators())
+    @pytest.mark.parametrize("olm_operator", ["lso", "cnv"])
+    def test_olm_operator(self, nodes: Nodes, cluster, olm_operator):
```
omertuc, line 24: Maybe also add test to test all at the same time?

YuviGold: At the moment OCS isn't supported, hence only LSO and CNV are tested.
CNV already depends on LSO then this logic happens.
Running all the future permutations is resource-consuming. We don't need to verify this right now.

omertuc: I only meant to suggest running all supported operators (so no need for OCS, if it doesn't work now).

I didn't suggest running all permutations, just adding a test that runs all currently supported operators together.

For now maybe you can add it like this:
```python
TESTED_OPERATORS=["lso", "cnv"]

@pytest.mark.parametrize("olm_operator", TESTED_OPERATORS)
def test_olm_operator(self, nodes: Nodes, cluster, olm_operator):
	...

@pytest.mark.skipif(TESTED_OPERATORS == ["lso", "cnv"],
                    reason=f"Since CNV depends on LSO, this already gets ran as part of test_olm_operator")
def test_all_olm_operators(self, nodes: Nodes, cluster):
    ...
```

YuviGold:
@pkliczewski @eliorerz 
Any idea if something got broken?
```
        "status_info": "Host cannot be installed due to following failing validation(s): Require at least 8 CPU cores for master role, found only 4",
```

discovery-infra/tests/test_install.py
YuviGold: I see. This could be nice but it won't add anything for now. When OCS would be back I can add this test.

osherdp (approved):

PR #752, opened 2021-04-29
NO-ISSUE: Fixed OCS typo that caused to tests failure

priyanka19-98:
Signed off by: Priyanka Jiandani <pjiandan@redhat.com>


priyanka19-98:
Fix already covered in https://github.com/openshift/assisted-test-infra/pull/730

PR #754, opened 2021-05-01
NO-ISSUE another fix for sosreport on ibip flow

osherdp:
It turns out that ``Nodes`` class doesn't set SSH private keys on its nodes.
As a fix, using the controller class to make access to the nodes.


YuviGold (approved):
LGTM

PR #768, opened 2021-05-05
MGMT-6049 Wait for operators to finish first wait for builtin then to OLM

YuviGold:
Builtin operators have to succeed. Otherwise, the cluster isn't
accessible. OLM could fail and then the cluster would be declared as
degraded.

/cc @lalon4

### DIFF
--- a/discovery-infra/test_infra/helper_classes/cluster.py
+++ b/discovery-infra/test_infra/helper_classes/cluster.py
@@ -21,6 +21,8 @@ from test_infra.helper_classes.config import BaseClusterConfig
 from test_infra.helper_classes.nodes import Nodes
 from test_infra.tools import static_network, terraform_utils
 
+from assisted_service_client.models.operator_type import OperatorType
+
 
 class Cluster:
     MINIMUM_NODES_TO_WAIT = 1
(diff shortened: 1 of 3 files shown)


discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -580,6 +582,17 @@ def wait_for_hosts_to_install(
...
     def wait_for_operators_to_finish(self, timeout=consts.CLUSTER_INSTALLATION_TIMEOUT, fall_on_error_status=True):
+        operators = self.get_operators()
+        utils.wait_till_all_operators_are_in_status(
+            client=self.api_client,
+            cluster_id=self.id,
+            operators_count=len(utils.filter_operators_by_type(operators, OperatorType.BUILTIN)),
+            operator_type=OperatorType.BUILTIN,
+            statuses=[consts.OperatorStatus.AVAILABLE],
+            timeout=timeout,
+            fall_on_error_status=fall_on_error_status,
```
lalon4, line 593: we shouldn't have fall_on_error_status=True for built-in

YuviGold: I switched the statuses.
In case someone wants to use the API and fail on error (when only available is acceptable for builtins) then it should work. (for example, to test failures)
Regarding OLMs we accept anyway both of them.

PR #769, opened 2021-05-05
MGMT-5395: Support multiple installations per namespace

RazRegev:
By passing CLUSTER_INDEX (by default 0), test-infra will support
multiple installations for a specific namespace.


discovery-infra/test_infra/consts/consts.py
```diff
@@ -40,6 +40,8 @@
 DEFAULT_NAMESPACE = 'assisted-installer'
 DEFAULT_TEST_INFRA_DOMAIN = f".{CLUSTER_PREFIX}-{DEFAULT_NAMESPACE}.{DEFAULT_BASE_DNS_DOMAIN}"
 TEST_TARGET_INTERFACE = "vnet3"
+RANGE_8BIT = 256
```
omertuc, line 43: `2**8`

discovery-infra/test_infra/consts/consts.py
```diff
@@ -40,6 +40,8 @@
...
+RANGE_16BIT = 65536
```
omertuc, line 44: `2**16`

scripts/assisted_deployment.sh
```diff
@@ -6,24 +6,33 @@ function destroy_all() {
 
 function set_dns() {
     NAMESPACE_INDEX=${1:-0}
+    CLUSTER_INDEX=${2:-0}
+    NETWORK_BRIDGE=tt$NAMESPACE_INDEX
```
omertuc, line 10: Use `${..}` rather than `$..` (also in the `if` below) to make this a bit more readable

discovery-infra/test_infra/consts/consts.py
RazRegev: 2**8 + 1, I want it to effect the second bit

discovery-infra/test_infra/consts/consts.py
RazRegev: 2**16 + 1

discovery-infra/test_infra/consts/consts.py
RazRegev: I added the +1 to the calculation itself

omertuc: But 256 is exactly 2**8, not 2**8+1

2**8+1 is 257

RazRegev: oh ops, you are right, fixing

tsorya:
@RazRegev can't we use e2e kube-api test to spawn this? we already support parallel run there.

YuviGold:
I would like to avoid introducing another INDEX variable flooding all of our functions. We already have _NAMESPACE_INDEX_ and I don't understand why we also need _CLUSTER_INDEX_

PR #772, opened 2021-05-07
MGMT-4792 add ocs support in olm_operators

pkliczewski:
In order to provision cluster with OCS deployed we need to add enough memory and more disks.


pkliczewski:
@YuviGold still requires some work but adding dynamically disks works fine

discovery-infra/test_infra/utils.py
```diff
@@ -938,3 +938,10 @@ def resource_param(base_value: int, resource_name: str, operators: List):
         return sum((consts.OperatorResource.values()[operator][resource_name] for operator in operators), base_value)
     except KeyError as e:
         raise ValueError(f"Unknown operator name {e.args[0]}")
+
+def disk_names(base: int, count: int):
```
YuviGold, line 942: Is that really needed? You already do a for loop in `main.tf` where you define the full name of the disks. You can create the prefix over there.

discovery-infra/test_infra/consts/olm_operators.py
```diff
@@ -19,25 +19,29 @@ class OperatorResource:
...
         return {
             cls.WORKER_MEMORY_KEY: worker_memory,
             cls.MASTER_MEMORY_KEY: master_memory,
             cls.WORKER_VCPU_KEY: worker_vcpu,
             cls.MASTER_VCPU_KEY: master_vcpu,
             cls.WORKER_COUNT_KEY: worker_count,
-            cls.WORKER_DISK_KEY: worker_disk
+            cls.WORKER_DISK_KEY: worker_disk,
+            cls.MASTER_DISK_KEY: master_disk,
+            cls.DISK_COUNT_KEY: disk_count
```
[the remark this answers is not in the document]
pkliczewski: ATM it is for both. Will split it.

discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
```diff
@@ -67,15 +71,17 @@ def _terraform_params(self, **kwargs):
...
+                kwargs.get('master_disk', 128849018880), consts.OperatorResource.MASTER_DISK_KEY, operators),
             "libvirt_secondary_network_name": consts.TEST_SECONDARY_NETWORK + self.cluster_suffix,
             "libvirt_storage_pool_path": kwargs.get('storage_pool_path', os.path.join(os.getcwd(), "storage_pool")),
             # TODO change to namespace index
             "libvirt_secondary_network_if": self.network_conf.libvirt_secondary_network_if,
             "provisioning_cidr": self.network_conf.provisioning_cidr,
             "running": True,
             "single_node_ip": kwargs.get('single_node_ip', ''),
-            }
+            "disk_names": utils.disk_names(kwargs.get("disk_count", 1), consts.OperatorResource.MASTER_DISK_KEY)
```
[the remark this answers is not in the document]
pkliczewski: Yeah, right it should be `DISK_COUNT_KEY` not `MASTER_DISK_KEY`. Will fix. Thanks

discovery-infra/test_infra/utils.py
pkliczewski: It took me some time to figure out how to do it. I use `setproduct` function to find all of the possible combinations of disk names based on worker # and `disk_name` with #.  "You can't dynamically generated nested blocks using for expressions" so we would need to use local variable and reuse it to build `disk_names` but it is not semantically different to what we have now.

pkliczewski: On the other hand by using range and disk_number I could improve existing expression. Will explore.

pkliczewski:
@eliorerz Is there anything else we are waiting for or can we unhold?

pkliczewski:
@YuviGold Please review again

YuviGold:
LGTM. Please update BaseTerraformConfig with the new resources members (that's how we would be able to override the configuration on TerraformController)

YuviGold (approved):
LGTM

PR #773, opened 2021-05-09
NO-ISSUE Adding support to update and set SNO specific HW details

tsorya:


discovery-infra/update_assisted_service_cm.py
```diff
@@ -38,7 +38,9 @@
     ("OCM_BASE_URL", ""),
     ("PUBLIC_CONTAINER_REGISTRIES", ""),
     ("CHECK_CLUSTER_VERSION", ""),
-    ("HW_VALIDATOR_REQUIREMENTS", "")
+    ("HW_VALIDATOR_REQUIREMENTS", ""),
+    ("HW_VALIDATOR_MIN_CPU_CORES_SNO", ""),
```
omertuc, line 42: Where is this used?

YuviGold, line 42: Why there's an SNO specific variable? I don't think we should support it. It goes the same for any mode specific variables

omertuc: It's already in the service https://github.com/openshift/assisted-service/blob/f4fc64b827be1213f7c98a9c4b207c0aaf1de32b/internal/hardware/validator.go#L58

omertuc: Also the service has to support both SNO and non-SNO, and it needs different requirements when there's only one node (because it has much more burden put on it), how else could you do it without separate variables?

YuviGold: I see. I think it's a bad habit to have that many environment variables just for different modes.
FYI @jakub-dzon

YuviGold: @omertuc There's _HW_VALIDATOR_REQUIREMENTS_ that holds all this data. So there should be different dictionaries for SNO and multi-node maybe. But having more variables for MIN_CPU and MIN_RAM looks bad. Especially when there could be more future aspects and modes to keep track.

omertuc: > @omertuc There's _HW_VALIDATOR_REQUIREMENTS_ that holds all this data. So there should be different dictionaries for SNO and multi-node maybe. But having more variables for MIN_CPU and MIN_RAM looks bad. Especially when there could be more future aspects and modes to keep track.

Sounds reasonable

PR #785, opened 2021-05-11
MGMT-5347- deprecate deploy_default_secret on kube_helpers secret

eliorerz:


YuviGold:
/hold
Until understanding why this change is needed

eliorerz:
After discussion the decision is to leave secret file as is

PR #791, opened 2021-05-12
Add DNS records to provisioned TF libvirt network

YuviGold:


terraform_files/baremetal/variables-libvirt.tf
```diff
@@ -203,4 +203,10 @@ variable "provisioning_cidr6" {
   description = "IPv6 provisioning network from network pool for automated tests"
   type = string
   default = ""
-}
\ No newline at end of file
+}
+
+variable "libvirt_dns_records" {
+  type        = map(string)
```
osherdp, line 209: this is map between string to string?

osherdp (approved):

terraform_files/baremetal/variables-libvirt.tf
YuviGold: Yes. It's host name points to the resolved name

osherdp (approved):

osherdp (approved):

PR #792, opened 2021-05-12
SSH through all available node IPs

osherdp:
When configuring a SSH connection, we need to decide which IP address to use.
Currently we get all DHCP leases from libvirt, and use the first provided address. It will work on IPv4 because DHCP will always return the same address to the hosts, but on IPv6 environment the returned addresses are not constant.
This change is making a complete traversal over all IP addresses of the node, over all networks. This will ensure that as long as the node is up, we'll be able to reach it.

### DIFF
--- a/discovery-infra/test_infra/controllers/node_controllers/node.py
+++ b/discovery-infra/test_infra/controllers/node_controllers/node.py
@@ -1,6 +1,8 @@
 import logging
 from typing import Callable
 
+from scp import SCPException
+
 from test_infra import consts
 from test_infra.controllers.node_controllers import ssh
 from test_infra.controllers.node_controllers.disk import Disk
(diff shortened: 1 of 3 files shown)


osherdp:
/cc @YuviGold @tsorya 
It's already seems like we'll need to fix kube-api afterwards, but at least this will fix all ipv6 builds

discovery-infra/test_infra/controllers/node_controllers/node.py
```diff
@@ -49,9 +51,26 @@ def macs(self):
...
+            exception = None
+            try:
+                connection = ssh.SshConnection(
+                    ip,
+                    private_ssh_key_path=self.private_ssh_key_path,
+                    username=self.username)
+                connection.connect()
+                return connection
+
+            except (TimeoutError, SCPException) as e:
```
tsorya, line 68: sure you don't want to catch generic exception too? I think much better to have logs than to right code nicer

osherdp: See the CI channel, we always have the logs coming after sosreport step
If we'll see it's getting on our way we should consider changing that

osherdp:
failure in ``ci/prow/e2e-metal-assisted-edge-ready`` is relevant to https://github.com/openshift/release/pull/18546

PR #796, opened 2021-05-13
NO-ISSUE deploy service kube-api: added rhcos_rootfs to versions

danielerez:
Following to https://github.com/openshift/assisted-service/pull/1697,
'rhcos_rootfs' property should now be included in OPENSHIFT_VERSIONS
env var for the kube-api flow.


osherdp (approved):

carbonin:
Thanks for catching this.
But also, why do we need this code? Why can't we just use the 4.8 entry in the original json?

PR #797, opened 2021-05-16
Remove old functions from assisted_deployment.sh and print net-dhcp-leases

YuviGold:


scripts/assisted_deployment.sh
```diff
@@ -1,20 +1,21 @@
 #!/usr/bin/env bash
 
-function destroy_all() {
```
tsorya, line 3: Why?

scripts/assisted_deployment.sh
```diff
@@ -43,32 +45,4 @@ function update_vips() {
     virsh net-update test-infra-net-${NAMESPACE} add dns-host "<host ip='${INGRESS_VIP}'><hostname>assisted-service-assisted-installer.apps.${CLUSTER_NAME}-${NAMESPACE}.${BASE_DOMAIN}</hostname></host>"
 }
 
-# Delete after pushing fix to dev-scripts
-function wait_for_cluster() {
-    echo "Nothing to do"
-}
-
-#TODO ADD ALL RELEVANT OS ENVS
-function run() {
```
tsorya, line 52: Why?

scripts/assisted_deployment.sh
YuviGold: it isn't used

scripts/assisted_deployment.sh
YuviGold: it isn't used

YuviGold:
Not relevant anymore.

PR #799, opened 2021-05-16
NO-ISSUE catch RuntimeError when node has no IPs

osherdp:
I changed the returned exception where no IPs have been assigned to the node, and forgot to change it in the using function.
/cc @YuviGold 
/hold


YuviGold (approved):
LGTM

PR #805, opened 2021-05-18
Support cpu virtualization flag as a node config

lalon4:
CPU mode is now parameterized in order to be able to test CPU virtualization validation flag in CNV


discovery-infra/test_infra/consts/consts.py
```diff
@@ -9,6 +9,7 @@
 IMAGE_NAME = "installer-image.iso"
 STORAGE_PATH = "/var/lib/libvirt/openshift-images"
 SSH_KEY = "ssh_key/key.pub"
+VM_CPU_MODE = "host-passthrough"
```
pkliczewski, line 12: What do you think about using more descriptive name like `MASTER_NODE_PASS_CPU_FEATURES` or `WORKER_NODE_PASS_CPU_FEATURES`?

discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
```diff
@@ -90,6 +90,8 @@ def _terraform_params(self, **kwargs):
                 kwargs.get("master_disk_count", 1), consts.OperatorResource.MASTER_DISK_COUNT_KEY, operators),
             "worker_disk_count": utils.resource_param(
                 kwargs.get("worker_disk_count", 1), consts.OperatorResource.WORKER_DISK_COUNT_KEY, operators),
+            "worker_cpu_mode": kwargs.get("worker_cpu_mode", consts.VM_CPU_MODE),
+            "master_cpu_mode": kwargs.get("master_cpu_mode", consts.VM_CPU_MODE)
```
pkliczewski, line 94: Why not to use 2 variables. one for worker and the other for master?

discovery-infra/test_infra/consts/consts.py
lalon4: yeah, OK, i agree the name is not informative enough. But MASTER_NODE_PASS_CPU_FEATURES also not clear for anyone that is out of this context.... how about MASTER_TF_CPU_MODE?

discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
lalon4: i assumed that this will never used except of very specific flows, do you think there is a need to have 2 different?

discovery-infra/test_infra/consts/consts.py
pkliczewski: the name you proposed sounds good

discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
pkliczewski: ATM it is only this use case but in the future someone could change it. For clarity I suggest to use 2.

discovery-infra/test_infra/consts/consts.py
lalon4: done

discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
lalon4: sure, done

discovery-infra/test_infra/consts/consts.py
```diff
@@ -9,6 +9,8 @@
 IMAGE_NAME = "installer-image.iso"
 STORAGE_PATH = "/var/lib/libvirt/openshift-images"
 SSH_KEY = "ssh_key/key.pub"
+MASTER_TF_CPU_MODE = "host-passthrough"
```
eliorerz, line 12: `host-passthrough` string is duplicated for master and worker cpu mode.
Maybe add `HOST_PASSTHROUGH_CPU_MODE` and set it to both of them? Another option is to leave only `HOST_PASSTHROUGH_CPU_MODE` and set on `conftest`?

lalon4: done

PR #806, opened 2021-05-18
MGMT-6408: Added DynamicNetworkAssets class

RazRegev:
An assets class that stores values based on the current available
resources, in order to allow multiple (but not parallel) installations
while avoiding conflicts.


discovery-infra/test_infra/tools/assets.py
```diff
@@ -47,3 +65,83 @@ class NetworkAssets(Assets):
...
+        self._load_current_allocated_host_ips()
+
+        for ip_network_field in self.IP_NETWORK_FIELDS:
+            ip_network = IPNetwork(getattr(asset, ip_network_field))
+            self._set_next_available_ip_network(ip_network)
+            self._allocate_ip_network(ip_network)
+            setattr(asset, ip_network_field, str(ip_network))
+
+    def _load_current_allocated_host_ips(self):
+        cmd_output = sp.check_output(['hostname', '--all-ip-addresses'])
```
eliorerz, line 104: Change to double quotes

discovery-infra/test_infra/tools/assets.py
```diff
@@ -47,3 +65,83 @@ class NetworkAssets(Assets):
...
+            self._allocate_network_bridge(f'tt{index}')
+            index += 1
+
+        return f'tt{index}'
+
+    def _is_network_bridge_allocated(self, network_bridge: str) -> bool:
+        if network_bridge in self._allocated_bridges:
+            return True
+
+        proc = sp.Popen(['ip', 'addr', 'show', network_bridge], stderr=sp.PIPE)
```
eliorerz, line 142: Change to double quotes

discovery-infra/test_infra/tools/assets.py
```diff
@@ -47,3 +65,83 @@ class NetworkAssets(Assets):
...
     def __init__(self):
         super().__init__(assets_file=consts.TF_NETWORK_POOL_PATH)
+
+
+class DynamicNetworkAssets(NetworkAssets):
+    """ An assets class that stores values based on the current available
+        resources, in order to allow multiple (but not parallel) installations
+        while avoiding conflicts. """
+
+    IP_NETWORK_FIELDS = (
```
eliorerz, line 75: We might need this in consts so that we can use the strings elsewhere in the project.

discovery-infra/test_infra/tools/assets.py
```diff
@@ -1,14 +1,26 @@
...
 
 from munch import Munch
+from netaddr import IPNetwork, IPAddress
+
 from test_infra import consts, utils
 
 
 class Assets:
     ASSETS_LOCKFILE_DEFAULT_PATH = "/tmp"
+    REQUIRED_FIELDS = (
```
eliorerz, line 16: We might need this in consts so that we can use the strings elsewhere in the project.

discovery-infra/test_infra/tools/assets.py
```diff
@@ -47,3 +65,83 @@ class NetworkAssets(Assets):
 
     def __init__(self):
         super().__init__(assets_file=consts.TF_NETWORK_POOL_PATH)
+
+
+class DynamicNetworkAssets(NetworkAssets):
```
eliorerz, line 70: Why adding new class and not editing `NetworkAssets` class?

YuviGold: +1

tsorya:
@RazRegev you are breaking all the idea of assets class.

discovery-infra/test_infra/tools/assets.py
```diff
@@ -1,49 +1,158 @@
...
+
+    def _find_current_allocated_virsh_net_resources(self):
+        for uid in self._virsh_net_resources_uids_generator():
+            net_xml_cmd_output = sp.check_output(["virsh", "net-dumpxml", uid])
+            net_xml = xmltodict.parse(net_xml_cmd_output)
+            self._store_allocated_net_resources_from_net_xml(net_xml)
+
+    @staticmethod
+    def _virsh_net_resources_uids_generator() -> Generator[str, None, None]:
+        cmd_output = sp.check_output(["virsh", "net-list", "--uuid", "--all"])
```
tsorya, line 72: use python libvirt

discovery-infra/test_infra/consts/consts.py
```diff
@@ -47,6 +47,22 @@
 DEFAULT_TEST_INFRA_DOMAIN = f".{CLUSTER_PREFIX}-{DEFAULT_NAMESPACE}.{DEFAULT_BASE_DNS_DOMAIN}"
 TEST_TARGET_INTERFACE = "vnet3"
 SUFFIX_LENGTH = 8
+NODES_DEPLOYMENT_LOCK_FILE = "/tmp/nodes_deployment.lock"
+
+REQUIRED_ASSET_FIELDS = (
```
eliorerz, line 52: That not what I meant.
You are writing `machine_cidr` `machine_cidr6` and more for both `REQUIRED_ASSET_FIELDS` and `IP_NETWORK_ASSET_FIELDS` (and probably other places on test infra)
We might want them to be on class/module and create `REQUIRED_ASSET_FIELDS` using that class/module.
Later on we will be able to use those consts on all over the project.

discovery-infra/tests/base_test.py
```diff
@@ -34,33 +36,34 @@ def get_nodes(self) -> Callable:
...
+        def get_nodes_func(config: Optional[TerraformConfig] = None):
+            if not config:
+                config = TerraformConfig()
+
             if "nodes" in nodes_data:
                 return nodes_data["nodes"]
 
-            net_asset = NetworkAssets()
-            config.net_asset = net_asset.get()
+            with file_lock_context(consts.NODES_DEPLOYMENT_LOCK_FILE):
```
tsorya, line 46: filelock should be part of asset and not base test as it will allow to use asset in other places if needsed

discovery-infra/test_infra/tools/assets.py
```diff
@@ -1,49 +1,158 @@
...
-            with open(self._assets_file) as _file:
-                all_assets = json.load(_file)
-            all_assets.extend([Munch.toDict(asset) for asset in self._took_assets])
-            with open(self._assets_file, "w") as _file:
-                json.dump(all_assets, _file)
-
-    def release_all(self) -> None:
-        logging.info("Returning all %d assets", len(self._took_assets))
-        self._release()
+        with open(self._assets_file) as _file:
```
tsorya, line 23: not sure i understand we need asset file in this case?

discovery-infra/test_infra/tools/assets.py
```diff
@@ -1,49 +1,158 @@
...
+    """ An assets class that stores values based on the current available
+        resources, in order to allow multiple installations while avoiding
+        conflicts. """
+
     def __init__(self):
         super().__init__(assets_file=consts.TF_NETWORK_POOL_PATH)
+        self._allocated_ips_objects = []
+        self._allocated_bridges = []
+
+    def get(self) -> Munch:
```
tsorya, line 44: this logic should have lock

discovery-infra/test_infra/tools/assets.py
```diff
@@ -1,49 +1,158 @@
...
+        return asset
+
+    @staticmethod
+    def _verify_asset_fields(asset: Munch):
+        for field in consts.REQUIRED_ASSET_FIELDS:
+            assert hasattr(asset, field), f"missing field {field}"
+
+    def _find_current_allocated_virsh_net_resources(self):
+        for uid in self._virsh_net_resources_uids_generator():
+            net_xml_cmd_output = sp.check_output(["virsh", "net-dumpxml", uid])
```
tsorya, line 66: use python libvirt

discovery-infra/tests/base_test.py
eliorerz, line 46: Why not inside the Assets object?
Every time someone will use it he needs to remember to lock the file

discovery-infra/test_infra/tools/assets.py
```diff
@@ -1,49 +1,158 @@
...
+        # move asset to the end of the list
+        asset_data = all_assets.pop(0)
+        all_assets.append(asset_data)
+        with open(self._assets_file, "w") as _file:
+            json.dump(all_assets, _file)
+        logging.info("Taken asset: %s", asset_data)
+        return Munch.fromDict(asset_data)
 
 
 class NetworkAssets(Assets):
```
tsorya, line 34: maybe LibvirtNetworkAssets ?

discovery-infra/tests/base_test.py
```diff
@@ -34,33 +36,34 @@ def get_nodes(self) -> Callable:
...
 
         _nodes: Nodes = nodes_data.get("nodes")
-        try:
-            if EnvConfig.get("test_teardown"):
-                logging.info('--- TEARDOWN --- node controller\n')
-                _nodes.destroy_all_nodes()
-            if _nodes.is_nat_active():
-                _nodes.unconfigure_nat()
-        finally:
-            _nodes.release_net_asset()
```
eliorerz, line 63: We don't need to release it?

discovery-infra/test_infra/tools/assets.py
```diff
@@ -1,49 +1,158 @@
...
 
-    def __init__(self, assets_file: str, lock_file: str = None):
+    def __init__(self, assets_file: str):
         self._assets_file: str = assets_file
-        self._took_assets: List[Munch] = list()
-        self._lock_file: str = lock_file or os.path.join(
-            self.ASSETS_LOCKFILE_DEFAULT_PATH, os.path.basename(assets_file) + ".lock"
-        )
 
     def get(self) -> Munch:
```
tsorya, line 21: i think you can override thode functions and not to delete them?

discovery-infra/test_infra/tools/assets.py
```diff
@@ -1,49 +1,158 @@
...
     def __init__(self):
         super().__init__(assets_file=consts.TF_NETWORK_POOL_PATH)
+        self._allocated_ips_objects = []
+        self._allocated_bridges = []
+
+    def get(self) -> Munch:
+        asset = super().get()
+        self._verify_asset_fields(asset)
+
+        self._allocated_bridges.clear()
```
tsorya, line 48: not sure i understand why we need to clean it every time we ran get?

discovery-infra/test_infra/tools/assets.py
```diff
@@ -1,49 +1,158 @@
...
+            pass
+
+        try:
+            start_ip = net_xml["network"]["ip"]["dhcp"]["range"]["@start"]
+            end_ip = net_xml["network"]["ip"]["dhcp"]["range"]["@end"]
+            self._add_allocated_ip(IPRange(start_ip, end_ip))
+        except KeyError:
+            pass
+
+    def _override_ip_networks_values_if_not_free(self, asset: Munch):
```
tsorya, line 94: why to override ? you should just find one that is free

discovery-infra/test_infra/tools/assets.py
```diff
@@ -1,49 +1,158 @@
...
+        self._allocated_bridges = []
+
+    def get(self) -> Munch:
+        asset = super().get()
+        self._verify_asset_fields(asset)
+
+        self._allocated_bridges.clear()
+        self._allocated_ips_objects.clear()
+
+        self._find_current_allocated_virsh_net_resources()
```
tsorya, line 51: i think name is not really descriptive. 
It should describe that this function is going to fill _allocated_ips_objects and _allocated_bridges, if i understand right the logic.

tsorya:
@RazRegev it is very cool idea.
but i think we have some logic mismatch here. You are adding dynamic libvirt resource assets but still save previous file base logic.
It causes some strange logic. I think you should just move to dynamic asset creation without preserving previous logic. 
Just take some json as starting point generate resource accordingly. 
All the asset generating code should be locked inside asset itself and not inside base test.

discovery-infra/test_infra/tools/assets.py
RazRegev: because allocated resources may be freed or newly allocated resources will be discovered until the next check.
I don't think we need to count on previous check.

discovery-infra/test_infra/tools/assets.py
```diff
@@ -1,49 +1,158 @@
...
+        resources, in order to allow multiple installations while avoiding
+        conflicts. """
+
     def __init__(self):
         super().__init__(assets_file=consts.TF_NETWORK_POOL_PATH)
+        self._allocated_ips_objects = []
+        self._allocated_bridges = []
+
+    def get(self) -> Munch:
+        asset = super().get()
```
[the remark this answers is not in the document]
RazRegev: the file holds the default base values. can be replaced by env as well, not sure if needed.

discovery-infra/test_infra/tools/assets.py
RazRegev: yes and then override the current value of the allocated resource, no?

discovery-infra/tests/base_test.py
RazRegev: I must have a lock here. If two NetAssets are created simultaneously they can try to allocate the same resources.

RazRegev: there should be also a file lock in the base test since the parallel test will fail in case two processes allocate the same resources at the same time.

discovery-infra/test_infra/tools/assets.py
tsorya: but then what's the point to have this list as class object? just return value from function

discovery-infra/test_infra/tools/assets.py
tsorya: that's the main issue i was talking about. I think that idea is to find free, so you start from some starting point and search for free one, i mean always. In this case logically you are not overriding something but you are searching for free one

discovery-infra/test_infra/tools/assets.py
```diff
@@ -44,6 +50,110 @@ def release_all(self) -> None:
...
+        logging.info("IPs in use: %s", self._allocated_ips_objects)
+
+        for ip_network_field in consts.IP_NETWORK_ASSET_FIELDS:
+            ip_network = IPNetwork(getattr(asset, ip_network_field))
+            self._set_next_available_ip_network(ip_network)
+            self._add_allocated_ip(ip_network)
+            setattr(asset, ip_network_field, str(ip_network))
+
+    def _find_current_allocated_host_ips(self):
+        cmd_output = sp.check_output(["hostname", "--all-ip-addresses"])
```
tsorya, line 107: don't python have something like tihs?

discovery-infra/test_infra/tools/assets.py
```diff
@@ -44,6 +50,110 @@ def release_all(self) -> None:
...
+
+        if self._is_net_bridge_allocated(asset.libvirt_secondary_network_if):
+            net_bridge = self._get_next_available_net_bridge(is_secondary=True)
+            asset.libvirt_secondary_network_if = net_bridge
+
+    def _get_next_available_net_bridge(
+        self,
+        *,
+        is_secondary: bool = False
+    ) -> str:
```
tsorya, line 140: partially you set return value from the function and partially not. Please decide which syntax you want to stick to it

discovery-infra/tests/base_test.py
tsorya: it should be handled by lock inside assets file.

discovery-infra/test_infra/tools/assets.py
```diff
@@ -44,6 +50,110 @@ def release_all(self) -> None:
...
+    """ An assets class that stores values based on the current available
+        resources, in order to allow multiple installations while avoiding
+        conflicts. """
+
     def __init__(self):
         super().__init__(assets_file=consts.TF_NETWORK_POOL_PATH)
+        self._allocated_ips_objects = []
+        self._allocated_bridges = []
+
+    def get(self) -> Munch:
```
tsorya, line 63: i still think that this function must run with filelock

discovery-infra/test_infra/tools/assets.py
```diff
@@ -44,6 +50,110 @@ def release_all(self) -> None:
...
+        asset = deepcopy(super().get())
```
tsorya, line 64: from my point of view you don't need this, you need json object with starting points. Not sure what it gives to you to use asset class here? there is no point to release object even as it is released by removing vms.
I think using here assets.get is making this more complicated

discovery-infra/test_infra/tools/assets.py
```diff
@@ -44,6 +50,110 @@ def release_all(self) -> None:
...
+        self._verify_asset_fields(asset)
```
tsorya, line 65: i can you move to use json that will be consts you will not need this function

discovery-infra/test_infra/tools/assets.py
RazRegev: what will the filelock gives here?

discovery-infra/test_infra/tools/assets.py
RazRegev: I actually not set return values only when it is None. 
Will add it explicitly.

discovery-infra/tests/base_test.py
RazRegev: but then we are missing the whole point of this PR.

RazRegev: Two assets from two processes can allocate the same resources even with the internal lock.
It looks on the current free resources and until the nodes are not up the same resources will be free.

RazRegev: I will change it to write the allocated resources to a shared file.

discovery-infra/test_infra/tools/assets.py
tsorya: it will block any other get action. Just use const filename and it will block same thing that you are trying to block in base test

discovery-infra/test_infra/tools/assets.py
tsorya: as for me it looks a little bit strange but i can live with it.

discovery-infra/test_infra/tools/assets.py
```diff
@@ -44,6 +50,110 @@ def release_all(self) -> None:
...
+            self._add_allocated_net_bridge(f"{prefix}{index}")
+            index += 1
+
+        return f"{prefix}{index}"
+
+    def _is_net_bridge_allocated(self, net_bridge: str) -> bool:
+        if net_bridge in self._allocated_bridges:
+            return True
+
+        proc = sp.Popen(["ip", "addr", "show", net_bridge], stderr=sp.PIPE)
```
tsorya, line 154: i think all this logic can be done with netifaces module, no?

discovery-infra/test_infra/tools/assets.py
RazRegev: not the same thing, prepare nodes is the part that actually allocates the resources, but I changed the logic so it is not relevant anymore.

discovery-infra/test_infra/consts/consts.py
```diff
@@ -158,3 +170,12 @@ class Platforms:
 class HighAvailabilityMode:
     FULL = 'Full'
     NONE = 'None'
+
+
+class BaseAsset:
+    MACHINE_CIDR = "192.168.127.0/24"
```
tsorya, line 176: why 127? i think our default is 126 no?

discovery-infra/test_infra/consts/consts.py
```diff
@@ -158,3 +170,12 @@ class Platforms:
...
+    MACHINE_CIDR6 = "1001:db9::/120"
+    PROVISIONING_CIDR = "192.168.145.0/24"
```
tsorya, line 178: 144?

discovery-infra/test_infra/consts/consts.py
RazRegev: to avoid conflicts in case someone runs discovery flow and then tests (or in parallel)

discovery-infra/test_infra/consts/consts.py
RazRegev: same reason

discovery-infra/test_infra/consts/consts.py
tsorya: no one will run discovery flow and tests in parallel, it is not supported.
In any other case we should not fail. Lets have same starting point as before

YuviGold (approved):

discovery-infra/test_infra/consts/consts.py
RazRegev: https://github.com/openshift/assisted-test-infra/pull/806#discussion_r648935526

RazRegev: see run operator - we deploy operator and then run tests against it

PR #813, opened 2021-05-19
MGMT-6521 - Create operators utils module

eliorerz:


discovery-infra/test_infra/utils/olm_operators_utils.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def get_env(env, default=None):
+    res = os.environ.get(env, "").strip()
+    if not res or res == '""':
+        res = default
+    return res
+
+
+class OlmOperatorsUtils:
```
YuviGold, line 19: Those functions aren't only OLM related

YuviGold, line 19: I don't see any reason it should be a class.

discovery-infra/test_infra/utils/utils.py
```diff
@@ -930,11 +908,12 @@ def download_iso(image_url, image_path):
 
 
 def parse_olm_operators_from_env():
```
YuviGold, line 910: You should check, but I think most of the functions here are internals, and we don't need to keep the old function. Just move their references in case they are in this repository.

eliorerz: `start_discovery` and ``conftest` are using this method.

YuviGold: So change the calling functions as well

eliorerz: Soon `start_discovery`, `conftest` and other scripts will be deprecated (or deleted) so changing those scripts is a bit duplicate job. Also some of those function are used by the QE and I don't want to be blocked by them (or scan their project for every change)

PR #818, opened 2021-05-20
fix events.html fetching events JSON

osherdp:
In the case of a path in the server that contains underscore, we might miscalculate the UUID of the cluster.
So this logic should first extract the directory name of the UUID, and afterwards parse the directory name.
/cc @YuviGold @rollandf 
/hold


YuviGold (approved):

osherdp:
(we need openshift/assisted-service#1770 to fix this test)

PR #820, opened 2021-05-23
NO-ISSUE define REST client after gathering sosreport

osherdp:
On kube-api we cannot create the REST client, which makes a too early failure before gathering sosreport.
/cc @YuviGold @filanov 
/hold


YuviGold (approved):
LGTM

PR #821, opened 2021-05-23
NO-ISSUE add a label to the infraenv to match labels of nmstate-config

osherdp:
It seems like we don't label infraenv with the cluster-name (as we do with nmstate-config), making the cluster not configuring static IPs.
/cc @YuviGold @RazRegev @filanov 
/hold


YuviGold (approved):
LGTM

PR #823, opened 2021-05-23
MGMT-6587 Include disk_index as part of disk name on format_disk

YuviGold:
On https://github.com/openshift/assisted-test-infra/pull/772 disks names
were changed in order to support creation of multiple disks.
Now instead of the disk name to be the same as the node name
<cluster_name>-<node_role>-<node_index>, it is now
<cluster_name>-<node_role>-<node_index>-disk-<disk_index>

For now, format_disk will receive the disk index as a parameter.

cc @lalon4 @patrickdillon @tsorya  @omertuc

### DIFF
--- a/discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
+++ b/discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
@@ -197,9 +197,9 @@ class TerraformController(LibvirtController):
         else:
             return super().start_all_nodes()
 
-    def format_node_disk(self, node_name):
+    def format_node_disk(self, node_name: str, disk_index: int = 0):
         logging.info("Formating disk for %s", node_name)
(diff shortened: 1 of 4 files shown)


discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
```diff
@@ -197,9 +197,9 @@ def start_all_nodes(self):
         else:
             return super().start_all_nodes()
 
-    def format_node_disk(self, node_name):
+    def format_node_disk(self, node_name: str, disk_index: int = 0):
         logging.info("Formating disk for %s", node_name)
-        self.format_disk(f'{self.params.libvirt_storage_pool_path}/{self.cluster_name}/{node_name}')
+        self.format_disk(f'{self.params.libvirt_storage_pool_path}/{self.cluster_name}/{node_name}-disk-${disk_index}')
```
lalon4, line 202: without $

PR #824, opened 2021-05-23
NO-ISSUE utility to check if logs are not uploaded

slaviered:
- The utility checks that logs from the input category
     (worker, master, bootstrap) were not uploaded
    
   - This utility is needed for a fix in the log tests
     that currently check this is an implicit way that
         is unreliable


YuviGold:
Please delete the mistakenly commited rpm

PR #832, opened 2021-05-24
MGMT-6534 - Fix PEP8 violations on tools and add type hints annotations

eliorerz:


discovery-infra/test_infra/tools/terraform_utils.py
```diff
@@ -36,12 +36,12 @@ def change_variables(self, variables, refresh=True):
             json.dump(tfvars, _file)
         self.apply(refresh=refresh)
 
-    def get_state(self):
+    def get_state(self) -> None:
```
YuviGold, line 39: This doesn't look right

discovery-infra/test_infra/tools/concurrently.py
```diff
@@ -15,13 +16,17 @@ def _safe_run(job, job_id, done_handler):
             done_handler(job_id)
 
 
-def run_concurrently(jobs, done_handler=None, max_workers=5, timeout=2 ** 31):
+def run_concurrently(
+    jobs: Union[List, Dict, Tuple],
+    done_handler: Callable[[int], None] = None,
+    max_workers: int = 5,
+    timeout: float = 2 ** 31,
+) -> dict:
```
YuviGold, line 24: What `dict`? Please use `Dict[KT, VT]` notation

discovery-infra/test_infra/tools/concurrently.py
```diff
@@ -1,8 +1,9 @@
 import logging
 from concurrent.futures import ThreadPoolExecutor
+from typing import Callable, Dict, List, Tuple, Union
 
 
-def _safe_run(job, job_id, done_handler):
+def _safe_run(job, job_id: int, done_handler: Callable[[int], None]):
```
YuviGold, line 6: Use `Optional`

eliorerz: Where? `Callable[[int], None]` annotate the function type so done handler signature must be `handler(x: int) -> None`

YuviGold (approved):

PR #836, opened 2021-05-25
MGMT-6544 Create API VIP DNS

empovit:
For running conformance tests, API VIP must be accessible by
DNS name 'api.test-infra-cluster-assisted-installer.redhat.com'.

In order for that to work
1. Make sure the libvirt network has a dns-host entry for API,
   (created by Terraform).
2. Update the entry with the correct IP address as soon as it
   becomes known by updating a Terraform variable. This is
   already done in the Python code, so removing the Shell function
   because the logic there is incomplete.
3. Make dnsmasq use the libvirt network's address to handle DNS
   resolution, so that the DNS entry for API can be resolved from
   the host (already implemented by `set_dns()`).

Tested scenarios:
* multinode + IPv4
* SNO + IPv4
* multinode + static network config + IPv4
* SNO + static network config + IPv4
* multinode + IPv6
* SNO + IPv6
* multinode + static network config + IPv6
* SNO + static network config + IPv6

### DIFF
--- a/scripts/assisted_deployment.sh
+++ b/scripts/assisted_deployment.sh
@@ -19,8 +19,6 @@ function set_dns() {
       exit 1
     fi
 
-    update_vips
-
     FILE="/etc/NetworkManager/conf.d/dnsmasq.conf"
     if ! [ -f "${FILE}" ]; then
         echo -e "[main]\ndns=dnsmasq" | sudo tee $FILE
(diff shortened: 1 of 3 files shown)


YuviGold:
/approve
/lgtm

Please merge only in case all modes pass (ipv6, single node)
/hold

PR #847, opened 2021-05-26
MGMT-6548 - Fix kubeconfig is missing for pytest jobs

eliorerz:


YuviGold:
We need to find a solution how the naming of the kubeconfig would fit for openshift-tests expected KUBECONFIG path.

Makefile
```diff
@@ -81,7 +81,7 @@ IMAGE_NAME=test-infra
 IMAGE_REG_NAME=quay.io/itsoiref/$(IMAGE_NAME)
 
 # oc deploy
-KUBECONFIG := $(or $(KUBECONFIG),${HOME}/.kube/config)
+#KUBECONFIG := $(or $(KUBECONFIG),${HOME}/.kube/config)
```
YuviGold, line 84: ?

discovery-infra/tests/conftest.py
```diff
@@ -2,19 +2,48 @@
...
 from test_infra.utils import operators_utils
 
 
-def _get_cluster_name():
-    cluster_name = utils.get_env('CLUSTER_NAME', f'{consts.CLUSTER_PREFIX}')
-    if cluster_name == consts.CLUSTER_PREFIX:
-        cluster_name = cluster_name + '-' + str(uuid.uuid4())[:consts.SUFFIX_LENGTH]
-    return cluster_name
+@dataclass
+class ClusterName:
```
YuviGold, line 16: please rebase after the clustername PR would be merged

discovery-infra/tests/conftest.py
```diff
@@ -2,19 +2,48 @@
...
+        if self.prefix == consts.CLUSTER_PREFIX:
+            return self.prefix + '-' + self.suffix
+        return self.prefix
+
+
+cluster_name = ClusterName()
+
+
+def get_kubeconfig_path():
+    kubeconfig_dir = Path("/tmp/kubeconfig")
```
YuviGold, line 39: why `/tmp` and not `/build` directory like we do in the old workflow?

Makefile
eliorerz: It defined in the code, no need to set it also here

YuviGold (approved):

PR #848, opened 2021-05-26
WIP: MGMT-6675 Hyperthreading support for test-infra

slaviered:


slaviered:
cc: @yevgeny-shnaidman

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -54,6 +54,7 @@ def _update_day2_config(self, api_client: InventoryClient, cluster_id: str):
                 high_availability_mode=day2_cluster.high_availability_mode,
                 olm_operators=day2_cluster.monitored_operators,
                 base_dns_domain=day2_cluster.base_dns_domain,
+                hyperthreading=day2_cluster.hyperthreading,
```
yevgeny-shnaidman, line 57: this is meaningless, since day2 worker receives its configuration from the installed OCP cluster.

discovery-infra/test_infra/helper_classes/kube_helpers/agent_cluster_install.py
```diff
@@ -146,9 +146,32 @@ def _get_spec_dict(
...
 
+        if "hyperthreading" in kwargs:
+            self._set_hyperthreading(spec=spec, mode=kwargs.pop("hyperthreading"))
+
         spec.update(kwargs)
         return spec
 
+    def _set_hyperthreading(self, spec: Dict[str, Any], mode: str) -> None:
+        # if hypethreading is not configured, let the service choose the default setup
+        if not mode:
```
yevgeny-shnaidman, line 157: i think you can remove it, since you check that it is present in line 149

skipper.env
```diff
@@ -79,7 +79,9 @@ OLM_OPERATORS
 KUBE_API
 INSTALLER_KUBECONFIG
 DEBUG
+FOCUS
```
yevgeny-shnaidman, line 82: what does this variable means?

discovery-infra/test_infra/helper_classes/kube_helpers/agent_cluster_install.py
slaviered: No, In line 149 I check that the --hyperthreading option is there but not its value. It is true that now the default value of this option will not be None if we are reaching this point from day1_flow() but since it is going to be removed I left this check to cover all options.

skipper.env
slaviered: It is the variable that allows focusing on a single ginko test. It has nothing to do with HYPERTHRADING but since I am changing the file already it is an important addition to ease developing cycle

discovery-infra/test_infra/helper_classes/cluster.py
slaviered: FIXED

eliorerz (approved):
LGTM

eliorerz (approved):
LGTM

PR #849, opened 2021-05-27
MGMT-6602 Make set_dns run after each installation

empovit:
Setting DNS must be done as part of every `deploy_nodes_with_install`,
this way it will cover all possible installation flows.

If DNS is not set, API VIP of the installed cluster cannot be reached
by its DNS name from the host.

In case of `deploy_nodes`, which doesn't install a cluster automatically
(with installation from the UI / API afterwards), API VIP is still
unknown. Therefore `make set_dns` and changes to dns-host of the
libvirt network must be done manually.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -231,11 +231,11 @@ validate_namespace:
 
 run: validate_namespace deploy_assisted_service deploy_ui
 
-run_full_flow: run deploy_nodes set_dns
+run_full_flow: run deploy_nodes
 
 redeploy_all: destroy run_full_flow
 
-run_full_flow_with_install: run deploy_nodes_with_install set_dns
+run_full_flow_with_install: run deploy_nodes_with_install
 
 redeploy_all_with_install: destroy run_full_flow_with_install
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -318,7 +318,7 @@ _deploy_nodes:
 deploy_nodes_with_install: start_load_balancer
 	bash scripts/utils.sh local_setup_before_deployment $(PLATFORM) $(NAMESPACE) $(OC_FLAG)
 	skipper make $(SKIPPER_PARAMS) _deploy_nodes NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE) $(OC_FLAG)) NAMESPACE=$(NAMESPACE) ADDITIONAL_PARAMS="'-in ${ADDITIONAL_PARAMS}'" $(SKIPPER_PARAMS) DAY1_PARAMS=--day1-cluster
-
+	make set_dns
```
YuviGold, line 321:
```suggestion
	$(MAKE) set_dns
```

Makefile
```diff
@@ -235,7 +235,7 @@ run_full_flow: run deploy_nodes set_dns
```
YuviGold, line 235: Please remove from `run_full_flow` as well

empovit: `run_full_flow` invokes `deploy_nodes`, which doesn't have `set_dns`. 

Actually, I'm not sure how it's supposed to work because deploying nodes without installation won't have an API VIP yet. Maybe that's exactly the catch in #836 - there won't be API VIP connectivity after deploying nodes using `make` and then installing them manually from the GUI or API. On the other hand, it didn't work correctly anyway. 

Maybe we can open an enhancement to re-introduce updating dns-host in libvirt from `set_dns`, but this time with a more complex logic that gets API VIP from the cluster.

empovit: For now I added an explanation to the commit message. This issue can be considered a limitation and won't affect automated flows.

YuviGold (approved):

PR #856, opened 2021-05-31
MGMT-6744 Use hostnames from TF if hosts have IPv6

empovit:
It looks like libvirt won't have a hostname in the DHCP lease of a VM
that has IPv6, even if it also has an IPv4 address. In this case we
need to get the hostname from Terraform, like we do with IPv6-only
hosts.

This is in preparation for dual stack support (`IPv4=yes IPv6=yes`).


ori-amizur (approved):
/lgtm

PR #862, opened 2021-06-02
MGMT-5276 No network_conf attribute

michaellevy101:
The network_conf attribute is not part of the base controller but implemented only for the terraform controller
Nodes class shouldn't deal with the networking because it's not related so extracting this code to the BaseTest class

### DIFF
--- a/discovery-infra/tests/base_test.py
+++ b/discovery-infra/tests/base_test.py
@@ -11,10 +11,10 @@ import pytest
 import test_infra.utils as infra_utils
 import waiting
 from assisted_service_client.rest import ApiException
+
 from download_logs import download_logs
 from junit_report import JunitFixtureTestCase, JunitTestCase
 from paramiko import SSHException
-from test_infra import consts
 from tests.config import ClusterConfig, EnvConfig
 from test_infra.controllers.proxy_controller.proxy_controller import ProxyController
 from test_infra.helper_classes.cluster import Cluster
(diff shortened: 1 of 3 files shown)


discovery-infra/tests/base_test.py
```diff
@@ -40,27 +42,44 @@ def get_nodes_func(config: TerraformConfig = TerraformConfig()):
...
+
         try:
             if EnvConfig.get("test_teardown"):
                 logging.info('--- TEARDOWN --- node controller\n')
                 _nodes.destroy_all_nodes()
-            if _nodes.is_nat_active():
-                _nodes.unconfigure_nat()
+
+            if _config.platform == consts.Platforms.NONE:
+                if _config.net_asset:
```
eliorerz: Duplicate

discovery-infra/tests/base_test.py
```diff
@@ -40,27 +42,44 @@ def get_nodes_func(config: TerraformConfig = TerraformConfig()):
...
+                    _config.net_asset.release_all()
+
+                if _nat:
```
eliorerz, line 69: Always `remove_nat_rules` not only on None-Platform

discovery-infra/tests/base_test.py
```diff
@@ -40,27 +42,44 @@ def get_nodes_func(config: TerraformConfig = TerraformConfig()):
...
+                    _nat.remove_nat_rules()
         finally:
-            _nodes.release_net_asset()
+            if _config.net_asset:
+                _config.net_asset.release_all()
+
+    @classmethod
+    def nat_interfaces(cls, config: TerraformConfig):
+        return (config.net_asset.libvirt_network_if,
```
eliorerz, line 77: Remove `(`

PR #863, opened 2021-06-02
OCPBUGSM-26780 api_vip_dns_name resolves to incorrect IP leading to API VIP connection validation failure

ori-amizur:
When VIP allocation is done using DHCP, the API VIP DNS has to be published after it is allocated.

This flow is irrelevant for non DHCP allocation since the VIPs are static when this flow is used
so they are set before running terraform the first time.

/cc @eliorerz

### DIFF
--- a/discovery-infra/test_infra/helper_classes/cluster.py
+++ b/discovery-infra/test_infra/helper_classes/cluster.py
@@ -923,6 +923,11 @@ class Cluster:
     def get_worker_ips(client, cluster_id, network):
         return Cluster.get_ips_for_role(client, cluster_id, network, consts.NodeRoles.WORKER)
 
+    @staticmethod
+    def get_vips_from_cluster(client, cluster_id):
+        cluster_info = client.cluster_get(cluster_id)
+        return dict(api_vip=cluster_info.api_vip, ingress_vip=cluster_info.ingress_vip)
+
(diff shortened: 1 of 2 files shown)


discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -923,6 +923,14 @@ def get_master_ips(client, cluster_id, network):
     def get_worker_ips(client, cluster_id, network):
         return Cluster.get_ips_for_role(client, cluster_id, network, consts.NodeRoles.WORKER)
 
+    @staticmethod
+    def get_vips_from_cluster(client, cluster_id):
+        ret = dict()
+        cluster_info = client.cluster_get(cluster_id).to_dict()
+        ret["api_vip"] = cluster_info["api_vip"]
+        ret["ingress_vip"] = cluster_info["ingress_vip"]
+        return ret
```
eliorerz, line 932:
```suggestion
    def get_vips_from_cluster(client, cluster_id):
        cluster_info = client.cluster_get(cluster_id)
        return dict(api_vip=cluster_info.api_vip, ingress_vip=cluster_info.ingress_vip)
```

ori-amizur: OK

eliorerz (approved):
LGTM

PR #866, opened 2021-06-02
MGMT-5276 No network_conf attribute MGMT-6805 None platform test has failed 

michaellevy101:
The network_conf attribute is not part of the base controller but implemented only for the terraform controller
Nodes class shouldn't deal with the networking because it's not related so extracting this code to the BaseTest class


YuviGold:
/lgtm

/test system-test
/test e2e-metal-assisted-none
/test e2e-metal-assisted-none-ipv6

Unhold only if pass
/hold

eliorerz (approved):
LGTM

YuviGold:
```
waiting.exceptions.TimeoutExpired: Timeout of 120 seconds expired waiting for Waiting for load balancer to be active
```

Now with https://github.com/openshift/release/pull/19198 we should be able to read podman logs
/retest

YuviGold (approved):

YuviGold:
e2e-metal-assisted-none job is passing
/unhold

eliorerz (approved):
LGTM

PR #877, opened 2021-06-09
NO-ISSUE: Default downloading all events will include only user events

oshercc:


discovery-infra/test_infra/assisted_service_api.py
```diff
@@ -306,19 +306,19 @@ def get_events(self, cluster_id: str, host_id: Optional[str] = "") -> dict:
 
         return json.loads(response.data)
 
-    def get_all_events(self, cluster_id: str, host_id: Optional[str] = "") -> dict:
```
YuviGold, line 309: No need to `get_all_events` anymore. Use only `get_events` with provided `categories` as a parameter

YuviGold (approved):
LGTM

PR #885, opened 2021-06-14
Update for missing events on ES DB

oshercc:


discovery-infra/log_scrap.py
```diff
@@ -103,6 +104,33 @@ def elastefy_events(self, cluster, event_list):
         cluster_bash_data = process_metadata(metadata_json)
         event_names = get_cluster_object_names(cluster_bash_data)
 
+        self.process_and_log_events(cluster_bash_data, event_list, event_names)
+
+
+        if self.dose_cluster_needs_full_update(cluster_id, event_list):
```
gamli75, line 110: please change it to does_cluster_needs_full_update

discovery-infra/log_scrap.py
```diff
@@ -103,6 +104,33 @@ def elastefy_events(self, cluster, event_list):
...
+            log.info(f"Cluster {cluster_id} logged events are not same as the event count, logging all clusters events")
```
gamli75, line 111: please add the number of events that are missing to the log

discovery-infra/log_scrap.py
```diff
@@ -103,6 +104,33 @@ def elastefy_events(self, cluster, event_list):
...
+            return False
+        else:
+            cluster_events_count_from_db = self.get_cluster_event_count_on_db(cluster_id)
+            self.cache_event_count_per_cluster[cluster_id] = cluster_events_count_from_db
+        if cluster_events_count_from_db < relevant_event_count:
+            return True
+        else:
+            return False
+
+    def get_cluster_event_count_on_db(self, cluster_id):
```
gamli75, line 129: this method is used for getting the events count from elastic DB, right? If the answer is yes - please change the method name to get_cluster_event_count_from_elastic_db - it's confusing.

discovery-infra/log_scrap.py
```diff
@@ -103,6 +104,33 @@ def elastefy_events(self, cluster, event_list):
...
+
+    def dose_cluster_needs_full_update(self, cluster_id, event_list):
+        # check if cluster is missing past events
+        cluster_events_count = self.cache_event_count_per_cluster.get(cluster_id, None)
+        relevant_event_count = len([event for event in event_list if not process.is_event_skippable(event)])
+
+        if cluster_events_count and cluster_events_count == relevant_event_count:
+            return False
+        else:
+            cluster_events_count_from_db = self.get_cluster_event_count_on_db(cluster_id)
```
gamli75, line 122: Do we know how much time this takes? What will happen if we need to do it for all clusters in DB?

oshercc: can vary, 
if we need it for all clusters that's what we need to do

discovery-infra/log_scrap.py
oshercc: why?

oshercc: added on 126

gamli75 (approved):

PR #887, opened 2021-06-14
Add assisted service DNS record for libvirt networks to be resolved to SNO ip or ingress VIP

YuviGold:
/cc @tsorya  @empovit

### DIFF
--- a/discovery-infra/install_cluster.py
+++ b/discovery-infra/install_cluster.py
@@ -197,8 +197,8 @@ def update_vip_from_tf(client, cluster_id, tf, pull_secret):
         cluster_info.api_vip = helper_cluster.get_api_vip_from_cluster(
             client, cluster_info, pull_secret)
 
-    log.info("Setting new vip=%s", cluster_info.api_vip)
-    tf.set_new_vip(cluster_info.api_vip)
+    log.info(f"Setting new API VIP {cluster_info.api_vip} and Ingress VIP {cluster_info.ingress_vip}")
+    tf.set_new_vips(api_vip=cluster_info.api_vip, ingress_vip=cluster_info.ingress_vip)
 
 
 def run_installation_flow_kube_api(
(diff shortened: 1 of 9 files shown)


discovery-infra/install_cluster.py
```diff
@@ -197,8 +197,8 @@ def update_vip_from_tf(client, cluster_id, tf, pull_secret):
         cluster_info.api_vip = helper_cluster.get_api_vip_from_cluster(
             client, cluster_info, pull_secret)
 
-    log.info("Setting new vip=%s", cluster_info.api_vip)
-    tf.set_new_vip(cluster_info.api_vip)
+    log.info(f"Setting new API VIP {cluster_info.api_vip} and Ingress VIP {cluster_info.ingress_vip}")
+    tf.set_new_vip(cluster_info.api_vip, cluster_info.ingress_vip)
```
empovit, line 201: I'd want to be more explicit about which VIPs are handled, especially that they can be easily confused. What about something like`tf.set_new_vips(api_vip=..., ingress_vip=...)`?

discovery-infra/test_infra/tools/terraform_utils.py
```diff
@@ -40,8 +40,8 @@ def get_state(self) -> str:
         self.tf.read_state_file(self.STATE_FILE)
         return self.tf.tfstate
 
-    def set_new_vip(self, api_vip: str) -> None:
-        self.change_variables(variables={"api_vip": api_vip}, refresh=True)
+    def set_new_vip(self, api_vip: str, ingress_vip: str) -> None:
```
tsorya, line 43: maybe we should have function that will set list of dns_names to ip? 
i mean we will not add those values to tf in advance.
Sort of dynamic tf update?

YuviGold: Actually, I already introduced dns_records field on https://github.com/openshift/assisted-test-infra/pull/791
This is currently not used. I guess we should take advantage of it.

YuviGold: I opened a ticket since it's a bigger task
https://issues.redhat.com/browse/MGMT-6928

PR #888, opened 2021-06-15
NO-ISSUE - Fetch all masters from primary and secondary networks when IPv6 is set with None platform.

eliorerz:
/cc @tsorya @YuviGold

### DIFF
--- a/discovery-infra/test_infra/helper_classes/nodes.py
+++ b/discovery-infra/test_infra/helper_classes/nodes.py
@@ -167,7 +167,9 @@ class Nodes:
             # set to valid hostname
             # TODO - NodeController has no `params` and `tf` attributes
             network_name = self.controller.params.libvirt_network_name
-            libvirt_nodes = utils.get_libvirt_nodes_from_tf_state(network_name, self.controller.tf.get_state())
+            secondary_network_name = self.controller.params.libvirt_secondary_network_name
+            libvirt_nodes = utils.get_libvirt_nodes_from_tf_state([network_name, secondary_network_name],
+                                                                  self.controller.tf.get_state())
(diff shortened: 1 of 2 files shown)


eliorerz:
@YuviGold e2e-metal-assisted-none-ipv6 is not supported at the moment.
/override e2e-metal-assisted-none-ipv6

discovery-infra/test_infra/helper_classes/nodes.py
```diff
@@ -167,7 +167,10 @@ def set_hostnames(self, cluster, nodes_count: int, is_ipv6: bool, is_static_ip:
             # set to valid hostname
             # TODO - NodeController has no `params` and `tf` attributes
             network_name = self.controller.params.libvirt_network_name
+            secondary_network_name = self.controller.params.libvirt_secondary_network_name
             libvirt_nodes = utils.get_libvirt_nodes_from_tf_state(network_name, self.controller.tf.get_state())
+            libvirt_nodes.update(utils.get_libvirt_nodes_from_tf_state(secondary_network_name,
```
ori-amizur, line 172: Seems there is a mismatch. _get_libvirt_nodes_from_tf_state_ receives network_names collection as first parameter, and here we pass a single network.

ori-amizur (changes requested):

discovery-infra/test_infra/helper_classes/nodes.py
eliorerz: @ori-amizur You right, didn't noticed it. The original implementation expect it to be an iterable object but here we pass string (luckily strings in python are iterables so it worked so far).

PR #893, opened 2021-06-17
NO-ISSUE Always add hosts due to hosts being removed from cluster lis…

oshercc:
…t payload

### DIFF
--- a/discovery-infra/download_logs.py
+++ b/discovery-infra/download_logs.py
@@ -117,8 +117,7 @@ def is_update_needed(output_folder: str, update_on_events_update: bool, client: 
 def download_logs(client: InventoryClient, cluster: dict, dest: str, must_gather: bool,
                   update_by_events: bool = False, retry_interval: int = RETRY_INTERVAL, pull_secret=""):
 
-    # TODO remove when cluster_list has param to include download hosts
-    if "hosts" not in cluster:
+    if "hosts" not in cluster or len(cluster["hosts"]) == 0:
         cluster["hosts"] = client.get_cluster_hosts(cluster_id=cluster["id"])
 
     output_folder = get_logs_output_folder(dest, cluster)


gamli75 (approved):

discovery-infra/download_logs.py
```diff
@@ -117,9 +117,8 @@ def is_update_needed(output_folder: str, update_on_events_update: bool, client:
 def download_logs(client: InventoryClient, cluster: dict, dest: str, must_gather: bool,
                   update_by_events: bool = False, retry_interval: int = RETRY_INTERVAL, pull_secret=""):
 
-    # TODO remove when cluster_list has param to include download hosts
-    if "hosts" not in cluster:
-        cluster["hosts"] = client.get_cluster_hosts(cluster_id=cluster["id"])
+    # TODO remove when cluster_list has param to include download hosts, task MGMT-6961
+    cluster["hosts"] = client.get_cluster_hosts(cluster_id=cluster["id"])
```
ronniel1, line 121: Please  do this only under `if "hosts" not in cluster or len(cluster["hosts"]) = 0:`
Then this will also work then we do get the hosts in the get_cluster reply

discovery-infra/download_logs.py
```diff
@@ -117,8 +117,8 @@ def is_update_needed(output_folder: str, update_on_events_update: bool, client:
 def download_logs(client: InventoryClient, cluster: dict, dest: str, must_gather: bool,
                   update_by_events: bool = False, retry_interval: int = RETRY_INTERVAL, pull_secret=""):
 
-    # TODO remove when cluster_list has param to include download hosts
-    if "hosts" not in cluster:
+    # TODO remove when cluster_list has param to include download hosts, task MGMT-6961
```
ronniel1, line 120: Lets remove this todo.
I think  that with the current code, it's irrelevant

ronniel1 (approved):
/lgtm

PR #897, opened 2021-06-20
[WIP] MGMT-6417- Deprecate deploy_nodes Makefile target

eliorerz:
- This PR is a POC for replacing the old Makefile targets with a new logic.
- Basically, the targets will no longer use the scripts under `discovery-infra` and for each targets group there will be module/class.
- Some of the Makefile logic will be moved to a readable python code instead of Makefile targets.
- Using classes rather than modules allow us to use inheritance for creating new targets.
- As part of this example, factory module is created, the goal is to create a base shared code for the targets and the tests(`BaseTest`) packages
- In this implementation, each target module is inherits from `Target` class that define all basic arguments and parameters that will probably be used an every other target.
- Here, 2 targets were created: `DeployNodes` and `DeployNodesWithInstall`, all env vars are handled within the `Target` class and part of the shared code created in previous tasks (Therefore, there is no need for an infinite number of variables).

Note that the code in factory is partially duplicated with base test code and will be merged to single implementation

/cc @tsorya @osherdp @YuviGold

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -322,19 +322,16 @@ install_cluster:
 #########
 
 _deploy_nodes:
-	discovery-infra/start_discovery.py -i $(ISO) -n $(NUM_MASTERS) -p $(STORAGE_POOL_PATH) -k '$(SSH_PUB_KEY)' -md $(MASTER_DISK) -wd $(WORKER_DISK) -mm $(MASTER_MEMORY) -wm $(WORKER_MEMORY) -nw $(NUM_WORKERS) -ps '$(PULL_SECRET)' -bd $(BASE_DOMAIN) -cN $(CLUSTER_NAME) -vN $(NETWORK_CIDR) -nM $(NETWORK_MTU) -iU $(REMOTE_SERVICE_URL) -id $(CLUSTER_ID) -mD $(BASE_DNS_DOMAINS) -ns $(NAMESPACE) -pX $(HTTP_PROXY_URL) -sX $(HTTPS_PROXY_URL) -nX $(NO_PROXY_VALUES) --service-name $(SERVICE_NAME) --vip-dhcp-allocation $(VIP_DHCP_ALLOCATION) --ns-index $(NAMESPACE_INDEX) --deploy-target $(DEPLOY_TARGET) $(DAY1_PARAMS) $(OC_PARAMS) $(KEEP_ISO_FLAG) $(ADDITIONAL_PARAMS) $(DAY2_PARAMS) -ndw $(NUM_DAY2_WORKERS) --ipv4 $(IPv4) --ipv6 $(IPv6) --platform $(PLATFORM) --proxy $(PROXY) --iso-image-type $(ISO_IMAGE_TYPE) --hyperthreading $(HYPERTHREADING) --kube-api $(KUBE_API)
+	python3 ${DEBUG_FLAGS} discovery-infra/targets/deploy_nodes.py $(ADDITIONAL_PARAMS)
 
 deploy_nodes_with_install: start_load_balancer
-	bash scripts/utils.sh local_setup_before_deployment $(PLATFORM) $(NAMESPACE) $(OC_FLAG)
-	skipper make $(SKIPPER_PARAMS) _deploy_nodes NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE) $(OC_FLAG)) NAMESPACE=$(NAMESPACE) ADDITIONAL_PARAMS="'-in ${ADDITIONAL_PARAMS}'" $(SKIPPER_PARAMS) DAY1_PARAMS=--day1-cluster
-	$(MAKE) set_dns
+	skipper make $(SKIPPER_PARAMS) _deploy_nodes ADDITIONAL_PARAMS="-i"
 
 deploy_static_network_config_nodes_with_install:
 	make deploy_nodes_with_install ADDITIONAL_PARAMS="'--with-static-network-config'"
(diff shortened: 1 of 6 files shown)


discovery-infra/targets/target.py
```diff
@@ -0,0 +1,30 @@
+from abc import ABC, abstractmethod
+
+from test_infra import utils
+from test_infra.assisted_service_api import InventoryClient, ClientFactory
+from test_infra.tools.assets import LibvirtNetworkAssets
+from test_infra.utils.global_variables import GlobalVariables
+from tests.config import TerraformConfig, ClusterConfig
+
+
+def get_api_client(namespace: str, remote_service_url: str, offline_token=None, **kwargs) -> InventoryClient:
```
YuviGold, line 10: Merge with conftest
https://github.com/openshift/assisted-test-infra/blob/28ca2b0b281e10dda55a6849c69cf4c4bfa10a6f/discovery-infra/tests/conftest.py#L99-L106

discovery-infra/targets/deploy_nodes.py
```diff
@@ -0,0 +1,38 @@
+import argparse
```
YuviGold, line 1: Make executable with shabang

Makefile
```diff
@@ -322,19 +322,16 @@ install_cluster:
...
+	python3 ${DEBUG_FLAGS} discovery-infra/targets/deploy_nodes.py $(ADDITIONAL_PARAMS)
 
 deploy_nodes_with_install: start_load_balancer
-	bash scripts/utils.sh local_setup_before_deployment $(PLATFORM) $(NAMESPACE) $(OC_FLAG)
-	skipper make $(SKIPPER_PARAMS) _deploy_nodes NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE) $(OC_FLAG)) NAMESPACE=$(NAMESPACE) ADDITIONAL_PARAMS="'-in ${ADDITIONAL_PARAMS}'" $(SKIPPER_PARAMS) DAY1_PARAMS=--day1-cluster
-	$(MAKE) set_dns
+	skipper make $(SKIPPER_PARAMS) _deploy_nodes ADDITIONAL_PARAMS="-i"
```
osherdp, line 328: maybe use the longer version of ``-i`` option, to have it easier to understand

discovery-infra/targets/deploy_nodes.py
```diff
@@ -0,0 +1,41 @@
...
+                          self._terraform_config.net_asset.libvirt_secondary_network_if)
+        self._nat = NatController(nat_interfaces)
+        self._nodes: Nodes = nodes_factory(self._terraform_config, self._nat)
+        self._cluster: Cluster = cluster_factory(self._nodes, self._api_client, self._cluster_config)
+
+    def run(self):
+        self._cluster.prepare_for_installation()
+
+
+class DeployNodesWithInstall(DeployNodes):
```
osherdp, line 27: I don't know what I feel about those levels of inheritance and about transforming makefile concepts to python classes
I of course agree with the need to simplify our code, just not sure about the proposed solution

discovery-infra/targets/deploy_nodes.py
```diff
@@ -0,0 +1,41 @@
...
+    def run(self):
+        super().run()
+        self._cluster.start_install_and_wait_for_installed()
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description="Deploy nodes")
+    parser.add_argument("-i", "--install", help="Install flag", type=bool, nargs='?', default=False, const=True)
```
osherdp, line 35: please don't write ``Install flag``, I'm not presenting myself as "Osher some kind of human being" :)

discovery-infra/targets/deploy_nodes.py
```diff
@@ -0,0 +1,41 @@
...
+
+    args = parser.parse_args()
+    if args.install:
+        DeployNodesWithInstall().run()
+    else:
+        DeployNodes().run()
```
osherdp, line 41: again, it's hard to my eyes
if using inheritance then at least use polymorphism

osherdp (changes requested):
partial review

discovery-infra/test_infra/factory.py
```diff
@@ -0,0 +1,46 @@
...
+from test_infra.controllers.nat_controller import NatController
+from test_infra.controllers.node_controllers import TerraformController
+from test_infra.controllers.proxy_controller.proxy_controller import ProxyController
+from test_infra.helper_classes.cluster import Cluster
+from test_infra.helper_classes.nodes import Nodes
+from test_infra.tools.assets import LibvirtNetworkAssets
+from tests.config import TerraformConfig, ClusterConfig
+
+
+def nodes_factory(config: TerraformConfig, nat: NatController):
```
osherdp, line 14: I remember us having a similar code, no?
Or maybe it just creates the controller and not Nodes

osherdp (changes requested):
full review

tsorya:
@osherdp @YuviGold @eliorerz to say the truth i thought that the idea is to use 100% same logic as with tests.
I am not sure why we need different logic for running specific target, it is duplication of the test from my point of view.
We can have generic test with support of envs or we can have test per target but i don't think we need another code that we have already in test.  
make deploy_nodes -> make test TEST=deploy_nodes or make test TEST=generic DONT_INSTALL=True ...
What do you think?

eliorerz:
This PR replaced with https://github.com/openshift/assisted-test-infra/pull/990

PR #903, opened 2021-06-21
MGMT-7048: retry if minikube creation failed

osherdp:
Sometimes ``minikube start`` fails.
Let's do a retry several times until it succeed.

/cc @YuviGold 
/hold


osherdp:
failed ``kube-api`` on purpose, let's see if I can retest and make ``minikube start fail``

osherdp:
/unhold
works just fine by testing it locally

PR #904, opened 2021-06-21
MGMT-6806: Enable remote debugging on Beaker hosts

michaellevy101:
1. Patch the image with the debug information
2. Expose debug port in the firewall
3. Expose debug port in the host (xinit) - Support multiple ports recognition
4. Enable registry add-on - a completion to https://github.com/openshift/assisted-service/pull/1611 PR


scripts/deploy_assisted_service.sh
```diff
@@ -112,13 +115,29 @@ EOF
...
 
     add_firewalld_port $SERVICE_PORT
 
     print_log "Starting port forwarding for deployment/${SERVICE_NAME} on port $SERVICE_PORT"
-    wait_for_url_and_run ${SERVICE_BASE_URL} "spawn_port_forwarding_command $SERVICE_NAME $SERVICE_PORT $NAMESPACE $NAMESPACE_INDEX $KUBECONFIG minikube"
+    wait_for_url_and_run ${SERVICE_BASE_URL} "spawn_port_forwarding_command $SERVICE_NAME $SERVICE_PORT $NAMESPACE $NAMESPACE_INDEX $KUBECONFIG minikube undeclared $SERVICE_INTERNAL_PORT"
+
+    if [ "${DEBUG_SERVICE}" == "true" ]; then
+        print_log "Removing liveness Probe to prevent rebooting while debugging"
+        kubectl patch deployment assisted-service -n $NAMESPACE --type json   -p='[{"op": "remove", "path": "/spec/template/spec/containers/0/livenessProbe"}]'
```
YuviGold, line 134: why?

scripts/deploy_assisted_service.sh
```diff
@@ -17,13 +19,14 @@ export ENABLE_KUBE_API=${ENABLE_KUBE_API:-false}
 export ENABLE_KUBE_API_CMD="ENABLE_KUBE_API=${ENABLE_KUBE_API}"
 export OPENSHIFT_VERSIONS=${OPENSHIFT_VERSIONS:-}
 export OPENSHIFT_VERSIONS_CMD=""
+export DEBUG_SERVICE=${DEBUG_SERVICE:-}
```
YuviGold, line 22: I think it would be easier to use the same `DEBUG` as for debugging python

scripts/deploy_assisted_service.sh
```diff
@@ -112,13 +115,29 @@ EOF
...
+    fi
+
     skipper run discovery-infra/update_assisted_service_cm.py
     (cd assisted-service/ && skipper --env-file ../skipper.env run "make deploy-all" ${SKIPPER_PARAMS} $ENABLE_KUBE_API_CMD $OPENSHIFT_VERSIONS_CMD DEPLOY_TAG=${DEPLOY_TAG} DEPLOY_MANIFEST_PATH=${DEPLOY_MANIFEST_PATH} DEPLOY_MANIFEST_TAG=${DEPLOY_MANIFEST_TAG} NAMESPACE=${NAMESPACE} AUTH_TYPE=${AUTH_TYPE})
 
     add_firewalld_port $SERVICE_PORT
 
     print_log "Starting port forwarding for deployment/${SERVICE_NAME} on port $SERVICE_PORT"
-    wait_for_url_and_run ${SERVICE_BASE_URL} "spawn_port_forwarding_command $SERVICE_NAME $SERVICE_PORT $NAMESPACE $NAMESPACE_INDEX $KUBECONFIG minikube"
+    wait_for_url_and_run ${SERVICE_BASE_URL} "spawn_port_forwarding_command $SERVICE_NAME $SERVICE_PORT $NAMESPACE $NAMESPACE_INDEX $KUBECONFIG minikube undeclared $SERVICE_INTERNAL_PORT"
```
YuviGold, line 130: this is changing the behavior without running DEBUG. you meant for it to happen?

scripts/deploy_assisted_service.sh
```diff
@@ -112,13 +115,29 @@ EOF
...
+
+    if [ "${DEBUG_SERVICE}" == "true" ]; then
+        print_log "Removing liveness Probe to prevent rebooting while debugging"
+        kubectl patch deployment assisted-service -n $NAMESPACE --type json   -p='[{"op": "remove", "path": "/spec/template/spec/containers/0/livenessProbe"}]'
+
+        add_firewalld_port $DEBUG_SERVICE_PORT
```
YuviGold, line 136: Please use parenthesis wherever you can :) 
```suggestion
        add_firewalld_port ${DEBUG_SERVICE_PORT}
```

scripts/deploy_assisted_service.sh
```diff
@@ -1,14 +1,16 @@
 #!/usr/bin/env bash
 set -euo pipefail
-
 source scripts/utils.sh
 
 export SERVICE_NAME=assisted-service
+export DEBUG_SERVICE_NAME=assisted-service-debug
```
YuviGold, line 6: Group the debug flags together

scripts/deploy_assisted_service.sh
michaellevy101: When you debug the service it the liveness prob failed and the server reboots so i disabled it

scripts/deploy_assisted_service.sh
michaellevy101: why? I give the control to the developer to define exactly what do debug

scripts/deploy_assisted_service.sh
michaellevy101: This function redirects the service node port to the host port.
For example: The assisted-service service(k8s service) serves on 8090 but its node port is different and determine in runtime. 
To access the service from outside we should forward the minikube service node port to the host(beaker machine) port.
If debug is enabled the k8s has two ports, 8090 for the service itself and 40000 for the debug.
The old behavior assumed that the service has only one port but it's not true any more.
I changed the spawn_port_forwarding_command function to resolve the node port by the service port.
Here i just added the service port to the call.

scripts/deploy_assisted_service.sh
YuviGold: Why does it fail? :)

scripts/deploy_assisted_service.sh
YuviGold: I understand that. I just think that when a contributor comes to a repo he would think about `DEBUG` and not specifically what to debug. The same goes for debugging the python that it's `DEBUG` and not `DEBUG_PYTHON`

scripts/deploy_assisted_service.sh
michaellevy101: don't know... don't see why to investigate it

scripts/deploy_assisted_service.sh
michaellevy101: I am not sure that setting one parameter to debug them all it's a good idea...
IMO the service should be always debuggable (add it to my bashrc or somthing) but not the E2E.

scripts/deploy_assisted_service.sh
YuviGold: We count on the service being healthy. I don't think that ignoring that, even for debugging, is a good idea.

scripts/deploy_assisted_service.sh
YuviGold: Could the e2e debugging be running in the background as well? I guess that's the desirable behavior. Then whoever would want to debug, just turns on the `DEBUG` flag and the entities would be debuggable, no strings attached :)

scripts/deploy_assisted_service.sh
michaellevy101: This is not in production just in your beaker machine and only when debugging.
We have no choice you can't debug with it...

scripts/deploy_assisted_service.sh
michaellevy101: It's possible but not desired, the test might failed until you connected.

YuviGold: Then the developer should put the breakpoint before running :)
Honestly, I think this debug behavior is preferred

scripts/deploy_assisted_service.sh
YuviGold: Would you mind investigating it? I am afraid it happens because of some internal logic we should fix, and we would get ricochets from disabling it in the future. (a developer would debug and see x while in prod it's y)

Otherwise, please at least add a detailed comment about the failure and why it happens.

scripts/deploy_assisted_service.sh
michaellevy101: no, If you want to debug the beginning of the test you never succeed.
I debugged it so many times and i don't agree but we can give it a try and change it in the future.

scripts/deploy_assisted_service.sh
michaellevy101: ok. found the issue:
https://github.com/go-delve/delve/issues/777
Delve stops all the threads on a breakpoint hit

scripts/deploy_assisted_service.sh
YuviGold: After discussion, it seems that python behavior can't be aligned to service behavior.
Breakpoints can't be predefined unless attached, and it takes time to attach to a remote.

So we will keep the 2 flags for flexibility and the behavior by design.

scripts/deploy_assisted_service.sh
michaellevy101: Agreed to keep the probs but change its restartPolicy to Never

YuviGold (approved):

YuviGold (approved):

PR #905, opened 2021-06-21
NO-ISSUE - Rearrange nodes data on get nodes fixture

eliorerz:
Prevent case where `prepare_nodes` is failing and `_nodes` is not defined on teardown


michaellevy101 (commented):
Looks good!!! Please consider deal with the actual failiour

YuviGold (approved):
LGTM

PR #912, opened 2021-06-23
MGMT-7071 - Fix PEP8 violations on utils

eliorerz:


YuviGold:
```
discovery-infra/test_infra/utils/operators_utils.py:11: in get_env
    res = os.environ.get(env, "").strip()
E   NameError: name 'os' is not defined
```

PR #914, opened 2021-06-24
NO-ISSUE Add grace of 1 sec for ES doc to flush and add host the event metadata

oshercc:
* due to race option when checking DB available events vs events to log waiting for all events to flush on the es db 
this will prevent getting a mismatch between the two numbers and save logging time in that case

* add host information to metadata if missing


gamli75 (approved):

gamli75 (approved):

oshercc:
dup of 916

PR #918, opened 2021-06-27
MGMT-6732 Log download decision will be by a fail event and not by s…

oshercc:
…tatus

### DIFF
--- a/discovery-infra/download_logs.py
+++ b/discovery-infra/download_logs.py
@@ -65,7 +65,8 @@ def main():
             return
 
         for cluster in clusters:
-            if args.download_all or should_download_logs(cluster):
+            cluster_events = client.get_events(cluster_id=cluster["id"], categories=["metrics"])
+            if args.download_all or should_download_logs(cluster_events):
                 download_logs(client, cluster, args.dest, args.must_gather, args.update_by_events,
                               pull_secret=args.pull_secret)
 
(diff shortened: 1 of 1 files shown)


discovery-infra/download_logs.py
```diff
@@ -80,9 +81,11 @@ def get_clusters(client, all_cluster):
     return client.clusters_list()
 
 
-def should_download_logs(cluster: dict):
-    return cluster['status'] in [ClusterStatus.ERROR]
-
+def should_download_logs(cluster_events: list):
```
omertuc, line 84: nit: Using `list` directly as a type annotation is a Python 3.9+ feature, for older versions `from typing import List` must be used. Also please specify the list type with `[]`

discovery-infra/download_logs.py
```diff
@@ -80,9 +81,11 @@ def get_clusters(client, all_cluster):
     return client.clusters_list()
 
 
-def should_download_logs(cluster: dict):
-    return cluster['status'] in [ClusterStatus.ERROR]
```
omertuc, line 84: We should probably add a `add_triage_signature.py` signature to detect non-error state cluster and give them a special label so they can be easily filtered in triage dashboards

omertuc: @osherdp

PR #919, opened 2021-06-27
NO-ISSUE Remove AgentLabelSelector from InfraEnv

rollandf:
The `agentLabelSelector` is not part of the InfraEnv Spec anymore.


osherdp:
Removing because scenario in test-infra doesn't need it right now or because it's not needed on any level?

rollandf:
It was removed from the Spec.

osherdp (approved):

PR #920, opened 2021-06-27
NO-ISSUE Log download continuously to prevent missing fails

oshercc:
since the log collection is properly run and scans for failed status we miss clusters that failed and moved from this state. 
if we increase the sampling we will not solve this 100% but we will miss fewer clusters since the sampling will give a closer display of the real state of the cluster and will give them less time to change

### DIFF
--- a/Jenkinsfile.download_logs
+++ b/Jenkinsfile.download_logs
@@ -5,7 +5,7 @@ pipeline {
         string(name: 'REMOTE_SERVICE_URL', defaultValue: 'https://api.openshift.com', description: 'Service URL')
     }
 
-    triggers { cron('H/30 * * * *') }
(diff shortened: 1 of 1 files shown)


omertuc:
Why would running it less often miss fails?

omertuc:
> that failed and moved from this state.

What moves them out of the failed state? User manually re-trying the installation?

osherdp (approved):

osherdp:
/override ci/prow/images

this PR have no connection to those jobs

YuviGold:
I am concerned about spamming... Our Slack channel would be flooded once the job is broken.
/lgtm cancel

YuviGold:
We decided to move to a different solution https://issues.redhat.com/browse/MGMT-7098

PR #928, opened 2021-06-29
MGMT-7115: When running test_parallel, VIPs DNS cannot be resolved from host

ori-amizur:
- Created DNS forwarding file(s) in /etc/NetworkManager/dnsmasq.d by using terraform.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -478,6 +478,7 @@ _test: $(REPORTS) _test_setup
 test_parallel:
 	$(MAKE) start_load_balancer START_LOAD_BALANCER=true
 	skipper make $(SKIPPER_PARAMS) _test_parallel
+	scripts/assisted_deployment.sh set_all_vips_dns
 
 _test_setup:
 	rm -rf /tmp/assisted_test_infra_logs
--- a/discovery-infra/test_infra/helper_classes/cluster.py
+++ b/discovery-infra/test_infra/helper_classes/cluster.py
@@ -7,6 +7,7 @@ import re
 import time
 from collections import Counter
 from typing import List, Union, Optional
+from textwrap import dedent
 
 import requests
 import waiting
@@ -621,6 +622,21 @@ class Cluster:
             timeout=timeout,
         )
 
+    def _set_dns(self):
+        cluster = self.api_client.cluster_get(self.id)
+        cluster_name = cluster.name
+        base_domain = cluster.base_dns_domain
+        main_cidr = self.nodes.controller.get_machine_cidr()
+        nameserver_ip = str(IPNetwork(main_cidr).ip + 1)
+        fname = f"/etc/NetworkManager/dnsmasq.d/openshift-{cluster_name}.conf"
+        contents = dedent(f"""
+            server=/api.{cluster_name}.{base_domain}/{nameserver_ip}
+            server=/.apps.{cluster_name}.{base_domain}/{nameserver_ip}
+            """)
+        self.nodes.controller.tf.change_variables(
+            {"dns_forwarding_file": contents, "dns_forwarding_file_name": fname}
+        )
+
     @JunitTestCase()
     def prepare_for_installation(self, static_network_config=None, **kwargs):
         self.update_config(**kwargs)
@@ -655,6 +671,7 @@ class Cluster:
 
         if self._config.platform == consts.Platforms.NONE:
             self._configure_load_balancer()
+        self._set_dns()
 
     def download_kubeconfig_no_ingress(self, kubeconfig_path: str = None):
         self.api_client.download_kubeconfig_no_ingress(self.id, kubeconfig_path or self._config.kubeconfig_path)
--- a/scripts/assisted_deployment.sh
+++ b/scripts/assisted_deployment.sh
@@ -4,6 +4,13 @@ function destroy_all() {
     make destroy
 }
 
+function update_conf_file() {
+    FILE="/etc/NetworkManager/conf.d/dnsmasq.conf"
+    if ! [ -f "${FILE}" ]; then
+        echo -e "[main]\ndns=dnsmasq" | sudo tee $FILE
+    fi
+}
+
 function set_dns() {
     NAMESPACE_INDEX=${1:-0}
     if [ "${BASE_DNS_DOMAINS}" != '""' ]; then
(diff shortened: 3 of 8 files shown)


discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -621,6 +621,18 @@ def wait_for_install(self, timeout=consts.CLUSTER_INSTALLATION_TIMEOUT):
...
+        cluster = self.api_client.cluster_get(self.id)
+        cluster_name = cluster.name
+        base_domain = cluster.base_dns_domain
+        main_cidr = self.nodes.controller.get_machine_cidr()
+        nameserver_ip = str(IPNetwork(main_cidr).ip + 1)
+        fname = f"/etc/NetworkManager/dnsmasq.d/openshift-{cluster_name}.conf"
+        contents = f"server=/api.{cluster_name}.{base_domain}/{nameserver_ip}\n" + f"server=/.apps.{cluster_name}.{base_domain}/{nameserver_ip}\n"
+        self.nodes.controller.tf.change_variables(
+            {"dns_forwarding_file": contents, "dns_forwarding_file_name": fname}
+        )
```
osherdp, line 634: What's the reason for managing those files in terraform?

osherdp (changes requested):

skipper.yaml
```diff
@@ -35,6 +35,9 @@ volumes:
   - /tmp:/tmp/
   - /dev/:/dev
   - /run/udev:/run/udev
+
+  # Network manager dnsmasq
+  - /etc/NetworkManager/dnsmasq.d:/etc/NetworkManager/dnsmasq.d
```
YuviGold, line 40: Please add a comment on why we're mounting it

discovery-infra/test_infra/helper_classes/cluster.py
YuviGold: +1
Isn't managing it with TF just make it more complex?

Makefile
```diff
@@ -478,6 +478,7 @@ _test: $(REPORTS) _test_setup
 test_parallel:
 	$(MAKE) start_load_balancer START_LOAD_BALANCER=true
 	skipper make $(SKIPPER_PARAMS) _test_parallel
+	scripts/assisted_deployment.sh set_all_vips_dns
```
YuviGold, line 481: why is that needed if `set_dns` is called as part of `preparing_for_installation`?

discovery-infra/test_infra/helper_classes/cluster.py
ori-amizur: Terraform treats the file as resource.  It will be destroyed during teardown.

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -621,6 +621,18 @@ def wait_for_install(self, timeout=consts.CLUSTER_INSTALLATION_TIMEOUT):
...
         )
 
+    def _set_dns(self):
+        cluster = self.api_client.cluster_get(self.id)
+        cluster_name = cluster.name
+        base_domain = cluster.base_dns_domain
+        main_cidr = self.nodes.controller.get_machine_cidr()
+        nameserver_ip = str(IPNetwork(main_cidr).ip + 1)
+        fname = f"/etc/NetworkManager/dnsmasq.d/openshift-{cluster_name}.conf"
+        contents = f"server=/api.{cluster_name}.{base_domain}/{nameserver_ip}\n" + f"server=/.apps.{cluster_name}.{base_domain}/{nameserver_ip}\n"
```
YuviGold, line 631: Using multi-line with `textwrap.dedent` would make it readable.

```suggestion
        from textwrap import dedent
        ....
        
        contents = dedent(f"""
        server=/api.{cluster_name}.{base_domain}/{nameserver_ip}
        server=/.apps.{cluster_name}.{base_domain}/{nameserver_ip}""")
```

Makefile
ori-amizur: To ask NM to reload

discovery-infra/test_infra/helper_classes/cluster.py
ori-amizur: Is it more readable?

scripts/assisted_deployment.sh
```diff
@@ -32,6 +32,16 @@ function set_dns() {
     echo "Finished setting dns"
 }
 
+function set_all_vips_dns() {
+    FILE="/etc/NetworkManager/conf.d/dnsmasq.conf"
+    if ! [ -f "${FILE}" ]; then
+        echo -e "[main]\ndns=dnsmasq" | sudo tee $FILE
+    fi
```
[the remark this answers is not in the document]
ori-amizur: OK. Will refactor

skipper.yaml
ori-amizur: OK

discovery-infra/test_infra/helper_classes/cluster.py
ori-amizur: For cleanup after use.

Makefile
YuviGold: Could we reload inside `_set_dns()` function?

discovery-infra/test_infra/helper_classes/cluster.py
YuviGold: Yes. You don't have one long line with strings concatenation nor new-line backslash.

discovery-infra/test_infra/helper_classes/cluster.py
osherdp: and yet configuration will still hold until doing ``systemctl NetworkManager reload``, no?
I think it will become hard to investigate network problems

Makefile
ori-amizur: No.  It has to be done outside the container

discovery-infra/test_infra/helper_classes/cluster.py
ori-amizur: It will be destroyed only when **_terraform destroy_** is called.
If you decide to leave the cluster alive, the configuration will remain in place.
During reload, if the cluster was destroyed, the operation will not have any effect (as expected).

ori-amizur: I think that leaving these files in this directory (part of the global NM) is incorrect behavior.

osherdp: yes, we might become inconsistent either way
anyway, I think it should be ok as is. It should only affect local environments and let's see what will actually happen on those environments

Makefile
YuviGold: Why? Can we make NetworkManager changes be applied from within as we do to the dnsmasq conf? It requires some socket maybe?

discovery-infra/test_infra/helper_classes/cluster.py
YuviGold: I agree with @osherdp it might become a bit complicated to investigate... But we can try. sure, the TF cleanup is a great feature.

Makefile
ori-amizur: It is systemd interface.  Not sure we want to deal with it.  I don't think it is in the scope of this PR

YuviGold (approved):
LGTM

PR #931, opened 2021-07-01
NO-ISSUE - Fix SNO lack of ram and cpu

eliorerz:
Pytest SNO is failing due to lack of CPU and RAM


discovery-infra/test_infra/utils/global_variables/global_variables.py
```diff
@@ -33,14 +39,11 @@ def __post_init__(self):
...
         for (env, expected), values in _triggers.items():
-            is_triggered = False
             if getattr(self, env) == expected:
                 for k, v in values.items():
                     self._set(k, v)
-                    is_triggered = True
-
-            if is_triggered:
                 self._triggered.append(env)
+                logging.info(f"{env.upper()} is triggered. Updating global variables: {values}")
```
YuviGold, line 46: You should check `values` isn't an empty list

eliorerz: @YuviGold Why will the trigger be empty? It is being manually added

YuviGold: TL;DR nvm :) 

You don't verify it anywhere. From this function POV you previously made a variable flag to mark when there is any trigger. So I just assumed you want the same behavior. 
Not critical, since mentioning "`env` was triggered" regardless of any actual changes (=no values), but you print it as well. So we would see there are no values at all...

YuviGold (approved):

PR #933, opened 2021-07-01
Fix path for nginx config files

trewest:
Creates correct path so we do not see issues when running load_balancer container


trewest:
@YuviGold should this line be updated too? 
https://github.com/openshift/assisted-test-infra/blob/0dd21d8af776614cbd739070e9f554b6f54bc77f/Makefile#L210

Looks like it points to the directory the script was previously creating

YuviGold:
> @YuviGold should this line be updated too?
> https://github.com/openshift/assisted-test-infra/blob/0dd21d8af776614cbd739070e9f554b6f54bc77f/Makefile#L210
> 
> Looks like it points to the directory the script was previously creating

Yes. Please update this one as well.

PR #937, opened 2021-07-04
NO-ISSUE - Fix invalid openshift_version on test_install

eliorerz:
if `make test OPENSHIFT_VERSION=4.7 NUM_MASTERS=1` the triggers mechanism supposed to override the openshift version and set it to 4.8.
This PR is fixing that conflict.
/cc @YuviGold


YuviGold:
I have to disagree. 
IMO when a user (ci) explicitly specifies what he wants - I would expect it to be taken. otherwise, it could be very misleading that stuff is happening under the hood.

/hold

YuviGold:
Not supported.

PR #942, opened 2021-07-05
MGMT-7155: ignore SCPException when sosreport gathering fails

osherdp:
It's another one of those exceptions that we don't have a lot to do about.
/cc @YuviGold 
/hold


YuviGold (approved):

YuviGold:
/retitle MGMT-7155: ignore SCPException when sosreport gathering fails

osherdp:
(infra problems)

PR #945, opened 2021-07-07
Support SNO in update_oc_config

lalon4:


YuviGold:
We already configure the DNS as part of `preparing_for_installation`
https://github.com/openshift/assisted-test-infra/blob/8cd997a241fc572f07638dbc14fe16074fef03c7/discovery-infra/test_infra/helper_classes/cluster.py#L694-L706
why do we need it in this function as well?

lalon4:
@YuviGold didn't know that dns for sno is set on `preparing_for_installation`..
In any case, `update_oc_config` is also setting KUBECONFIG environment variable
and to support cases where prepare_for_installation is not being used during test

YuviGold:
> @YuviGold didn't know that dns for sno is set on `preparing_for_installation`..
> In any case, `update_oc_config` is also setting KUBECONFIG environment variable
> and to support cases where prepare_for_installation is not being used during test

I don't think should support overriding an environment variable.
WDYT? @tsorya

lalon4:
why not? its the environment variable of the test container, and envs are per test process, so no affect on other tests.
BTW its not the scope of this PR, i mean, i only add support for SNO to an existing method

eliorerz:
@YuviGold I'm also don't think we should override the environment variables but as @lalon4 wrote, it's was there before.
Note that @tsorya is on a vacation so we can discuss it when he will be back, how and if to fix it.
Meanwhile we can open a task for changing it :)

YuviGold (approved):

PR #958, opened 2021-07-14
Adding vsphere terraform

michaellevy101:
Upgrading terraform to 0.14
Adding hcl2 library to parse a terraform files
Add vsphere env variables


terraform_files/vsphere/variables-vsphere.tf
```diff
@@ -0,0 +1,91 @@
...
+
+variable "iso_download_path" {
+  type        = string
+  description = "The ISO path (We have to upload this file to the vsphere)"
+  default = "dummy"
+}
+
+variable "vsphere_cluster" {
+  type        = string
+  description = "vSphere cluster name"
```
YuviGold, line 44: Explain what's the different usage between the AI cluster name and vSphere cluster name

terraform_files/vsphere/variables-vsphere.tf
```diff
@@ -0,0 +1,91 @@
...
+
+variable "vsphere_username" {
+  type        = string
+  description = "vSphere vcenter server username"
+}
+
+variable "vsphere_password" {
+  type        = string
+  description = "vSphere vcenter server username"
+}
```
YuviGold, line 18: Rearrange these variables into groups. It's not clear what's mandatory for vSphere login and what's used for other APIs

terraform_files/vsphere/variables-vsphere.tf
```diff
@@ -0,0 +1,91 @@
...
+  description = "The size of the master's virtual machine's memory, in MB"
+}
+
+variable "master_disk_size_gib" {
+  type = number
+  default = 120
+  description = "The size of the master's disk, in GB"
+}
+
+variable "master_vcpu" {
```
YuviGold, line 77: Some of these variables are duplicated.
Do we really need to redefine these common variables?
https://github.com/openshift/assisted-test-infra/blob/51d5d1c28822dba48ac56dd4210117d0b606a272/terraform_files/baremetal/variables-libvirt.tf#L131-L135
Could it be a shared variables file?

terraform_files/vsphere/variables-vsphere.tf
```diff
@@ -0,0 +1,91 @@
...
+  type = number
+  default = 4
+  description = "The total number of virtual processor cores to assign to the master virtual machine."
+}
+
+variable "vsphere_control_plane_cores_per_socket" {
+  type = number
+  default = 1
+  description = <<EOF
+The number of cores per socket in this virtual machine.
```
YuviGold, line 87: What does a socket represents?

terraform_files/vsphere/main.tf
```diff
@@ -0,0 +1,117 @@
...
+  required_providers {
+    vsphere = {
+      source = "hashicorp/vsphere"
+      version = "2.0.2"
+    }
+  }
+}
+
+locals {
+  hasISO = var.iso_download_path != "dummy" && var.iso_download_path != null
```
YuviGold, line 11: In what case would we want a `dummy` download path?

terraform_files/vsphere/main.tf
```diff
@@ -0,0 +1,117 @@
...
+
+data "vsphere_network" "network" {
+  name          = var.vsphere_network
+  datacenter_id = data.vsphere_datacenter.datacenter.id
+}
+
+# Tagging every VirtualMachine, ResourcePool, Folder with the cluster name to easy recognition and add the description:
+# "Created by the test-infra, do not delete manually"
+resource "vsphere_tag_category" "category" {
+  name        = "assisted-test-infra-${var.cluster_name}"
```
YuviGold, line 43: `cluster_name` already includes `test-infra`

terraform_files/vsphere/variables-vsphere.tf
michaellevy101: vsphere cluster is a cluster of hosts not applicative cluster.

terraform_files/vsphere/variables-vsphere.tf
michaellevy101: In the terraform controller we translate between the config and the terraform variables. 
This controller just pass the config as a terraform variables like this:
config = {**self._config.get_all(), **self._cluster_config.get_all(), "cluster_name": self.cluster_name}
and then filter the irrelevant variables. IMO more elegant.

michaellevy101: Need to refactor the terraform controller as well but not as part of this epic

terraform_files/vsphere/main.tf
michaellevy101: never it's just a marker

terraform_files/vsphere/main.tf
michaellevy101: no, it's not

terraform_files/vsphere/main.tf
michaellevy101: please read the documentation below

terraform_files/vsphere/variables-vsphere.tf
michaellevy101: You shouldn't set this variable manually just a default value...
 The number of cores per socket in this virtual machine. The number of vCPUs on the virtual machine will be num_cpus divided by num_cores_per_socket. If specified, the value supplied to num_cpus must be evenly divisible by this value. Default: 1.

michaellevy101: did you read the entire section? is from the vsphere official documentation

terraform_files/vsphere/variables-vsphere.tf
YuviGold: It's still unclear to me what's required for API operations and what's only for login.
In addition, `iso_download_path` shouldn't be under `vSphere variables` category.

terraform_files/vsphere/variables-vsphere.tf
YuviGold: Please open a ticket

terraform_files/vsphere/variables-vsphere.tf
YuviGold: I read. I still don't understand what does a socket represents.

discovery-infra/test_infra/tools/terraform_utils.py
```diff
@@ -18,15 +21,35 @@ def __init__(self, working_dir: str):
...
+                                                  input=False, skip_plan=True)
         if return_value != 0:
             message = f"Terraform apply failed with return value {return_value}, output {output} , error {err}"
             logging.error(message)
             raise Exception(message)
 
+    def set_and_apply(self, refresh: bool = True, **kwargs) -> None:
+        defined_variables = self.select_defined_variables(**kwargs)
+        self.change_variables(defined_variables)
+        self.apply(refresh=refresh)
```
YuviGold, line 51: How those modifications are related to VSphere?

YuviGold (commented):
Add a README providing a holistic overview about using vsphere provider.
You should direct to the original docs https://registry.terraform.io/providers/hashicorp/vsphere/latest/docs

but also explain how it integrates with our current infra.

terraform_files/vsphere/variables-vsphere.tf
michaellevy101: https://issues.redhat.com/browse/MGMT-7232?filter=-2

discovery-infra/test_infra/tools/terraform_utils.py
michaellevy101: This code allow us to take the configuration objects, select only the relevant terraform properties and then apply.
It's used by the controller but basically it's a utility method.
I'm trying to splitting out things

michaellevy101:
@YuviGold you asked for a documentation for the vsphere terraform  
It's used internally by the controller you shouldn't use it manually. it's just a code which has a documentation inside it.
The provider name is there so you can just google it and i'm expecting a basic terraform knowledge

terraform_files/vsphere/variables-vsphere.tf
YuviGold: Don't explain it to me.. Write it on the description

michaellevy101: I think we shouldn't describe vsphere object types but i did

YuviGold:
> @YuviGold Readme file isn't necessary because vsphere terraform files are used internally by the controller you shouldn't use it manually. it's just a code which has a documentation inside it.
> The provider name is there so you can just google it and i'm expecting a basic terraform knowledge
> IMO we should document manual steps or non obvious code something that requires a second look
> 
> We shouldn't document vsphere objects like cluster, vm, folder and tag nor terraform vsphere provider just our usage which is a trivial (self documenting code) expect the special cases that i documented in the code.
> 
> if you feel that something isn't trivial please add a comment there and will try to document it better

I think it could be nice to have an overview of the integration process between AI and vSphere somewhere.

terraform_files/vsphere/variables-vsphere.tf
michaellevy101: please open a ticket for vmware:) https://registry.terraform.io/providers/hashicorp/vsphere/latest/docs/resources/virtual_machine#num_cores_per_socket

michaellevy101:
https://github.com/openshift/assisted-test-infra/pull/958#issuecomment-879825028
agreed, after the whole process would work i will create a short readme but no need to document the terraform internally.
We should document the controller itself and how to to create a multi-platform test

michaellevy101:
> > @YuviGold Readme file isn't necessary because vsphere terraform files are used internally by the controller you shouldn't use it manually. it's just a code which has a documentation inside it.
> > The provider name is there so you can just google it and i'm expecting a basic terraform knowledge
> > IMO we should document manual steps or non obvious code something that requires a second look
> > We shouldn't document vsphere objects like cluster, vm, folder and tag nor terraform vsphere provider just our usage which is a trivial (self documenting code) expect the special cases that i documented in the code.
> > if you feel that something isn't trivial please add a comment there and will try to document it better
> 
> I think it could be nice to have an overview of the integration process between AI and vSphere somewhere.
agreed, after the whole process would work i will create a short readme but no need to document the terraform internally.
We should document the controller itself and how to to create a multi-platform test

YuviGold (approved):
LGTM

YuviGold (approved):

PR #962, opened 2021-07-15
Set fixed versions on requirements.txt

YuviGold:


YuviGold:
This is a blocker.
hcl version `python-hcl2-3.0.0` is taken and it broke the current API.

/override e2e-metal-assisted
/override e2e-metal-assisted-kube-api
/override images

PR #970, opened 2021-07-18
MGMT-7289 - Change basePath in python client

eliorerz:


discovery-infra/test_infra/assisted_service_api.py
```diff
@@ -36,6 +46,15 @@ def __init__(self, inventory_url: str, offline_token: Union[str, None], pull_sec
         self.domains = api.ManagedDomainsApi(api_client=self.api)
         self.operators = api.OperatorsApi(api_client=self.api)
 
+    @classmethod
+    def _get_host_url(cls, inventory_url: str, configs: Configuration):
+        url = urljoin(inventory_url, cls.BASE_API_PATH)
+
+        api_version = Path(configs.host).name
```
ravidbro, line 53: What is the expected value of 'configs.host'?

eliorerz: `http://api.openshift.com/api/assisted-install/v1` or `http://api.openshift.com/api/assisted-install`

YuviGold: I think we should always use the `configs.host` regardless of the version, without the base default host `http://api.openshift.com`, then we would be compatible for every base path.

https://github.com/openshift/assisted-service/blob/48cdd53a6ec96f39b67c92d24ad0cf3e5ab2a8cd/swagger.yaml#L9-L10

ravidbro: I agree with @YuviGold , in every swagger always the basePath and the paths of the specific APIs will be coordinated regardless if there is some /vX in the path

eliorerz: @YuviGold The client is not containing the default host and the base_path separately. They are joined to configuration.host

YuviGold: Yes I see, it looks like
```python
        # Default Base url
        self.host = "http://api.openshift.com/api/assisted-install/v1"
```

So I'd prefer that we will have a default host, and remove it. This way you can get the changeable base path

eliorerz: I already updated it. it will parse the host url and take the base_path from it

YuviGold: The current solution still expects an API version as part of the `config.hosts` field. I'd prefer to have a default host that isn't supposed to change than a default base path that might change as it happened so far.

PR #973, opened 2021-07-18
MGMT-7293 Add networkType as cluster create params

oshercc:


discovery-infra/start_discovery.py
```diff
@@ -307,6 +307,11 @@ def _get_host_ip_from_cidr(cidr):
 def _get_proxy_ip(cidr):
     return IPNetwork(cidr).ip + 1
 
+def _get_network_type(ipv4, ipv6):
+    ipv4_only = ipv4 and not ipv6
+    if ipv4_only:
+        return "OVNKubernetes"
```
gamli75, line 313: the default for IPv4 should be OpenShiftSDN and not OVNKubernetes

gamli75 (commented):
Please add a task or bug number for this PR.
If this PR fixes the CI problem - please add it in the PR description

ori-amizur (commented):
The way this is written - the change is not necessary since the assisted service already handles the network-type by default.  The only reason to add it to test-infra is if we want explicitly to set the network-type but it is not part of the change.

PR #976, opened 2021-07-20
MGMT-7298: retry creating minikube with the registry addon

michaellevy101:


osherdp:
/retitle MGMT-7298: retry creating minikube with the registry addon

osherdp (approved):

PR #978, opened 2021-07-20
MGMT-7293: Add networkType as cluster create params to discovery

oshercc:


discovery-infra/start_discovery.py
```diff
@@ -391,6 +393,12 @@ def _get_provisioning_cidr6(cidr, ns_index):
     return str(provisioning_cidr)
 
 
+def _get_network_type(ipv4, ipv6):
+    ipv4_only = ipv4 and not ipv6
+    if ipv4_only:
```
michaellevy101, line 398: if not ipv6:

michaellevy101: what about 'auto-assign'? shouldn't it be the default?

oshercc: no,

osherdp: I'm a bit confused here, but maybe I'm outdated about recent events
Don't we use OVNKubernetes for ipv6?
Because it's kind of reverse in here

osherdp:
/hold
until we understand if it should be like this or the opposite

osherdp (approved):

oshercc:
/hold
I see this is not an issue on the CI,

YuviGold:
The network type is already set as part of the controller's infrastructure on https://github.com/openshift/assisted-test-infra/pull/973

PR #985, opened 2021-07-23
Create post-install script for IBM Cloud machines

mkowalski:
This PR creates a post-install script that can be used on the IBM Cloud
machines so that they are deployed with a set of reasonable defaults.

### DIFF
--- a/scripts/ibmcloud_post_install.sh
+++ b/scripts/ibmcloud_post_install.sh
@@ -1,0 +1,61 @@
+#!/usr/bin/env bash
+
+cat > /usr/bin/install_complete.sh <<EOF
+#!/usr/bin/env bash
+set -euo pipefail
+
+export HOME=/root
+
+### Status flag
+LOGFILE=/root/ibmcloud-post-install.log
+log () {
+  echo "\$(date) +++ \$1" >> \$LOGFILE
+}
+
+log "Starting ibmcloud-post-install.sh"
+
+### Check if the script has been already run
+if [ -f /root/ibmcloud-post-install.log ]; then
+    log "Found existing log, exiting !"
+    exit 0
+fi
+
+{
+### Base prerequisites
+dnf install -y git
+
+### Get AI repo
+mkdir -p /home/test
+cd /home/test
+git clone https://github.com/openshift/assisted-test-infra.git
+
+### Provision test-infra
+cd /home/test/assisted-test-infra
+
+scripts/install_environment.sh
+scripts/install_environment.sh config_sshd
+
+} &>> \$LOGFILE
+
+### Status flag
+log "Finished ibmcloud-post-install.sh, thank you !"
+EOF
+chmod 755 /root/install_complete.sh
+
+# set timeoutsec explicitly to prevent timeout fail
+# ref. https://qiita.com/khayama/items/861243aed5cf95f318d1
+cat > /etc/systemd/system/install_complete.service <<EOF
+[Unit]
+Description=INSTALL_COMPLETE
+After=network.target
+[Service]
+ExecStart=/usr/bin/install_complete.sh
+Type=oneshot
+TimeoutSec=infinity
+[Install]
+WantedBy=multi-user.target
+EOF
+
+# run your script above without timeout fail bug
+systemctl daemon-reload
+systemctl restart install_complete.service
--- a/scripts/install_environment.sh
+++ b/scripts/install_environment.sh
@@ -181,6 +181,19 @@ EOF
   sudo firewall-cmd --zone=libvirt --add-port=80/tcp
 }
 
+function config_sshd() {
+  echo "Harden SSH daemon"
+
+  ### IBM Cloud machines by default allow SSH as root using the password. Given that the default one
+  ### is extremely simple, we disable this option.
+
+  sudo sed -i "s/.*RSAAuthentication.*/RSAAuthentication yes/g" /etc/ssh/sshd_config
+  sudo sed -i "s/.*PubkeyAuthentication.*/PubkeyAuthentication yes/g" /etc/ssh/sshd_config
+  sudo sed -i "s/.*PasswordAuthentication.*/PasswordAuthentication no/g" /etc/ssh/sshd_config
+
+  sudo systemctl restart sshd.service
+}
+
 function additional_configs() {
     if [ "${ADD_USER_TO_SUDO}" != "n" ]; then
         current_user=$(whoami)
@@ -225,4 +238,3 @@ if [ $# -eq 0 ]; then
 else
     $@
 fi
-


mkowalski:
A thing that may be spotted but is not really caused by this PR is the fact that the dnsmasq instance controlled by the NetworkManager does not start on `systemctl restart NetworkManager.service` nor on reboot, but only on `systemctl reload NetworkManager.service`.

For this reason, after rebooting the server a manual `systemctl reload NetworkManager.service` is needed, otherwise there is no `nameserver 127.0.0.1` entry in `/etc/resolv.conf` and there is no `dnsmasq` process running on the host.

Given that it's kinda out of scope of the provisioning script, I'm keeping it as it is until I find a proper place to plug a workaround for this.

Ref.: https://serverfault.com/questions/1054792/networkmanager-isnt-starting-dnsmasq-on-restart-only-reload
Ref.: https://www.reddit.com/r/CentOS/comments/lrv20x/networkmanager_isnt_starting_dnsmasq_on_restart/

scripts/ibmcloud_post_install.sh
```diff
@@ -0,0 +1,72 @@
...
+
+log "Starting ibmcloud-post-install.sh"
+
+### SSH hardening
+### IBM Cloud machines by default allow SSH as root using the password. Given that the default one
+### is extremely simple, we disable this option.
+
+sed -i "s/.*RSAAuthentication.*/RSAAuthentication yes/g" /etc/ssh/sshd_config
+sed -i "s/.*PubkeyAuthentication.*/PubkeyAuthentication yes/g" /etc/ssh/sshd_config
+sed -i "s/.*PasswordAuthentication.*/PasswordAuthentication no/g" /etc/ssh/sshd_config
```
celebdor, line 19: Could pass the multiple replacements to a single sed command.

scripts/ibmcloud_post_install.sh
```diff
@@ -0,0 +1,72 @@
...
+systemctl restart sshd.service
+
+log "SSH configuration applied successfuly"
+
+### SSH local key
+
+mkdir -p /root/.ssh
+chmod 700 /root/.ssh
+ssh-keygen -t rsa -b 4096 -N "" -f /root/.ssh/id_rsa
+cat /root/.ssh/id_rsa.pub >> /root/.ssh/authorized_keys
```
celebdor, line 30: What do we need to authorize the local root key for?

scripts/ibmcloud_post_install.sh
```diff
@@ -0,0 +1,72 @@
...
+
+dnf install -y dnsmasq
+
+mkdir -p /etc/NetworkManager/conf.d/
+cat <<EOF >/etc/NetworkManager/conf.d/dnsmasq.conf
+[main]
+dns = dnsmasq
+EOF
+
+cat /etc/resolv.conf  | grep -v '^#' | grep nameserver | awk '{print "server="$2}' > /etc/NetworkManager/dnsmasq.d/ibm-cloud.conf
```
celebdor, line 58: Why shouldn't it just be:

```
awk '/^nameserver / {print "server="$2}' /etc/resolv.conf >> /etc/NetworkManager/dnsmasq.d/ibm-cloud.conf

File: scripts/ibmcloud_post_install.sh (hunk 2 above)
Comment (mkowalski) [reply scripts/ibmcloud_post_install.sh:30]: I often find myself doing `ssh localhost`

File: scripts/ibmcloud_post_install.sh (hunk 3 above)
Comment (mkowalski) [reply scripts/ibmcloud_post_install.sh:58]: Yes it could and should. Fixed :)

File: scripts/ibmcloud_post_install.sh (hunk 1 above)
Comment (mkowalski) [reply scripts/ibmcloud_post_install.sh:19]: Purely for readability, using multiple expressions in one will create a big one-liner difficult to modify in the future if someone wants to add one more parameter to change. With this here, you can just copy-paste and change a single one so `git blame` and stuff like this becomes easier

**Comment by YuviGold** [comment]
Could you elaborate on what's the difference between this script and the one running with `make create_full_environment`? This is the script we use for Packet instances, if it's not generic enough let's adjust it

**Comment by mkowalski** [comment]
> Could you elaborate on what's the difference between this script and the one running with `make create_full_environment`? This is the script we use for Packet instances, if it's not generic enough let's adjust it

The whole of `./create_full_environment.sh` is too big because provisioning script can be one script, it cannot be a bunch of files sourcing each other. If we take `./install_environment.sh` instead it seems to be like it's a bit too heavy because it's designed specifically for assisted-test-infra and does stuff that I do not need (or even, must not do) if I'm using the server as a workstation or for dev-infra or for any other work (e.g. `setenforce 0` is something I won't do unless really needed).

**Comment by romfreiman** [comment]
@mkowalski as I see it, the purpose of those servers are for assisted development, where 100% of the people use assisted-test-infra.
I dont think we should maintain 2 scripts. Most of the people dont care about setenforce 0. So this is not an issue for them. If it's really important for you, feel free top change this post deployement. IMO it's still less maintenance than having 2 scripts

**Comment by mkowalski** [comment]
Sure, but this is not the main point of the discussion and I don't want to diverge into parts which are implementation details (yet). The problem here is not *what* the script is going to do but *how* to run it. In the current form you *cannot* use `./create_full_environment.sh`.

If we use only `./install_environment.sh` as a provisioning script, then you would still need to run `make create_full_environment` afterwards because e.g. minikube has not been installed.

**Comment by mkowalski** [comment]
@romfreiman @YuviGold I have reorganized the structure base on what we have discussed

/hold "it needs a test"

**Comment by mkowalski** [comment]
The first non-interactive run gave me

```
  Installing       : seabios-bin-1.13.0-2.module_el8.4.0+547+a85d02ba    85/123 
  Running scriptlet: radvd-2.17-15.el8.x86_64                            86/123 
error: %prein(radvd-2.17-15.el8.x86_64) scriptlet failed, signal 15

  Installing       : mesa-filesystem-20.3.3-2.el8.x86_64                 87/123 
error: radvd-2.17-15.el8.x86_64: install failed
```

File: scripts/ibmcloud_post_install.sh [hunk 4]
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
+
+cat > /root/install_complete.sh <<EOF
```
tsorya, line 3: why not /usr/bin?

scripts/ibmcloud_post_install.sh
```diff
@@ -0,0 +1,56 @@
...
+}
+
+log "Starting ibmcloud-post-install.sh"
+
+{
+### Base prerequisites
+dnf install -y git
+
+### Get AI repo
+cd /tmp
```
tsorya, line 22: maybe better to clone it to /home ?

tsorya: i mean why to remove it? if i understand right those machines will be used for test-infra in any case right?

scripts/ibmcloud_post_install.sh
mkowalski: Done

PR #987, opened 2021-07-25
Use default ssh keys for SSH connections

osherdp:
It will be much easier doing ``ssh core@<vm>`` than ``ssh -i ssh_key/key core@<vm>``.
/cc @YuviGold @tsorya 
/hold


GUIDE.md
```diff
@@ -19,7 +19,7 @@ Procedure
 
 As `$USER` user with `sudo` privileges,
 
-1.  Generate an SSH key if not already available (`ssh_key/key`)
+1.  Generate an SSH key if not already available (`~/.ssh/id_rsa`)
```
tsorya, line 22: it doesn't exits on fresh beaker as if i remember right.

scripts/install_environment.sh
```diff
@@ -202,7 +202,7 @@ function additional_configs() {
     sudo chmod ugo+rx "$(dirname "$(pwd)")"
     echo "disabling selinux by setenforce 0"
     sudo setenforce 0 || true
-    sudo chmod 600 ssh_key/key
+    sudo chmod 600 ~/.ssh/id_rsa
```
tsorya, line 205: it is needed? maybe better to generate it if doesn't exists ?

osherdp: you're right
take another look

GUIDE.md
osherdp: ack
creating it on ``install_environment``

scripts/install_environment.sh
```diff
@@ -202,7 +202,12 @@ function additional_configs() {
     sudo chmod ugo+rx "$(dirname "$(pwd)")"
     echo "disabling selinux by setenforce 0"
     sudo setenforce 0 || true
-    sudo chmod 600 ssh_key/key
+
+    if [ ! -f ~/.ssh/id_rsa ]; then
+        ssh-keygen -t rsa -f ~/.ssh/id_rsa -P ''
```
tsorya, line 207: ssh-keygen always exists ? or maybe we should install it?

osherdp: it's from ``openssh`` which seems to always be installed

YuviGold (commented):
The most important thing - we won't publish SSH key on a public repository :)

I see that the `SSH_KEY` const is redefined 3 times in different ways on 
* https://github.com/openshift/assisted-test-infra/pull/987/files#diff-52120aa2d8051444a0504d63fc39abed0cdebc7710de5ed821276f622cde3e93R34
* https://github.com/openshift/assisted-test-infra/pull/987/files#diff-e16325b38e537e1345e71800ddf18a639608f63bf3ab0d8b4e51ed5fba43cde8R20
* https://github.com/openshift/assisted-test-infra/pull/987/files#diff-c0116e7e491f5cef6315ef89dc73503c57a4a6717ed4b2c18dbe45b0a1b08182R10

Could you merge them?

discovery-infra/bootstrap_in_place.py
```diff
@@ -31,7 +31,7 @@
 INSTALLER_GATHER_DIR = os.path.join(IBIP_DIR, "installer-gather")
 INSTALLER_GATHER_DEBUG_STDOUT = os.path.join(INSTALLER_GATHER_DIR, "gather.stdout.log")
 INSTALLER_GATHER_DEBUG_STDERR = os.path.join(INSTALLER_GATHER_DIR, "gather.stderr.log")
-SSH_KEY = os.path.join("ssh_key", "key")
+SSH_KEY = os.path.join(os.path.expanduser("~"), ".ssh", "id_rsa")
```
tsorya, line 34: read from os env?

discovery-infra/test_infra/consts/consts.py
```diff
@@ -17,7 +17,7 @@ class OpenshiftVersion(Enum):
 BASE_IMAGE_FOLDER = "/tmp/images"
 IMAGE_NAME = "installer-image.iso"
 STORAGE_PATH = "/var/lib/libvirt/openshift-images"
-SSH_KEY = "ssh_key/key.pub"
+SSH_KEY = "~/.ssh/id_rsa.pub"
```
tsorya, line 20: get from os env?

osherdp:
@YuviGold one of them is public key, but sure. merged the other two

discovery-infra/test_infra/consts/consts.py
osherdp: not today porfavor

osherdp: I mean, it seems like we're already doing it, but from the makefile
We can do a bit bigger refactoring, but I think we should afterwards

YuviGold (approved):
LGTM

PR #988, opened 2021-07-25
Supporting modifying the configuration objects by custom fixtures

michaellevy101:


discovery-infra/tests/base_test.py
```diff
@@ -63,6 +63,16 @@ def prepare_controller_configuration(self, controller_configuration: BaseNodeCon
         yield controller_configuration
         net_asset.release_all()
 
+    @pytest.fixture
+    def post_controller_configuration(self, request: pytest.FixtureRequest,
```
lalon4, line 67: can we rename to something like _override_controller_configuration_?

discovery-infra/tests/base_test.py
```diff
@@ -73,37 +83,47 @@ def cluster_configuration(self) -> ClusterConfig:
         return ClusterConfig()
 
     @pytest.fixture
-    def controller(self, cluster_configuration: ClusterConfig,
-                   prepare_controller_configuration: BaseNodeConfig) -> NodeController:
+    def post_cluster_configuration(self, request: pytest.FixtureRequest,
```
lalon4, line 86: can we rename to something like override_cluster_configuration?

discovery-infra/tests/base_test.py
```diff
@@ -73,37 +83,47 @@ def cluster_configuration(self) -> ClusterConfig:
...
                 logging.info('--- TEARDOWN --- node controller\n')
                 nodes.destroy_all_nodes()
-                logging.info(f'--- TEARDOWN --- deleting iso file from: {cluster_configuration.iso_download_path}\n')
-                infra_utils.run_command(f"rm -f {cluster_configuration.iso_download_path}", shell=True)
+                logging.info(f'--- TEARDOWN --- deleting iso file from: {post_cluster_configuration.iso_download_path}\n')
+                infra_utils.run_command(f"rm -f {post_cluster_configuration.iso_download_path}", shell=True)
 
     @pytest.fixture
-    def prepare_network(self, prepare_nodes: Nodes, prepare_controller_configuration: BaseNodeConfig) -> Nodes:
+    def prepare_network(self, prepare_nodes: Nodes, post_controller_configuration: BaseNodeConfig) -> Nodes:
```
lalon4, line 121: can we rename to something like _get_nodes_?

discovery-infra/tests/base_test.py
```diff
@@ -63,6 +63,16 @@ def prepare_controller_configuration(self, controller_configuration: BaseNodeCon
         yield controller_configuration
         net_asset.release_all()
 
+    @pytest.fixture
+    def post_controller_configuration(self, request: pytest.FixtureRequest,
+                                      prepare_controller_configuration: BaseNodeConfig) -> BaseNodeConfig:
+        """
+        Allows the test to modify the controller configuration by registering a custom fixture.
+        To register the custom fixture you have to mark the test with "post_controller_configuration" marker.
+        @pytest.mark.post_controller_configuration(FIXTURE1_NAME.__name__, FIXTURE2_NAME.__name__)
```
lalon4, line 72: should get only one fixture

discovery-infra/tests/base_test.py
michaellevy101: changed it to controller_configuration because this is the final result. and first fixture which creates it to new_controller_configuration

PR #998, opened 2021-07-28
MGMT-4987 - Move test directory to project root

eliorerz:
Move tests package to project root

/cc @YuviGold @tsorya @lalon4 @nshidlin


tsorya:
not sure what was decided regarding moving those files but why? (just asking)

eliorerz:
This task has been postponed due to its complexity.

PR #999, opened 2021-07-28
NO-ISSUE - Rename target tests names

eliorerz:
Prevent running `test_install_with_deploy_nodes` when running `test_install`
/cc @YuviGold


YuviGold (approved):
LGTM

PR #1000, opened 2021-07-28
MGMT-4492: Removing WITH_AMS_SUBSCRIPTIONS env variable.

ybettan:
This variable was originally introduced in order to be able to easily
disable in on the transition period.

OCM integration should be enabled as long as we are running with any OCM
(AMS) instance, therefore, this configuration is redundant.

Signed-off-by: Yoni Bettan <ybettan@redhat.com>


gamli75:
related to this PR: https://github.com/openshift/assisted-service/pull/2294

gamli75 (approved):

PR #1003, opened 2021-07-29
NO-ISSUE - Fix single openshift version when running without specific openshift version

eliorerz:


osherdp (approved):

osherdp:
/hold
(to be able to check the result)

osherdp:
@eliorerz I think we can force-push and investigate the failure offline

eliorerz:
image

PR #1005, opened 2021-07-30
NO-ISSUE - Fix logs failing while trying to download custom_manifests.yaml file

eliorerz:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-openshift-release-master-nightly-4.9-e2e-metal-assisted-ipv6/1420661063333974016/artifacts/e2e-metal-assisted-ipv6/baremetalds-assisted-gather/build-log.txt

```
2021-07-29 09:52:04,989 ERROR      - 140416225086336 - <class 'assisted_service_client.rest.ApiException'> 	(/home/assisted/discovery-infra/logger.py:37)
  File "./discovery-infra/download_logs.py", line 143, in download_logs
    os.path.join(output_folder, "cluster_files", cluster_file))
  File "/home/assisted/discovery-infra/test_infra/assisted_service_api.py", line 253, in download_and_save_file
    cluster_id=cluster_id, file_name=file_name, _preload_content=False
  File "/usr/local/lib/python3.6/site-packages/assisted_service_client/api/installer_api.py", line 580, in download_cluster_files
    (data) = self.download_cluster_files_with_http_info(cluster_id, file_name, **kwargs)  # noqa: E501
  File "/usr/local/lib/python3.6/site-packages/assisted_service_client/api/installer_api.py", line 668, in download_cluster_files_with_http_info
    collection_formats=collection_formats)
  File "/usr/local/lib/python3.6/site-packages/assisted_service_client/api_client.py", line 330, in call_api
    _preload_content, _request_timeout)
  File "/usr/local/lib/python3.6/site-packages/assisted_service_client/api_client.py", line 161, in __call_api
    _request_timeout=_request_timeout)
  File "/usr/local/lib/python3.6/site-packages/assisted_service_client/api_client.py", line 351, in request
    headers=headers)
  File "/usr/local/lib/python3.6/site-packages/assisted_service_client/rest.py", line 238, in GET
    query_params=query_params)
  File "/usr/local/lib/python3.6/site-packages/assisted_service_client/rest.py", line 228, in request
    raise ApiException(http_resp=r)
assisted_service_client.rest.ApiException: (409)
Reason: Conflict
HTTP response headers: HTTPHeaderDict({'Content-Type': 'application/octet-stream', 'Date': 'Thu, 29 Jul 2021 09:52:04 GMT', 'Content-Length': '362'})
HTTP response body: b'{"code":"409","href":"","id":409,"kind":"Error","reason":"Failed to fetch metadata for object bc186278-222b-4673-b253-565a0171146b/custom_manifests.yaml in bucket test: Failed to fetch metadata for object bc186278-222b-4673-b253-565a0171146b/custom_manifests.yaml in bucket test: NotFound: Not Found\\n\\tstatus code: 404, request id: 16963A376862D81E, host id: "}'
```

### DIFF
--- a/discovery-infra/download_logs.py
+++ b/discovery-infra/download_logs.py
@@ -2,15 +2,17 @@
 
 import filecmp
 import json
-import logging
 import os
 import shutil
 import tempfile
 import time
+import traceback
 from argparse import ArgumentParser
 from collections import Counter
 from contextlib import suppress
 from datetime import datetime
+from types import TracebackType
+from typing import Type
 
 import assisted_service_client
 import requests
@@ -20,7 +22,7 @@ from junit_report import JunitTestCase, JunitTestSuite
 from paramiko.ssh_exception import SSHException
 from scp import SCPException
 
-from logger import log, suppressAndLog, add_log_file_handler
+from logger import log, add_log_file_handler
 from test_infra import warn_deprecate
 from test_infra.assisted_service_api import InventoryClient, create_client
 from test_infra.consts import ClusterStatus, HostsProgressStages, env_defaults
@@ -51,6 +53,18 @@ urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
 warn_deprecate()
 
 
+class SuppressAndLog(suppress):
+    def __exit__(self, exctype: Type[Exception], excinst: Exception, exctb: TracebackType):
+        res = super().__exit__(exctype, excinst, exctb)
+
+        if res:
+            with suppress(BaseException):
+                tb_data = traceback.extract_tb(exctb, 1)[0]
+                log.warning(f"Suppressed {exctype.__name__} from {tb_data.name}:{tb_data.lineno} : {excinst}")
+
+        return res
+
+
 def main():
     args = handle_arguments()
 
@@ -79,7 +93,6 @@ def main():
 
 def download_cluster_logs(client: InventoryClient, cluster: dict, dest: str, must_gather: bool,
                           update_by_events: bool = False, retry_interval: int = RETRY_INTERVAL, pull_secret=""):
-
     @JunitTestSuite(custom_filename=f"junit_download_report_{cluster['id']}")
     def download_logs_suite():
         return download_logs(client, cluster, dest, must_gather, update_by_events, retry_interval, pull_secret)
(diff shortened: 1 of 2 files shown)


discovery-infra/download_logs.py
```diff
@@ -117,6 +118,18 @@ def is_update_needed(output_folder: str, update_on_events_update: bool, client:
     return need_update
 
 
+def download_and_save_files(files_list: list, client: InventoryClient, cluster: dict,
+                            output_folder: str, raise_exception:bool = True):
+    try:
+        for cluster_file in files_list:
+            with suppressAndLog(assisted_service_client.rest.ApiException):
```
YuviGold, line 125: it the exception is suppressed then it would never be raised and be caught

discovery-infra/download_logs.py
```diff
@@ -137,10 +150,11 @@ def download_logs(client: InventoryClient, cluster: dict, dest: str, must_gather
...
 
-        for cluster_file in ("bootstrap.ign", "master.ign", "worker.ign", "install-config.yaml", "custom_manifests.yaml"):
-            with suppressAndLog(assisted_service_client.rest.ApiException):
-                client.download_and_save_file(cluster['id'], cluster_file,
-                                              os.path.join(output_folder, "cluster_files", cluster_file))
+        download_and_save_files(["bootstrap.ign", "master.ign", "worker.ign", "install-config.yaml"],
+                                client, cluster, output_folder)
+
+        # best effort
+        download_and_save_files(["custom_manifests.yaml"], client, cluster, output_folder, raise_exception=False)
```
YuviGold, line 157: I think that actually there are more files that aren't necessarily available.. for example if there are no workers there won't be an ignition file.

It comes back to the point where the service should provide a list of all the avilable files to be downloadef
cc @osherdp

discovery-infra/download_logs.py
eliorerz: I see.. So if it is suppressed, how does it happen in the first place?

YuviGold: That's why the function is named as "supress and log" :) you see the log output. the flow of the calling function isn't affected.

eliorerz: I just saw it .. :) 
Maybe suppress and log is not the best approach here. I don't want to see the whole traceback in log if it's not an actual error

eliorerz: @YuviGold What do you think change `suppressAndLog` to print a warning instead of full tarceback?

YuviGold: I think that most of the time when there's an exception thrown I'd like to know where it comes from. So the traceback is necessary, otherwise, I might think it was something critical.
Suggestion - What do you thing about `suppressAndLog` writing `Suppressed exception from {func}: {message}`?

eliorerz: @YuviGold  I don't really mind the syntax.  If we decide to suppress the exception, I think we should write it down so anyone who reads it will know that it was suppressed  :)

YuviGold: So let's emphasize it's a suppressed exception. :)

YuviGold:
@eliorerz Can you use the `suppressAndLog` function and emphasize it's a suppressed exception?

osherdp (approved):

PR #1006, opened 2021-08-01
Get libvirt domain mac and ip from agent instead of lease

nshidlin:
Static IP flow was broken by https://github.com/openshift/assisted-test-infra/pull/960/files
Currently libvirt node's IP and mac are taken from dhcp lease file. In the case of static IP this list is empty. This PR queries libvirt agent for addresses instead of the dhcp leases

### DIFF
--- a/discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
+++ b/discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
@@ -9,6 +9,7 @@ from contextlib import contextmanager, suppress
 from pathlib import Path
 from typing import List, Callable
 from xml.dom import minidom
+from netaddr import IPAddress
 
 import libvirt
 import waiting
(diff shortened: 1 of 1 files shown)


osherdp:
@nshidlin can you take a look afterwards to see if sosreports are still generated properly?
because I think they depend on the same code

nshidlin:
> @nshidlin can you take a look afterwards to see if sosreports are still generated properly?
> because I think they depend on the same code

So, basically see that the nodes can be ssh'd into with `run_command`?

osherdp:
@nshidlin you can also look at the artifacts of ``e2e-metal-assisted``, we should have it in a directory named ``sosreport``

osherdp:
but yes, that would work also I think

osherdp (approved):

osherdp:
@nshidlin why did you close it?

PR #1015, opened 2021-08-04
MGMT-7105 - create multiple clusters on vsphere (multiple AIs might h…

michaellevy101:
…ave the same cluster name)


discovery-infra/test_infra/tools/terraform_utils.py
```diff
@@ -13,12 +13,14 @@ class TerraformUtils:
     VAR_FILE = "terraform.tfvars.json"
     STATE_FILE = "terraform.tfstate"
 
-    def __init__(self, working_dir: str):
+    def __init__(self, working_dir: str, init: bool = True):
```
eliorerz, line 16: Give a batter name for init, and maybe add a short doc why do we need it

eliorerz (approved):
LGTM

PR #1016, opened 2021-08-04
Revert "Bump junit-report from 0.1.24 to 0.1.32 (#1013)"

eliorerz:
This reverts commit 28876018 due to some issues on version > v0.1.24
/cc @YuviGold 

junit-report 0.1.32 uses new mechanism for handling pytest fixtures. Is cases some of the reports to show partially reports or no report at all.
It will probably will fix in the next version of junit-report

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ ipdb==0.13.9
 ipython==7.11.1
 Jinja2===3.0.1
 jira==3.0.1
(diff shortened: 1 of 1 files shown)


YuviGold:
/lgtm

Please add details why do you revert this change. It seems that the JUNIT of this PR and previous runs look the same.
/hold

PR #1025, opened 2021-08-10
NO-ISSUE: Removed some minikube configurations.

ybettan:
Those removed configurations don't exist in newer versions of minikube.

Signed-off-by: Yoni Bettan <ybettan@redhat.com>


omertuc:
Do they have alternatives in the newer version? Why were they added in the first place?

ybettan:
Probably to have some more verbose printing @omertuc.

Couldn't find the commit message related to that code for some reason.

PR #1027, opened 2021-08-11
Update agent api calls to use v2_* according to swagger

nshidlin:


osherdp:
/retitle Update agent api calls to use v2_* according to swagger

osherdp (approved):

PR #1007, opened 2021-08-01
NO-ISSUE - Add junit reports for download_logs.py

eliorerz:

PR #1029, opened 2021-08-11
MGMT-7474 - Add download_logs junit report

eliorerz:
Old PR - https://github.com/openshift/assisted-test-infra/pull/1007


eliorerz:
/unhold 
image

scripts/download_logs.sh
```diff
@@ -45,7 +46,7 @@ function download_cluster_logs() {
     fi
   fi
 
-  skipper run ./discovery-infra/download_logs.py ${SERVICE_URL} ${LOGS_DEST} --cluster-id ${CLUSTER_ID} ${ADDITIONAL_PARAMS}
+  skipper run -e JUNIT_REPORT_DIR ./discovery-infra/download_logs.py ${SERVICE_URL} ${LOGS_DEST} --cluster-id ${CLUSTER_ID} ${ADDITIONAL_PARAMS}
```
YuviGold, line 49: Add `JUNIT_REPORT_DIR` to `skipper.env`

discovery-infra/download_logs.py
```diff
@@ -50,31 +52,37 @@
 
 def main():
     args = handle_arguments()
+    report_name = f"junit_download_report_{args.cluster_id or str(uuid.uuid4())[:8]}"
```
YuviGold, line 55: You shouldn't generate a UUID. Cluster ID should always be available either manually from the `args` or during the download.
The `report_name` should be generated as part of the main function itself

eliorerz: what do you mean by `as part of the main function`?

YuviGold: First, UUID shouldn't be generated. download_logs module would always work with some cluster_id. It's just not necessarily comes from an argument.
So I think that the definition of the `report_name` should be in a more internal scope.

BTW, this should work with multiple clusters. The `assisted-gather` step now generated 1 report while there are 3 clusters available. There should be 3 reports per cluster

YuviGold (approved):

PR #1034, opened 2021-08-16
Wait for extended time for must-gather logs when hosts are done installation

osherdp:
Right now we're waiting with 3 retries when hosts are in "done" state, and 15 retries when hosts are in "joined" state.
This change will retry more with those cases that we actually more probable to gather must-gather logs.
/cc @rollandf @michaellevy101 @YuviGold


YuviGold (approved):

osherdp:
/retitle Wait for extended time for must-gather logs when hosts are done installation

PR #1039, opened 2021-08-17
MGMT-7577: Retry must-gather collection also when hosts are in Done state

rollandf:
When collecting logs, we make sure that must-gather is present and retry if not only if 2 nodes are in Joined state.

We need to support also when 2 nodes are in Done or Joined.


osherdp (approved):

osherdp:
/retitle MGMT-7577: Retry must-gather collection also when hosts are in Done state

PR #1042, opened 2021-08-18
NO-ISSUE: race in reset hosts

slaviered:


eliorerz:
@slaviered Is that PR still needed?

PR #1046, opened 2021-08-22
NO-ISSUE: raise an error for only relevant error codes

osherdp:
Asserting on return codes which are not 200 is probably not right and assertion message is not indicative enough about the error.
/cc @YuviGold 
/cc @nmagnezi 
/hold


YuviGold (approved):
LGTM

PR #1047, opened 2021-08-22
Fixing destroy_nodes , none platform deploy_node and adding console to none platform vms

tsorya:


scripts/install_environment.sh
```diff
@@ -98,7 +98,7 @@ function install_runtime_container() {
 
 function install_packages() {
     echo "Installing dnf packages"
-    sudo dnf install -y make python3 python3-pip git jq bash-completion xinetd
+    sudo dnf install -y make python3 python3-pip git jq bash-completion xinetd aicli
```
YuviGold, line 101: ?
I think it's a python package.

terraform_files/none/consolemodel.xsl
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" ?>
```
YuviGold, line 1: What does it do? Please elaborate on what's the benefit from adding this feature as part of the PR description

scripts/install_environment.sh
tsorya: oops, you are right

terraform_files/none/consolemodel.xsl
tsorya: there is nothing new , it is just copy paste from BM platform functionality

YuviGold (approved):

tsorya:
/test e2e-metal-assisted-none
--


<br class="Apple-interchange-newline">

PR #1049, opened 2021-08-22
Support dynamic OCP default version

YuviGold:
Return the openshift version that needs to be handled
according to the following process:

1. In case env var `OPENSHIFT_INSTALL_RELEASE_IMAGE` is defined to override the release image -
extract its OCP version.
2. In case env var `OPENSHIFT_VERSION` is defined - return it.
3. In case of `allow_default` is enabled, return the default supported version by assisted-service.
3.1 If a client is provided, request the versions from the service (supports remote service).
3.2 Otherwise, Get from the JSON file in assisted-service repository.

/cc @eliorerz


discovery-infra/test_infra/utils/utils.py
```diff
@@ -685,7 +685,14 @@ def get_kubeconfig_path(cluster_name: str) -> str:
     return kubeconfig_path
 
 
-def get_openshift_version(default=consts.DEFAULT_OPENSHIFT_VERSION):
+def get_default_openshift_version() -> str:
+    stdout, _, _ = run_command(
+        f"jq -r 'to_entries | .[] | select(.value.default) | .key' assisted-service/data/default_ocp_versions.json", shell=True
```
tsorya, line 690: if you run test-infra on remote-url, without spawning minikube and deploying service, you will not get this fille.
I think you should validate it before.
Why to use jq? You are using python , just json.load it

osherdp, line 690: just use python and ``json.load``

discovery-infra/test_infra/utils/utils.py
```diff
@@ -697,13 +704,20 @@ def get_openshift_version(default=consts.DEFAULT_OPENSHIFT_VERSION):
             )
         return stdout
 
-    return get_env("OPENSHIFT_VERSION", default)
+    version = get_env("OPENSHIFT_VERSION")
+    if version:
+        return version
 
+    if allow_default:
+        return get_default_openshift_version()
```
osherdp, line 713: at the end of the function, return ``None`` explicitly

osherdp (changes requested):

discovery-infra/test_infra/utils/utils.py
YuviGold: why? In case no return statement, `None` is returned

osherdp: I know, that's why I said explicit
That's the styling choice that I prefer

discovery-infra/test_infra/utils/utils.py
YuviGold: That's true. So maybe we should use the client versions API instead.

YuviGold:
/hold
In order to support remote clusters, it would be better to fetch the versions from the client.

discovery-infra/tests/conftest.py
```diff
@@ -13,25 +12,14 @@ def api_client():
     yield get_api_client()
 
 
-def get_api_client(offline_token=None, **kwargs) -> InventoryClient:
```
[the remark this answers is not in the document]
YuviGold: https://gitlab.cee.redhat.com/ocp-edge-qe/kni-assisted-installer-auto/-/merge_requests/476

YuviGold:
Everything passed :tada: 
/unhold

discovery-infra/test_infra/utils/global_variables/env_variables_utils.py
```diff
@@ -16,7 +15,7 @@ class _EnvVariablesUtils(ABC):
     remote_service_url: str = get_env("REMOTE_SERVICE_URL")
     pull_secret: str = get_env("PULL_SECRET")
     offline_token: str = get_env("OFFLINE_TOKEN")
-    openshift_version: str = get_openshift_version()
+    openshift_version: str = ''
```
tsorya, line 18: maybe None?

discovery-infra/test_infra/utils/global_variables/global_variables.py
```diff
@@ -44,6 +47,11 @@ class GlobalVariables(_EnvVariablesUtils):
 
     def __post_init__(self):
         super().__post_init__()
+        client=None
+        if not self.is_kube_api:
+            with suppress(RuntimeError, TimeoutError):
```
tsorya, line 52: why it is ok to fail ? please comment

discovery-infra/test_infra/utils/utils.py
```diff
@@ -740,7 +735,31 @@ def get_kubeconfig_path(cluster_name: str) -> str:
     return kubeconfig_path
 
 
-def get_openshift_version(default=consts.DEFAULT_OPENSHIFT_VERSION):
+def get_default_openshift_version(client=None) -> str:
```
tsorya, line 738: not sure regarding utils. Why it is not part of cluster configuration classes?

PR #1057, opened 2021-08-25
MGMT-7473: Add support for infra-env (Late binding)

ori-amizur:
- Add support for infra-env with V2 API.  The infra-env creates nodes that are expected to
  reach 'known-unbound' status
- No bind/unbind support yet.

/cc @yevgeny-shnaidman 
/cc @omertuc


discovery-infra/test_infra/helper_classes/infra_env.py
```diff
@@ -0,0 +1,97 @@
...
+    def prepare_infraenv(self, static_network_config=None, **kwargs):
+        self.update_config(**kwargs)
+        logging.info(f"Preparing for installation with infra-env configurations: infraenv_config={self._config}")
+        self.nodes.controller.log_configuration()
+        if self._config.download_image:
+            if static_network_config:
+                static_network_config = static_network. \
+                    generate_static_network_data_from_tf(self.nodes.controller.tf_folder)
+            else:
+                static_network_config = None
```
yevgeny-shnaidman, line 59: where static_network_config is passed to infra-env?

ori-amizur: fixed

PR #1062, opened 2021-08-29
Update rhcos-live-iso version to 4.8.2 and add retry

eranco74:


discovery-infra/bootstrap_in_place.py
```diff
@@ -74,8 +75,8 @@ def download_live_image(download_path):
     logging.info("Downloading iso to %s", download_path)
     # TODO: enable fetching the appropriate rhcos image
     utils.run_command(
-        f"curl https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/pre-release/"
-        f"4.7.0-rc.2/rhcos-4.7.0-rc.2-x86_64-live.x86_64.iso --retry 5 -o {download_path}")
+        f"curl https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.8/4.8.2/rhcos-live.x86_64.iso"
```
omertuc, line 78: Why is this path even hardcoded? Shouldn't it be received as a parameter?

discovery-infra/bootstrap_in_place.py
```diff
@@ -74,8 +75,8 @@ def download_live_image(download_path):
...
+        f" --retry 5 -o {download_path}")
```
omertuc, line 79: Add `--continue-at -` to retry from last position. Increase retry attempts to around ~10-12. Retries happen with 1 second exponential backoff, so 5 retries is just 1 + 2 + 4 + 8 + 16 (31) seconds, not enough time IMO to let a server recover from failure

discovery-infra/bootstrap_in_place.py
eranco74: Parameter from where?
The issue is that we don't really know what rhcos-live-iso version to use.
We know the OCP version and while it;s reletivly simple to get the rhcos version from the ocp release image, sometimes there isn't a matching live-iso of that version.
This is why it's currently hard-coded.

omertuc: Ack

We should really improve that (not necessarily now) to at-least try and find the closest matching RHCOS version, maybe take some inspiration from how the version bump  https://github.com/openshift-assisted/assisted-installer-deployment/blob/86b3b48479ff046b18e38b7054d0309ded8bd927/tools/update_default_release_versions_to_latest.py does it

Otherwise we might see unexpected failures from this mismatch...

Lemme know if you think it's a good idea and I'll open a ticket

YuviGold: Why not use the versions we use for all the other jobs?
https://github.com/openshift/assisted-service/blob/master/data/default_ocp_versions.json

eranco74: Let's start with this (I mainly care about the retries so we won't just fail).

@YuviGold I had the same idea in mind.
e.g.
`curl -s https://raw.githubusercontent.com/openshift/assisted-service/master/data/default_ocp_versions.json | jq '."4.8".rhcos_image'`

But now that @omertuc suggested to use the bump script directly I'm conflicted,
The script will definitely give us a closer image (when possible) but might cause some noise:
1. The script may fail
2. The new image may change without anyone knowing causing some test failures

So far we used 4.7.0-rc.2 (super old) for the ISO and things worked without issues, so I guess getting the ISO the assisted-installer os using for the OCP release should be good enough.

PR #1064, opened 2021-08-29
Adding status_info validation for wrong boot order and reboot timeout

lalon4:


discovery-infra/test_infra/consts/consts.py
```diff
@@ -189,6 +189,11 @@ class Events:
     PROXY_SETTINGS_CHANGED = "Proxy settings changed"
 
 
+class HostStatusInfo:
+    WRONG_BOOT_ORDER = "Expected the host to boot from disk, but it booted the installation image"
+    REBOOT_TIMEOUT = "Host failed to reboot within timeout"
```
eranco74, line 194: This is just part of the status info:
Full status info is:
```
"Host failed to reboot within timeout, please boot the host from the the OpenShift installation disk $INSTALLATION_DISK. The installation will resume once the host reboot"
```
Same for the wrong boot order, full status info is:
```
Expected the host to boot from disk, but it booted the installation image - please reboot and fix boot order to boot from disk $INSTALLATION_DISK
```
Consider renaming to REBOOT_TIMEOUT_PREFIX

discovery-infra/test_infra/utils/utils.py
```diff
@@ -171,8 +171,8 @@ def set_tfvars(tf_folder, tfvars_json):
         json.dump(tfvars_json, _file)
 
 
-def are_hosts_in_status(hosts, nodes_count, statuses, fall_on_error_status=True):
-    hosts_in_status = [host for host in hosts if host["status"] in statuses]
+def are_hosts_in_status(hosts, nodes_count, statuses, status_info="", fall_on_error_status=True):
+    hosts_in_status = [host for host in hosts if (host["status"] in statuses and status_info in host["status_info"])]
```
eranco74, line 175: This should work.
Consider
```suggestion
    hosts_in_status = [host for host in hosts if (host["status"] in statuses and host["status_info"].startswith(status_info))]
```

discovery-infra/test_infra/consts/consts.py
lalon4: I dont have a way to pass the disk parameter, and anyway i dont think we need to validate the entire string, less maintenance in case wording is changes (same as we do with events). Note that there is also going to be a UI test for it. WDYT?

discovery-infra/test_infra/utils/utils.py
lalon4: this one works well when running test with SNO, however its fail when running multi-node (both tests using same method and strings). I need to deep dive it to understand why, but before doing so, is there a big difference between `startswith` and `status_info in` ? how critical is it?

lalon4: done

PR #1065, opened 2021-08-30
NO-ISSUE: Removed --retry-all-errors due to old curl version in the CI

eranco74:
machine
Added retry decorator in order to overcome SSL_ERROR_SYSCALL

### DIFF
--- a/discovery-infra/bootstrap_in_place.py
+++ b/discovery-infra/bootstrap_in_place.py
@@ -66,6 +66,8 @@ def installer_gather(ip, ssh_key, out_dir):
     os.remove(bundle_file_path) if os.path.exists(bundle_file_path) else None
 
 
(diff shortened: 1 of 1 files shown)


YuviGold (approved):

discovery-infra/bootstrap_in_place.py
```diff
@@ -66,6 +66,7 @@ def installer_gather(ip, ssh_key, out_dir):
     os.remove(bundle_file_path) if os.path.exists(bundle_file_path) else None
 
 
+@utils.retry(exceptions=Exception, tries=3, delay=30)
```
omertuc, line 69: # We retry both with Python and with curl because sometimes `curl` doesn't retry specific errors

osherdp (approved):

PR #1077, opened 2021-09-01
Adding service restart method

cdvultur:
Adding service restart method, for tests that require service restarts on non-persistent changes

### DIFF
--- a/discovery-infra/test_infra/controllers/node_controllers/node.py
+++ b/discovery-infra/test_infra/controllers/node_controllers/node.py
@@ -102,6 +102,10 @@ class Node:
         self.shutdown()
         self.start()
 
+    def restart_service(self, service):
+        logging.info("Restarting service: %s on host %s", service, self.name)
+        self.run_command(f'sudo systemctl restart {service}.service')
+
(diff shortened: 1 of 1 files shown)


discovery-infra/test_infra/controllers/node_controllers/node.py
```diff
@@ -102,6 +102,10 @@ def restart(self):
         self.shutdown()
         self.start()
 
+    def restart_service(self, service):
+        logging.info("Restarting service: %s on host %s", service, self.name)
+        self.run_command(f'sudo systemctl restart {service}.service || true')
```
YuviGold, line 107: I know that is already being used by `is_service_active` and `kill_service` but  I don't like the `|| true`. In case there was a fatal error I'd like to know about it. In case a calling function would like to ignore it, then we can wrap it with a corresponding `supress`.
For example, I'd like to know if the service name I provided isn't available at all.

cdvultur: Fixed it. Thanks!

YuviGold (approved):

PR #1085, opened 2021-09-02
Update OWNERS

YuviGold:
Lately, several times a new code was added without consideration of styling, 
other sections of the code, and other clients that might get affected.

Update the OWNERS file to have approvers that are familiar with the whole assisted-test-infra ecosystem
Still, every OpenShift org member would be able to LGTM but it would require an approver in order to get merged.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -3,13 +3,6 @@
 approvers:
 - tsorya
 - yuvigold
-- romfreiman
-- eranco74
 - osherdp
-- omertuc
-- ori-amizur
-- gamli75
 - michaellevy101
 - eliorerz
-- ybettan
-- lranjbar


eliorerz (approved):
LGTM

flaper87 (commented):
Were there previous conversations about this?
Why do you think we need this PR?

I find the need for this PR a bit counterintuitive as I think everyone in the team should be responsible for providing as accurate reviews as possible and keeping in mind guidelines.

I know there is good intention behind this PR but I think it's solving the wrong issue. I would rather spend time  answering the following questions:

- Are there guidelines to follow? 
- How can we bring folks up-to-speed on those guidelines and styling requirements?
- Is this a review issue? Code issue? Human issue?

Let's talk about it and try to come up with a different approach that we can use for similar issues in the future.

YuviGold:
> Were there previous conversations about this?
> Why do you think we need this PR?

Yes. I think it's pretty much explained in the PR description.

> I find the need for this PR a bit counterintuitive as I think everyone in the team should be responsible for providing as accurate reviews as possible and keeping in mind guidelines.
> 

IMO ideally, regardless of the recent breakages, I think that the approvers group of any repository should include only heavy contributors that are familiar with various processes that the repository is managing.
Not all the people in the team fit these criteria.
It is well defined under the approver definition - https://github.com/kubernetes/community/blob/master/community-membership.md#approver
```
Code approvers are able to both review and approve code contributions.  While
code review is focused on code quality and correctness, approval is focused on
holistic acceptance of a contribution including: backwards / forwards
compatibility, adhering to API and flag conventions, subtle performance and
correctness issues, interactions with other parts of the system, etc.
```

> I know there is good intention behind this PR but I think it's solving the wrong issue. I would rather spend time answering the following questions:
> 
> * Are there guidelines to follow?
> * How can we bring folks up-to-speed on those guidelines and styling requirements?
> * Is this a review issue? Code issue? Human issue?
> 
> Let's talk about it and try to come up with a different approach that we can use for similar issues in the future.

These are good points.
Still, I don't think that having 20 people as approvers is a good approach.

Some actions to follow:
* Having an **up-to-date** `CONTRIBUTING.md`/maintainer docs for every repository.
Explaining the responsibilities and the things that need to be taken into account (from styling, to "holistic acceptance")
* Splitting the code and having a different `OWNERS` per directory.
Currently, we only define approvers&reviewers at the root directory which takes into effect the entire repository. But we can and should split the responsibility per field of expertise. For example, SNO source code can have different approvers than IPv6.

flaper87:
> Yes. I think it's pretty much explained in the PR description.

I'm trying to find where the conversations happened. Whether I agree or not with the motivation behind this cleanup doesn't matter. I'm mostly trying to improve the way we are managing these repos, how we are communicating, and making sure expectations and goals are clear.

I'm not arguing against a tidy `OWNERS` file, I agree with the principle. To apply this principle, however, we have to make sure that all the expectations are clear. That doesn't seem to be the case in this repo.

We have talked multiple times about granular `OWNERS` files, and being more strict with permissions. I just think we are far from that and that the extra process that will add can be solved with a proper set of guidelines and trusting folks in the team won't be merging things they don't have enough knowledge on.

Sure, we are humans, we make mistakes and there is no way around that.

lranjbar:
I think that we do need to have a conversation about what concerns we have and address them as a team. The project is maturing and has needs to meet. This set of approvers does not meet the needs of the project. Work must continue regardless of any specific holidays and thus locking our approvers to one country will not be sufficient. I'm positive management will understand how this will negatively impact timelines.

There are plenty of things we can improve on but that must be communicated and worked through as a team. I will be willing to participate in any discussions around what we need in the future. I do not agree that creating more silos within our team is a healthy course of action.

I understand that change is difficult but one must adapt to change to survive.

YuviGold:
Thank you for your feedback. I was talking with @celebdor about it and we shall be more transparent and communicative about changes like this (additions or subtractions. Any modification to repository ownership)

PR #1088, opened 2021-09-02
Use kubeconfig path that always exists

osherdp:
The previous one we used doesn't always exist: only on the bootstrap and during bootstrapping phase.
/cc @eranco74 @tsorya 

/test e2e-metal-assisted-ipv6
/hold


eranco74 (commented):
/lgtm

osherdp:
/unhold
all looking good with the artifacts

PR #1090, opened 2021-09-02
MGMT-6699 - Move utils libvirt functions to controller. Move unused utils functions to deprecated_utils.py.

eliorerz:
This PR came to do a bit of order in the utils module.
Changes in this PR:

- Move libvirt methods to `LibvirtController` and delete the extra connection from `utils.py`
- Move scripts only functions to separate file - need further discussion about what do we want to do with them.
- New utils module created - `waiting.py` for all the `wait_for_*` methods  - I believe that in the future (after remove their dependencies in the scripts) we will be able to move them inside the `NodeControllers`

Note: Some of the functions left without change due to QE/scripts dependencies.

/cc @YuviGold @tsorya


tsorya:
why we need to remove utils?

eliorerz:
> why we need to remove utils?

We don't. There are many unused utils and utils that are being used only on the old workflow.

tsorya:
so we can delete those that are unused and leave all others? maybe i miss something but i don't understand why to rename file to deprecated utils ?

eliorerz:
> so we can delete those that are unused and leave all others? maybe i miss something but i don't understand why to rename file to deprecated utils ?

I moved the deprecated utilities to new file only because they are still in use on the old scripts (e.g.  `start_discovery.py`).
After replacing all the makefile targets we will be able to delete this file.

PR #1099, opened 2021-09-09
Fix static IPs config option and use a new name

osherdp:
Fixed ``IS_STATIC_IP`` configuration option, and changed its name to just ``STATIC_IPS``.

Seems like ``static_network_config`` option not delivered to ``prepare_for_installation`` so that we always just configured dynamic networking.
Also, we only gather IPs via DHCP server, although libvirt is capable to get it via ARP (which is useful for static IPs).

/cc @eliorerz @YuviGold 
/hold


discovery-infra/test_infra/utils/global_variables/env_variables_utils.py
```diff
@@ -0,0 +1,81 @@
+from abc import ABC
```
YuviGold, line 1: should be removed

YuviGold (approved):

osherdp:
It's still not fixing it all, need to keep working on the fix
/retitle [WIP] Fix static IPs config option and use a new name

osherdp:
/retitle Fix static IPs config option and use a new name

YuviGold (approved):

PR #1101, opened 2021-09-12
MGMT-7173: podman support

michaellevy101:
upgrade podman to 3.2.3 and start podman socket
upgrade skipper to the latest version
work on skipper with podman-remote only

PR #1104, opened 2021-09-13
Revert "MGMT-7173: podman support (#1101)"

michaellevy101:
This reverts commit f4d6c4847b02b1deb803cefa106d8c3ee45f49be.


rollandf:
Why do we revert this?
Please add a reason.

michaellevy101:
> Please add a reason.

@slaviered told me that we have an issue with the logs collection

PR #1105, opened 2021-09-13
Use rpmfind.net/linux/centos/8 path to install libgcrypt

nshidlin:
https://rpmfind.net/linux/centos/8.3.2011/ path is [depricated](https://rpmfind.net/linux/centos/8.3.2011/readme). Using https://rpmfind.net/linux/centos/8 instead


YuviGold (approved):
LGTM

PR #1106, opened 2021-09-13
MGMT-7786: remove openshift_version in kube-api flow

danielerez:
kube-api flow requires only the OS images list (instead of openshift_versions map).
Since assisted-service is using defaults values if OS_IMAGES is missing,
pass empty objects for OPENSHIFT_VERSIONS and RELEASE_IMAGES.

### DIFF
--- a/scripts/deploy_assisted_service.sh
+++ b/scripts/deploy_assisted_service.sh
@@ -14,8 +14,6 @@ export OCP_SERVICE_PORT=$(( 7000 + $NAMESPACE_INDEX ))
 export OPENSHIFT_INSTALL_RELEASE_IMAGE=${OPENSHIFT_INSTALL_RELEASE_IMAGE:-}
 export ENABLE_KUBE_API=${ENABLE_KUBE_API:-false}
 export ENABLE_KUBE_API_CMD="ENABLE_KUBE_API=${ENABLE_KUBE_API}"
-export OPENSHIFT_VERSIONS=${OPENSHIFT_VERSIONS:-}
-export OPENSHIFT_VERSIONS_CMD=""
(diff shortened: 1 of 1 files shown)


danielerez:
/hold
depends on: https://github.com/openshift/assisted-service/pull/2598

rollandf:
:+1:

scripts/deploy_assisted_service.sh
```diff
@@ -25,15 +23,14 @@ export REGISTRY_SERVICE_PORT=80
...
-    OPENSHIFT_VERSIONS_CMD="OPENSHIFT_VERSIONS=$(printf "${json_template}" "${OPENSHIFT_VERSIONS}")"
+if [[ "${ENABLE_KUBE_API}" == "true" || "${DEPLOY_TARGET}" == "operator" ]]; then
+    # Only OS_IMAGES list is required in kube-api flow (assisted-service is using defaults if missing)
+    if [ -z "${OPENSHIFT_VERSIONS:-}" ]; then
+        export OPENSHIFT_VERSIONS='{}'
+    fi
+    if [ -z "${RELEASE_IMAGES:-}" ]; then
+        export RELEASE_IMAGES='[]'
+    fi
 fi
```
YuviGold, line 34: Could this section be deleted entirely?

rollandf: We want to test that dynamic version using ClusterImageSet is working, this is the reason for clearing these.

danielerez: Right, it's for simulating the kube-api/operator scenario, in which only OS_IMAGES var is available.

PR #1112, opened 2021-09-14
MGMT-6763: Add IPv4 config

YuviGold:
/cc @mkowalski 
/cc @eliorerz


mkowalski:
The error seems legit

```
TypeError: Can't instantiate abstract class LibvirtController with abstract methods is_ipv4, is_ipv6, notify_iso_ready, set_dns, set_dns_for_user_managed_network
```

(https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_assisted-test-infra/1112/pull-ci-openshift-assisted-test-infra-master-e2e-metal-assisted/1437806321373024256/build-log.txt)

discovery-infra/test_infra/utils/global_variables/global_variables.py
```diff
@@ -53,12 +52,25 @@ def __post_init__(self):
...
-                log.info(f"{env.upper()} is triggered. Updating global variables: {values}")
+        for conditions, values in _triggers.items():
+            # A dictionary doesn't support lists so the only way to verify
+            # multi-level is by checking the inner value is also a tuple
+            if isinstance(conditions[0], tuple):
+                if all(map(lambda condition: self.is_set(condition[0], condition[1]), conditions)):
+                    self._handle_trigger(conditions, values)
+            else:
+                if self.is_set(conditions[0], conditions[1]):
+                    self._handle_trigger(conditions, values)
```
osherdp, line 63: In case of not making it all the same type (see comment above), just generalizing it a bit:
```suggestion
            if not isinstance(conditions[0], tuple):
                conditions = (conditions,)

            if all(map(lambda condition: self.is_set(condition[0], condition[1]), conditions)):
                self._handle_trigger(conditions, values)
```

discovery-infra/test_infra/utils/global_variables/global_variables.py
```diff
@@ -53,12 +52,25 @@ def __post_init__(self):
...
+
+    def is_set(self, var, expected):
+        return getattr(self, var) == expected
+
+    def _handle_trigger(self, conditions, values):
+        for k, v in values.items():
+            self._set(k, v)
+
+        self._triggered.append(conditions)
+        log.info(f"{conditions} is triggered. Updating global variables: {values}")
```
osherdp, line 73: It took me a while to understand what happens here
Your changes are better than the previous code, so I don't think more than that is needed to be done in the scope of the current PR

discovery-infra/test_infra/utils/global_variables/global_variables.py
```diff
@@ -53,12 +52,25 @@ def __post_init__(self):
...
 
-        for (env, expected), values in _triggers.items():
-            if getattr(self, env) == expected:
-                for k, v in values.items():
-                    self._set(k, v)
-                self._triggered.append(env)
-                log.info(f"{env.upper()} is triggered. Updating global variables: {values}")
+        for conditions, values in _triggers.items():
+            # A dictionary doesn't support lists so the only way to verify
+            # multi-level is by checking the inner value is also a tuple
```
osherdp, line 57: We can make all the conditions as tuple of tuples, no?

osherdp (changes requested):

discovery-infra/test_infra/utils/global_variables/global_variables.py
YuviGold: made the same type

YuviGold:
> The error seems legit
> 
> ```
> TypeError: Can't instantiate abstract class LibvirtController with abstract methods is_ipv4, is_ipv6, notify_iso_ready, set_dns, set_dns_for_user_managed_network
> ```
> 
> (https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_assisted-test-infra/1112/pull-ci-openshift-assisted-test-infra-master-e2e-metal-assisted/1437806321373024256/build-log.txt)

@mkowalski 
Fixed

skipper.env
```diff
@@ -60,6 +60,8 @@ E2E_TESTS_MODE
 LOG_FOLDER
 IPv4
 IPv6
+IS_IPV4
+IS_IPV6
```
osherdp, line 64: I actually don't like that there's a couple of ways to do the same thing
Can we conform with one convention?
Also, I don't like the addition of the ``IS`` which feels redundant

osherdp: btw, moving to all caps is a really good thing you're doing here, which might be a good convention to stick to everywhere

osherdp (changes requested):

skipper.env
YuviGold: I removed the latter. Let's keep `IPv4` notation and discuss whether to change it to all-caps.

osherdp (approved):

PR #1113, opened 2021-09-19
Ignore libvirt errors when trying to get domain ips and macs

YuviGold:
Sometimes the interfaces aren't available

```
  File "/home/assisted/discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py", line 459, in _get_domain_ips_and_macs
    interfaces = dict(domain.interfaceAddresses(libvirt.VIR_DOMAIN_INTERFACE_ADDRESSES_SRC_LEASE))
  File "/usr/local/lib64/python3.6/site-packages/libvirt.py", line 1657, in interfaceAddresses
    raise libvirtError('virDomainInterfaceAddresses() failed')
libvirt.libvirtError: Requested operation is not valid: domain is not running
```

We should try to get all the available IPs and ignore failures in that case.

/cc @osherdp

### DIFF
--- a/discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
+++ b/discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
@@ -455,10 +455,18 @@ class LibvirtController(NodeController, ABC):
 
     @staticmethod
     def _get_domain_ips_and_macs(domain):
-        # getting all DHCP leases IPs
-        interfaces = dict(domain.interfaceAddresses(libvirt.VIR_DOMAIN_INTERFACE_ADDRESSES_SRC_LEASE))
-        # getting static IPs via ARP
-        interfaces.update(domain.interfaceAddresses(libvirt.VIR_DOMAIN_INTERFACE_ADDRESSES_SRC_ARP))
+        interfaces_sources = [
+            # getting all DHCP leases IPs
+            libvirt.VIR_DOMAIN_INTERFACE_ADDRESSES_SRC_LEASE,
+
+            # getting static IPs via ARP
+            libvirt.VIR_DOMAIN_INTERFACE_ADDRESSES_SRC_ARP,
+        ]
+
+        interfaces = {}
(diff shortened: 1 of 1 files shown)


discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
```diff
@@ -455,10 +455,11 @@ def get_node_ips_and_macs(self, node_name):
 
     @staticmethod
     def _get_domain_ips_and_macs(domain):
-        # getting all DHCP leases IPs
-        interfaces = dict(domain.interfaceAddresses(libvirt.VIR_DOMAIN_INTERFACE_ADDRESSES_SRC_LEASE))
-        # getting static IPs via ARP
-        interfaces.update(domain.interfaceAddresses(libvirt.VIR_DOMAIN_INTERFACE_ADDRESSES_SRC_ARP))
+        interfaces = dict()
```
osherdp, line 458: nit
```suggestion
        interfaces = {}
```

discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py
```diff
@@ -455,10 +455,11 @@ def get_node_ips_and_macs(self, node_name):
...
+
+        for addresses_source in (libvirt.VIR_DOMAIN_INTERFACE_ADDRESSES_SRC_LEASE, libvirt.VIR_DOMAIN_INTERFACE_ADDRESSES_SRC_ARP):
```
osherdp, line 460: copying comment from the other PR
I think we should leave the comments in some form, because it's not that immediate as for why we need those two methods
libvirt documentation is also not that great, so we should leave the original intent in the code imho

osherdp (changes requested):

osherdp (approved):

osherdp:
/test e2e-metal-assisted-static-ip
/hold
(to explore produced sosreports)

osherdp:
/unhold
(all looking good)

PR #1128, opened 2021-10-03
MGMT-7483: use release images defaults

danielerez:
Use default_release_images instead of default_ocp_versions for extracting ocp version and release image URL.


danielerez:
/hold
pending on https://github.com/openshift/assisted-service/pull/2690

discovery-infra/override_release_images.py
```diff
@@ -0,0 +1,38 @@
...
+
+    # Update release image
+    release_image["openshift_version"] = ocp_version
+    release_image["url"] = os.getenv("OPENSHIFT_INSTALL_RELEASE_IMAGE")
+    release_image["version"] = ocp_full_version
+    release_images[release_image_index] = release_image
+
+    # Save release images
+    with open(consts.OCP_RELEASE_IMAGES_PATH, 'w') as f:
+        f.write(json.dumps(release_images))
```
osherdp, line 34: I find it problematic that we need to change source code of assisted-service
For example, someone that overrides the first time, will have to remember to remove changes on the next run that doesn't require override 
Can't we have the value stored in OPENSHIFT_VERSIONS instead of changing the original values?

osherdp (changes requested):

discovery-infra/test_infra/utils/utils.py
```diff
@@ -479,17 +477,28 @@ def get_openshift_version(allow_default=True, client=None) -> str:
     return None
 
 
+def get_full_openshift_version_from_release(release_image: str) -> str:
+        with pull_secret_file() as pull_secret:
```
osherdp, line 481: Too much indentation

osherdp (changes requested):

discovery-infra/override_release_images.py
danielerez: Yeah, added the result to RELEASE_IMAGES env var instead.

discovery-infra/test_infra/utils/utils.py
danielerez: done

osherdp (approved):

osherdp (approved):

PR #1134, opened 2021-10-05
MGMT-7900: Test infra proxy may fail if install-config contains Proxy (Capitalized) instead of proxy

ori-amizur:
- Add support for both forms
NO-ISSUE: Add acl for all IPv6 addresses (::/0) in addition to the existing 0.0.0.0/0
- For squid to allow proxying of IPv6 traffic

/cc @eliorerz 
/cc @YuviGold


YuviGold:
Why would the name be different?

YuviGold (approved):

PR #1135, opened 2021-10-05
MGMT-7902: Test infra IPv6 static ips not working

ori-amizur:
When using test infra, in order to match the hostname the actual host in
the cluster, test-infra is using interfaceAddresses method on libvirt
domain object, which provides the macs and ips for the domain (VM).
Libirt supports 2 methods to gather this information: DHCP and ARP.  ARP
only exists for IPv4, and no DHCP information exists when using static
ips.
To retrieve this information, the tfstate file is parsed.  It already
contains the hostname and the mac addresses of the interfaces.  By
matching the mac addresses in the inventory, the hostname and roles can
be set correctly for this configuration.

/cc @YuviGold 
/cc @eliorerz


discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -703,11 +703,11 @@ def _set_hostnames_and_roles(self):
             if host.has_hostname():
                 continue
 
-            node = Cluster.find_matching_node(host, nodes)
-            assert node is not None, f"Failed to find matching node for host with mac address {host.macs()} nodes: {[(n.name, n.ips, n.macs) for n in nodes]}"
```
YuviGold, line 707: Why? It's more verbose to know what was available

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -740,7 +740,10 @@ def prepare_for_installation(self, **kwargs):
             )
 
         self.nodes.notify_iso_ready()
-        self.nodes.start_all()
+        if self._config.is_static_ip and self._config.is_ipv6:
+            self.nodes.start_all_no_check_ips()
+        else:
+            self.nodes.start_all()
```
YuviGold, line 746: What about changing the `start_all` to `check_ips=False` by default? we should try to minimize the different behavior or different configurations.

discovery-infra/test_infra/tools/static_network.py
```diff
@@ -71,6 +71,22 @@ def generate_static_network_data_from_tf(tf_folder: str) -> List[Dict[str, List[
     return network_config
 
 
+def get_name_to_mac_addresses_mapping(tf_folder: str) -> Dict[str, Set[str]]:
+    tfstate_json_file = os.path.join(tf_folder, "terraform.tfstate")
```
YuviGold, line 75:
```suggestion
    tfstate_json_file = os.path.join(tf_folder, TerraformUtils.STATE_FILE)
```

discovery-infra/test_infra/tools/static_network.py
```diff
@@ -71,6 +71,22 @@ def generate_static_network_data_from_tf(tf_folder: str) -> List[Dict[str, List[
...
+    with open(tfstate_json_file) as _file:
+        tfstate = json.load(_file)
+    ret = dict()
+    for resource in tfstate["resources"]:
+        if resource["mode"] == "managed" and resource["type"] == "libvirt_domain":
+            for domain in resource["instances"]:
+                attributes = domain["attributes"]
+                macs = set()
+                for intf in attributes["network_interface"]:
+                    macs.add(intf["mac"].lower())
```
YuviGold, line 85: I think that instead of returning a lower version here, you should validate in the calling function that both are lower case
```
            for mac in map(str.lower, host.macs()):
                for name, macs in mappings.items():
                    if mac in map(str.lower, macs):
                        return name
```

discovery-infra/test_infra/helper_classes/cluster.py
ori-amizur: Didn't want to change default behavior as part of this PR

discovery-infra/test_infra/tools/static_network.py
ori-amizur: OK

discovery-infra/test_infra/tools/static_network.py
ori-amizur: [](https://github.com/openshift/assisted-test-infra/blob/223158e520161517ac2c68b88db486a357c014da/discovery-infra/test_infra/helper_classes/cluster_host.py#L31)

discovery-infra/test_infra/helper_classes/cluster.py
eliorerz: @ori-amizur You don't have to.
Instead of:

```python
    def start_all(self):
        self.run_for_all_nodes("start")

    def start_all_no_check_ips(self):
        self.run_for_all_nodes("start", False)
```
You can:

```python
    def start_all(self, check_ips=True):
        self.run_for_all_nodes("start", check_ips)
```

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -1127,13 +1130,20 @@ def _wait_for_api_vip(self, hosts, timeout=180):
             lambda: self.get_kube_api_ip(hosts=hosts), timeout_seconds=timeout, sleep_seconds=5, waiting_for="API's IP"
         )
 
-    @staticmethod
-    def find_matching_node(host: ClusterHost, nodes: List[Node]):
+    def find_matching_node_name(self, host: ClusterHost, nodes: List[Node]):
```
eliorerz, line 1133: Please add return value type annotation

discovery-infra/test_infra/helper_classes/nodes.py
```diff
@@ -99,6 +99,9 @@ def notify_iso_ready(self):
     def start_all(self):
         self.run_for_all_nodes("start")
 
+    def start_all_no_check_ips(self):
+        self.run_for_all_nodes("start", False)
+
```
eliorerz, line 104:
```suggestion
    def start_all(self, check_ips=True):
        self.run_for_all_nodes("start", check_ips)
```

discovery-infra/test_infra/helper_classes/cluster.py
ori-amizur: OK

mkowalski:
/approve

For LGTM let's address the pending reviews and we are good to go 🥐

PR #1136, opened 2021-10-06
Support REST API dual-stack

YuviGold:
A lot of changes in order to support `IPv4=true IPv6=true`.

- Remove env vars for `SERVICE_CIDR`, `CLUSTER_CIDR`, `CLUSTER_HOST_PREFIX` as
  no one uses them.
- Support sending multiple values for `cluster_networks`,
  `service_networks`, and `machine_networks`.
- Add constants for IPv4, IPv6, and IPv4v6 subnets.

Screenshot from 2021-10-06 18-32-55

/cc @eliorerz @mkowalski @ori-amizur  @tsorya


eliorerz (approved):
LGTM

PR #1140, opened 2021-10-10
Move general cluster APIs to use v2

nshidlin:


discovery-infra/test_infra/assisted_service_api.py
```diff
@@ -429,14 +429,15 @@ def get_openshift_versions(self) -> models.OpenshiftVersions:
     def get_supported_operators(self) -> List[str]:
         return self.operators.list_supported_operators()
 
+# TODO remove
```
osherdp, line 432: we should do it pretty close in time, but anyway add a reason as to "why" we want to remove it

osherdp (changes requested):

PR #1142, opened 2021-10-10
MGMT-7920: enable testing TPM disk encryption mode

osherdp:
This should include virtual TPM device which is accessible from all hosts.
The selection of disk encryption mode and on which nodes it should apply is via environment variables ``DISK_ENCRYPTION_MODE`` and ``DISK_ENCRYPTION_ROLES`` respectively.

I couldn't find a way to include TPM only on selected roles, especially because of libvirt terraform provider.

/cc @ybettan @eliorerz 
/assign @eliorerz 
/hold


README.md
```diff
@@ -132,6 +132,8 @@ Check the [Install Guide](GUIDE.md) for installation instructions.
 | `NUM_WORKERS`                 | number of VMs to spawn as workers, default: 0 |
 | `OPENSHIFT_VERSION`           | OpenShift version to install, default: "4.7" |
 | `HYPERTHREADING`              | Set node's CPU hyperthreading mode. Values are: all, none, masters, workers. default: all |
+| `DISK_ENCRYPTION_MODE`        | Set disk encryption mode. Right now assisted-test-infra only supports "tpmv2". |
```
ybettan, line 135: If we use tpmv2 as default then we should mention what is the default (I am not sure we want a default in that case though)

discovery-infra/test_infra/consts/env_defaults.py
```diff
@@ -27,4 +27,6 @@
 DEFAULT_IMAGE_FOLDER: Path = Path(consts.IMAGE_FOLDER)
 DEFAULT_IMAGE_FILENAME: str = "installer-image.iso"
 DEFAULT_NETWORK_TYPE: str = consts.NetworkType.OpenShiftSDN
+DEFAULT_DISK_ENCRYPTION_MODE: str = consts.DiskEncryptionMode.TPM_VERSION_2
```
ybettan, line 30: Does that mean we will enable TPM in bios even when we run a non-encrypted installation flow?

scripts/install_environment.sh
```diff
@@ -13,8 +13,31 @@ function version_is_greater() {
...
-    sudo dnf install -y libvirt libvirt-devel libvirt-daemon-kvm qemu-kvm libgcrypt
+    sudo setenforce enforcing && sudo dnf install -y \
+        libvirt \
+        libvirt-devel \
+        libvirt-daemon-kvm \
+        qemu-kvm \
+        libgcrypt \
+        swtpm \
+        swtpm-tools \
+            && sudo setenforce permissive
```
ybettan, line 39: Same as https://github.com/openshift/assisted-test-infra/pull/1142/files#r725829881

osherdp: still under investigation, I hope to find a proper solution for the installation part

README.md
osherdp: I see
I can add a new mode ``"none"`` that will be a default, and that keeps this object empty then. I'll implement that when your PR is in
Also I added one comment on that PR, which I'm not sure is an issue but I'll let you check it

scripts/install_environment.sh
```diff
@@ -13,10 +13,36 @@ function version_is_greater() {
...
+        qemu-kvm \
+        libgcrypt
+
     sudo systemctl enable libvirtd
 
+    echo "Installing vTPM packages..."
+    sudo setenforce enforcing && sudo dnf install -y \
+        swtpm \
+        swtpm-tools \
+            && sudo setenforce permissive
```
osherdp, line 44: validate with QE it works for them

ybettan:
> have it by default and risk any user failing because of missing dependencies

Why should users fail? you have added the dependencies to `scripts/install_environment.sh`

> Wait for libvirt terraform provider to release a new version that supports it (it's already existing on master, just need to wait for the release)

Another option will be to fork it and have our own release until we can switch back to their new release that contains the module.

> Give up on the console logs that QE added with the XML transformation, because we can either use the transformers for getting the console log into a file or for the vTPM

Not sure I understand this one

mkowalski:
As we are using dmacvicar/libvirt, why not just

```
      tpm {
        backend_type    = "emulator"
        backend_version = "2.0"
      }
```

directly in the resource instead of all this custom transformation ?

The commit introducing it (https://github.com/dmacvicar/terraform-provider-libvirt/commit/f3edad343c6d97cf2eeee8c7116f8624de274558) does not seem to be released yet, but I think it's better to wait for the provider and use the clean way instead of applying the custom transformation

osherdp:
> As we are using dmacvicar/libvirt, why not just
> 
> ```
>       tpm {
>         backend_type    = "emulator"
>         backend_version = "2.0"
>       }
> ```
> 
> directly in the resource instead of all this custom transformation ?
> 
> The commit introducing it ([dmacvicar/terraform-provider-libvirt@f3edad3](https://github.com/dmacvicar/terraform-provider-libvirt/commit/f3edad343c6d97cf2eeee8c7116f8624de274558)) does not seem to be released yet, but I think it's better to wait for the provider and use the clean way instead of applying the custom transformation



> As we are using dmacvicar/libvirt, why not just
> 
> ```
>       tpm {
>         backend_type    = "emulator"
>         backend_version = "2.0"
>       }
> ```
> 
> directly in the resource instead of all this custom transformation ?
> 
> The commit introducing it ([dmacvicar/terraform-provider-libvirt@f3edad3](https://github.com/dmacvicar/terraform-provider-libvirt/commit/f3edad343c6d97cf2eeee8c7116f8624de274558)) does not seem to be released yet, but I think it's better to wait for the provider and use the clean way instead of applying the custom transformation

yes, I'm aware of this change (also joined the discussion on the PR to have it merged :) ) and also prefer this
but I decided to leave things working until we have this feature in a released version

terraform_files/baremetal/main.tf
```diff
@@ -1,8 +1,13 @@
...
   required_providers {
+    # TODO: revert to upstream package when newer version than v0.6.11 has been released
+    # libvirt = {
+    #   source = "dmacvicar/libvirt"
+    #   version = "0.6.9"
+    # }
     libvirt = {
-      source = "dmacvicar/libvirt"
-      version = "0.6.9"
+      source = "osherdp/libvirt"
```
ybettan, line 9: Great!!

discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
```diff
@@ -52,6 +52,22 @@ def _create_tf_folder(self, name: str, platform: str):
 
         return os.path.join(tf_folder, consts.Platforms.BARE_METAL)
 
+    def _get_disk_encryption_appliance(self):
+        assert (
+            self._entity_config.disk_encryption_mode == consts.DiskEncryptionMode.TPM_VERSION_2
```
eliorerz, line 57: It's only exists on `BaseClusterConfig`.
This class can also get `BaseInfraEnvConfig`.
To my opinion you can:

1. Add it to `BaseInfraEnvConfig` (e.g. `BaseEntityConfig`) if it's also related to InfraEnv
2. Verify it has this property first in this function (not in `_get_specific_tf_entity_params`)
3. Pass `BaseInfraEnvConfig` to the function instead of using it via self

discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
```diff
@@ -52,6 +52,22 @@ def _create_tf_folder(self, name: str, platform: str):
...
+        ), "Currently only supporting TPMv2"
+
+        master_vtpm2 = worker_vtpm2 = False
+
+        if self._entity_config.disk_encryption_roles == consts.DiskEncryptionRoles.ALL:
```
eliorerz, line 62: Same here

discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
```diff
@@ -52,6 +52,22 @@ def _create_tf_folder(self, name: str, platform: str):
...
+            master_vtpm2 = worker_vtpm2 = True
+        elif self._entity_config.disk_encryption_roles == consts.DiskEncryptionRoles.MASTERS:
```
eliorerz, line 64: Same here

discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
```diff
@@ -52,6 +52,22 @@ def _create_tf_folder(self, name: str, platform: str):
...
+            master_vtpm2 = True
+        elif self._entity_config.disk_encryption_roles == consts.DiskEncryptionRoles.WORKERS:
```
eliorerz, line 66: Same here

discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
```diff
@@ -52,6 +52,22 @@ def _create_tf_folder(self, name: str, platform: str):
 
         return os.path.join(tf_folder, consts.Platforms.BARE_METAL)
 
+    def _get_disk_encryption_appliance(self):
```
eliorerz, line 55: What about `VSphereController`? We don't want to support it in [vsphere](https://docs.vmware.com/en/VMware-vSphere/7.0/com.vmware.vsphere.security.doc/GUID-3D39CBA6-E5B2-43E2-A596-B9A69B094558.html)?

eliorerz (changes requested):

discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
osherdp: The emulator only works on libvirt

discovery-infra/test_infra/controllers/node_controllers/terraform_controller.py
osherdp: I'll make it agnostic to the config it uses
thanks!

osherdp:
/test e2e-metal-assisted-none
seems like we sometimes have machines that don't boot, will have to look at it at some point

eliorerz (approved):
LGTM

PR #1150, opened 2021-10-13
MGMT-7957 - `test_kube_api` fix to use newer `ClusterConfig` members

omertuc:
aec453f5fcf1da37e6171940df71fef1bc96cf66 changed the format of the `ClusterConfig`
class but forgot to update all usages of it.

For now this is a quick fix just to get the kube-api CI working again, it is
not an entirely correct fix because it has poor support for dual network stack
clusters.

This may be improved in future PRs

### DIFF
--- a/discovery-infra/test_infra/utils/global_variables/global_variables.py
+++ b/discovery-infra/test_infra/utils/global_variables/global_variables.py
@@ -29,6 +29,10 @@ _triggers = frozendict(
             "master_memory": resources.DEFAULT_MASTER_SNO_MEMORY,
             "master_vcpu": resources.DEFAULT_MASTER_SNO_CPU,
         },
+        (("is_ipv4", True), ("is_ipv6", False),): {
+            "cluster_networks": consts.DEFAULT_CLUSTER_NETWORKS_IPV4,
+            "service_networks": consts.DEFAULT_SERVICE_NETWORKS_IPV4,
+        },
         (("is_ipv4", False), ("is_ipv6", True),): {
             "cluster_networks": consts.DEFAULT_CLUSTER_NETWORKS_IPV6,
             "service_networks": consts.DEFAULT_SERVICE_NETWORKS_IPV6,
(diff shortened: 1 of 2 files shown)


discovery-infra/test_infra/utils/global_variables/global_variables.py
```diff
@@ -57,12 +62,14 @@ def __post_init__(self):
         self._set("openshift_version", utils.get_openshift_version(allow_default=True, client=client))
 
         for conditions, values in _triggers.items():
-            assert isinstance(conditions, tuple) and isinstance(conditions[0], tuple), f"Key {conditions} must be tuple of tuples"
-            if all(map(lambda condition: self.is_set(condition[0], condition[1]), conditions)):
+            assert isinstance(conditions, tuple) and all(isinstance(condition, tuple)
```
omertuc, line 65: this change is just some refactoring / cleaning up, no interesting bug fixes here

discovery-infra/test_infra/utils/global_variables/global_variables.py
```diff
@@ -29,6 +29,11 @@
             "master_memory": resources.DEFAULT_MASTER_SNO_MEMORY,
             "master_vcpu": resources.DEFAULT_MASTER_SNO_CPU,
         },
+        (("is_ipv4", True), ("is_ipv6", False),): {
```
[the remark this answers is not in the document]
eliorerz: I think those are the default values, No?

discovery-infra/tests/test_kube_api.py
```diff
@@ -213,9 +213,9 @@ def kube_api_test_prepare_late_binding_cluster(kube_api_context, cluster_config:
     agent_cluster_install.create(
         cluster_deployment_ref=cluster_deployment.ref,
         image_set_ref=deploy_image_set(cluster_name, kube_api_context),
-        cluster_cidr=cluster_config.cluster_network_cidr,
-        host_prefix=cluster_config.cluster_network_host_prefix,
-        service_network=cluster_config.service_network_cidr,
+        cluster_cidr=cluster_config.cluster_networks[0].cidr,
+        host_prefix=cluster_config.cluster_networks[0].host_prefix,
+        service_network=cluster_config.service_networks[0].cidr,
```
eliorerz, line 218: I think it is better to add getters on `cluster_config.py` and use them on all the other places:
e.g. 

```python
class BaseClusterConfig(BaseEntityConfig, ABC):
    ...

    @property
    def network_cidr(self):
        return self.cluster_networks[0].cidr if self.cluster_networks else None

    @property
    def network_host_prefix(self):
        return self.cluster_networks[0].host_prefix if self.cluster_networks else None

    @property
    def service_network_cidr(self):
        return self.service_networks[0].cidr if self.service_networks else None

```

discovery-infra/test_infra/utils/global_variables/global_variables.py
```diff
@@ -29,6 +29,11 @@
...
+            "cluster_networks": consts.DEFAULT_CLUSTER_NETWORKS_IPV4,
+            "service_networks": consts.DEFAULT_SERVICE_NETWORKS_IPV4,
+            "network_type": consts.NetworkType.OVNKubernetes
```
osherdp, line 35: Isn't it a change from the previous default?

osherdp (changes requested):

osherdp (approved):

omertuc:
\o/

PR #1153, opened 2021-10-14
Moving events to v2

iovadia:

/cc @osherdp @eliorerz @lalon4 @nshidlin

### DIFF
--- a/discovery-infra/test_infra/helper_classes/cluster.py
+++ b/discovery-infra/test_infra/helper_classes/cluster.py
@@ -926,8 +926,8 @@ class Cluster:
         cluster.wait_until_hosts_are_discovered()
         cluster.wait_for_ready_to_install()
 
-    def get_events(self, host_id=""):
-        return self.api_client.get_events(cluster_id=self.id, host_id=host_id)
+    def get_events(self, host_id="", infra_env_id=""):
+        return self.api_client.get_events(cluster_id=self.id, host_id=host_id, infra_env_id=infra_env_id)
 
     def _configure_load_balancer(self):
         main_cidr = self.get_primary_machine_cidr()
@@ -950,8 +950,8 @@ class Cluster:
         matcher = re.match(r"^tt(\d+)$", libvirt_network_if)
         return int(matcher.groups()[0]) if matcher is not None else 0
 
-    def _find_event(self, event_to_find, reference_time, params_list, host_id):
-        events_list = self.get_events(host_id=host_id)
+    def _find_event(self, event_to_find, reference_time, params_list, host_id, infra_env_id):
+        events_list = self.get_events(host_id=host_id, infra_env_id=infra_env_id)
         for event in events_list:
             if event_to_find in event["message"]:
                 # Adding a 2 sec buffer to account for a small time diff between the machine and the time on staging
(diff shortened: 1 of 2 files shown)


discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -950,8 +950,8 @@ def _get_namespace_index(cls, libvirt_network_if):
         matcher = re.match(r"^tt(\d+)$", libvirt_network_if)
         return int(matcher.groups()[0]) if matcher is not None else 0
 
-    def _find_event(self, event_to_find, reference_time, params_list, host_id):
-        events_list = self.get_events(host_id=host_id)
+    def _find_event(self, event_to_find, reference_time, params_list, host_id, infra_env_id):
```
eliorerz, line 953: Add type hint annotations and return type annotations

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -963,19 +963,19 @@ def _find_event(self, event_to_find, reference_time, params_list, host_id):
         else:
             return False
 
-    def wait_for_event(self, event_to_find, reference_time, params_list=None, host_id="", timeout=10):
+    def wait_for_event(self, event_to_find, reference_time, params_list=None, host_id="",infra_env_id="", timeout=10):
```
eliorerz, line 966: Add type hint annotations

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -926,8 +926,8 @@ def reset_cluster_and_wait_for_ready(cls, cluster):
         cluster.wait_until_hosts_are_discovered()
         cluster.wait_for_ready_to_install()
 
-    def get_events(self, host_id=""):
-        return self.api_client.get_events(cluster_id=self.id, host_id=host_id)
+    def get_events(self, host_id="", infra_env_id=""):
```
eliorerz, line 929: Add type hint annotations and return type annotations

eliorerz:
I'm not sure events should stay under `Cluster` ass it was before.
On v2 `events` endpoint moved to independent endpoint and no longer is under `cluster`.
I think we should create new object to handle the events.
@nmagnezi @tsorya @nshidlin WDYT?

tsorya:
> I'm not sure events should stay under `Cluster` as it was before. On v2 `events` endpoint moved to independent endpoint and it's no longer under `cluster`. I think we should create new object to handle the events.
> 
> e.g.
> 
> ```python
> class EventsHandler:
>     def __init__(self, api_client: InventoryClient, cluster_id: str, infra_env_id: str):
>         self.api_client = api_client
>         self.infra_env_id = infra_env_id
>         self.cluster_id = cluster_id
> 
>     def _find_event(self, event_to_find, reference_time, params_list, host_id):
>         ...
>     def get_events(self, host_id=""):
>         ...
> 
>     def wait_for_event(self, event_to_find, reference_time, params_list=None, host_id="", timeout=10):
>         ...
> ```
> 
> @nmagnezi @tsorya @nshidlin WDYT?

I don't think we need events out of the cluster scope so i don't see any reason to move 2 small api out of it right now

nmagnezi (changes requested):
> I'm not sure events should stay under `Cluster` as it was before. On v2 `events` endpoint moved to independent endpoint and it's no longer under `cluster`. I think we should create new object to handle the events.
> 
> e.g.
> 
> ```python
> class EventsHandler:
>     def __init__(self, api_client: InventoryClient, cluster_id: str, infra_env_id: str):
>         self.api_client = api_client
>         self.infra_env_id = infra_env_id
>         self.cluster_id = cluster_id
> 
>     def _find_event(self, event_to_find, reference_time, params_list, host_id):
>         ...
>     def get_events(self, host_id=""):
>         ...
> 
>     def wait_for_event(self, event_to_find, reference_time, params_list=None, host_id="", timeout=10):
>         ...
> ```
> 
> @nmagnezi @tsorya @nshidlin WDYT?

IMO, you make a good point. In V1 all events had `cluster-id`, but now this is no longer the case. For example, host events that are linked to `infra_env`, but might not be bounded to a cluster at this point.

https://github.com/openshift/assisted-service/blob/cc39df66bb6431e7843d3be828fdc9e4a731939c/docs/events.yaml#L35-L41

eliorerz:
> In that case lets move them with another jira ticket. Right now there is no use of events without cluster id in the code. Lets finish this task and move them to new object when we will need it

OK,  don't think it requires much of a change but I opened a ticket for doing it https://issues.redhat.com/browse/MGMT-7977

eliorerz (approved):

PR #1139, opened 2021-10-10
MGMT-7928: Use infrenv api to generate and download image

nshidlin:
This PR:

- uses infraenv to create and download cluster image
- moves api client to use v2_register_cluster to create a new cluster
- use infraenv to update host roles and names

PR #1154, opened 2021-10-15
Pass set cluster proxy config when generating infraenv, to fix ipv6 flow

nshidlin:
This PR is to fix ipv6 test that was broken in https://github.com/openshift/assisted-test-infra/pull/1139, as the nodes would not be discovered if proxy values are not set in infraenv.

/cc @eliorerz @mkowalski


discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -152,8 +157,16 @@ def download_infra_env_image(self, iso_download_path = None) -> None:
         self._infra_env.download_image(iso_download_path=iso_download_path)
 
     @JunitTestCase()
-    def generate_and_download_infra_env(self, iso_download_path = None, static_network_config=None, iso_image_type=None, ssh_key=None, ignition_info=None):
-        self.generate_infra_env(static_network_config=static_network_config, iso_image_type=iso_image_type, ssh_key=ssh_key, ignition_info=ignition_info)
+    def generate_and_download_infra_env(
```
eliorerz, line 160: I think we should add type annotations here

discovery-infra/test_infra/helper_classes/config/base_entity_config.py
```diff
@@ -24,3 +25,5 @@ class BaseEntityConfig(_BaseConfig, ABC):
     is_ipv6: bool = None
     base_dns_domain: str = None
     entity_name: BaseName = None
+    proxy: models.Proxy = None
```
eliorerz, line 28: Why do we need it here? Why not pass it when it's needed

discovery-infra/test_infra/helper_classes/cluster.py
mkowalski: The already defined function had no annotations, so I'm eager to say that we should just make this PR as small as possible in order to get the stuff working again and do the refactoring in the follow-up PR. With the CI broken already for a week I'd rather fix it as soon as possible with as small changes as possible

mkowalski:
/test e2e-metal-assisted-ipv6

The job that we are fighting for with this PR has failed with 'SyncLoopRefreshProgressing: Working toward version 4.8.12, 2 replicas available'. It is different than the initial error `waiting.exceptions.TimeoutExpired: Timeout of 1200 seconds expired waiting for Nodes to be in of the statuses ['pending-for-input', 'known', 'insufficient']` so let's retry and see if it was an intermittent failure or if the same error appears again

discovery-infra/test_infra/helper_classes/cluster.py
nshidlin: ack. will add annotations in seperate PR

discovery-infra/test_infra/helper_classes/config/base_entity_config.py
nshidlin: It's used both in Cluster and Infraenv configs. but i'll add it to each explicitly

osherdp (approved):

osherdp:
/hold
(for letting other reviews to be done)

discovery-infra/test_infra/helper_classes/config/base_entity_config.py
nshidlin: @eliorerz moved back into BaseEntityConfig

PR #1157, opened 2021-10-17
Reverting test_parallel has 4 test processes

eliorerz:
Since changing `TEST_WORKERS_NUM` to 4 we observed flaky behavior (Timeout) that can be caused by lack of resources on the executing machines.
This PR reverting `TEST_WORKERS_NUM` to run with 3 test workers

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -492,4 +492,4 @@ _test_setup:
 	rm -f /tmp/tf_network_pool.json
 
 _test_parallel: $(REPORTS) _test_setup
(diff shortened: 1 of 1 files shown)


osherdp:
interesting theory, let's see how it runs
/lgtm
/hold

eliorerz:
@osherdp what do you think about using `dist=load` on pytest?

osherdp:
But looks like it's still failing, no?

eliorerz:
There is no indication that e2e-metal-assisted-multiversion is failing due to lack of resources

PR #1159, opened 2021-10-18
revert PR #1136 (dual stack support)

osherdp:
Trying to make jobs pass again by reverting the breaking changes
/cc @osherdp 
/hold


eliorerz:
Why removing it?

PR #1173, opened 2021-10-20
NO-ISSUE: define build-root-image in the repository itself

osherdp:
So hopefully, this should allow us to try and use the Dockerfile as it's defined in the PR, and not just take it from master.
It should be followed with changing configuration in ``release`` to use build-root configuration from the repository.

This follows configuration from:
https://docs.ci.openshift.org/docs/architecture/ci-operator/#build-root-image

/cc @eliorerz 
/hold

### DIFF
--- a/.ci-operator.yaml
+++ b/.ci-operator.yaml
@@ -1,0 +1,4 @@
+build_root_image:
+  namespace: openshift
+  name: release
+  tag: golang-1.16


osherdp:
/override ci/prow/e2e-metal-assisted-olm
(it has nothing to do with those jobs, so just overriding them)

PR #1175, opened 2021-10-20
MGMT-8030: Updating host disk selection to use infraenv

nshidlin:


discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -244,8 +244,8 @@ def set_cluster_name(self, cluster_name: str):
         self.update_config(cluster_name=ClusterName(prefix=cluster_name, suffix=None))
         self.api_client.update_cluster(self.id, {"name": cluster_name})
 
-    def select_installation_disk(self, hosts_with_disk_paths):
-        self.api_client.select_installation_disk(self.id, hosts_with_disk_paths)
+    def select_installation_disk(self, host_id: str, disk_paths: List[dict]) -> None:
```
eliorerz, line 247: It doesn't need to be moved to `infra_env`?

nshidlin: the api call was moved into infraenv. but i as long as cluster holds infra-env it should call `self._infra_env.select_host_installation_disk`

eliorerz (approved):

PR #1195, opened 2021-10-28
MGMT-8087 - Show k8s events failures on prow spyglass

eliorerz:
/cc @osherdp 

/hold waiting for https://github.com/openshift/assisted-test-infra/pull/1194


discovery-infra/junit_log_parser.py
```diff
@@ -83,6 +84,40 @@ def export_service_logs_to_junit_suites(source_dir: Path, report_dir: Path):
...
+def export_service_events_to_junit_suite(source_dir: Path, report_dir: Path, events_file_name="k8s_events.json"):
+    test_cases = list()
+    with open(source_dir.joinpath(events_file_name)) as f:
+        events_data = json.load(f)
+
+    log.info(f"Creating test suite from service events json file - {events_file_name}")
+    for event in events_data["items"]:
+        test_case = get_event_test_case(event)
+        if test_case:
+            test_cases.append(test_case)
```
osherdp, line 111:
```
test_cases = list(get_test_cases(events_data["items"]))
```
and let ``get_test_cases`` be a generator that yields only relevant entries

discovery-infra/junit_log_parser.py
```diff
@@ -83,6 +84,40 @@ def export_service_logs_to_junit_suites(source_dir: Path, report_dir: Path):
...
+
+    log.info(f"Generating events xml file")
+    test_suite = [TestSuite(name="EVENTS", test_cases=test_cases)]
+    xml_report = to_xml_report_string(test_suites=test_suite)
+    with open(report_dir.joinpath(f"junit_events_parser_{str(uuid.uuid4())[:8]}.xml"), "w") as f:
+        log.info(f"Exporting events xml-report with {len(xml_report)} characters to {f.name}")
```
osherdp, line 117: that characters say anything?
maybe change with this?
```suggestion
        log.info(f"Exporting events xml-report with {len(test_cases)} events to {f.name}")
```

discovery-infra/junit_log_parser.py
```diff
@@ -83,6 +84,40 @@ def export_service_logs_to_junit_suites(source_dir: Path, report_dir: Path):
...
+        test_case = TestCase(name=obj, classname="EVENT", category=reason, timestamp=timestamp)
+        test_case.failures.append(CaseFailure(message=message, output=message, type=event_type))
+        return test_case
+    return None
+
+
+def export_service_events_to_junit_suite(source_dir: Path, report_dir: Path, events_file_name="k8s_events.json"):
+    test_cases = list()
+    with open(source_dir.joinpath(events_file_name)) as f:
+        events_data = json.load(f)
```
osherdp, line 105: also here, why should we save all ``events_data`` in memory?

discovery-infra/junit_log_parser.py
```diff
@@ -83,6 +84,40 @@ def export_service_logs_to_junit_suites(source_dir: Path, report_dir: Path):
...
+    event_type = event["type"]
+    if event_type.lower() in EXPORTED_EVENT_LEVELS:
+        log.info(f"Adding {event_type} event as test-case")
+        reason = event["reason"]
+        involved_object = event["involvedObject"]
+        obj = involved_object["kind"].lower() + "/" + involved_object["name"]
+        message = event["message"]
+        timestamp = event["firstTimestamp"]
+        test_case = TestCase(name=obj, classname="EVENT", category=reason, timestamp=timestamp)
+        test_case.failures.append(CaseFailure(message=message, output=message, type=event_type))
```
osherdp, line 97: this doesn't yet treats them as flakes, right?

osherdp (changes requested):

discovery-infra/junit_log_parser.py
eliorerz: From a quick look it's doesn't look like there are a lot of failures on events as in the service log. If there will be more then expected we will treats them as flaky. WDYT?

discovery-infra/junit_log_parser.py
eliorerz: It's cannot be done. Only with `JSON lines format` (we can't convert it without open the entire file)

discovery-infra/junit_log_parser.py
eliorerz: I can't see the benefit for setting `get_test_cases` to be a generator. To my opinion its making the code less readable.   Anyway, after rebaseing the logs_parser the file has changed a bit.

discovery-infra/junit_log_parser.py
```diff
@@ -43,66 +44,111 @@ class LogEntry:
...
+    @classmethod
+    def export_service_events_to_junit_suite(cls, source_dir: Path, report_dir: Path, events_file_name="k8s_events.json"):
+        with open(source_dir.joinpath(events_file_name)) as f:
+            events_data = json.load(f)
+
+        log.info(f"Creating test suite from service events json file - {events_file_name}")
+        test_cases = cls.get_event_test_cases(events_data)
+
+        log.info(f"Generating events xml file")
+        test_suite = [TestSuite(name="EVENTS", test_cases=test_cases)]
```
osherdp, line 147:
```suggestion
        test_suites = [TestSuite(name="EVENTS", test_cases=test_cases)]
```

discovery-infra/junit_log_parser.py
```diff
@@ -43,66 +44,111 @@ class LogEntry:
...
-        suites.append(TestSuite(name=suite_name, test_cases=test_cases, timestamp=timestamp))
-
-    log.info(f"Generating xml file for {len(suites)} suites")
-    xml_report = to_xml_report_string(suites)
-    with open(report_dir.joinpath(f"junit_log_parser_{str(uuid.uuid4())[:8]}.xml"), "w") as f:
-        log.info(f"Exporting {len(suites)} suites xml-report with {len(xml_report)} characters to {f.name}")
-        f.write(xml_report)
+class LogsConverter:
+
+    @staticmethod
```
osherdp, line 49: I'm in favor to make this ``classmethod`` instead
Even if it doesn't require the ``cls`` parameter, basically ``staticmethod``s are not something I think should exist

discovery-infra/junit_log_parser.py
```diff
@@ -43,66 +44,111 @@ class LogEntry:
...
+    def is_duplicate_entry(entry: LogEntry, entry_message: str, fail_cases: Dict[str, List[TestCase]]) -> bool:
```
osherdp, line 50:
```suggestion
    def _is_duplicate_entry(entry: LogEntry, entry_message: str, fail_cases: Dict[str, List[TestCase]]) -> bool:
```
it will make it more clear which methods should be used from outside and which shouldn't 
same for other methods in this file

discovery-infra/junit_log_parser.py
```diff
@@ -43,66 +44,111 @@ class LogEntry:
...
+        for case in fail_cases.get(entry.func, []):
+            if case.failures and case.failures[0].message == entry_message:
+                return True
+        return False
+
+    @classmethod
+    def get_log_entry_case(cls, entry: LogEntry, fail_cases: Dict[str, List[TestCase]], suite_name: str) -> \
+            List[TestCase]:
+        fail_case: List[TestCase] = list()
+        message = f"{entry.msg}\n{entry.error if entry.error else ''}"
```
osherdp, line 60:
```suggestion
        message = f"{entry.msg}\n{entry.error}"
```
if it's empty then it will have the same effect with this shortened syntax

discovery-infra/junit_log_parser.py
```diff
@@ -43,66 +44,111 @@ class LogEntry:
...
+
+        with open(log_file_name) as f:
+            for line in f:
+                values = re.match(LOG_FORMAT, line)
+                if values is None:
+                    continue
+
+                entry = LogEntry(**values.groupdict())
+                if entry.level not in EXPORTED_LOG_LEVELS:
+                    continue
```
osherdp, line 88: nit: add blank line

discovery-infra/junit_log_parser.py
```diff
@@ -43,66 +44,111 @@ class LogEntry:
...
+                if entry.func not in fail_cases:
+                    fail_cases[entry.func] = list()
```
osherdp, line 90: nit: add blank line

discovery-infra/junit_log_parser.py
```diff
@@ -43,66 +44,111 @@ class LogEntry:
...
+            f.write(xml_report)
+
+
+class EventsConverter:
+    @classmethod
+    def get_event_test_cases(cls, events_data: dict) -> List[TestCase]:
+        test_cases = list()
+        for event in events_data["items"]:
+            test_case = cls.get_event_test_case(event)
+            test_cases.append(test_case) if test_case else None
```
osherdp, line 119:
```suggestion
            if test_case is not None:
                test_cases.append(test_case)
```

discovery-infra/junit_log_parser.py
```diff
@@ -43,66 +44,111 @@ class LogEntry:
...
+        if event_type.lower() in EXPORTED_EVENT_LEVELS:
+            log.info(f"Adding {event_type} event as test-case")
+            reason = event["reason"]
+            involved_object = event["involvedObject"]
+            obj = involved_object["kind"].lower() + "/" + involved_object["name"]
+            message = event["message"]
+            timestamp = event["firstTimestamp"]
+            test_case = TestCase(name=f"{reason}: {obj}", classname="EVENT", category=reason, timestamp=timestamp)
+            test_case.failures.append(CaseFailure(message=message, output=message, type=event_type))
+            return test_case
```
osherdp, line 135: nit: add blank line

osherdp: or rather have the ``if event_type.lower() in EXPORTED_EVENT_LEVELS`` short-circuited

osherdp (changes requested):

discovery-infra/junit_log_parser.py
eliorerz: If entry.error is empty it's None by default and that will cause  `str(None) --> "None"`

discovery-infra/junit_log_parser.py
eliorerz: Me too, but in the past I was told to set it to static :)

osherdp (approved):

PR #1202, opened 2021-10-31
NO-ISSUE - Fix junit log parser adding multiple success cases

eliorerz:


discovery-infra/junit_log_parser.py
```diff
@@ -58,10 +59,11 @@ def get_failure_cases(log_file_name: Path, suite_name: str) -> List[TestCase]:
             fail_cases.append(test_case)
 
             # test is flaky
-            if entry.level != "fatal":
+            if entry.level != "fatal" and entry.func not in test_case_keys:
```
osherdp, line 62: let's do something else
define ``LogEntry`` as follows:
```
@dataclass
class LogEntry:
    time: str = None
    level: str = None
    msg: str = None
    func: str = dataclasses.field(hash=True)
    file: str = None
```
then you can have just ``fail_cases = set()``, adding items to it without fearing duplications

osherdp (changes requested):

discovery-infra/junit_log_parser.py
eliorerz: I don't want that fail cases will be set, I just want to add the success case once (because that it is only for marking the test as flaky).
Note that the same function can fail multiple times and each failure is test case.

eliorerz: Also, I thin for using set I have to also override `__hash__`

osherdp: ``dataclasses.field(hash=True)`` automatically implements ``__hash__``
let me take another look at the code, I think we can still make it simpler

eliorerz: Sure .. simpler is better  :)

```python
import dataclasses

@dataclasses.dataclass
class A:
    a: str = dataclasses.field(hash=True)
    b: str
    c: str

a = A("1", "2", "3")
b = A("5", "2", "3")
c = A("5", "2", "3")

print({a, b, c})
>>>>> TypeError: unhashable type: 'A'
```

discovery-infra/junit_log_parser.py
```diff
@@ -40,8 +40,34 @@ class LogEntry:
 EXPORTED_EVENT_LEVELS = ("critical", "error")
 
 
+def is_duplicate_entry(entry: LogEntry, fail_cases: Dict[str, List[TestCase]]):
```
osherdp, line 43: return type annotation

discovery-infra/junit_log_parser.py
```diff
@@ -40,8 +40,34 @@ class LogEntry:
...
+    if entry.func in fail_cases:
+        for case in fail_cases[entry.func]:
+            if case.failures and case.failures[0].message == entry.msg:
+                return True
+    return False
+
+
+def get_log_entry_case(entry: LogEntry, fail_cases: Dict[str, List[TestCase]], suite_name: str):
```
osherdp, line 51: return type annotation

discovery-infra/junit_log_parser.py
```diff
@@ -40,8 +40,34 @@ class LogEntry:
 EXPORTED_EVENT_LEVELS = ("critical", "error")
 
 
+def is_duplicate_entry(entry: LogEntry, fail_cases: Dict[str, List[TestCase]]):
+    if entry.func in fail_cases:
+        for case in fail_cases[entry.func]:
```
osherdp, line 45:
```suggestion
    for case in fail_cases.get(entry.func, []):
```

osherdp (changes requested):

discovery-infra/junit_log_parser.py
osherdp: ah, it needs ``@dataclasses.dataclass(frozen=True)`` apparently

osherdp (approved):

PR #1203, opened 2021-11-01
NO-ISSUE: disable VIP DHCP allocation when using OVNKubernetes

osherdp:
When using ``NETWORK_TYPE=OVNKubernetes`` the cluster cannot become ready.
That leaves the cluster in a state that has no apparent affect, so we should change the default mode we're working with.

/cc @eliorerz


eliorerz (approved):
LGTM

PR #1213, opened 2021-11-03
Adding a retry on tf init

lalon4:
From time to time we are failing to init tf during regression runs i.e
"Failed to\nrequest discovery document: Get\n"https://registry.terraform.io/.well-known/terraform.json": dial tcp: lookup\nregistry.terraform.io on 127.0.0.1:53: read udp 127.0.0.1:53546->127.0.0.1:53:\ni/o timeout\n\x1b[0m\x1b[0m\n'"

This PR add a retry on tf init.
We have same retry on other flows in our QE Jenkins and seems like this retry helps to reduce the occurrence of such failure

### DIFF
--- a/discovery-infra/test_infra/tools/terraform_utils.py
+++ b/discovery-infra/test_infra/tools/terraform_utils.py
@@ -6,7 +6,8 @@ from builtins import list
 from typing import Any, Dict, List
 
 import hcl2
-from python_terraform import IsFlagged, Terraform, Tfstate
+from retry import retry
+from python_terraform import IsFlagged, Terraform, Tfstate, TerraformCommandError
 
 
 class TerraformUtils:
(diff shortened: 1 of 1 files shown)


discovery-infra/test_infra/tools/terraform_utils.py
```diff
@@ -22,6 +23,7 @@ def __init__(self, working_dir: str, terraform_init: bool = True):
         if terraform_init:
             self.init_tf()
 
+    @retry(exceptions=Exception, tries=10, delay=10)
```
osherdp, line 26:
```suggestion
    @retry(exceptions=python_terraform.TerraformCommandError, tries=10, delay=10)
```
I'm in favor of being more specific at first, so in case we have more basic error (syntax, or just another bug of ours) we will fail immediately

osherdp (changes requested):
Just one little thing

discovery-infra/test_infra/tools/terraform_utils.py
lalon4: done!

osherdp (approved):

discovery-infra/test_infra/tools/terraform_utils.py
osherdp: Thank you! ☺️

PR #1214, opened 2021-11-04
MGMT-7977 - Move test-infra events from Cluster to separate object

eliorerz:
See https://github.com/openshift/assisted-test-infra/pull/1153 for more information and motivation
/cc @osherdp

FYI @nshidlin @lalon4 
Deprecated methods on this PR:

1. `Cluster.get_events` - Now can be used as part of `EventsHandler`, note that cluster_id is no longer part of this class.
2. `Cluster.wait_for_event` - Now can be used as part of `EventsHandler`, note that cluster_id is no longer part of this class.
3. `Cluster.EVENTS_THRESHOLD` - Will permanently be removed after will no longer be in use (if needed on specific test - it can be declared there)

### DIFF
--- a/discovery-infra/test_infra/helper_classes/cluster.py
+++ b/discovery-infra/test_infra/helper_classes/cluster.py
@@ -24,6 +24,7 @@ from test_infra.controllers.node_controllers import Node
 from test_infra.helper_classes.cluster_host import ClusterHost
 from test_infra.helper_classes.config import BaseClusterConfig, BaseInfraEnvConfig
 from test_infra.helper_classes.entity import Entity
+from test_infra.helper_classes.events_handler import EventsHandler
 from test_infra.helper_classes.infra_env import InfraEnv
 from test_infra.helper_classes.nodes import Nodes
 from test_infra.tools import static_network, terraform_utils
@@ -33,7 +34,7 @@ from test_infra.utils.entity_name import ClusterName
 
 class Cluster(Entity):
     MINIMUM_NODES_TO_WAIT = 1
-    EVENTS_THRESHOLD = 500
+    EVENTS_THRESHOLD = 500  # TODO - remove EVENTS_THRESHOLD after removing it from kni-assisted-installer-auto
 
     _config: BaseClusterConfig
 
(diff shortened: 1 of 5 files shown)


discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -33,6 +34,9 @@
 
 class Cluster(Entity):
     MINIMUM_NODES_TO_WAIT = 1
+
+    # todo - remove EVENTS_THRESHOLD after removing it from
```
osherdp, line 38:
```suggestion
    # TODO - remove EVENTS_THRESHOLD after removing it from
```

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -33,6 +34,9 @@
...
+    #  https://gitlab.cee.redhat.com/ocp-edge-qe/kni-assisted-installer-auto/-/blob/master/api_tests/test_events.py#L255
```
osherdp, line 39:
```suggestion
```
not available from outside, so I think we should remove it from here

discovery-infra/test_infra/helper_classes/events_handler.py
```diff
@@ -0,0 +1,58 @@
...
+        try:
+            waiting.wait(
+                lambda: self._find_event(event_to_find, reference_time, params_list, host_id, infra_env_id, cluster_id),
+                timeout_seconds=timeout,
+                sleep_seconds=2,
+                waiting_for=f"Event: {event_to_find}",
+            )
+        except waiting.exceptions.TimeoutExpired:
+            log.error(f"Event: {event_to_find} didn't found")
+            raise
```
osherdp, line 58:
```suggestion
        waiting.wait(
            lambda: self._find_event(event_to_find, reference_time, params_list, host_id, infra_env_id, cluster_id),
            timeout_seconds=timeout,
            sleep_seconds=2,
            waiting_for=f"event {event_to_find}",
        )
```

osherdp (changes requested):

osherdp (approved):

PR #1229, opened 2021-11-09
MGMT-8193 - Remove download_image logic from assisted_service_api.py and use infra_env.download_url to download it

eliorerz:
From [MGMT-8193](https://issues.redhat.com/browse/MGMT-8193):

> Currently whenever we need to download the discovery iso during a CI test we're using some API client which assumes we're using the download endpoint on assisted service.
> 
> When we move to the image service, that download url will no longer work and our tests will fail even though the flow for users would be unchanged.
> 
> We should be relying on the same url we present to users when we run CI (the one in the cluster image info and infraEnv).
> 

Deprecated methods in this PR:

1. `assisted_service_client._download`
2. `assisted_service_client.download_infraenv_image`
3. `assisted_service_client.download_image`

/cc @carbonin @osherdp 

FYI @lalon4 @nshidlin

### DIFF
--- a/discovery-infra/test_infra/assisted_service_api.py
+++ b/discovery-infra/test_infra/assisted_service_api.py
@@ -166,6 +166,12 @@ class InventoryClient(object):
 
     @classmethod
     def _download(cls, response: HTTPResponse, file_path: str, verify_file_size=False) -> None:
+        warnings.warn(
+            "_download is deprecated and soon will be deleted. "
+            "Use infra_env.download_image or cluster.download_image instead.",
+            DeprecationWarning,
+        )
+
         with open(file_path, "wb") as f:
             shutil.copyfileobj(response, f)
         if verify_file_size:
@@ -194,6 +200,12 @@ class InventoryClient(object):
 
     @retry(exceptions=RuntimeError, tries=2, delay=3)
     def download_image(self, cluster_id: str, image_path: str) -> None:
+        warnings.warn(
+            "download_image is deprecated and soon will be deleted. "
+            "Use infra_env.download_image or cluster.download_image instead.",
+            DeprecationWarning,
+        )
+
         log.info("Downloading image for cluster %s to %s", cluster_id, image_path)
         response = self.client.download_cluster_iso_with_http_info(cluster_id=cluster_id, _preload_content=False)
         response_obj = response[0]
@@ -207,6 +219,12 @@ class InventoryClient(object):
         image_type: str = consts.ImageType.FULL_ISO,
         static_network_config: Optional[list] = None,
     ):
+        warnings.warn(
+            "generate_and_download_image is deprecated and soon will be deleted. "
+            "Use infra_env.download_image or cluster.generate_and_download_image instead.",
+            DeprecationWarning,
+        )
+
         self.generate_image(
             cluster_id=cluster_id, ssh_key=ssh_key, image_type=image_type, static_network_config=static_network_config
         )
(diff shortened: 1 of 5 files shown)


nshidlin:
asked on the jira, but asking here again. is the `/v2/infra-envs/{infra_env_id}` path being removed all together?

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -167,22 +167,23 @@ def generate_and_download_infra_env(
         ssh_key=None,
         ignition_info=None,
         proxy=None,
-    ):
+    ) -> Path:
+        if self._config.is_static_ip:
```
nshidlin, line 171: strange behavior to except `static_network_config` and then override it. should be `self._config.is_static_ip and not static_network_config`

eliorerz: Yeah, probably missed that. It was late 😅

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -751,17 +752,14 @@ def _ha_not_none(self):
...
-            static_network_config = None
-
-        self.generate_and_download_infra_env(
-            iso_download_path=iso_download_path or self._config.iso_download_path,
-            iso_image_type=self._config.iso_image_type,
-            static_network_config=static_network_config,
-        )
+    def download_image(self, iso_download_path: str = None) -> Path:
+        if self._infra_env is None:
+            log.warning("No infra_env found. Generating infra_env and downloading ISO")
```
carbonin, line 757: I don't think finding or not finding the infraEnv should change this behavior.
It's still valid to call generate for v1 and update the infraEnv in v2.

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -167,22 +167,23 @@ def generate_and_download_infra_env(
         ssh_key=None,
         ignition_info=None,
         proxy=None,
-    ):
+    ) -> Path:
+        if self._config.is_static_ip and static_network_config is None:
```
carbonin, line 171: What does this have to do with changing how we download the image?

discovery-infra/test_infra/assisted_service_api.py
```diff
@@ -166,6 +166,12 @@ def get_infra_env_by_cluster_id(self, cluster_id: str) -> List[models.infra_env.
 
     @classmethod
     def _download(cls, response: HTTPResponse, file_path: str, verify_file_size=False) -> None:
+        warnings.warn(
```
carbonin, line 169: Instead of deprecating these, does it make sense to just change their implementation?

discovery-infra/test_infra/helper_classes/cluster.py
eliorerz: I moved that from `download_image` to here.
It's needed for generating the `infra_env`

discovery-infra/test_infra/assisted_service_api.py
eliorerz: I thought about it but we will anyway need to change the function signature (to pass infra_env or infra_env arguments).
Also, it's not really needed here, now we can download image directly from `infra_env`  using `infra_env.download_image`.
I would just delete it if no one was using it

discovery-infra/test_infra/helper_classes/cluster.py
carbonin: Not sure I fully understand, but I don't think it can hurt :man_shrugging:

carbonin:
Looks good, but did you want that last test to pass? Do you think those failures are caused by this change?

eliorerz:
> Looks good, but did you want that last test to pass? Do you think those failures are caused by this change?

Thanks, I don't think that issue is related..
see https://search.ci.openshift.org/?search=+Timeout+of+360+seconds+expired+waiting+for+Waiting+for+Ips&maxAge=168h&context=1&type=bug%2Bjunit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

discovery-infra/test_infra/helper_classes/cluster.py
eliorerz: It does generate the image. As far as I know, creating infra_env generates also an ISO

PR #1196, opened 2021-10-28
NO-ISSUE: Update minikube to 1.23.2 and pin OCP client to stable-4.8

jordigilh:
* Updates minikube to the latest version [1.23.2](https://github.com/kubernetes/minikube/releases/tag/v1.23.2) as I started facing the known issue where the IP of the host was <pending> when waiting for the assisted-service to be ready:
```
python3 ./tools/wait_for_assisted_service.py --target local --namespace "assisted-installer" \
	--domain "" --apply-manifest True
Wait for http://<pending>:8090/ready
Wait for http://<pending>:8090/ready
```

Using the 1.23.2 version I am now able to see the IP reflected:
```
python3 ./tools/wait_for_assisted_service.py --target local --namespace "assisted-installer" \
	--domain "" --apply-manifest True
Wait for http://10.99.47.174:8090/ready
Wait for http://10.99.47.174:8090/ready
```

* Pins the OCP client to use the [stable-4.8](https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/stable-4.8/) branch. Currently it points to `4.8.0-rc.0`. This change guarantees that the build will install the latest stable 4.8 client when not found. 

@osherdp @tsorya can you help to review at your discretion?

PR #1231, opened 2021-11-10
NO-ISSUE: fix installation of oc with the right artifact

osherdp:
Seems like on #1196 we installed by accident openshift-installer instead of oc.
/cc @eliorerz @eranco74 @jordigilh 
/hold


osherdp:
/unhold
seems to work great

(those are unrelated to the issue)

PR #1232, opened 2021-11-14
MGMT-7983 - Fix networking job

eliorerz:


osherdp (approved):

osherdp:
/hold
(just so we can see networking job status)

PR #1234, opened 2021-11-15
MGMT-8181: added day2 cloud flow to e2e tests

danielerez:
As e2e flows are now invoked using pytest, fixed day2 on cloud flow to accommodate for the new infra:
* Added test_day2.py - initialize and execute day2_cloud_flow
* Note: For simplicity, used day2.py mostly as-is. We can rewrite and move its logic into test_day2 later on. Also, should add support for the ocp flow.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -239,7 +239,7 @@ _run_terraform:
 		terraform apply -auto-approve -input=false -state=terraform.tfstate -state-out=terraform.tfstate -var-file=terraform.tfvars.json
 
 _apply_terraform:
-		cd build/terraform/$(CLUSTER_NAME)__$(NAMESPACE) && \
+		cd build/terraform/$(CLUSTER_NAME)/$(PLATFORM) && \
 		terraform apply -auto-approve -input=false -state=terraform.tfstate -state-out=terraform.tfstate -var-file=terraform.tfvars.json
 
 destroy_terraform:
--- a/discovery-infra/day2.py
+++ b/discovery-infra/day2.py
@@ -33,9 +33,7 @@ def execute_day2_ocp_flow(cluster_id, args, has_ipv6):
 def execute_day2_flow(cluster_id, args, day2_type_flag, has_ipv6):
     utils.recreate_folder(consts.IMAGE_FOLDER, force_recreate=False)
 
-    client = ClientFactory.create_client(url=utils.get_assisted_service_url_by_args(args=args),
-                                         offline_token=utils.get_env("OFFLINE_TOKEN"))
-
+    client = args.api_client
     cluster = client.cluster_get(cluster_id=cluster_id)
     cluster_name = cluster.name
     openshift_version = cluster.openshift_version
@@ -46,26 +44,24 @@ def execute_day2_flow(cluster_id, args, day2_type_flag, has_ipv6):
         terraform_cluster_dir_prefix = f"{consts.CLUSTER_PREFIX}-{consts.DEFAULT_NAMESPACE}"
     else:
         cluster_id = str(uuid.uuid4())
-        copy_proxy_from_cluster = cluster
         cluster = client.create_day2_cluster(
             cluster_name + "-day2", cluster_id, **_day2_cluster_create_params(openshift_version, api_vip_dnsname)
         )
         set_cluster_pull_secret(client, cluster_id, args.pull_secret)
-        set_cluster_proxy(client, cluster_id, copy_proxy_from_cluster, args)
+        set_cluster_proxy(client, cluster_id, args)
 
     config_etc_hosts(api_vip_ip, api_vip_dnsname)
     image_path = os.path.join(
         consts.IMAGE_FOLDER,
-        f'{args.namespace}-installer-image.iso'
+        f'{cluster_name}-installer-image.iso'
     )
 
-    tf_folder = os.path.join(utils.get_tf_folder(terraform_cluster_dir_prefix, args.namespace),
-                             consts.Platforms.BARE_METAL)
-    set_day2_tf_configuration(tf_folder, args.number_of_day2_workers, api_vip_ip, api_vip_dnsname)
+    tf_folder = os.path.join(utils.get_tf_folder(terraform_cluster_dir_prefix), consts.Platforms.BARE_METAL)
+    set_day2_tf_configuration(tf_folder, args.num_day2_workers, api_vip_ip, api_vip_dnsname)
 
     static_network_config = None
     if args.with_static_network_config:
-        static_network_config = static_network.generate_day2_static_network_data_from_tf(tf_folder, args.number_of_day2_workers)
+        static_network_config = static_network.generate_day2_static_network_data_from_tf(tf_folder, args.num_day2_workers)
 
     client.generate_and_download_image(
         cluster_id=cluster.id,
@@ -80,9 +76,7 @@ def execute_day2_flow(cluster_id, args, day2_type_flag, has_ipv6):
         tf_folder,
         cluster,
         has_ipv6,
-        args.number_of_day2_workers,
-        api_vip_ip,
-        api_vip_dnsname,
+        args.num_day2_workers,
         args.install_cluster,
         day2_type_flag,
         args.with_static_network_config,
@@ -96,17 +90,18 @@ def day2_nodes_flow(client,
                     cluster,
                     has_ipv_6,
                     num_worker_nodes,
-                    api_vip_ip,
-                    api_vip_dnsname,
                     install_cluster_flag,
                     day2_type_flag,
                     with_static_network_config,
                     base_cluster_name):
     tf_network_name, total_num_nodes = get_network_num_nodes_from_tf(tf_folder)
-    with utils.file_lock_context():
-        utils.run_command(
-            f'make _apply_terraform CLUSTER_NAME={terraform_cluster_dir_prefix}'
-        )
+
+    # Running twice as a workaround for an issue with terraform not spawning a new node on first apply.    
+    for _ in range(2):
+        with utils.file_lock_context():
+            utils.run_command(
+                f'make _apply_terraform CLUSTER_NAME={terraform_cluster_dir_prefix} PLATFORM={consts.Platforms.BARE_METAL}'
+            )
     time.sleep(5)
 
     if day2_type_flag == "ocp":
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -296,6 +296,12 @@ run_full_flow_with_ipv6:
 # Day 2 #
 #########
 
+install_day1_and_day2_cloud: start_load_balancer
+	TEST_TEARDOWN=no TEST=./discovery-infra/tests/test_targets.py TEST_FUNC=test_target_install_with_deploy_day2_nodes_cloud $(MAKE) test
+
+deploy_day2_nodes: start_load_balancer
+	TEST_TEARDOWN=no TEST=./discovery-infra/tests/test_targets.py TEST_FUNC=test_target_deploy_day2_nodes_cloud $(MAKE) test
+
```
osherdp, line 304: we're going lately for total deprecation of specialized makefile targets
meaning just using ``make test`` or ``make test_parallel`` with their matching arguments

Makefile
```diff
@@ -355,9 +361,6 @@ deploy_nodes: start_load_balancer
 deploy_static_network_config_nodes:
 	make deploy_nodes ADDITIONAL_PARAMS="'--with-static-network-config'"
 
-deploy_day2_nodes:
-	skipper make $(SKIPPER_PARAMS) _deploy_nodes NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE) $(OC_FLAG)) NAMESPACE=$(NAMESPACE) $(SKIPPER_PARAMS) ADDITIONAL_PARAMS="'--day2-cloud-cluster'"
-
 deploy_day2_cloud_nodes_with_install:
 	skipper make $(SKIPPER_PARAMS) _deploy_nodes NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE) $(OC_FLAG)) NAMESPACE=$(NAMESPACE) $(SKIPPER_PARAMS) ADDITIONAL_PARAMS="'-in --day2-cloud-cluster ${ADDITIONAL_PARAMS}'" DEPLOY_TARGET=minikube
```
osherdp, line 365: we need this?

Makefile
```diff
@@ -370,9 +373,6 @@ deploy_static_network_config_day2_nodes:
 deploy_static_network_config_day2_nodes_with_install:
 	skipper make $(SKIPPER_PARAMS) _deploy_nodes NAMESPACE_INDEX=$(shell bash scripts/utils.sh get_namespace_index $(NAMESPACE) $(OC_FLAG)) NAMESPACE=$(NAMESPACE) $(SKIPPER_PARAMS) ADDITIONAL_PARAMS="'-in --day2-cloud-cluster --with-static-network-config'"
```
osherdp, line 374: we need this?
it seems like the original target with ``IS_STATIC_IP=true``

discovery-infra/test_infra/helper_classes/cluster.py
```diff
@@ -69,6 +69,10 @@ def iso_download_path(self):
     def enable_image_download(self):
         return self._config.download_image
 
+    @property
+    def config(self):
+        return self._config
+
```
osherdp, line 75: Shouldn't it be a property of the parent class?

discovery-infra/day2.py
```diff
@@ -300,15 +296,17 @@ def _day2_cluster_create_params(openshift_version, api_vip_dnsname):
...
+def set_cluster_proxy(client, cluster_id, args):
     """
     Set cluster proxy - copy proxy configuration from another (e.g. day 1) cluster,
     or allow setting/overriding it via command arguments
     """
-    http_proxy = args.http_proxy if args.http_proxy else copy_proxy_from_cluster.http_proxy
-    https_proxy = args.https_proxy if args.https_proxy else copy_proxy_from_cluster.https_proxy
-    no_proxy = args.no_proxy if args.no_proxy else copy_proxy_from_cluster.no_proxy
-    client.set_cluster_proxy(cluster_id, http_proxy, https_proxy, no_proxy)
+    http_proxy = args.http_proxy if hasattr(args, "http_proxy") else None
```
osherdp, line 304:
```suggestion
    http_proxy = getattr(args, "http_proxy", None)
```

discovery-infra/day2.py
```diff
@@ -96,16 +90,18 @@ def day2_nodes_flow(client,
...
     tf_network_name, total_num_nodes = get_network_num_nodes_from_tf(tf_folder)
     with utils.file_lock_context():
         utils.run_command(
-            f'make _apply_terraform CLUSTER_NAME={terraform_cluster_dir_prefix}'
+            f'make _apply_terraform CLUSTER_NAME={terraform_cluster_dir_prefix} PLATFORM={consts.Platforms.BARE_METAL}'
+        )
+    with utils.file_lock_context():
+        utils.run_command(
+            f'make _apply_terraform CLUSTER_NAME={terraform_cluster_dir_prefix} PLATFORM={consts.Platforms.BARE_METAL}'
         )
```
osherdp, line 105: why do we do it twice?

discovery-infra/day2.py
```diff
@@ -96,16 +90,18 @@ def day2_nodes_flow(client,
...
     time.sleep(5)
```
osherdp, line 106: is it possible to wait for something more significant instead of sleeping?

discovery-infra/tests/test_day2.py
```diff
@@ -0,0 +1,37 @@
...
+        return self.override_cluster_configuration()
+
+    def override_cluster_configuration():
+        config = ClusterConfig()
+        config.cluster_id = global_variables.cluster_id
+        return config
+
+    @JunitTestSuite()
+    def test_deploy_nodes_cloud(self, cluster):
+        args = SimpleNamespace()
```
osherdp, line 24: we must provide arguments in some other way than mimicking cli options
can't ``execute_day2_cloud_flow`` be made to have those arguments directly?

discovery-infra/tests/test_targets.py
```diff
@@ -10,3 +14,19 @@ def test_target_deploy_nodes(self, cluster):
...
+    @JunitTestSuite()
+    def test_target_deploy_day2_nodes_cloud(self, cluster):
+        # Deploy day2 nodes on an installed cluster (should specify CLUSTER_ID env var)
+        Day2().test_deploy_nodes_cloud(cluster)
+
+    @JunitTestSuite()
+    def test_target_install_with_deploy_day2_nodes_cloud(self, prepared_cluster):
+        # Install day1 cluster and deploy day2 nodes (cloud flow)
+        self.test_target_install_with_deploy_nodes(prepared_cluster)
+        Day2().test_deploy_nodes_cloud(prepared_cluster)
```
osherdp, line 32: shouldn't it work with ``make test TEST=./discovery-infra/tests/test_day2.py``?
I think it's better to not duplicate it

osherdp (changes requested):

discovery-infra/tests/test_day2.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def override_cluster_configuration(self):
+        config = ClusterConfig()
+        config.cluster_id = global_variables.cluster_id
+        return config
+
+    # Install day1 cluster and deploy day2 nodes (cloud flow).
+    # Or, deploy day2 nodes on an installed cluster if CLUSTER_ID env var is specified.
+    @JunitTestSuite()
+    def test_deploy_nodes_cloud(self, cluster):
```
eliorerz, line 22:
```suggestion
    def test_deploy_nodes_cloud(self, cluster, new_cluster_configuration: ClusterConfig):
```

discovery-infra/tests/test_day2.py
```diff
@@ -0,0 +1,40 @@
...
+        args.num_day2_workers = global_variables.num_day2_workers
+        args.ssh_key = global_variables.ssh_public_key
+        args.install_cluster = True
+
+        if cluster.config.proxy:
+            args.http_proxy = cluster.config.proxy.http_proxy
+            args.https_proxy = cluster.config.proxy.https_proxy
+            args.no_proxy = cluster.config.proxy.no_proxy
+
+        day2.execute_day2_cloud_flow(cluster.config.cluster_id, args, cluster.config.is_ipv6)
```
eliorerz, line 40: `day2.py` is deprecated. Add this logic to another module or create a new one. Once this task is done we want to delete `day2.py` module
Maybe create a new class for `Day2Cluster` that inherits from `Cluster` and implement the extra logic

eliorerz: Also I think that this function signature should get cluster and cluster config

eliorerz:
>Note: For simplicity, used day2.py mostly as-is. We can rewrite and move its logic into test_day2 later on. Also, should add support for the ocp flow.

@osherdp What do you think about it?

discovery-infra/tests/test_day2.py
danielerez: Regarding the day2.py deprecation, added a note in the description that for used day2.py mostly as-is. Since it requires a rewrite of the logic and integration into tests, it's safer to handle it after we have a running e2e test.

eliorerz:
/hold
Lets add day2 job and run it on this PR to verify it's working

discovery-infra/tests/test_day2.py
eliorerz, line 22: Rename it to more unique name
e.g `test_day2_deploy_nodes_cloud`

osherdp:
> > Note: For simplicity, used day2.py mostly as-is. We can rewrite and move its logic into test_day2 later on. Also, should add support for the ocp flow.
> 
> @osherdp What do you think about it?

as long as it's a follow-up task and we get right into it, I don't mind first pushing it almost as is
but let's scope it only to one place, e.g. ``day2.py`` file and not all changes in this PR

discovery-infra/day2.py
osherdp: just add a TODO

discovery-infra/tests/test_day2.py
osherdp: add a TOOD for now

osherdp (changes requested):

osherdp:
> /hold Lets add day2 job and run it on this PR to verify it's working

good idea, we can add only presubmit job for test-infra for now

danielerez:
> > > Note: For simplicity, used day2.py mostly as-is. We can rewrite and move its logic into test_day2 later on. Also, should add support for the ocp flow.
> > 
> > 
> > @osherdp What do you think about it?
> 
> as long as it's a follow-up task and we get right into it, I don't mind first pushing it almost as is but let's scope it only to one place, e.g. `day2.py` file and not all changes in this PR

Sure, will create a task for it that could be prioritized later.

osherdp:
I wonder if it already works...
/test e2e-metal-assisted-day2

osherdp:
doesn't seem related to those changes
/test e2e-metal-assisted-day2

osherdp:
Now it should work
/test e2e-metal-assisted-day2

eliorerz:
/test e2e-metal-assisted-day2
Want to re-run the test to verify that it's not flaky

osherdp:
that's super cool!

eliorerz:
> that's super cool!

Indeed.. 😀

danielerez:
> /hold Lets add day2 job and run it on this PR to verify it's working

/unhold
day2 job passed :)

discovery-infra/day2.py
danielerez: no exception is raised, so just added a loop

discovery-infra/tests/test_day2.py
danielerez: added

danielerez: done

osherdp (approved):

PR #1237, opened 2021-11-15
NO-ISSUE: remove all elasticsearch-related code

osherdp:
Remove events-scraper code and ability to run elasticsearch locally, both should be practically dead code by now.
/cc @michaellevy101 @eliorerz 
/hold


eliorerz (approved):
LGTM

PR #1243, opened 2021-11-16
NO-ISSUE: remove jira from requirements.txt as it's not being used

osherdp:
It's probably a left-over from having triage creation in assisted-test-infra (it has been moved to assisted-installer-deployment)
/cc @eliorerz


eliorerz (approved):
LGTM

PR #1247, opened 2021-11-17
Fix for logs test on V1

iovadia:

/cc @lalon4 @osherdp @eliorerz

### DIFF
--- a/discovery-infra/test_infra/utils/logs_utils.py
+++ b/discovery-infra/test_infra/utils/logs_utils.py
@@ -14,7 +14,8 @@ OC_DOWNLOAD_LOGS_TIMEOUT = 60 * 60
 
 
 def verify_logs_uploaded(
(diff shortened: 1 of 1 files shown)


discovery-infra/test_infra/utils/logs_utils.py
```diff
@@ -26,7 +27,7 @@ def verify_logs_uploaded(
             )
             tar.extractall(tempdir)
             for gz in os.listdir(tempdir):
-                if "bootstrap" in gz:
+                if "bootstrap" in gz and verify_bootstrap_errors is True:
```
osherdp, line 30:
```suggestion
                if "bootstrap" in gz and verify_bootstrap_errors:
```

osherdp (changes requested):

discovery-infra/test_infra/utils/logs_utils.py
iovadia: Done

osherdp (approved):

PR #1253, opened 2021-11-24
NO-ISSUE: add pme-bot user for better jira integration

osherdp:
/cc @osherdp 
/hold


osherdp:

PR #1255, opened 2021-11-28
Dockerfile for assisted-test-infra which uses Python 3.9

osherdp:
This is step 1 to upgrade assisted-test-infra to Python 3.9.
This will allow us to ditch using ``src`` image on prow, which overrides top-level image which lacks Python 3.9 packages.
Step 2 will be to actually create this image in prow, not using src in the process, and let all jobs on master use this image.
/cc @eliorerz 
/hold


osherdp:
(prow...)

eliorerz (approved):
LGTM

PR #1261, opened 2021-11-29
Allow overriding dockerfile path

osherdp:
That's a temporal solution until we can move to the new test-infra image with Python 3.9.
For more information: #1255
/cc @eliorerz


eliorerz (approved):
LGTM

osherdp:
redundant

PR #1262, opened 2021-11-29
MGMT-7351 - Move deployment scripts from assisted-service into test-infra scripts

eliorerz:
As part of adding ZTP flow into test-infra we want to move all the deployment scripts into test-infra repository and reduce code duplication 

/cc @osherdp 

FYI @lranjbar


osherdp (approved):
all in all, looks good to me
holding to let @lranjbar have a look as well
/hold

osherdp:
```
cat: /home/assisted/data/default_os_images.json: No such file or directory
```

I guess those need to be adjusted

lranjbar:
LGTM as well. These files live in assisted-service, but the question is do they need to remain in assisted-service. If so, then assisted-test-infra stuff needs to be adjusted to not use them.

```
cat: /home/assisted/data/default_os_images.json: No such file or directory
cat: /home/assisted/data/default_release_images.json: No such file or directory
```

osherdp (approved):

PR #1269, opened 2021-11-30
Use container command variable

masayag:
The container command is being set before invoking the rules.
The Makefile rules should be using the variable instead of an explicit
container command, as it might not be avaiable in certain systems.


eliorerz (approved):
LGTM

PR #1272, opened 2021-12-02
MGMT-8411 - Allow download ISO without ssl verification on test-infra v1

eliorerz:
/cc @lalon4


eliorerz:
No need for that feature in v1. Added to master instead.

PR #1274, opened 2021-12-05
add a new parameter for setting assisted-installer-deployment branch

osherdp:
This should allow us testing triage-tickets automation with Jenkins.
The parameter will use master branch (as before) by default, but we'll be able to set it to other values when using "Build with Parameters".
/cc @eliorerz


osherdp:
(redundant)

eliorerz (approved):
LGTM

PR #1284, opened 2021-12-08
NO-ISSUE: Install Golang when setting up capi-provider env

eranco74:
Added install_golang.sh that will install golang in case go isn't installed already
Call install_golang.sh from setup_capi_provider_agent_env.sh

Golang is required for the setup_capi_provider_agent_env.sh to work

### DIFF
--- a/scripts/install_golang.sh
+++ b/scripts/install_golang.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+set -euo pipefail
+
+GO_VERSION="${GO_VERSION:-1.17.4}"
+
+function install_golang() {
+    if ! [ -x "$(command -v go)" ]; then
+        echo "Installing golang..."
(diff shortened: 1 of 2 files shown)


scripts/install_golang.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+set -euo pipefail
+
+GO_VERSION="${GO_VERSION:-1.17.4}"
+
+function install_golang() {
+    if ! [ -x "$(command -v go)" ]; then
+        echo "Installing golang..."
+        curl -s https://go.dev/dl/go$GO_VERSION.linux-amd64.tar.gz | tar -C /usr/local -xz
```
tsorya, line 9: local is enough? not always worked for me

osherdp:
@eranco74 can you retitle and provide more info in the description?

scripts/install_golang.sh
eranco74: It works

PR #1290, opened 2021-12-09
NO-ISSUE: Install golang in the build container instead on the host

eranco74:
Run the setup_capi_provider_agent_env.sh script with skipper


eranco74:
Required for this https://github.com/openshift/release/pull/24329

PR #1293, opened 2021-12-13
NO-ISSUE: updated the capi test

eranco74:
the test should end with a working hypershift cluster with at least one ready node
otherwise the minimal conformence test that runs afterwards fail due to
missing networks.operator.openshift.io "cluster"

### DIFF
--- a/discovery-infra/test_infra/helper_classes/hypershift.py
+++ b/discovery-infra/test_infra/helper_classes/hypershift.py
@@ -71,16 +71,18 @@ class HyperShift:
             body=body,
         )
 
-    def get_nodes(self) -> V1NodeList:
+    def get_nodes(self, ready: bool = False) -> V1NodeList:
         if self.hypershift_cluster_client is None:
             hypershift_cluter_kubeapi_client = create_kube_api_client(self.kubeconfig_path)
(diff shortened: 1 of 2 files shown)


discovery-infra/test_infra/helper_classes/hypershift.py
```diff
@@ -71,16 +71,18 @@ def set_nodepool_node_count(self, kube_api_client: ApiClient, node_count: int) -
...
             self.hypershift_cluster_client = CoreV1Api(hypershift_cluter_kubeapi_client)
+        nodes = self.hypershift_cluster_client.list_node()
+        if ready:
+            return filterNodeByReadyStatus(nodes)
+        return nodes
 
-        return self.hypershift_cluster_client.list_node()
-
-    def wait_for_nodes(self, node_count: int) -> V1NodeList:
+    def wait_for_nodes(self, node_count: int, ready: bool) -> V1NodeList:
```
tsorya, line 83: i think better to have it as wait_for_nodes_in_status and give status no? or it is complicated?

eranco74: I tried that, it's useless
the only status anyone cares about is ready

avishayt (approved):

avishayt (approved):

PR #1310, opened 2021-12-16
NO-ISSUE - Fix arkade not found in PATH on some machines. Add /usr/local/bin to PATH.

eliorerz:
```
...
New version of arkade installed to /usr/local/bin
/usr/bin/which: no ark in (/usr/local/sbin:/sbin:/bin:/usr/sbin:/usr/bin:/root/bin)
main: line 176: arkade: command not found
...
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,9 @@ CONTAINER_COMMAND = $(shell if [ -x "$(shell command -v docker)" ];then echo "do
 PULL_PARAM=$(shell if [ "${CONTAINER_COMMAND}" = "podman" ];then echo "--pull-always" ; else echo "--pull";fi)
 
 ROOT_DIR = $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
-PYTHONPATH=$(shell scripts/utils.sh get_pythonpath ${ROOT_DIR})
+PYTHONPATH := "${PYTHONPATH}:${ROOT_DIR}/src"
+PATH := "${PATH}:/usr/local/bin"
+
 
 REPORTS = $(ROOT_DIR)/reports
 
--- a/scripts/utils.sh
+++ b/scripts/utils.sh
@@ -185,15 +185,4 @@ function configure_none_platform_iptables_rules() {
     iptables -t nat -A POSTROUTING ! -d 192.168.0.0/16 -j SNAT --source $sec_network --to-source $ip
 }
 
-function get_pythonpath() {
-  root_dir=$1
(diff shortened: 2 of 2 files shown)


scripts/utils.sh
```diff
@@ -187,13 +187,27 @@ function configure_none_platform_iptables_rules() {
...
+  update_path_var "${PATH}" "$path"
+}
+
+function update_path_var() {
+  var_value=$1
+  path=$2
+
+  if [[ ":${var_value}:" != *":${path}:"* ]] && [[ ":${var_value}:" != *":${path}/:"* ]];then
+    echo "${var_value:+"$var_value:"}$path" ;
+  else echo "${var_value}";
```
lranjbar, line 208: I don't have the context for this change but it seems like a lot to have here instead of:

`export PATH=$PATH:/usr/local/bin`

Are the changes to the PYTHONPATH variable intended?

lranjbar (commented):
For when you get back in on Sunday.

scripts/utils.sh
eliorerz: We often call make multiple times and if we export PATH as you suggested it Will append endlessly. We need to check if exists first.
We already doing that for PYTHONPAT, I just created a generic function for both of those vars

Makefile
```diff
@@ -8,6 +8,8 @@ PULL_PARAM=$(shell if [ "${CONTAINER_COMMAND}" = "podman" ];then echo "--pull-al
 
 ROOT_DIR = $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
 PYTHONPATH=$(shell scripts/utils.sh get_pythonpath ${ROOT_DIR})
+PATH=$(shell scripts/utils.sh get_path)
```
osherdp, line 11:
```suggestion
PATH := "${PATH}:/usr/local/bin"
```

``./create_full_environment.sh`` is either way being called in the same shell, and no reason for it to pollute outside shell scope

scripts/utils.sh
osherdp: I'm with @lranjbar with this one
Check above suggestion, it can save a lot of code (I tried running it multiple times, as expected it's not appending endlessly)

osherdp (changes requested):

osherdp (approved):

eliorerz:
I don't like appending without checking the value first but it will save all the current logic code.
I removed that utils and append the values on the Makefie to both PATH and PYTHONPATH

PR #1311, opened 2021-12-19
NO-ISSUE: wait_for_nodes should pass the ready option to the get_nodes method

eranco74:
Without passing the arg the wait_for_nodes return regardless of the
nodes ready status


eliorerz (approved):
LGTM

eranco74:
/override e2e-metal-assisted-kube-api-late-binding-single-node

PR #1312, opened 2021-12-19
NO-ISSUE - Allow multiple assisted-test-infra logs

eliorerz:
Create random test-infra log name so on tests with POSTSUBMIT tests we will not override test_infra.log.
Clear logs on local environment on every test execution.

/cc @osherdp


eliorerz:
We currently appending the logs on the `FileHandler`.

PR #1315, opened 2021-12-20
MGMT-8423: remove hack for disable day2 validation

rollandf:
Update README, no need to disable the api-vip-connected validation

### DIFF
--- a/README.md
+++ b/README.md
@@ -467,8 +467,6 @@ export MINIKUBE_RAM_MB=12288
 Setup minikube with assisted-installer (kube-api enabled)
 ```bash
 export PULL_SECRET=<your pull secret>
-# This is a temporary hack - MGMT-8265
-export DISABLED_HOST_VALIDATIONS=api-vip-connected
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -468,7 +468,7 @@ Setup minikube with assisted-installer (kube-api enabled)
 ```bash
 export PULL_SECRET=<your pull secret>
 # This is a temporary hack - MGMT-8265
-export DISABLED_HOST_VALIDATIONS=api-vip-connected
+export DISABLED_HOST_VALIDATIONS=ignition-downloadable
```
eranco74, line 471: Since you fixed MGMT-8265, and after I updated the agent to decode the cert these lines are no longer necessary.
I validated it locally and it works without it.
Let's remove it

rollandf: Ack

eranco74 (commented):
/lgtm

eliorerz (approved):

PR #1254, opened 2021-11-25
MGMT-7920: enable testing TPM disk encryption mode

nmagnezi:
This PR continues @osherdp's work from https://github.com/openshift/assisted-test-infra/pull/1142

Co-authored-by: Osher De Paz <odepaz@redhat.com>

PR #1325, opened 2021-12-21
Go back to using upstream terraform-libvirt-provider

osherdp:
Back in #1254, we decided to use forked libvirt-provider until we have a newer version.
Now that there's a new version, we can use the upstream provider again.
/cc @nmagnezi @eliorerz


eliorerz (approved):
LGTM

PR #1328, opened 2021-12-22
MGMT-8769- Fail olm test on operator failure.

eliorerz:
Use cluster fixture in test_olm_operator instead of outdated fixture get_cluster.
Raise error on new_cluster_configuration on missing ClusterConfig attribute

/cc @osherdp

### DIFF
--- a/src/assisted_test_infra/test_infra/helper_classes/config/base_config.py
+++ b/src/assisted_test_infra/test_infra/helper_classes/config/base_config.py
@@ -1,6 +1,6 @@
 from abc import ABC, abstractmethod
 from dataclasses import asdict, dataclass
-from typing import Any, Dict, Tuple
+from typing import Any, Dict, List, Tuple
 
 from service_client import log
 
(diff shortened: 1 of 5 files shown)


src/assisted_test_infra/test_infra/helper_classes/cluster.py
```diff
@@ -738,7 +738,7 @@ def wait_for_operators_to_finish(self, timeout=consts.CLUSTER_INSTALLATION_TIMEO
             cluster_id=self.id,
             operators_count=len(operators_utils.filter_operators_by_type(operators, OperatorType.OLM)),
             operator_types=[OperatorType.OLM],
-            statuses=[consts.OperatorStatus.AVAILABLE, consts.OperatorStatus.FAILED],
+            statuses=statuses,
```
lalon4, line 741: why this change is needed? we have fall_on_error_status if you expect failure

eliorerz: But we don't use those here. It always wait for failures regardless of `fall_on_error_status` value.
Also, If we don't mind if the operator is success or failure we can now use the `fall_on_error_status`

lalon4: why not using it? it does considered in _are_operators_in_status

PR #1330, opened 2021-12-23
allow gathering from the new name of assisted-service container

osherdp:
On openshift/assisted-service#3096, we will use a new container name for assisted-service.
This should gather logs from both names, until we'll do the transition.
/cc @eliorerz


eliorerz:
Need to change the rest of the images names
```
ui -> assisted-installer-ui
db -> postgres
image-service -> assisted-image-service
installer -> assisted-service
```
But we should wait to https://github.com/openshift/assisted-service/pull/3100 before changing it
/hold

osherdp:
Fixed!

eliorerz (approved):
LGTM

PR #1339, opened 2021-12-29
Fixing none-platform network interfaces

osherdp:
On some occasions, the none-platform environment test is getting into the following situation:
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-assisted-test-infra-master-e2e-metal-assisted-none-periodic/1474880968832585728

In this case, the libvirt domains have an unintended network device attached, which is of type ``user``. When declaring part of the terraform fields for a network, it seems like it creates those kind of networks of type ``Userspace SLIRP stack``, and it probably adds some flakiness. This yielded the unintended generation of those networks.

This change should only declare whole networks, with the added benefit of being more general and provide ``networks`` (and not ``primary_network`` and ``secondary_network``).

### DIFF
--- a/terraform_files/none/variables-libvirt.tf
+++ b/terraform_files/none/variables-libvirt.tf
@@ -68,6 +68,11 @@ variable "libvirt_secondary_master_ips" {
   description = "the list of desired master second interface ips. Must match master_count"
 }
 
+variable "libvirt_master_macs" {
+  type        = list(string)
+  description = "the list of the desired macs for master interface"
+}
+
 variable "libvirt_worker_ips" {
   type        = list(list(string))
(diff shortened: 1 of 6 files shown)


terraform_files/none/variables-libvirt.tf
```diff
@@ -68,6 +68,16 @@ variable "libvirt_secondary_master_ips" {
   description = "the list of desired master second interface ips. Must match master_count"
 }
 
+variable "libvirt_master_macs" {
```
ori-amizur, line 71: No need for 2 lists in none platform

osherdp: fixed

PR #1340, opened 2021-12-29
MGMT-8705: Collect capi-provider-agent logs

eranco74:
Assisted-test-infra can run cluster-api-provider-agent e2e tests.
This PR adds log gathering of the provider pod and the relevant CRs content

### DIFF
--- a/scripts/download_logs.sh
+++ b/scripts/download_logs.sh
@@ -11,7 +11,8 @@ ADDITIONAL_PARAMS=${ADDITIONAL_PARAMS:-""}
 KUBECTL=${KUBECTL:-kubectl}
 LOGS_DEST=${LOGS_DEST:-build}
 JUNIT_REPORT_DIR=${JUNIT_REPORT_DIR:-"reports/"}
-KUBE_CRS=( clusterdeployment infraenv agentclusterinstall agent )
+KUBE_CRS=(clusterdeployment infraenv agentclusterinstall agent)
+CAPI_PROVIDER_CRS=(agentmachine agentcluster)
 
 function download_service_logs() {
   mkdir -p ${LOGS_DEST} || true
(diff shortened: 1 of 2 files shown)


avishayt (approved):

eranco74:
/test e2e-metal-assisted-capi
Let's see if we get the cluster logs

osherdp:
> /test e2e-metal-assisted-capi /test e2e-metal-assisted-kube-api-late-binding-single-node Let's see if we get the cluster logs

when running optional jobs, always hold
otherwise, it will just continue with merging it without waiting for those jobs
/hold

scripts/download_logs.sh
```diff
@@ -58,5 +59,25 @@ function collect_kube_api_resources() {
...
+    ${KUBECTL} get pods -n ${NAMESPACE} -o=custom-columns=NAME:.metadata.name --no-headers | xargs -r -I {} sh -c "${KUBECTL} logs {} -n ${NAMESPACE} --all-containers > ${LOGS_DEST}/k8s_{}.log" || true
+    skipper run ./src/junit_log_parser.py --src "${LOGS_DEST}" --dst "${JUNIT_REPORT_DIR}"
+}
+
+function get_pod_namespace() {
+  ${KUBECTL} get pods --no-headers -A  -o jsonpath='{range .items[*]}{@.metadata.name}{" "}{@.metadata.namespace}{"\n"}' | grep $1 | awk -F " " '{print $2}'
+}
+
+function collect_capi_provider_resources() {
+    JSONPATH=
```
osherdp, line 74: ?

osherdp (changes requested):

osherdp:
can you add to PR description this resources you're gathering?

scripts/download_logs.sh
```diff
@@ -58,5 +59,25 @@ function collect_kube_api_resources() {
...
+    for CR in "${CAPI_PROVIDER_CRS[@]}"
+    do
+      ${KUBECTL} get ${CR} --all-namespaces -o jsonpath='{range .items[*]}{@.metadata.name}{" -n "}{@.metadata.namespace}{"\n"}{end}' --no-headers |\
```
osherdp, line 77: what does this line do?

osherdp (changes requested):

scripts/download_logs.sh
eranco74: Get the CR from all namespaces.
Unlike the assisted-service CRs (though I'm pretty sure that's not entirely true) the provider CRs get created on different namespaces.
This line get all relevant CRs name and namespaces and pass it on the the `kubectl get -o json ...` command

eranco74: Added a comment in the code

osherdp: hmm, but the code is still unreadable I'm afraid
can you make it in a for loop?

osherdp (changes requested):

osherdp (approved):

scripts/download_logs.sh
eranco74: Done,
Also updated the existing `collect_kube_api_resources` to use for loop

PR #1341, opened 2022-01-01
Use skipper interactive env-var

osherdp:
Seems like there's a simpler way to make skipper interactive every time it's being called, which is the ``SKIPPER_INTERACTIVE`` environment variable.
It's not enabling to customize skipper params, but I don't really think it's in wide use.
/cc @eliorerz


eliorerz (approved):
LGTM

PR #1344, opened 2022-01-02
install go on py39 Dockerfile

osherdp:
Seems like it's missing from the new Dockerfile, since #1290
/cc @eliorerz


eliorerz (approved):
LGTM

PR #1345, opened 2022-01-02
fix SNO installations always picking OCP 4.8

osherdp:
Seems like we always override SNO installations with OCP 4.8, instead of going with what the user chose on ``OPENSHIFT_VERSION`` or ``OPENSHIFT_INSTALL_RELEASE_IMAGE``.
/cc @eliorerz


osherdp:
In addition to this, we probably need to understand why this triggering mechanism overrides something that the user already decided to customize.
I guess it's not the intended behavior, right?

eliorerz (approved):
We need to change the triggers mechanism to be more sophisticated

osherdp:
right. I'll open a task/bug around that
/unhold

PR #1347, opened 2022-01-03
MGMT-8666 - Fix onprem logs destination in prow

eliorerz:
/cc @osherdp

### DIFF
--- a/scripts/download_logs.sh
+++ b/scripts/download_logs.sh
@@ -21,7 +21,7 @@ function download_service_logs() {
     podman ps -a || true
 
     for service in "assisted-service" "assisted-image-service" "assisted-installer-ui" "postgres"; do
-      podman logs ${service} >${LOGS_DEST}/onprem_${service}.log || true
+      podman logs ${service} >${LOGS_DEST}/logs_${service}_${DEPLOY_TARGET}.log || true
     done
   else
     CRS=node,pod,svc,deployment,pv,pvc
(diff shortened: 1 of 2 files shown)


scripts/download_logs.sh
```diff
@@ -21,7 +21,7 @@ function download_service_logs() {
     podman ps -a || true
 
     for service in "assisted-service" "assisted-image-service" "assisted-installer-ui" "postgres"; do
-      podman logs ${service} >${LOGS_DEST}/onprem_${service}.log || true
+      podman logs ${service} >${LOGS_DEST}/k8s_${service}_onprem.log || true
```
osherdp, line 24: it will be misleading, because this deployment method has nothing to do with kubernetes, no?

osherdp (changes requested):

scripts/download_logs.sh
eliorerz: Right.. I think I didn't thought all the way through this.
I wanted to make the logs with the same prefix as the minikube deployments.
I'll change them both to start with a new prefix

scripts/download_logs.sh
```diff
@@ -21,7 +21,7 @@ function download_service_logs() {
     podman ps -a || true
 
     for service in "assisted-service" "assisted-image-service" "assisted-installer-ui" "postgres"; do
-      podman logs ${service} >${LOGS_DEST}/onprem_${service}.log || true
+      podman logs ${service} >${LOGS_DEST}/logs${service}_${DEPLOY_TARGET}.log || true
```
osherdp, line 24:
```suggestion
      podman logs ${service} >${LOGS_DEST}/logs_${service}_${DEPLOY_TARGET}.log || true
```

osherdp: same for other stuff, just make sure it's displaying nicely

osherdp (changes requested):

osherdp (approved):

PR #1349, opened 2022-01-03
Assign pip packages maintainers when bumping dependencies

osherdp:
Seems reasonable we're mostly concerned about those packages than others.
/cc @eliorerz


eliorerz (approved):

osherdp:
Way too flaky

PR #1356, opened 2022-01-06
MGMT-8363 - Join flaky tests results on prow spyglass

eliorerz:
Currently, errors on assisted-service logs are shown as separate JUNIT test-case, causing PROW to "split" each test case to new row under `Flaky` section (multiple entries in the Flaky table)
Here instead of 17 lines of failures on `github.com/openshift/assisted-service/internal/controller/controllers.(*ClusterDeploymentsReconciler).Reconcile` with the error `error setting owner reference`, we can see it under single table entry (as single string) without the need to expand multiple entries.
image

### DIFF
--- a/src/junit_log_parser.py
+++ b/src/junit_log_parser.py
@@ -55,16 +55,18 @@ class LogsConverter:
         return False
 
     @classmethod
-    def get_log_entry_case(cls, entry: LogEntry, fail_cases: Dict[str, List[TestCase]], suite_name: str) -> \
-            List[TestCase]:
+    def get_log_entry_case(cls, entry: LogEntry,
+                           fail_cases: Dict[str, List[TestCase]],
+                           suite_name: str,
+                           failure_message: str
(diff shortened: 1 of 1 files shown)


osherdp:
interestingly it doesn't follow ``black`` format
we disclosed scripts from being linted with this format? (not that I think it's critical, just asking)

osherdp:
@eliorerz please add a pic to the PR description, when you have one

osherdp (approved):

eliorerz:
> interestingly it doesn't follow `black` format we disclosed scripts from being linted with this format? (not that I think it's critical, just asking)

Currently scripts are excluded from being formatted and to be checked by flake8
https://github.com/openshift/assisted-test-infra/blob/36ff2155de88fae0de3c2c3c1d443c7faa74f6a9/Makefile#L28

osherdp (approved):

PR #1370, opened 2022-01-12
Unbound variable

eranco74:
Also adds a space to the hypershift create command


eranco74:
/hold
Wait for capi test...

PR #1371, opened 2022-01-13
MGMT-8900: Update e2e test to use capi-provider-agent image from env

eranco74:


eranco74:
The PR changes are specific to capi test, test failure isn't related to the change

PR #1379, opened 2022-01-17
NO-ISSUE - Reorganize test_kube_api.py file and remove duplicated code

eliorerz:
Reorganize test_kube_api.py file. Remove duplicated code and add type hint annotation.

/cc @osherdp

### DIFF
--- a/src/tests/test_kube_api.py
+++ b/src/tests/test_kube_api.py
@@ -3,12 +3,13 @@ import json
 import os
 import tempfile
 import uuid
-from typing import List
+from typing import Callable, List, Optional
 
 import openshift as oc
 import pytest
 import waiting
 from junit_report import JunitFixtureTestCase, JunitTestCase, JunitTestSuite
+from kubernetes.client import ApiClient
 from netaddr import IPNetwork
 
 from assisted_test_infra.download_logs import collect_debug_info_from_cluster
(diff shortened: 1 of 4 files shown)


src/tests/test_kube_api.py
```diff
@@ -83,14 +75,344 @@ def _get_vips(cls, cluster_config, nodes: Nodes):
...
+            get_nodes(tf_config, cluster_config),
+            cluster_config,
+            global_variables.spoke_namespace,
+            proxy_server,
+            is_ipv4=False,
+        )
+
+    @classmethod
+    def start_nodes(cls, nodes: Nodes, infra_env: InfraEnv, entity_config: BaseEntityConfig, is_ipv4) -> List[Agent]:
+        infra_env.status()
```
osherdp, line 103: What this does?
And if it just to wait for the status, we should add some comment about that

src/tests/test_kube_api.py
```diff
@@ -83,14 +75,344 @@ def _get_vips(cls, cluster_config, nodes: Nodes):
...
+        cls.download_iso_from_infra_env(infra_env, entity_config.iso_download_path)
+
+        log.info("iso downloaded, starting nodes")
+        nodes.start_all()
+
+        log.info("waiting for host agent")
+        agents = infra_env.wait_for_agents(len(nodes))
+        for agent in agents:
+            agent.approve()
+            cls.set_agent_hostname(nodes[0], agent, is_ipv4)  # Currently only supports single node
```
osherdp, line 113: maybe we should replace the comment with an assertion or something

src/tests/test_kube_api.py
```diff
@@ -83,14 +75,344 @@ def _get_vips(cls, cluster_config, nodes: Nodes):
...
+            cluster_cidr=cluster_config.cluster_networks[0].cidr,
+            host_prefix=cluster_config.cluster_networks[0].host_prefix,
+            service_network=cluster_config.service_networks[0].cidr,
+            ssh_pub_key=cluster_config.ssh_public_key,
+            hyperthreading=cluster_config.hyperthreading,
+            control_plane_agents=nodes.masters_count,
+            worker_agents=nodes.workers_count,
+            machine_cidr=machine_cidr,
+        )
+        agent_cluster_install.wait_to_be_ready(False)
```
osherdp, line 156:
```suggestion
        agent_cluster_install.wait_to_be_ready(ready=False)
```

it's still a bit weird, but at least more clear as to what should happen

src/tests/test_kube_api.py
```diff
@@ -83,14 +75,344 @@ def _get_vips(cls, cluster_config, nodes: Nodes):
...
+        )
+        agents = self.start_nodes(nodes, infra_env, cluster_config, is_ipv4)
+
+        if len(nodes) == 1:
+            self.set_single_node_ip(cluster_deployment, nodes, is_ipv4)
+
+        log.info("Waiting for agent status verification")
+        Agent.wait_for_agents_to_install(agents)
+
+        agent_cluster_install.wait_to_be_ready(True)
```
osherdp, line 180:
```suggestion
        agent_cluster_install.wait_to_be_ready(ready=True)
```

src/tests/test_kube_api.py
osherdp: Or maybe ``wait_to_be_ready`` should be changed to ``wait_for_status`` or ``wait_to_state``

src/tests/test_kube_api.py
```diff
@@ -83,14 +75,344 @@ def _get_vips(cls, cluster_config, nodes: Nodes):
...
+        log.info("waiting for agent-cluster-install to be in installing state")
+        agent_cluster_install.wait_to_be_installing()
+
+        try:
+            log.info("installation started, waiting for completion")
+            agent_cluster_install.wait_to_be_installed()
+            log.info("installation completed successfully")
+        except Exception:
+            log.exception("Failure during kube-api installation flow:")
+            collect_debug_info_from_cluster(cluster_deployment, agent_cluster_install)
```
osherdp, line 191: hmm, why we don't raise the exception?

src/tests/test_kube_api.py
```diff
@@ -101,76 +423,29 @@ def unbound_highly_available_infraenv(
...
         yield infra_env, nodes
 
     @pytest.fixture
     @JunitFixtureTestCase()
     def unbound_single_node_cluster(self, kube_test_configs_single_node, kube_api_context):
         cluster_config, _ = kube_test_configs_single_node
-        yield kube_api_test_prepare_late_binding_cluster(
-            kube_api_context=kube_api_context, cluster_config=cluster_config, num_controlplane_agents=1
-        )
+        yield self.kube_api_test_prepare_late_binding_cluster(kube_api_context.api_client, cluster_config, 1)
```
osherdp, line 435: I think it's better to leave usage of keyword arguments on those kind places, even if it produces longer lines
No one can know what ``1`` means in this context without it, it's mainly a magic number

osherdp: here and elsewhere

src/tests/test_kube_api.py
```diff
@@ -101,76 +423,29 @@ def unbound_highly_available_infraenv(
...
-        agent_cluster_install.wait_to_be_installing()
-        Agent.wait_for_agents_to_install(agents)
-        agent_cluster_install.wait_to_be_installed()
-
-    @staticmethod
-    def _set_agent_cluster_install_machine_cidr(agent_cluster_install, nodes):
-        machine_cidr = nodes.controller.get_primary_machine_cidr()
-        agent_cluster_install.set_machinenetwork(machine_cidr)
+        api_client = kube_api_context.api_client
+        yield self.kube_api_test_prepare_late_binding_cluster(api_client, cluster_config, 1, hold_installation=True)
```
osherdp, line 448: also on those kind of places. ``hold_installation=True`` is great, ``1`` is just not understandable

osherdp (changes requested):

src/tests/test_kube_api.py
eliorerz: Not sure. I kept the current flow

src/tests/test_kube_api.py
eliorerz: From the function  documentation:


```
        Status is a section in the CRD that is created after registration to
        assisted service and it defines the observed state of InfraEnv.
        Since the status key is created only after resource is processed by the
        controller in the service, it might take a few seconds before appears.
```

src/tests/test_kube_api.py
osherdp: okay

src/tests/test_kube_api.py
osherdp: I'd suggest raising the exception in that case
Recent events just taught us the severity of failing silently :joy:

src/tests/test_kube_api.py
osherdp: Just like you did here: https://github.com/openshift/assisted-test-infra/blob/2581b64f4ce4b095f2dafd65d6d7ccf34d462c25/discovery-infra/test_infra/helper_classes/kube_helpers/infraenv.py#L282-L284

In the scope of this file it's a weird looking line of code that seems to have no side-effect

src/tests/test_kube_api.py
```diff
@@ -238,402 +406,66 @@ def test_kube_api_late_binding_ipv4_highly_available(
...
+            cluster_cidr=cluster_config.cluster_networks[0].cidr,
+            host_prefix=cluster_config.cluster_networks[0].host_prefix,
+            service_network=cluster_config.service_networks[0].cidr,
+            ssh_pub_key=cluster_config.ssh_public_key,
+            hyperthreading=cluster_config.hyperthreading,
+            control_plane_agents=num_controlplane_agents,
+            hold_installation=hold_installation,
+            worker_agents=0,
+        )
+        agent_cluster_install.wait_to_be_ready(False)
```
osherdp, line 443: here also

osherdp (changes requested):

src/tests/test_kube_api.py
eliorerz: I'm actually not sure we need `wait_to_be_ready` function. It just calls `wait_for_condition` with 2 extra parameters (timeout is the default anyway).
The only think I don't understand is why when calling to `wait_for_condition` we `str(ready : bool)`

eliorerz: I think that currently I'll do as you suggested `agent_cluster_install.wait_to_be_ready(ready=False)` and will open a ticket for cleanup kubeapi infra code

src/tests/test_kube_api.py
eliorerz: Lets do it on a different PR? I don't want that capi or late-binding will fail in this PR

src/tests/test_kube_api.py
eliorerz: You are giving me too much credit, that was there when I arrived :)

osherdp (approved):

PR #1385, opened 2022-01-18
Remove redundant Jenkinsfiles

osherdp:
I'm pretty sure those are not really being used.
Mentioning everyone I think might be relevant, just to make sure of it.
/cc @michaellevy101 @eliorerz @lalon4 
/hold


eliorerz (approved):
LGTM

PR #1389, opened 2022-01-19
MGMT-8105 - Make kube_api target work with the new workflow

eliorerz:
This PR come to fix new-flow (pytest) KubeAPI tests.

Changes in this PR:

- Add NMState and static IP to kube_api test
- Pass proxy to AgentClusterInstall
- Join KubeAPI IPv4 and IPv6 to single test - run with parameterize
- Join late-binding and late-unbinding to single test - run with parameterize on `hold_installation=False/True`
- Remove `spoke_namespace` variable from `global_variables` - using single spoke-namespace on parallel run, caused destruction of all other resources during test execution. `spoke_namespace` is now fixture under `BaseKubeAPI` class, Add `spoke_namespace` to `KubeAPIContext`.
- Remove all usage of `get_nodes` fixture

Note: Currently none of the test won't work due to changes in test functions names.

/cc @osherdp @eranco74 @tsorya


src/tests/base_kubeapi_test.py
```diff
@@ -165,3 +179,29 @@ def deploy_image_set(cls, cluster_name: str, api_client: ApiClient):
         image_set.create(openshift_release_image)
 
         return ClusterImageSetReference(image_set_name)
+
+    @classmethod
+    @JunitTestCase()
+    def apply_static_network_config(cls, kube_api_context: KubeAPIContext, nodes: Nodes, cluster_name: str):
+        if not global_variables.is_static_ip:
+            return None
+
```
osherdp, line 188: I would delete this (or delete the condition of the calling function), because changes in this condition should preferably happen on one place only

```suggestion
        if not global_variables.is_static_ip:
            return None

```

src/tests/test_kube_api.py
```diff
@@ -2,13 +2,15 @@
 import json
 import os
 import tempfile
+from contextlib import suppress
 from typing import Callable, List, Optional
 
 import openshift as oc
 import pytest
 import waiting
+from _pytest.fixtures import FixtureLookupError, FixtureRequest
```
osherdp, line 11:
```suggestion
from pytest.fixtures import FixtureLookupError, FixtureRequest
```

src/tests/test_kube_api.py
```diff
@@ -18,49 +20,71 @@
...
 
 
 class TestKubeAPI(BaseKubeAPI):
-    @JunitTestSuite()
-    @pytest.mark.kube_api
-    def test_kube_api_ipv4(self, kube_test_configs_single_node, kube_api_context, get_nodes):
-        namespace = global_variables.spoke_namespace
-        cluster_config, tf_config = kube_test_configs_single_node
-        self.kube_api_test(kube_api_context.api_client, get_nodes(tf_config, cluster_config), cluster_config, namespace)
+    KUBEAPI_IP_OPTIONS = [(False, True), (True, False)]
```
osherdp, line 36: I might prefer to embed this in the parametrize call
because otherwise it loses its meaning having it just boolean values

src/tests/test_kube_api.py
```diff
@@ -18,49 +20,71 @@
...
+        cluster_config.trigger(get_default_triggers())
+        terraform_config.trigger(get_default_triggers())
+
+        yield cluster_config, terraform_config
 
-    @JunitTestSuite()
     @pytest.mark.kube_api
-    def test_kube_api_ipv6(self, kube_test_configs_single_node, kube_api_context, proxy_server, get_nodes):
+    @JunitTestSuite()
+    @pytest.mark.parametrize("is_ipv4, is_ipv6", KUBEAPI_IP_OPTIONS)
```
osherdp, line 59:
```suggestion
    @pytest.mark.parametrize("is_ipv4, is_ipv6", [(False, True), (True, False)])
```

src/tests/test_kube_api.py
```diff
@@ -18,49 +20,71 @@
...
+    def test_kubeapi(self, kube_test_configs_single_node, kube_api_context, proxy_server, get_nodes, is_ipv4, is_ipv6):
```
osherdp, line 60: some places can benefit from type annotations I think

src/tests/test_kube_api.py
```diff
@@ -120,35 +148,31 @@ def kube_api_test(
             log.info("installation started, waiting for completion")
             agent_cluster_install.wait_to_be_installed()
             log.info("installation completed successfully")
-        except Exception:
+        except TimeoutExpired:
             log.exception("Failure during kube-api installation flow:")
             collect_debug_info_from_cluster(cluster_deployment, agent_cluster_install)
+            raise
```
osherdp, line 154: :+1:

src/tests/test_kube_api.py
```diff
@@ -120,35 +148,31 @@ def kube_api_test(
             log.info("installation started, waiting for completion")
             agent_cluster_install.wait_to_be_installed()
             log.info("installation completed successfully")
-        except Exception:
+        except TimeoutExpired:
             log.exception("Failure during kube-api installation flow:")
```
osherdp, line 152:
```suggestion
            log.exception("Failure during kube-api installation flow. Collecting debug info...")
```

osherdp (changes requested):

src/tests/test_kube_api.py
eliorerz: It's not available for import from `pytest`

osherdp (approved):

PR #1390, opened 2022-01-20
Describing kubernetes objects on creation failure

michaellevy101:
The purpose of this PR is to debug the "e2e-metal-assisted-kube-api-late-binding-single-node" flaky test.
Looks like the problem was on the setup stage while deploying the assisted-service pod on the hub cluster (before running the actual test) we didn't find any relevant data, so we added an "oc describe" command to check why the pod couldn't become ready.
https://issues.redhat.com/browse/MGMT-8805

### DIFF
--- a/scripts/operator/utils.sh
+++ b/scripts/operator/utils.sh
@@ -35,7 +35,7 @@ function wait_for_pod() {
     namespace="${2:-}"
     selector="${3:-}"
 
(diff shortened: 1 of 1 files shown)


osherdp:
@michaellevy101 can you please supply some PR description? Also attach Jira tickets if there are any
Thanks in advance!

eliorerz:
This PR is no longer relevant due to  https://github.com/openshift/assisted-test-infra/pull/1395

PR #1391, opened 2022-01-21
NO-ISSUE - Add colors to logging

eliorerz:
/cc @osherdp 

image
image
image

### DIFF
--- a/src/service_client/logger.py
+++ b/src/service_client/logger.py
@@ -32,28 +32,70 @@ class SensitiveFormatter(logging.Formatter):
         return self._filter(original)
 
 
+class ColorizingStreamHandler(logging.StreamHandler):
+    BLUE = "\033[0;34m"
+    LIGHT_RED = "\033[1;31m"
+    LIGHT_YELLOW = "\033[1;33m"
+    LIGHT_BLUE = "\033[1;34m"
+    LIGHT_PURPLE = "\033[1;35m"
+    LIGHT_CYAN = "\033[1;36m"
+    WHITE = "\033[1;37m"
+    RESET = "\033[0m"
+
+    def __init__(self, *args, **kwargs):
+        self._colors = {
+            logging.DEBUG: self.BLUE,
+            logging.INFO: self.RESET,
+            logging.WARNING: self.LIGHT_YELLOW,
+            logging.ERROR: self.LIGHT_RED,
+            logging.CRITICAL: self.LIGHT_PURPLE,
+        }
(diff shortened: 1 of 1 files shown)


src/service_client/logger.py
```diff
@@ -32,28 +32,72 @@ def format(self, record):
...
+
+    def __init__(self, *args, **kwargs):
+        self._colors = {
+            logging.DEBUG: self.BLUE,
+            logging.INFO: self.RESET,
+            logging.WARNING: self.LIGHT_YELLOW,
+            logging.ERROR: self.LIGHT_RED,
+            logging.CRITICAL: self.LIGHT_PURPLE,
+        }
+        super(ColorizingStreamHandler, self).__init__(*args, **kwargs)
```
osherdp, line 53: To allow future-compatibility for when they'll disallow ``super`` with arguments:
```suggestion
        super().__init__(*args, **kwargs)
```

src/service_client/logger.py
```diff
@@ -32,28 +32,72 @@ def format(self, record):
...
+            if not self.is_tty:
+                stream.write(message)
+            else:
+                message = self._colors[record.levelno] + message + self.RESET
+                stream.write(message)
+            stream.write(getattr(self, "terminator", "\n"))
+            self.flush()
+        except (KeyboardInterrupt, SystemExit):
+            raise
+        except BaseException:
```
osherdp, line 73: This should be pretty much identical:
```suggestion
        except Exception:
```

(https://docs.python.org/3/library/exceptions.html#exception-hierarchy)

osherdp (changes requested):

src/service_client/logger.py
osherdp: I meant you can also remove this:
```
        except (KeyboardInterrupt, SystemExit):
            raise
```

as those are not inheriting from ``Exception``

src/service_client/logger.py
```diff
@@ -32,28 +32,72 @@ def format(self, record):
...
+            stream = self.stream
+            if not self.is_tty:
+                stream.write(message)
+            else:
+                message = self._colors[record.levelno] + message + self.RESET
+                stream.write(message)
+            stream.write(getattr(self, "terminator", "\n"))
+            self.flush()
+        except (KeyboardInterrupt, SystemExit):
+            raise
```
osherdp, line 72:
```suggestion
```

osherdp (changes requested):

osherdp (approved):

PR #1393, opened 2022-01-23
MGMT-8992 - Separate environment variables from defaults values

eliorerz:
Separate env vars from it's default value.
Add the ability to know if the user set the value and select the overrided value.
Remove env vars defaults from Makefile so that the defaults values will be identified as set by the user.
Few minor bug fixes that related to environment variables and triggers.

/cc @osherdp


Makefile
```diff
@@ -46,39 +48,10 @@ UI_SERVICE_NAME := $(or $(UI_SERVICE_NAME),assisted-installer-ui)
...
-STORAGE_POOL_PATH := $(or $(STORAGE_POOL_PATH), $(PWD)/storage_pool)
-CLUSTER_ID := $(or $(CLUSTER_ID), "")
-CLUSTER_NAME := $(or $(CLUSTER_NAME),test-infra-cluster)
-REMOTE_SERVICE_URL := $(or $(REMOTE_SERVICE_URL), "")
-WORKER_DISK ?= 21474836480
-MASTER_DISK ?= 128849018880
-DISABLED_HOST_VALIDATIONS := $(or $(DISABLED_HOST_VALIDATIONS),"")
-
-#hyperthreading
-HYPERTHREADING := $(or $(HYPERTHREADING), all)
```
eliorerz, line 65: Those defaults are already part of the environment variables in test infra

Makefile
```diff
@@ -46,39 +48,10 @@ UI_SERVICE_NAME := $(or $(UI_SERVICE_NAME),assisted-installer-ui)
...
-NETWORK_MTU := $(or $(NETWORK_MTU), 1500)
-HTTP_PROXY_URL := $(or $(HTTP_PROXY_URL), "")
-HTTPS_PROXY_URL := $(or $(HTTPS_PROXY_URL), "")
-NO_PROXY_VALUES := $(or $(NO_PROXY_VALUES), "")
-VIP_DHCP_ALLOCATION := $(or $(VIP_DHCP_ALLOCATION),yes)
-ISO_IMAGE_TYPE := $(or $(ISO_IMAGE_TYPE), full-iso)
-
-#day2 params
-API_VIP_IP := $(or $(API_VIP_IP),"")
-API_VIP_DNSNAME := $(or $(API_VIP_DNSNAME),"")
```
eliorerz, line 81: Those defaults are already part of the environment variables in test infra

src/tests/base_test.py
```diff
@@ -29,28 +29,61 @@
...
 
 
 class BaseTest:
+    @classmethod
+    def _get_parameterized_keys(cls, request: pytest.FixtureRequest):
+        parameterized_keys = []
+        optional_keys = [m.args[0] for m in request.keywords.node.own_markers if m and m.name == "parametrize"]
+
+        for key in optional_keys:
+            if "," in key:
```
eliorerz, line 49: No need to check for ","

src/tests/conftest.py
```diff
@@ -23,7 +23,7 @@ def get_available_openshift_versions() -> List[str]:
 
     if override_version:
         if override_version == consts.OpenshiftVersion.MULTI_VERSION.value:
-            return sorted(available_versions)
+            return sorted(available_versions, key=lambda s: list(map(int, s.split("."))))
```
eliorerz, line 26: Fix bug when sorting `4.x and 4.10 when x > 1`

Makefile
```diff
@@ -46,39 +48,10 @@ UI_SERVICE_NAME := $(or $(UI_SERVICE_NAME),assisted-installer-ui)
...
-MASTER_DISK ?= 128849018880
-DISABLED_HOST_VALIDATIONS := $(or $(DISABLED_HOST_VALIDATIONS),"")
-
-#hyperthreading
-HYPERTHREADING := $(or $(HYPERTHREADING), all)
-
 # network params
 NAMESPACE := $(or $(NAMESPACE),assisted-installer)
 BASE_DNS_DOMAINS := $(or $(BASE_DNS_DOMAINS), "")
 BASE_DOMAIN := $(or $(BASE_DOMAIN),redhat.com)
```
osherdp, line 54: what about this one?

Makefile
```diff
@@ -138,7 +109,6 @@ KUBE_API := $(or ${KUBE_API},no)
 ifeq ($(ENABLE_KUBE_API),true)
```
osherdp, line 109: I don't understand how ``ENABLE_KUBE_API`` here might be ``true`` and downwards can be ``no``

osherdp (changes requested):
partial review

eliorerz:
`e2e-metal-assisted-olm` and `e2e-metal-assisted-single-node` will pass after removing the job `NUM_WORKERS` default configurations from https://github.com/openshift/release/blob/1dd101cca3f703d2212ba98c5b8c399c7993d5b4/ci-operator/step-registry/baremetalds/assisted/setup/baremetalds-assisted-setup-commands.sh#L136

Makefile
eliorerz: I think it's relevant just for this `if` block

Makefile
eliorerz: We are using it on `assisted_deployment.sh` so I don't think it can be done (unless we change the service deployment from bash script to python script)

Makefile
osherdp: Not sure I understand what you're saying
So we can change it to yes/no values everywhere?

eliorerz: Not sure if that you meant but this is how it was configured on most of test-infra boolean variables.
We need to find a better way for setting boolean value

src/consts/env_defaults.py
```diff
@@ -4,9 +4,9 @@
 
 DEFAULT_NUMBER_OF_MASTERS: int = consts.NUMBER_OF_MASTERS
 DEFAULT_DAY2_WORKERS_COUNT: int = 1
-DEFAULT_WORKERS_COUNT: int = 0
+DEFAULT_WORKERS_COUNT: int = 2
```
osherdp, line 7: seems like you're changing the default that was until now, of using a "standalone" (3 nodes) clusters
I think it's better to revert it, so that it will not impact local environments that used to have enough resources for a basic installation in test-infra

src/tests/base_test.py
```diff
@@ -29,30 +29,60 @@
...
 from service_client import InventoryClient, SuppressAndLog, log
 from tests.config import ClusterConfig, InfraEnvConfig, TerraformConfig, global_variables
 from tests.config.global_configs import Day2ClusterConfig
+from triggers import get_default_triggers
+from triggers.env_trigger import Trigger
 
 
 class BaseTest:
+    @classmethod
+    def _get_parameterized_keys(cls, request: pytest.FixtureRequest):
```
osherdp, line 46: maybe some more explanation will be needed here

src/tests/base_test.py
```diff
@@ -29,30 +29,60 @@
...
+
+        for key in optional_keys:
+            keys = key.split(",")
+            for k in keys:
+                parameterized_keys.append(k.strip())
+
+        return parameterized_keys
+
+    @classmethod
+    def update_parameterized(cls, request: pytest.FixtureRequest, config: BaseConfig):
```
osherdp, line 58: also here, don't cheap on the docstring :)

src/tests/config/global_configs.py
```diff
@@ -17,22 +17,23 @@
...
-    def _get_iso_download_path(cluster_name: str):
+    def _get_iso_download_path(entity_name: str):
         return str(
-            Path(env_defaults.DEFAULT_IMAGE_FOLDER).joinpath(f"{cluster_name}-{env_defaults.DEFAULT_IMAGE_FILENAME}")
+            Path(env_defaults.DEFAULT_IMAGE_FOLDER).joinpath(f"{entity_name}-{env_defaults.DEFAULT_IMAGE_FILENAME}")
         ).strip()
 
+
+@dataclass
+class ClusterConfig(BaseClusterConfig, _EntityISOPath):
```
osherdp, line 29: we're having multiple inheritance just for using a utility function?
I think ``_get_iso_download_path`` can become something a lot simpler, not sure this complexity is justified

src/tests/config/global_configs.py
```diff
@@ -43,21 +44,14 @@ def __post_init__(self):
 
 
 @dataclass
-class InfraEnvConfig(BaseInfraEnvConfig):
+class InfraEnvConfig(BaseInfraEnvConfig, _EntityISOPath):
```
osherdp, line 47: same here
let's keep it simple

src/tests/conftest.py
osherdp, line 26:
```suggestion
            return sorted(available_versions, key=semver.VersionInfo.parse)
```

src/tests/global_variables/default_variables.py
```diff
@@ -1,39 +1,57 @@
...
+    __instance: ClassVar = None
+
+    def __new__(cls, *args, **kwargs):
+        """Prevent creating another env_var instance"""
+        if isinstance(cls.__instance, cls):
+            raise Exception("Can't initialized more then one global configuration object")
+        cls.__instance = object.__new__(cls, *args, **kwargs)
+        return cls.__instance
+
+    def __getattribute__(self, item):
```
osherdp, line 21: also here, possible ``__getattr__`` will be simpler to use

src/tests/global_variables/env_variables_defaults.py
```diff
@@ -1,95 +1,105 @@
...
-    day2_workers_count: int = int(get_env("NUM_DAY2_WORKERS", env_defaults.DEFAULT_DAY2_WORKERS_COUNT))
-    vip_dhcp_allocation: bool = bool(
-        strtobool(get_env("VIP_DHCP_ALLOCATION", str(env_defaults.DEFAULT_VIP_DHCP_ALLOCATION)))
+class _EnvVariables(DataPool, ABC):
+    ssh_public_key: EnvVar = EnvVar(["SSH_PUB_KEY"])
+    remote_service_url: EnvVar = EnvVar(["REMOTE_SERVICE_URL"])
+    pull_secret: EnvVar = EnvVar(["PULL_SECRET"])
+    deploy_target: EnvVar = EnvVar(["DEPLOY_TARGET"])
+    offline_token: EnvVar = EnvVar(["OFFLINE_TOKEN"])
+    openshift_version: EnvVar = EnvVar(["OPENSHIFT_VERSION"], default=consts.OpenshiftVersion.VERSION_4_9.value)
```
osherdp, line 19: default not taken from the service?

src/tests/test_bootstrap_in_place.py
```diff
@@ -141,12 +143,15 @@ def str_presenter(self, dumper, data):
     def new_cluster_configuration(self, request) -> ClusterConfig:
         return ClusterConfig(cluster_name=ClusterName(prefix="test-infra-cluster", suffix=""))
 
+    @pytest.fixture
+    def triggers(self):
```
osherdp, line 147: doc here is needed

src/triggers/default_triggers.py
```diff
@@ -0,0 +1,60 @@
...
+from triggers.olm_operators_trigger import OlmOperatorsTrigger
+
+_default_triggers = frozendict(
+    {
+        "none_platform": Trigger(
+            ("platform", consts.Platforms.NONE), user_managed_networking=True, vip_dhcp_allocation=False
+        ),
+        "vsphere_platform": Trigger(("platform", consts.Platforms.VSPHERE), user_managed_networking=False),
+        "sno": Trigger(
+            ("masters_count", 1),
```
osherdp, line 16: can we have it that way?
```suggestion
            condition=("masters_count", 1),
```

src/triggers/default_triggers.py
```diff
@@ -0,0 +1,60 @@
...
+        ),
+        "ipv6": Trigger(
+            (
+                ("is_ipv4", False),
+                ("is_ipv6", True),
+            ),
+            cluster_networks=consts.DEFAULT_CLUSTER_NETWORKS_IPV6,
+            service_networks=consts.DEFAULT_SERVICE_NETWORKS_IPV6,
+        ),
+        "ipv6_only": Trigger(
```
osherdp, line 40: maybe change the name to ``ipv6_required_configurations``, because it's confusing with the previous trigger which also seems like "ipv6 only"

src/triggers/env_trigger.py
```diff
@@ -0,0 +1,77 @@
...
+                self._set(k, v)
+            else:
+                log.warning(f"Skipping setting {k} to value {v} due that it already been set by the user")
+
+    @abstractmethod
+    def _set(self, key: str, value: Any):
+        pass
+
+
+class Trigger:
```
osherdp, line 46: class docstring

osherdp (changes requested):

src/assisted_test_infra/test_infra/helper_classes/config/base_config.py
```diff
@@ -1,35 +1,12 @@
...
-
-    @abstractmethod
-    def _set(self, key: str, value: Any):
-        pass
+from triggers.env_trigger import Triggerable
 
 
 @dataclass
-class _BaseConfig(Triggerable, ABC):
+class BaseConfig(Triggerable, ABC):
```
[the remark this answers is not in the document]
eliorerz: `_get_data_pool` is only exists on the configurations implementations and not in `BaseConfig`

src/assisted_test_infra/test_infra/utils/env_var.py
```diff
@@ -0,0 +1,61 @@
...
+    @property
+    def is_user_set(self):
+        return self.__is_user_set
+
+    def get(self, reload: bool = False):
+        # Try to load from cache
+        if self.__cached and not reload:
+            return self.__value
+
+        self.__cached = True
```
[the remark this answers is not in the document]
eliorerz: I wanted that we will have the ability to reload the value on demand, but didn't used that eventually so I'll change it

src/assisted_test_infra/test_infra/helper_classes/config/base_config.py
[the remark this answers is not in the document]
osherdp: so can you add it explicitly to this class?
```
@abstractmethod
def _get_data_pool(self, ...):
    pass
```

osherdp (changes requested):

src/consts/env_defaults.py
eliorerz: The default on CI was 2, and it already changes some of the behavior on CI (https://github.com/openshift/release/pull/25792)
 I just moved it to test-infra

src/tests/conftest.py
eliorerz: `ValueError: 4.6 is not valid SemVer string`

src/tests/global_variables/default_variables.py
eliorerz: Here we can't do it. We want to return the value of the env and not the env itself (and only if it's of type .
When loading the configurations on configuration.__init__ we try to load the defaults. 
I'll document it

src/tests/global_variables/env_variables_defaults.py
eliorerz: It does.
I'll probably had to add documentation here also.
When removing the default from the Makefile it cased some tests to fail. This is a temporary fix until we will remove that dependency.

src/assisted_test_infra/test_infra/helper_classes/config/base_config.py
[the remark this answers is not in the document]
eliorerz: No, because that `_get_data_pool` is more of implementation of this class and it's accessing the environment variables that we removed the access to it few months ago.

src/consts/env_defaults.py
osherdp: okay, so we glitched back then
I guess if no one complained then it's okay

osherdp (approved):

PR #1404, opened 2022-01-26
MGMT-9037:  Use a cheaper packet plan for baremetalds-sno-workflow

paul-maidment:
The purpose of this workflow is to save costs after this change as currently this test demands a 45GB virtual machine.
Coupled with work in [MGMT-8844](https://issues.redhat.com/browse/MGMT-8844) where the SNO validation limit has been reduced to 16GB, it would make sense to reduce this also.
So we have done this.

### DIFF
--- a/src/tests/test_bootstrap_in_place.py
+++ b/src/tests/test_bootstrap_in_place.py
@@ -146,7 +146,7 @@ class TestBootstrapInPlace(BaseTest):
             masters_count=1,
             workers_count=0,
             nodes_count=1,
-            master_memory=45 * 1024,  # in megabytes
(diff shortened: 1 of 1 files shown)


osherdp:
/ok-to-test
@paul-maidment I think you're not on openshift org, making you unauthorized to have your tests running
can you handle that?

osherdp (approved):

romfreiman:
how does this pr affect packet allocation? I think Im missing something here

osherdp:
> how does this pr affect packet allocation? I think Im missing something here

We wanted to move to a 32GB of RAM packet server, and it cannot happen when we provision a 45GB RAM VM

PR #1408, opened 2022-01-31
MGMT-9268: Fix kube-api conformance tests are failing

eliorerz:
/cc @osherdp

### DIFF
--- a/src/tests/base_kubeapi_test.py
+++ b/src/tests/base_kubeapi_test.py
@@ -42,7 +42,7 @@ class BaseKubeAPI(BaseTest):
     @JunitFixtureTestCase()
     def kube_api_context(self, kube_api_client, spoke_namespace):
         kube_api_context = KubeAPIContext(
-            kube_api_client, clean_on_exit=spoke_namespace, spoke_namespace=spoke_namespace
+            kube_api_client, clean_on_exit=global_variables.test_teardown, spoke_namespace=spoke_namespace
         )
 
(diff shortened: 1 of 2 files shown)


osherdp (approved):

osherdp:
/hold
all jobs are broken because of CentOS RPM repos being unreachable

osherdp:
@eliorerz unhold?

osherdp:
this just waist resources, because the nodes are getting deleted at the end and we cannot really do conformance tests against it
so I think the right thing here is to hold until fixing the nodes thing
/hold

eliorerz:
> this just waste resources, because the nodes are getting deleted at the end and we cannot really do conformance tests against it so I think the right thing here is to hold until fixing the nodes thing /hold

Correct me if I'm wrong but IIRC we said that we will hold the conformance tests until we will fix the teardown issue

osherdp:
> > this just waste resources, because the nodes are getting deleted at the end and we cannot really do conformance tests against it so I think the right thing here is to hold until fixing the nodes thing /hold
> 
> Correct me if I'm wrong but IIRC we said that we will hold the conformance tests until we will fix the teardown issue

We changed it to allow situations where no kubeconfig file exists. But once it finds any kubeconfig file, it starts testing the cluster
So we should first make nodes stay up after the test ends, then change kube-api job create the file

osherdp (approved):

PR #1248, opened 2021-11-17
NO-ISSUE: always upgrade libgcrypt package using regular sources

osherdp:
Basically a backport of #1245 (and for even more details on the issue please see #1244)

Some fedora-like distributions come with a pretty outdated version of libgcrypt, making use of kvm unsustainable
Previously we tried installing a specific version from some source, but this source seems to remove versions from time to time which is inconvenient

This change will just make sure we upgrade the package to its latest version by using the official repositories of the distribution, which should be fine for all cases

Failing build for example:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_assisted-service/2939/pull-ci-openshift-assisted-service-release-ocm-2.4-e2e-metal-assisted-static-ip/1461021226951512064

/cc @carbonin @tsorya @lranjbar

PR #1418, opened 2022-01-31
[release-ocm-2.4] MGMT-9102: always upgrade libgcrypt package using regular sources

eliorerz:
See https://github.com/openshift/assisted-test-infra/pull/1248 for more details


osherdp (approved):

lranjbar:
Is this a case we need to override the failing test? It isn't building but unsure if that is because it is so far behind now that we had to rollback

osherdp:
/hold
saving resources

eliorerz:
This change will be merged with https://github.com/openshift/assisted-test-infra/pull/1431

PR #1430, opened 2022-02-02
Supporting arbitrary user ids

michaellevy101:
Openshift runs containers as root group and assigns user IDs arbitrarily for security reasons.
Since this image has been built as root, the arbitrarily user does not have permission to write to the assisted-test-infra folder.
Allow root group write access to the folder, so that arbitrary users can write there
Please see: https://docs.openshift.com/container-platform/4.9/openshift_images/create-images.html#images-create-guide-openshift_create-images

### DIFF
--- a/Dockerfile.assisted-test-infra
+++ b/Dockerfile.assisted-test-infra
@@ -40,5 +40,8 @@ RUN go get -u github.com/vmware/govmomi/govc
 
 COPY . .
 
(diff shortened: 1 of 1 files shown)


eliorerz:
Please add a bit of documentation and not just link to the guide

lranjbar:
What are we trying to do or fix with this?

eliorerz (approved):
LGTM

PR #1433, opened 2022-02-03
MGMT-8707: Testing custom swtpm

nmagnezi:


nmagnezi:
/assign @osherdp 
/assign @eliorerz 

just an FYI, will watch for CI results.

scripts/install_environment.sh
```diff
@@ -43,6 +43,21 @@ function install_libvirt() {
         swtpm \
         swtpm-tools
 
+    # Just testing, do not merge.
+    sudo curl https://github.com/nmagnezi/swtpm/raw/PR-645/bin/swtpm > /usr/bin/swtpm
```
osherdp, line 47: maybe you'll need to do this, but let's see:
```suggestion
    sudo curl https://github.com/nmagnezi/swtpm/raw/PR-645/bin/swtpm -o /usr/bin/swtpm
```

osherdp:
in doubt, always hold :) 
/hold

scripts/install_environment.sh
nmagnezi: Tested this locally and it worked like that.  I did that to override the existing files deployed from the RPMs

scripts/install_environment.sh
```diff
@@ -43,6 +43,22 @@ function install_libvirt() {
         swtpm \
         swtpm-tools
```
osherdp, line 44: @nmagnezi BTW
with those still in place, are you sure the logic of the fix is actually being used?
because the binaries might still use dynamically linked libraries which are of the original code

nmagnezi: I left it in place because I tried to avoid deploying dependencies manually.but sure we can try to remove and see what happens.

filanov:
@nmagnezi isn't that the fix that you talked about?

nmagnezi:
> @nmagnezi isn't that the fix that you talked about?

@filanov yes. It looks good but as I wrote in https://github.com/stefanberger/swtpm/issues/644#issuecomment-1038985377, I want to give it a few more runs before we confirm it.

filanov:
@nmagnezi consider to reopen this PR so you will not forget about it.

nmagnezi:
@filanov, yup.
I'm not sure why, but it did not allow me to reopen it, so followed up with another PR to continue with the tests https://github.com/openshift/assisted-test-infra/pull/1467

PR #1438, opened 2022-02-06
NO-ISSUE - Change pytest default workers count to 3

eliorerz:
/cc @osherdp


osherdp (approved):
add some more information on why we're doing it

PR #1439, opened 2022-02-06
NO-ISSUE - Change openshift version set priority

eliorerz:
/cc @osherdp


osherdp (approved):
/hold
please add some explanation to the PR

PR #1441, opened 2022-02-06
Fix image_type param in infra_env and adding minimal iso event

lalon4:
Added a condition to get either cluster events or infra env events,
In addition - minimal iso flow is broken. Fixed by passing image_type to infra_env creation.

### DIFF
--- a/src/assisted_test_infra/test_infra/helper_classes/infra_env.py
+++ b/src/assisted_test_infra/test_infra/helper_classes/infra_env.py
@@ -42,6 +42,7 @@ class InfraEnv(Entity):
             static_network_config=self._config.static_network_config,
             ignition_config_override=ignition_config_override,
             proxy=self._config.proxy,
+            image_type=self._config.iso_image_type,
         )
         self._config.infra_env_id = infra_env.id
         return infra_env.id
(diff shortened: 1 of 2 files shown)


src/assisted_test_infra/test_infra/helper_classes/cluster.py
```diff
@@ -1022,8 +1022,9 @@ def wait_for_event(self, event_to_find, reference_time, params_list=None, host_i
             PendingDeprecationWarning,
         )
         handler = EventsHandler(self.api_client)
+        cluster_id = "" if infra_env_id else self.id
```
eliorerz, line 1025: Why?

lalon4: cause otherwise both cluster id && infra env are sent - the results are only events with both cluster id and infra env id. there are events with only infra env id

eliorerz: This is deprecated function.
Use `handler.get_events(host_id, self.id, infra_env_id)` in your test instead

eliorerz:
Run `make reformat` to fix lint failure

src/assisted_test_infra/test_infra/helper_classes/cluster.py
eliorerz: No need to use this method anymore. We stopped maintaining this code.
You can create `EventsHandler` or to get by calling `events_handler` fixture.
Next, instead of calling `cluster.wait_for_event` (while ignoring the cluster id) just call `events_handler.wait_for_event` with your desired keys (`cluster_id` or `infra_env_id`)

lalon4: using fixture in test: https://gitlab.cee.redhat.com/ocp-edge-qe/kni-assisted-installer-auto/-/merge_requests/689/diffs

eliorerz (approved):
LGTM

PR #1445, opened 2022-02-07
[release-ocm-2.4] MGMT-9197 - Fix kubeapi test

eliorerz:
- Remove default openshift version from configurations.
- Set CI flag to true as default

/cc @osherdp

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@
 SHELL=/bin/sh
 CONTAINER_COMMAND = $(shell if [ -x "$(shell command -v docker)" ];then echo "docker" ; else echo "podman";fi)
 PULL_PARAM=$(shell if [ "${CONTAINER_COMMAND}" = "podman" ];then echo "--pull-always" ; else echo "--pull";fi)
+CI ?= true
 
 ROOT_DIR = $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
 REPORTS = $(ROOT_DIR)/reports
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -5,6 +5,7 @@
 SHELL=/bin/sh
 CONTAINER_COMMAND = $(shell if [ -x "$(shell command -v docker)" ];then echo "docker" ; else echo "podman";fi)
 PULL_PARAM=$(shell if [ "${CONTAINER_COMMAND}" = "podman" ];then echo "--pull-always" ; else echo "--pull";fi)
+CI ?= true
```
osherdp, line 8: what effect does it have and how is it connected to kube-api test?

Makefile
```diff
@@ -499,4 +500,4 @@ _test_parallel: $(REPORTS) _test_setup
 
 # leave the target name the same to keep consistent behavior with master branch
 test_kube_api_parallel:
-	TEST=./discovery-infra/tests/test_kube_api.py make test
\ No newline at end of file
+	TEST=./discovery-infra/tests/test_kube_api.py make test_parallel
```
osherdp, line 503: can we please conform to adding blank line at end of file?

discovery-infra/tests/config/global_configs.py
```diff
@@ -14,6 +15,10 @@
 global_variables = GlobalVariables()
 
 
+log.debug(f"global_variables={global_variables}")
+log.debug(f"environ={os.environ}")
```
osherdp, line 19: seems a bit messy to have this kind of log on module level
can we add it anywhere else?

osherdp (changes requested):

Makefile
eliorerz: It's override the definition on assisted-service Makefile so that when the service is deployed it will have all the available images.
It's not related to kubeapi test directly but the test run on 4.8 and I wanted that test-infra will take the default value from assisted-service and not form other places

osherdp (approved):

osherdp (approved):

PR #1450, opened 2022-02-08
MGMT-8257: refactored day2 logic into test_day2

danielerez:
* Extract logic from day2.py to Day2Cluster class.
  * Used LibvirtController instead of deprecated code.
  * Removed unused 'ocp' day2 flow.
  * Deleted deprecated day2.py.


osherdp:
@danielerez suggesting improvements to the existing code is in the scope of this PR?
e.g. changing ``args`` to function parameters

danielerez:
> @danielerez suggesting improvements to the existing code is in the scope of this PR? e.g. changing `args` to function parameters

Can handle some in this context. You mean extracting to a struct as suggested here: https://github.com/openshift/assisted-test-infra/blob/master/src/tests/test_day2.py#L29 ?
It would be too many func params otherwise:)

osherdp:
> > @danielerez suggesting improvements to the existing code is in the scope of this PR? e.g. changing `args` to function parameters
> 
> Can handle some in this context. You mean extracting to a struct as suggested here: https://github.com/openshift/assisted-test-infra/blob/master/src/tests/test_day2.py#L29 ? It would be too many func params otherwise:)

Yes, that can work well too

danielerez:
> @danielerez suggesting improvements to the existing code is in the scope of this PR? e.g. changing `args` to function parameters

changed.

src/tests/test_day2.py
```diff
@@ -26,18 +39,299 @@ def test_deploy_day2_nodes_cloud(self, cluster, new_cluster_configuration: Clust
...
+        )
+
+    @classmethod
+    def execute_day2_flow(
+        cls,
+        api_client: Any,
+        proxy: models.Proxy,
+        cluster_id: str,
+        pull_secret: str,
+        ssh_key: str,
```
osherdp, line 61: in order to not strip meaning from this variable name, I think this should be better:
```suggestion
        ssh_public_key: str,
```

src/tests/test_day2.py
```diff
@@ -26,18 +39,299 @@ def test_deploy_day2_nodes_cloud(self, cluster, new_cluster_configuration: Clust
...
+        self.execute_day2_flow(
+            cluster.api_client,
+            new_cluster_configuration.proxy,
+            new_cluster_configuration.cluster_id,
+            new_cluster_configuration.pull_secret,
+            global_variables.ssh_public_key,
+            global_variables.num_day2_workers,
+            new_cluster_configuration.is_static_ip,
+            new_cluster_configuration.is_ipv6,
+            True,
```
osherdp, line 51:
```suggestion
            install_cluster=True,
```

src/tests/test_day2.py
```diff
@@ -26,18 +39,299 @@ def test_deploy_day2_nodes_cloud(self, cluster, new_cluster_configuration: Clust
...
+            ssh_public_key=ssh_key,
+            static_network_config=static_network_config,
+            pull_secret=pull_secret,
+            openshift_version=openshift_version,
+        )
+        # Download image
+        iso_download_url = infra_env.download_url
+        image_path = os.path.join(consts.IMAGE_FOLDER, f"{cluster_name}-installer-image.iso")
+        log.info(f"Downloading image {iso_download_url} to {image_path}")
+        utils.download_file(iso_download_url, image_path, False)
```
osherdp, line 110: what's the meaning of this ``False`` value?

src/tests/test_day2.py
```diff
@@ -26,18 +39,299 @@ def test_deploy_day2_nodes_cloud(self, cluster, new_cluster_configuration: Clust
...
+            new_cluster_configuration.cluster_id,
+            new_cluster_configuration.pull_secret,
+            global_variables.ssh_public_key,
+            global_variables.num_day2_workers,
+            new_cluster_configuration.is_static_ip,
+            new_cluster_configuration.is_ipv6,
+            True,
+        )
+
+    @classmethod
```
osherdp, line 54:
```suggestion
    @classmethod
```

just wondering about all those ``classmethod``s, do we gain anything from having a class that only has class methods, or maybe we can just transform to the less-meta use of just instance methods
In my opinion, we should choose the latter for simplicity and because pytest will eventually instantiate those classes

osherdp (changes requested):
added just a couple of comments for things that bothered me

src/tests/test_day2.py
danielerez: fixed

src/tests/test_day2.py
danielerez: yeah not much benefit, removed

src/tests/test_day2.py
danielerez: changed

src/tests/test_day2.py
danielerez: disables ssl verify

eliorerz (changes requested):
- Please don't put all the code in the test file.
- Keep the test function short. 
- Define any test preparations in pytest fixtures.

Note: Maybe you can create `Day2Cluster` object and use it instead of `Cluster` object (the same you did with `Day2ClusterConfig`)

danielerez:
> * Please don't put all the code in the test file.
> * Keep the test function short.
> * Define any test preparations in pytest fixtures.
> 
> Note: Maybe you can create `Day2Cluster` object and use it instead of `Cluster` object (the same you did with `Day2ClusterConfig`)

So the scope here actually is to avoid using any deprecated code, it's not meant for a full rewrite of the code:)
I can open a new task for that to tackle later...
Any crucial issue in this PR that doesn't require major changes? Do you think we should extract the helper functions to another file?

osherdp (approved):

eliorerz:
> So the scope here actually is to avoid using any deprecated code, it's not meant for a full rewrite of the code:)

Basically yes, I don't want it to be the same as kube-api test file 

> I can open a new task for that to tackle later...

If you think that this is really out of scope so OK

> Any crucial issue in this PR that doesn't require major changes? Do you think we should extract the helper functions to another file?

Whatever that you can remove from this file to external file that would be great

src/assisted_test_infra/test_infra/helper_classes/day2_cluster.py
```diff
@@ -0,0 +1,285 @@
...
+from assisted_test_infra.test_infra.controllers.node_controllers.libvirt_controller import LibvirtController
+from assisted_test_infra.test_infra.tools import static_network
+from assisted_test_infra.test_infra.utils.waiting import wait_till_all_hosts_are_in_status
+from service_client import log
+from service_client.assisted_service_api import InventoryClient
+from tests.config import ClusterConfig, TerraformConfig, global_variables
+from tests.config.global_configs import Day2ClusterConfig
+
+
+class Day2Cluster(ABC):
```
osherdp, line 22: maybe I missed something, but what's abstract about that class?

osherdp (changes requested):

src/assisted_test_infra/test_infra/helper_classes/day2_cluster.py
osherdp: other than that, code looks great!

src/assisted_test_infra/test_infra/helper_classes/day2_cluster.py
```diff
@@ -0,0 +1,285 @@
...
+
+import consts
+from assisted_test_infra.test_infra import utils
+from assisted_test_infra.test_infra.controllers.node_controllers.libvirt_controller import LibvirtController
+from assisted_test_infra.test_infra.tools import static_network
+from assisted_test_infra.test_infra.utils.waiting import wait_till_all_hosts_are_in_status
+from service_client import log
+from service_client.assisted_service_api import InventoryClient
+from tests.config import ClusterConfig, TerraformConfig, global_variables
+from tests.config.global_configs import Day2ClusterConfig
```
eliorerz, line 19: Do not import tests package into test-infra package.
You can use `BaseClusterConfig` and `BaseNodeConfig` instead.
`global_variables` can be used only outside test_infra package

eliorerz (changes requested):
Also, Please add type (and return) hint annotations

src/assisted_test_infra/test_infra/helper_classes/day2_cluster.py
```diff
@@ -0,0 +1,285 @@
...
+        self.config.day1_cluster_name = cluster.name
+        openshift_version = cluster.openshift_version
+        api_vip_dnsname = "api." + self.config.day1_cluster_name + "." + cluster.base_dns_domain
+        api_vip_ip = cluster.api_vip
+
+        openshift_cluster_id = str(uuid.uuid4())
+        day2_cluster_create_params = {
+            "openshift_version": openshift_version,
+            "api_vip_dnsname": api_vip_dnsname,
+        }
```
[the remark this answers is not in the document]
eliorerz: Max line length supposed to be 120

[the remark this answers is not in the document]
danielerez: yeah just trimmed the var name a bit, so squeezed into one line.

src/assisted_test_infra/test_infra/helper_classes/day2_cluster.py
```diff
@@ -0,0 +1,269 @@
...
+        return False
+
+    def are_libvirt_nodes_in_cluster_hosts(self, client, cluster_id, num_nodes):
+        try:
+            hosts_macs = client.get_hosts_id_with_macs(cluster_id)
+        except BaseException:
+            log.error("Failed to get nodes macs for cluster: %s", cluster_id)
+            return False
+        num_macs = len([mac for mac in hosts_macs if mac != ""])
+        return num_macs >= num_nodes
```
eliorerz, line 238: why not using `self._nodes.nodes_count`?

src/assisted_test_infra/test_infra/helper_classes/day2_cluster.py
```diff
@@ -0,0 +1,269 @@
...
+
+    def set_nodes_hostnames_if_needed(self, client, tf_folder, is_static_ip, has_ipv_6, network_name, cluster_id):
```
eliorerz, line 240: No need to pass arguments that you already have in the cluster configuration

src/assisted_test_infra/test_infra/helper_classes/config/day2_cluster_config.py
```diff
@@ -0,0 +1,13 @@
...
+from dataclasses import dataclass
+
+from assisted_test_infra.test_infra.helper_classes.config.cluster_config import BaseClusterConfig
+
+
+@dataclass
+class BaseDay2ClusterConfig(BaseClusterConfig, ABC):
+    day1_cluster_id: str = None
+    day1_cluster_name: str = None
+    num_day2_workers: int = None
```
eliorerz, line 11: Change to `day2_workers_count` or `workers_count_day2` to be consistent with `ClusterConfig.workers_count`

eliorerz (changes requested):

src/assisted_test_infra/test_infra/helper_classes/day2_cluster.py
```diff
@@ -0,0 +1,285 @@
...
+        for i, line in enumerate(hosts_lines):
+            if api_vip_dnsname in line:
+                hosts_lines[i] = api_vip_ip + " " + api_vip_dnsname + "\n"
+                break
+        else:
+            hosts_lines.append(api_vip_ip + " " + api_vip_dnsname + "\n")
+        with open("/etc/hosts", "w") as f:
+            f.writelines(hosts_lines)
+
+    def configure_terraform(self, tf_folder, num_worker_nodes, api_vip_ip):
```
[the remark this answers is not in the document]
eliorerz: Maybe create specific controller (that inherits from TerraformController) for day2? 
@osherdp WDYT?

src/assisted_test_infra/test_infra/helper_classes/day2_cluster.py
danielerez: it's the day2 nodes, can use self.config.num_day2_workers

src/assisted_test_infra/test_infra/helper_classes/day2_cluster.py
danielerez: done

src/assisted_test_infra/test_infra/helper_classes/config/day2_cluster_config.py
danielerez: done

src/assisted_test_infra/test_infra/helper_classes/day2_cluster.py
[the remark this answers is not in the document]
osherdp: let's inherit as less as we can, and instead parameterize the objects as needed

eliorerz (approved):
LGTM

PR #1452, opened 2022-02-08
NO-ISSUE - Separate download_logs logs from test_infra logs

eliorerz:
Due to that `make test` and `make download_cluster_logs` are writing to the same log file and both of them uses `download_logs.py` we need to find a way to identify which  log entry is "came from".
Also in this PR - Add the ability to debug download_logs package.

/cc @osherdp


eliorerz:
/unhold
```
2022-02-08 14:13:04,681 - root - INFO - 139736593461504 - Creating assisted-service client for url: http://10.104.9.2:8090

2022-02-08 16:12:35,625 - download_logs - INFO - 139814280069376 - Creating assisted-service client for url: http://192.168.39.196:31853
```

scripts/download_logs.sh
```diff
@@ -14,6 +14,8 @@ JUNIT_REPORT_DIR=${JUNIT_REPORT_DIR:-"reports/"}
 KUBE_CRS=(clusterdeployment infraenv agentclusterinstall agent)
 CAPI_PROVIDER_CRS=(agentmachine agentcluster cluster machine machinedeployment machineset)
 ENABLE_KUBE_API=${ENABLE_KUBE_API:-"false"}
+DEBUG_FLAGS=${DEBUG_FLAGS:-""}
```
osherdp, line 17: ``DEBUG_FLAGS`` is too generic. What specifically does it debug?

src/service_client/logger.py
```diff
@@ -72,30 +74,40 @@ def emit(self, record):
...
-def add_stream_handler():
-    fmt = SensitiveFormatter("%(asctime)s %(levelname)-10s - %(thread)d - %(message)s \t" "(%(pathname)s:%(lineno)d)")
-    ch = ColorizingStreamHandler(sys.stdout)
-    ch.setFormatter(fmt)
-    log.addHandler(ch)
-
-
-add_log_file_handler("test_infra.log")
-add_stream_handler()
+class Logger(ABC):
```
osherdp, line 77: what's abstract about this class?
there are no signs for anything that is abstract or that should be implemented by inheriting classes

osherdp (changes requested):

scripts/download_logs.sh
eliorerz: The IDE (pycharm/vscode)

src/service_client/logger.py
eliorerz: Just preventing from users to create instance from it

scripts/download_logs.sh
eliorerz: If we want to change it will do on another PR

osherdp: okay, but let's settle on the "next PR" (and not "another PR") :joy:

osherdp (changes requested):

src/service_client/logger.py
eliorerz: I restored it to its previous state

osherdp (approved):

osherdp:
/hold
lint fails

osherdp (approved):

osherdp (approved):

PR #1454, opened 2022-02-08
Support cluster & infra-env & hosts events

lalon4:
Deleting wait_for_event as it not being used anymore by kni repo.
We use events_handler fixture instead


eliorerz (approved):
LGTM

PR #1457, opened 2022-02-10
DRAFT: add get_preflight to cluster

talhil-rh:
/cc @lalon4


src/assisted_test_infra/test_infra/helper_classes/cluster.py
```diff
@@ -155,6 +155,9 @@ def get_host_assigned_roles(self):
     def get_operators(self):
         return self.api_client.get_cluster_operators(self.id)
 
+    def get_preflight_requirements(self):
+        self.api_client.get_preflight_requirements(self.id)
```
lalon4, line 159: return

PR #1459, opened 2022-02-13
add get_preflight_requirements to cluster

talhil-rh:
/cc @lalon4


talhil-rh:
get_preflight_requirements is used instead of get_cluster_host_requirements (v1 api). we use it in kni-assisted-installer-auto repo

PR #1460, opened 2022-02-13
Remove set of `bootstrap_in_place` in kubeapi test

eliorerz:
On the new libvirt version we will get error while trying to change the `bootstrap_in_place` variable.
Also we don't need to set bootstrap_in_place in kube-api test flow.

/cc @tsorya

### DIFF
--- a/src/tests/test_kube_api.py
+++ b/src/tests/test_kube_api.py
@@ -141,9 +141,6 @@ class TestKubeAPI(BaseKubeAPI):
         )
         agents = self.start_nodes(nodes, infra_env, cluster_config)
 
-        if len(nodes) == 1:
-            self.set_single_node_ip(cluster_deployment, nodes)
-
         log.info("Waiting for agent status verification")
         Agent.wait_for_agents_to_install(agents)
 


src/tests/base_kubeapi_test.py
```diff
@@ -169,7 +169,6 @@ def set_single_node_ip(cls, cluster_deployment: ClusterDeployment, nodes: Nodes)
         nodes.controller.tf.change_variables(
             {
                 "single_node_ip": single_node_ip,
```
tsorya, line 171: not sure why we need to run this function , it is relevant only for bootstrap_in_place test

eliorerz: Updated and removed the call for `set_single_node_ip` in kube-api function

osherdp:
> @osherdp we don't test ibip anymore?

We do, it's just got separated from ``start_discovery.py``: https://github.com/openshift/assisted-test-infra/blob/master/src/tests/test_bootstrap_in_place.py

osherdp:
@eliorerz you're sure it will work?
I thought it's needed to customize DNS entries to fit SNO networking

eliorerz:
> @eliorerz you're sure it will work? I thought it's needed to customize DNS entries to fit SNO networking

It's working without setting `bootstrap_in_place`, now testing it also without `single_node_ip`

osherdp (approved):

PR #1472, opened 2022-02-17
NO-ISSUE - Set python3.9 as the python environment default version

eliorerz:
/cc @osherdp

### DIFF
--- a/scripts/install_environment.sh
+++ b/scripts/install_environment.sh
@@ -130,12 +130,13 @@ function install_runtime_container() {
 
 function install_packages() {
     echo "Installing dnf packages"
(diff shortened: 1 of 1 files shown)


osherdp (changes requested):
I think this might be what breaks ``dnf`` and other python CLI tools of the OS
It might be dependent on the specific OS (because Fedora / CentOS / RHEL might default to different Pythohn versions) but it's worth making sure we don't break anything

eliorerz:
@osherdp I'm looking in to it locally, no need to retest again

eliorerz:
Not working on RHEL 8.2. Not prioritized for now.

PR #1483, opened 2022-02-22
Fixing terraform ci scsi controller type and vm name

michaellevy101:


osherdp:
@michaellevy101 can you add to the description how the traceback of this failure looks like?
And where you've seen it? (CI / local env / etc.)

michaellevy101:
> And where you've seen it? (CI / local env / etc.)

It's related to the vsphere CI effort. I created it months ago against a private environment and now i'm testing it against the CI env before creating the actual step on the the vsphere workflow

osherdp (approved):

PR #1484, opened 2022-02-22
Add basic CLI for test infra

eliorerz:
CLI tool for test-infra:
Supported commands:

- `help` - Print a list of all supported commands 
- `list clusters` - List available clusters (name and id)
- `test` - execute test function. Format: `test <test_file> <test_name>` - autocomplete all fields
- `clear` - clear CLI

/cc @eliorerz 

https://user-images.githubusercontent.com/16567084/155281517-daf788ef-8772-4bc5-b2a6-eddf0c04cc76.mp4


cli-requirements.txt
```diff
@@ -0,0 +1 @@
+assisted-service-client
```
osherdp, line 1: where it's being used?

scripts/install_environment.sh
```diff
@@ -130,7 +130,7 @@ function install_runtime_container() {
 
 function install_packages() {
     echo "Installing dnf packages"
-    sudo dnf install -y make python3 python3-pip git jq bash-completion xinetd
+    sudo dnf install -y make python3.9 python3-pip python39-devel git jq bash-completion xinetd
```
osherdp, line 133: bumping python is problematic on some of the versions we currently support

src/tests/config/global_configs.py
```diff
@@ -16,6 +16,11 @@
 global_variables = DefaultVariables()
 
 
+def reset_global_variables():
```
osherdp, line 19: Seems like we're doing something wrong here, rather than instantiating it somewhere top-level and propagating it anywhere needed 
Can we handle it some other way?

osherdp (changes requested):
partial review

src/tests/config/global_configs.py
eliorerz: We are, but we can't change `global_variables` behavior.
Our assumption was that the container is executed once per set of tests with n environment variables.
In this case we are triggering the test directly from code and not from pytest-cli-command, so when the test-infra CLI is triggers the test, it will not update the `global_variables` with the actual env vars values.

scripts/install_environment.sh
eliorerz: It's not needed anymore. I eventually used the test infra container instead of creating virtual environment

cli-requirements.txt
eliorerz: No need that anymore

src/cli/commands/commands_factory.py
```diff
@@ -0,0 +1,47 @@
...
+        "clear": EnvCommand,
+        "help": HelpCommand,
+    }
+
+    @classmethod
+    def get_command(cls, text: str) -> Union[Command, None]:
+        text = text if text else ""
+        factory = cls._supported_commands.get(text.split(" ")[0])
+        try:
+            return factory(text)
```
osherdp, line 34: where this ``factory`` function is defined?

src/cli/commands/env_command.py
```diff
@@ -0,0 +1,38 @@
+import os
+
+from prompt_toolkit.completion import NestedCompleter
+from tabulate import tabulate
+
+from .. import cli_utils
+from .command import Command
+
+
+class EnvCommand(Command):
```
osherdp, line 10: I'm missing a bit docstrings for the command classes

src/cli/commands/test_command.py
```diff
@@ -0,0 +1,69 @@
...
+from service_client import log
+
+from .command import Command
+
+
+class TestCommand(Command):
+    @classmethod
+    def get_completer(cls) -> Completer:
+        proc = subprocess.Popen(
+            "python3 -m pytest --collect-only -q", shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE
```
osherdp, line 19:
```suggestion
            ["python3", "-m", "pytest", "--collect-only", "-q"], stdout=subprocess.PIPE, stderr=subprocess.PIPE
```

src/cli/commands/test_command.py
```diff
@@ -0,0 +1,69 @@
...
+        stdout, _stderr = proc.communicate()
+
+        pattern = r"((?P<file>src/tests/.*\.py)::(?P<class>.*)::(?P<func>.*))"
+        groups = [
+            match.groupdict() for match in [re.match(pattern, line) for line in stdout.decode().split("\n")] if match
+        ]
+
+        groups_set = set()
+        for group in groups:
+            groups_set.add((group["file"], group.get("func", "").split("[")[0]))
```
osherdp, line 30:
```suggestion
        groups_set = set(
            (group["file"], group.get("func", "").split("[")[0])
            for group in groups
        )
```

osherdp, line 30: maybe add some comment to explain what's happening here with the parsing

src/cli/commands/test_command.py
```diff
@@ -0,0 +1,69 @@
...
+
+            test_options[file][func] = None
+
+        return NestedCompleter.from_nested_dict({"test": test_options})
+
+    def handle(self):
+        if not self._text:
+            return
+
+        environ_bak = deepcopy(os.environ)
```
osherdp, line 45: ``environ_bak`` -> ``environ_backup`` or ``original_environ``

src/cli/completers.py
```diff
@@ -0,0 +1,96 @@
...
+    def get_completions(self, document, complete_event):
+        text = document.text_before_cursor.lstrip()
+        separator_in_text = False
+
+        separators = ["=", "/"]
+        for sep in separators:
+            if not text.endswith(sep):
+                continue
+
+            args = re.findall(re.compile(r"([^\s=]+)=(.*?)(?=(?:\s[^\s=]+=|$))"), text)
```
osherdp, line 67: some explanation is needed here

osherdp (changes requested):

src/consts/consts.py
```diff
@@ -19,18 +19,30 @@ class NetworkType:
     OpenShiftSDN = "OpenShiftSDN"
     OVNKubernetes = "OVNKubernetes"
 
+    @classmethod
+    def __all__(cls):
+        return [cls.OpenShiftSDN, cls.OVNKubernetes]
```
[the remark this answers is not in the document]
eliorerz: I think we should changed that.
A while back I suggested that and there was some resistance to change those classes to Enum due to the need to access the value with extra action. e.g. `NetworkType.OpenShiftSDN.value`

src/cli/commands/commands_factory.py
eliorerz: On line 32: `factory = cls._supported_commands.get(text.split(" ")[0])`
It takes it from `CommandFactory._supported_commands`

src/cli/application.py
```diff
@@ -0,0 +1,40 @@
...
+        self._init_window()
+        while True:
+            try:
+                command = self._prompt_handler.get_prompt_results()
+            except InvalidCommandError:
+                print("\033[1;31mError, invalid command!\033[0m")
+                continue
+            if command is None:
+                break
+            if not command.is_valid:
```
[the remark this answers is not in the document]
eliorerz: I removed it from all the commands. It's no longer needed after creating the `DummyCommand` and raising `InvalidCommandError` on invalid command.

osherdp (approved):

PR #1485, opened 2022-02-23
add update_static_network_config to infra_env

nshidlin:
/cc @eliorerz @lalon4

### DIFF
--- a/src/assisted_test_infra/test_infra/helper_classes/infra_env.py
+++ b/src/assisted_test_infra/test_infra/helper_classes/infra_env.py
@@ -98,6 +98,11 @@ class InfraEnv(Entity):
         infra_env_update_params = models.InfraEnvUpdateParams(proxy=self._config.proxy)
         self.api_client.update_infra_env(infra_env_id=self.id, infra_env_update_params=infra_env_update_params)
 
+    def update_static_network_config(self, static_network_config: List[dict]) -> None:
+        self.update_config(static_network_config=static_network_config)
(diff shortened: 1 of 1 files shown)


src/assisted_test_infra/test_infra/helper_classes/infra_env.py
```diff
@@ -98,6 +98,11 @@ def update_proxy(self, proxy: models.Proxy) -> None:
         infra_env_update_params = models.InfraEnvUpdateParams(proxy=self._config.proxy)
         self.api_client.update_infra_env(infra_env_id=self.id, infra_env_update_params=infra_env_update_params)
 
+    def update_static_network_config(self, static_network_config: List[dict]) -> None:
+        self.update_config(static_network_config=static_network_config)
+        infra_env_update_params = models.InfraEnvUpdateParams(static_network_config=static_network_config)
```
eliorerz, line 103: Maybe instead of passing `InfraEnvUpdateParams` just change the function signature to be something like this 
```
    def update_infra_env(self, infra_env_id: str, **kwargs):
        infra_env_update_params = models.InfraEnvUpdateParams(**kwargs)
        log.info("Updating infra env %s with values %s", infra_env_id, infra_env_update_params)
        self.client.update_infra_env(infra_env_id=infra_env_id, infra_env_update_params=infra_env_update_params)
```
WDYT?

eliorerz (approved):
LGTM

PR #1486, opened 2022-02-24
NO-ISSUE - Fix missing default version on specific openshift version

eliorerz:
https://github.com/openshift/assisted-service/pull/3380 caused KeyError when trying to get the default version (if `OPENSHIFT_VERSION` is set to version different than the default version). 
In that case we tried to get the default version even if it's not exists in the service. 
The solution to this issue is: moving the `default_version` after the 'override_version' (`OPENSHIFT_VERSION`). 

/cc @osherdp


osherdp:
Can you add more information?
Tracebacks, links, explanation

osherdp (approved):

PR #660, opened 2021-04-04
Install minikube and kubectl using arkade

YuviGold:
Instead of maintaining different URLS for tools installations,
arkade provides a portable marketplace for downloading devops CLIs.

- Bumping minikube to v1.18.1
- Bumping kubectl to v1.20.0

stdout:
```bash
Installing minikube and oc
Installing arkade...
x86_64
Downloading package https://github.com/alexellis/arkade/releases/download/0.7.10/arkade as /tmp/arkade
Download complete.

Running with sufficient permissions to attempt to move arkade to /usr/local/bin
New version of arkade installed to /usr/local/bin
which: no ark in (/sbin:/bin:/usr/sbin:/usr/bin)
Creating alias 'ark' for 'arkade'.
main: line 176: arkade: command not found
Installing minikube...
Downloading minikube
https://github.com/kubernetes/minikube/releases/download/v1.18.1/minikube-linux-amd64
9.20 MiB / 57.02 MiB [----------->__________________________________________________________] 16.14%21.71 MiB / 57.02 MiB [-------------------------->__________________________________________] 38.07%34.01 MiB / 57.02 MiB [----------------------------------------->___________________________] 59.65%46.61 MiB / 57.02 MiB [-------------------------------------------------------->____________] 81.73%57.02 MiB / 57.02 MiB [--------------------------------------------------------------------] 100.00%Tool written to: /root/.arkade/bin/minikube

# Add (minikube) to your PATH variable
export PATH=$PATH:$HOME/.arkade/bin/

# Test the binary:
/root/.arkade/bin/minikube

# Or install with:
sudo mv /root/.arkade/bin/minikube /usr/local/bin/

Installing kubectl...
Downloading kubectl
https://storage.googleapis.com/kubernetes-release/release/v1.20.0/bin/linux/amd64/kubectl
20.00 MiB / 38.37 MiB [----------------------------------->_________________________________] 52.13%38.37 MiB / 38.37 MiB [--------------------------------------------------------------------] 100.00%Tool written to: /root/.arkade/bin/kubectl

# Add (kubectl) to your PATH variable
export PATH=$PATH:$HOME/.arkade/bin/

# Test the binary:
/root/.arkade/bin/kubectl

# Or install with:
sudo mv /root/.arkade/bin/kubectl /usr/local/bin/
```

PR #1488, opened 2022-02-27
NO-ISSUE - Remove arkade dependency

eliorerz:
Currently we are using `arkade` to install minikube and kubectl during environment setup.
After the change of `arkade` domains (that caused our CI to fail), we reconsidered the usage in this tool.

This PR also upgrade:
- `minikube` - from v1.20.0 to latest (v1.25.2)
- `kubectl` - from v1.20.1 to latest (v1.22.0)

/cc @osherdp

### DIFF
--- a/scripts/install_minikube.sh
+++ b/scripts/install_minikube.sh
@@ -2,41 +2,24 @@
 set -euxo pipefail
 export SUDO=$(if [ -x "$(command -v sudo)" ]; then echo "sudo"; else echo ""; fi)
 
+
 function install_minikube() {
-    if [ "${DEPLOY_TARGET}" != "minikube" ]; then
-        echo "Skips installing minikube when deployment target is ${DEPLOY_TARGET}..."
-        return
-    fi
-
-    minikube_version=v1.23.2
-    if ! [ -x "$(command -v minikube)" ]; then
-        echo "Installing minikube..."
-        download_minikube $minikube_version
-        ${SUDO} mv -f ${HOME}/.arkade/bin/minikube /usr/local/bin/
-    elif [ "$(minikube version | grep version | awk -F'version: *' '{print $2}')" != "$minikube_version" ]; then
-        echo "Upgrading minikube..."
-        download_minikube $minikube_version
-        ${SUDO} mv -f ${HOME}/.arkade/bin/minikube $(command -v minikube)
-    else
-        echo "minikube is already installed and up-to-date"
-    fi
+    minikube_version=v1.25.2
+    curl --retry 3 -Lo minikube https://storage.googleapis.com/minikube/releases/${minikube_version}/minikube-linux-amd64
+    ${SUDO} install minikube /usr/local/bin/
+    minikube version
 }
 
-function download_minikube() {
-    for i in {1..4}; do
-        arkade get minikube --version=$1 && break
-        echo "minikube installation failed. Retrying again in 5 seconds..."
-        sleep 5
-    done
-}
 function install_kubectl() {
-    if ! [ -x "$(command -v kubectl)" ]; then
-        echo "Installing kubectl..."
-        arkade get kubectl --version=v1.22.0
-        ${SUDO} mv ${HOME}/.arkade/bin/kubectl /usr/local/bin/
-    else
-        echo "kubectl is already installed"
-    fi
+    kubectl_version=v1.23.0
+    curl --retry 3 -LO https://dl.k8s.io/release/${kubectl_version}/bin/linux/amd64/kubectl
+    chmod +x kubectl
+
+    curl -LO https://dl.k8s.io/${kubectl_version}/bin/linux/amd64/kubectl.sha256
+    echo "$(<kubectl.sha256)  kubectl" | sha256sum --check
+    rm -f kubectl.sha256
+    ${SUDO} install kubectl /usr/local/bin/
+    kubectl
 }
 
 function install_oc() {
@@ -52,22 +35,7 @@ function install_oc() {
     fi
 }
 
-function install_arkade() {
-    if ! [ -x "$(command -v arkade)" ]; then
-        echo "Installing arkade..."
-        for i in {1..4}; do
-            curl --retry 3 -sLS https://get.arkade.dev | ${SUDO} sh && break
-            echo "arkade installation failed. Retrying again in 5 seconds..."
-            sleep 5
-        done
-        arkade version # Fail installation if not exist
-        echo "successfully installed arkade!"
-    else
-        echo "arkade is already installed"
-    fi
-}
 
-install_arkade
 install_minikube
 install_kubectl
 install_oc


osherdp (changes requested):
All PRs should come with a description which is not "cc osherdp"
Plus, what was the original rational of using arkade? Maybe we miss something here

eliorerz:
> All PRs should come with a description which is not "cc osherdp" Plus, what was the original rational of using arkade? Maybe we miss something here

LOL, I know, That is way

eliorerz:
> All PRs should come with a description which is not "cc osherdp" Plus, what was the original rational of using arkade? Maybe we miss something here

LOL, I'm already working on description, this is why the WIP is there :) 

And it doesn't seems that we miss something here, https://github.com/openshift/assisted-test-infra/pull/660, I think it was used for future easy installations but we ended up not using it.

osherdp:
> > All PRs should come with a description which is not "cc osherdp" Plus, what was the original rational of using arkade? Maybe we miss something here
> 
> LOL, I'm already working on description, this is why the WIP is there :)
> 
> And it doesn't seems that we miss something here, #660, I think it was used for future easy installations but we ended up not using it.

For WIP I recommend the good old "cc <author>"

Can't we use arkade for installation of oc as well?
I do think it looks better than the alternative

eliorerz:
>Can't we use arkade for installation of oc as well?

From a quick look I don't think we can do that 

>I do think it looks better than the alternative

It doesn't look better, but why do we need this extra dependency when we can install it directly from source?

osherdp:
> > Can't we use arkade for installation of oc as well?
> 
> From a quick look I don't think we can do that
> 
> > I do think it looks better than the alternative
> 
> It doesn't look better, but why do we need this extra dependency when we can install it directly from source?

No, I mean usage of arkade looks better and that's an advantage for using it
Additionally, I think we suffered from changing URLs for minikube or kubectl, and this was the catalyst for moving to arkade

scripts/install_minikube.sh
```diff
@@ -2,41 +2,18 @@
...
-        ${SUDO} mv -f ${HOME}/.arkade/bin/minikube /usr/local/bin/
-    elif [ "$(minikube version | grep version | awk -F'version: *' '{print $2}')" != "$minikube_version" ]; then
-        echo "Upgrading minikube..."
-        download_minikube $minikube_version
-        ${SUDO} mv -f ${HOME}/.arkade/bin/minikube $(command -v minikube)
-    else
-        echo "minikube is already installed and up-to-date"
-    fi
+function install_minikube() {
+    curl -Lo minikube https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64
```
romfreiman, line 7: maybe to control the version?

eliorerz: We can control the version when getting it from minikube directly, e.g.

```bash
    export MINIKUBE_VERSION=v1.20.0
    ... 
    curl -Lo minikube https://storage.googleapis.com/minikube/releases/${MINIKUBE_VERSION}/minikube-linux-amd64
```

osherdp:
> > > Can't we use arkade for installation of oc as well?
> > 
> > 
> > From a quick look I don't think we can do that
> > > I do think it looks better than the alternative
> > 
> > 
> > It doesn't look better, but why do we need this extra dependency when we can install it directly from source?
> 
> No, I mean usage of arkade looks better and that's an advantage for using it Additionally, I think we suffered from changing URLs for minikube or kubectl, and this was the catalyst for moving to arkade

That being said, I don't think one of the options is significantly better than the other one
So if you prefer to use direct sources and "skip the middle man", no problem at all

scripts/install_minikube.sh
osherdp: yes, let's control the version
it's usually better than the alternative

scripts/install_minikube.sh
```diff
@@ -2,41 +2,18 @@
...
-}
 function install_kubectl() {
-    if ! [ -x "$(command -v kubectl)" ]; then
-        echo "Installing kubectl..."
-        arkade get kubectl --version=v1.22.0
-        ${SUDO} mv ${HOME}/.arkade/bin/kubectl /usr/local/bin/
-    else
-        echo "kubectl is already installed"
-    fi
+    curl -LO https://dl.k8s.io/release/v1.23.0/bin/linux/amd64/kubectl
```
osherdp, line 14:
```suggestion
    curl --retry 3 -LO https://dl.k8s.io/release/v1.23.0/bin/linux/amd64/kubectl
```

scripts/install_minikube.sh
osherdp, line 7:
```suggestion
    curl --retry 3 -Lo minikube https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64
```

osherdp (changes requested):

scripts/install_minikube.sh
```diff
@@ -2,41 +2,23 @@
...
-        echo "Upgrading minikube..."
-        download_minikube $minikube_version
-        ${SUDO} mv -f ${HOME}/.arkade/bin/minikube $(command -v minikube)
-    else
-        echo "minikube is already installed and up-to-date"
-    fi
+function install_minikube() {
+    minikube_version=v1.25.2
+    curl --retry 3 -Lo minikube https://storage.googleapis.com/minikube/releases/${minikube_version}/minikube-linux-amd64
+    chmod +x minikube
```
osherdp, line 9: should be redundant, as ``install`` should set it the same way
```suggestion
```

scripts/install_minikube.sh
```diff
@@ -2,41 +2,23 @@
...
-    else
-        echo "kubectl is already installed"
-    fi
+    kubectl_version=v1.23.0
+    curl --retry 3 -LO https://dl.k8s.io/release/${kubectl_version}/bin/linux/amd64/kubectl
+    chmod +x kubectl
+
+    curl -LO https://dl.k8s.io/${kubectl_version}/bin/linux/amd64/kubectl.sha256
+    echo "$(<kubectl.sha256)  kubectl" | sha256sum --check
+    ${SUDO} mv kubectl /usr/local/bin/
```
osherdp, line 21: replace with ``install``

scripts/install_minikube.sh
```diff
@@ -2,41 +2,23 @@
...
-        arkade get kubectl --version=v1.22.0
-        ${SUDO} mv ${HOME}/.arkade/bin/kubectl /usr/local/bin/
-    else
-        echo "kubectl is already installed"
-    fi
+    kubectl_version=v1.23.0
+    curl --retry 3 -LO https://dl.k8s.io/release/${kubectl_version}/bin/linux/amd64/kubectl
+    chmod +x kubectl
+
+    curl -LO https://dl.k8s.io/${kubectl_version}/bin/linux/amd64/kubectl.sha256
```
osherdp, line 19: can you get rid of this file afterwards?

osherdp (changes requested):

osherdp (approved):

PR #1490, opened 2022-02-28
Terraform failed: Removing job_name from the vm name. Adding ansible …

michaellevy101:
…to the images

fixing "expected length of name to be in the range (1 - 80), got assisted-ci-rehearse-26612-pull-ci-openshift-assisted-test-infra-master-e2e-vsphere-assisted-1498344731409649664"

### DIFF
--- a/Dockerfile.assisted-test-infra
+++ b/Dockerfile.assisted-test-infra
@@ -21,6 +21,8 @@ RUN dnf -y install \
     && dnf config-manager --add-repo https://rpm.releases.hashicorp.com/RHEL/hashicorp.repo \
     && dnf -y install packer \
     && mv /usr/bin/packer /usr/bin/packer.io \
+    && dnf -y install epel-release \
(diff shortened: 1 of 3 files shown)


omertuc:
why ansible?

michaellevy101:
> why ansible?
Currently our CI uses bash and ansible. want to try using ansible for some tasks instead of bash. Just opening this option

omertuc:
If it's already using ansible why does it suddenly need to be installed? Why didn't it work before?

tsorya:
lets open the option when it will be really needed.

eliorerz:
>don't see any good reason for blocking this PR because of installing ansible

Regardless this PR, test-infra base image is already containing a lot more functionality than its should

michaellevy101:
> containing
right, It's a huge image need to minimize the image size by reducing layers or breaking to multiple images.
Please note that this image is internal and not use externally by customer or in production

PR #1500, opened 2022-03-01
NO-ISSUE: Bootstrap-in-place test flow should use RHCOS images fitting the OCP version tested

omertuc:
We've been hardcoding the URL for the 4.8 RHCOS, we should use an RHCOS image associated with the
OCP release being tested instead. We can get the URL for that RHCOS from
the installer

Also added retry loop to the ignition embed function, to deal with occasional
quay hiccups

### DIFF
--- a/src/assisted_test_infra/test_infra/utils/release_image_utils.py
+++ b/src/assisted_test_infra/test_infra/utils/release_image_utils.py
@@ -1,0 +1,64 @@
+import logging
+import json
+import semver
+
+from assisted_test_infra.test_infra import utils
+
+
+def extract_installer(release_image: str, dest: str):
+    """
+    Extracts the installer binary from the release image.
+
+    Args:
+        release_image: The release image to extract the installer from.
+        dest: The destination to extract the installer to.
+    """
+    logging.info("Extracting installer from %s to %s", release_image, dest)
+    with utils.pull_secret_file() as pull_secret:
+        utils.run_command(
+            f"oc adm release extract --registry-config '{pull_secret}'"
+            f" --command=openshift-install --to={dest} {release_image}"
+        )
+
+
+def extract_version(release_image):
+    """
+    Extracts the version number from the release image.
+
+    Args:
+        release_image: The release image to extract the version from.
+    """
+    logging.info(f"Extracting version number from {release_image}")
+    with utils.pull_secret_file() as pull_secret:
+        stdout, _, _ = utils.run_command(
+            f"oc adm release info --registry-config '{pull_secret}' '{release_image}' -ojson"
+        )
+
+    ocp_full_version = json.loads(stdout).get("metadata", {}).get("version", "")
+    ocp_semver = semver.VersionInfo.parse(ocp_full_version)
+    ocp_version = f"{ocp_semver.major}.{ocp_semver.minor}"
+
+    return ocp_version
+
+
+def extract_rhcos_url_from_ocp_installer(installer_binary_path: str):
+    """
+    Extracts the RHCOS download URL from the installer binary.
+
+    Args:
+        installer_binary_path: The path to the installer binary.
+    """
+    logging.info(f"Extracting RHCOS URL from {installer_binary_path}")
+    stdout, _, _ = utils.run_command(f"'{installer_binary_path}' coreos print-stream-json")
+
+    jsonpath = "architectures.x86_64.artifacts.metal.formats.iso.disk.location"
+    current_node = json.loads(stdout)
+    for element in jsonpath.split("."):
+        current_node = current_node.get(element, {})
+
+    if current_node == {}:
+        raise ValueError(f"Could not extract RHCOS URL from {installer_binary_path}, malformed JSON")
+
+    logging.info(f"Extracted RHCOS URL: {current_node}")
+
+    return current_node
--- a/src/deprecated_utils.py
+++ b/src/deprecated_utils.py
@@ -1,6 +1,3 @@
-import logging
-import os
-
 import libvirt
 import waiting
 import xml.dom.minidom as md
@@ -106,12 +103,3 @@ def are_libvirt_nodes_in_cluster_hosts(client, cluster_id, num_nodes):
         return False
     num_macs = len([mac for mac in hosts_macs if mac != ""])
     return num_macs >= num_nodes
-
-
-def extract_installer(release_image, dest):
-    logging.info("Extracting installer from %s to %s", release_image, dest)
-    with utils.pull_secret_file() as pull_secret:
-        utils.run_command(
-            f"oc adm release extract --registry-config '{pull_secret}'"
-            f" --command=openshift-install --to={dest} {release_image}"
-        )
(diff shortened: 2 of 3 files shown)


src/deprecated_utils.py
```diff
@@ -115,3 +116,19 @@ def extract_installer(release_image, dest):
             f"oc adm release extract --registry-config '{pull_secret}'"
             f" --command=openshift-install --to={dest} {release_image}"
         )
+
+
+def extract_version(release_image):
```
osherdp, line 121: Sounds weird to add it to a module named "deprecated_utils"

osherdp (changes requested):
Just one little thing that's bothering me

src/deprecated_utils.py
omertuc: It's almost identical to the `extract_installer` that's already used by the bootstrap test anyway, so it made sense to put it right next to it, even though the file is being deprecated.

Do you have any suggestions of where else it would be better to put it?

osherdp: @eliorerz can you advice?

eliorerz: We probably missed `extract_installer` when we moved bootstrap_in_place to the new workflow.
I think that both `extract_installer` and `extract_version` should be removed from `deprecated_utils.py` module.
If this utils are used only by `test_bootstrap_in_place.py` I think  that we can create a dedicated utils file (under utils package) for bootstrap_in_place, another option is placing it on `utils.py`

src/deprecated_utils.py
```diff
@@ -1,5 +1,6 @@
 import logging
-import os
+import json
+import semver
```
osherdp, line 3: seems redundant

src/tests/test_bootstrap_in_place.py
```diff
@@ -76,19 +93,22 @@ def installer_gather(self, ip: str, ssh_key: str, out_dir: str):
...
-        # TODO: enable fetching the appropriate rhcos image
         utils.run_command(
-            f"curl https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.8/4.8.2/rhcos-live.x86_64.iso"
-            f" --retry 10 --retry-connrefused -o {download_path} --continue-at -"
+            f"curl {download_image_url} --retry 10 --retry-connrefused -o {download_path} --continue-at -"
         )
 
+    # retry to avoid occassional quay hiccups
     @JunitTestCase()
+    @retry.retry(exceptions=Exception, tries=5, delay=30)
```
osherdp, line 111: what do you say about applying retries on the specific ``run_container`` action?
seems redundant (and maybe problematic) to redo other unintended failures in this method

osherdp (changes requested):

romfreiman:
@omertuc we can extract if from the release payload. There is a command for that

omertuc:
> @omertuc we can extract if from the release payload. There is a command for that

Extract the RHCOS URL?

src/tests/test_bootstrap_in_place.py
```diff
@@ -39,6 +39,23 @@
 INSTALLER_GATHER_DEBUG_STDERR = os.path.join(INSTALLER_GATHER_DIR, "gather.stderr.log")
 
 
+# TODO: Find a more elegant solution than hard-coding RHCOS URLs... Assisted
+# service maintains an up-to-date list over at:
+# https://github.com/openshift/assisted-service/blob/133ed9cf68e36590f19901d71813399bb86b165b/data/default_os_images.json#L1-L44  # noqa: E501,B950
+# maybe we can consume it somehow here? Although that would create a dependency
+# on assisted-service which we don't want for this job.
+class RHCOSImages:
+    _OPENSHIFT_RHCOS_MIRROR = "https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos"
```
romfreiman, line 48: it can be extracted from the payload

./openshift-install coreos print-stream-json | grep x86_64.iso

src/assisted_test_infra/test_infra/utils/release_image_utils.py
```diff
@@ -0,0 +1,62 @@
...
+        )
+
+    ocp_full_version = json.loads(stdout).get("metadata", {}).get("version", "")
+    ocp_semver = semver.VersionInfo.parse(ocp_full_version)
+    ocp_version = f"{ocp_semver.major}.{ocp_semver.minor}"
+
+    return ocp_version
+
+
+def extract_rhcos_url_from_ocp_installer(installer_binary_path: str):
```
romfreiman, line 44: from_release_payload? :)

src/assisted_test_infra/test_infra/utils/release_image_utils.py
```diff
@@ -0,0 +1,62 @@
...
+    stdout, _, _ = utils.run_command(f"'{installer_binary_path}' coreos print-stream-json")
+
+    jsonpath = "architectures.x86_64.artifacts.metal.formats.iso.disk.location"
+    current_node = json.loads(stdout)
+    for element in jsonpath.split("."):
+        current_node = current_node.get(element, {})
+
+    if current_node == {}:
+        raise ValueError(f"Could not extract RHCOS URL from {installer_binary_path}, malformed JSON")
+
```
romfreiman, line 61: let's print it for ease of debugging?

src/assisted_test_infra/test_infra/utils/release_image_utils.py
omertuc: Nope, it's from the installer (which we got from the release payload, but that's beside the point)

src/assisted_test_infra/test_infra/utils/release_image_utils.py
omertuc: GitHub Copilot wrote this error message from me, it'll be offended if I change it

omertuc: But in seriousness, the JSON is huge, I'm not sure printing it is a good idea

osherdp: If ``element`` is still in this scope, we can say we couldn't find ``{element}``
But no biggie, I think we can as well save it as an artifact or something under ``build/`` or just handle it if we'll have to

osherdp (approved):

omertuc:
```
Error: error creating build container: parsing image configuration: Get "https://quay.io/v2/edge-infrastructure/assisted-service/blobs/sha256:5b0b3a8628ea2dbc50fb56bafd972de8da24ad15180f7973a84d5649811e734d": net/http: TLS handshake timeout
```
sigh

https://status.quay.io/

src/assisted_test_infra/test_infra/utils/release_image_utils.py
romfreiman: I suggested only to print the final iso url, not the whole json

osherdp (approved):
don't forget to update the PR description

PR #1501, opened 2022-03-01
NO-ISSUE: Improve .flake8 rules and documentation

omertuc:
See comments in diff for more information

F403 (https://www.flake8rules.com/rules/F403.html) was removed because
we don't seem to be violate it anywhere anyway

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,5 +1,12 @@
 [flake8]
-ignore = E203, F403, W503, FS003
+ignore =
+    # E203 and W503 are rules that conflict with Black
+    E203,
+    W503,
+
+    # FS003 weirdly thinks things are supposed to be f-strings? Seems completely broken
+    FS003,
+
 max-line-length = 120
 max-complexity = 18
 select = B,C,E,F,W,T4,B9


.flake8
```diff
@@ -1,5 +1,28 @@
...
+    # Examples:
+    #   - Comments with huge URLs - can't be broken apart no matter what, unless you
+    #     want to break the URL itself.
+    #   - Strings with huge URLS, for the exact same reason.
+    # 
+    # Short lines are nice, but it's fine that once in a while some lines won't be short
+    # if it doesn't make sense. For lines where it's easy to split into multiple lines,
+    # Black will do so anyway, and we have a lint that makes sure the user ran Black -
+    # this will ensure that most lines in the project will not be too long when possible.
+    E501,
```
eliorerz, line 23: Sorry, but I don't think we should remove that.
Black can reformat most of the cases. For too long strings we can use multiline string. e.g.

```python
url = "https://github.com/openshift/assisted-test-infra/pull/1501/files#" \
      "diff-6951dbb399883798a226c1fb496fdb4183b1ab48865e75eddecf6ceb6cf46442R10-R23"
```

omertuc: You could do that, but then the URL is broken and cannot be copied.

What would you do if you want to put a long URL in a comment?

.flake8
```diff
@@ -1,5 +1,28 @@
 [flake8]
-ignore = E203, F403, W503, FS003
+ignore =
+    # E203 and W503 are rules that conflict with Black
+    E203,
+    W503,
+
+    # FS003 weirdly thinks things are supposed to be f-strings? Seems completely broken
+    FS003,
```
eliorerz, line 8: Where? This is works great on CI and locally on my machine

omertuc: Try to comment out this rule then run make flake8

.flake8
omertuc: >Black can reformat most of the cases

That's exactly my point. Things that black can format would still get flagged.

It's just cases where you have to awkwardly break a string a part that will no longer get flagged, and I think that would be more productive than striving for 100% perfect line limit

eliorerz: I don't think that it's worth removing the line length limitation. The case you described is not commonly used and for few cases (if any) I don't think we should remove it

>What would you do if you want to put a long URL in a comment?

I think we can split the line, but if you still want to leave the url as you can always add `# noqa` at the end of the line and preventing flake8 to yell about it

omertuc: Needed like 5 `noqa`s here, it's a bit ridiculous, that's why I made this PR: https://github.com/openshift/assisted-test-infra/pull/1500/files#diff-daa3406b01c75dca4c77c9e9a365c4d86020e0e65c8980981f12a6e1f63b909bR84

eliorerz: By the way, why not use the URLs from `assisted-service/data/release_os_images` instead?

omertuc: Because the job had nothing to do with assisted, the only reason it's in this repo is to abuse the existing scripts for libvirt / terraform, I don't want to create even more dependencies on assisted-service

eliorerz: I know, as long this is inside `test-infra` there will connection to the `assisted-service`.
I think that we are already cloning `assisted-service` during the test process so why not use the data files on `assisted-service`?

And if you insist about not adding that extra dependency  you can always add this urls to consts and remove the need of changing flake8 roles ignore, e.g.
```python
class RhcosImages:
    _OPENSHIFT_RHCOS_MIRROR = "https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos"

    RHCOS_4_6 = f"{_OPENSHIFT_RHCOS_MIRROR}/4.6/4.6.47/rhcos-4.6.47-x86_64-live.x86_64.iso",
    RHCOS_4_7 = f"{_OPENSHIFT_RHCOS_MIRROR}/4.7/4.7.33/rhcos-4.7.33-x86_64-live.x86_64.iso",
    RHCOS_4_8 = f"{_OPENSHIFT_RHCOS_MIRROR}/4.8/4.8.14/rhcos-4.8.14-x86_64-live.x86_64.iso",
    RHCOS_4_9 = f"{_OPENSHIFT_RHCOS_MIRROR}/4.9/4.9.0/rhcos-4.9.0-x86_64-live.x86_64.iso",
    RHCOS_4_10 = f"{_OPENSHIFT_RHCOS_MIRROR}/pre-release/4.10.0-rc.1/rhcos-4.10.0-rc.1-x86_64-live.x86_64.iso",
```

omertuc: >I know, as long this is inside test-infra there will connection to the assisted-service.
I think that we are already cloning assisted-service during the test process so why not use the data files on assisted-service?

I don't want to make a future removal of this job from this repo harder, so the less dependencies the better

>And if you insist about not adding that extra dependency you can always add this urls to consts and remove the need of changing flake8 roles ignore, e.g.

Yeah fair enough, but you're just getting lucky not hitting the 120 width limit when doing it like that, so it's kinda missing the point

Anyway removed the E501 / B950 ignoring

eliorerz: I understand your point, and you might be right. Removing this limitation (at least to my opinion) will cause more mess than order.
If we will reach a point when this will be a bigger limitation we can consider removing it.

PR #1505, opened 2022-03-03
NO-ISSUE - Fix vsphere controller not implementing all abstract methods.

eliorerz:
- Fix vsphere controller not implementing all abstract methods
- Rename vsphere controller to match the configurations naming convention

/cc @michaellevy101

### DIFF
--- a/src/tests/base_test.py
+++ b/src/tests/base_test.py
@@ -29,14 +29,14 @@ from assisted_test_infra.test_infra.controllers import (
 )
 from assisted_test_infra.test_infra.helper_classes import kube_helpers
 from assisted_test_infra.test_infra.helper_classes.cluster import Cluster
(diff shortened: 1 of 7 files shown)


src/tests/base_test.py
```diff
@@ -82,7 +82,7 @@ def new_controller_configuration(self, request: FixtureRequest) -> BaseNodeConfi
         :rtype: new node controller configuration
         """
         if global_variables.platform == consts.Platforms.VSPHERE:
-            config = VSphereControllerConfig()
+            config = BaseVSphereConfig()
```
osherdp, line 85: not sure I understand
if it's now abstract, how can we instantiate it?

osherdp (changes requested):

src/tests/base_test.py
eliorerz: Right, missed that. Fixed

osherdp (approved):

PR #1507, opened 2022-03-06
Archive must-gather directory

osherdp:
Following "KAS as a service", it became evident that it's better storing must-gather logs as a tar archive, than leaving it in a directory. This not only saves storage space, but also something that is more common among other jobs.
In addition, downloading the directory to the computer takes more time as it has to be done recursively (and GCS is rate-limitting it per-file)

/hold
/cc @osherdp


skipper.env
```diff
@@ -115,3 +115,4 @@ DEPLOY_CAPI_PROVIDER
 HOLD_INSTALLATION
 LOGGER_NAME
 AGENT_TIMEOUT_START
+ENABLE_KUBE_API
```
osherdp, line 118: seems like it is needed, otherwise the download_logs script tries to create a REST API client (and fails the complete gather step)

src/assisted_test_infra/download_logs/download_logs.py
```diff
@@ -329,16 +329,27 @@ def get_ui_url_from_api_url(api_url: str):
 
 @JunitTestCase()
 def download_must_gather(kubeconfig: str, dest_dir: str):
-    log.info(f"Downloading must-gather to {dest_dir}, kubeconfig {kubeconfig}")
+    must_gather_dir = f"{dest_dir}/must-gather-dir"
```
osherdp, line 332: also thought about using tempdir here, maybe it will look better

src/assisted_test_infra/download_logs/download_logs.py
```diff
@@ -329,16 +329,27 @@ def get_ui_url_from_api_url(api_url: str):
...
+        f" --dest-dir {must_gather_dir} > {must_gather_dir}/must-gather.log"
     )
     try:
         run_command(command, shell=True, raise_errors=True)
+
     except RuntimeError as ex:
         log.warning(f"Failed to run must gather: {ex}")
 
+    log.debug("Archiving %s...", must_gather_dir)
+    with tarfile.open(f"{dest_dir}/must-gather.tar", "w:gz") as tar:
```
omertuc, line 347: Did you make sure the tar file produced here is exactly the same as the one in regular OCP jobs?

1. Exactly same extension
2. Exactly same compression
3. Exactly same directory structure inside the tar

osherdp: yep, same thing as other jobs
there's one exception, but I think it's not significant which is regular OCP jobs also have an empty dir named ``.`` inside of the tar:
Screenshot from 2022-03-07 11-17-59

src/assisted_test_infra/download_logs/download_logs.py
```diff
@@ -260,7 +260,7 @@ def download_logs_kube_api(
 
 
 def _must_gather_kube_api(cluster_name, cluster_deployment, agent_cluster_install, output_folder):
-    kubeconfig_path = os.path.join(output_folder, "kubeconfig")
+    kubeconfig_path = os.path.join(output_folder, "kubeconfig", f"{cluster_name}_kubeconfig.yaml")
```
osherdp, line 263: Previous code led to:
```
read /home/assisted/build/kubeconfig: is a directory
```

when running ``oc adm must-gather```

eliorerz (approved):
LGTM

osherdp:
/test e2e-metal-assisted-kube-api-net-suite
(known and reported issue)

omertuc:
Live-iso still without tar https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_assisted-test-infra/1507/pull-ci-openshift-assisted-test-
infra-master-e2e-metal-single-node-live-iso/1500835058892673024

osherdp:
> Live-iso still without tar https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_assisted-test-infra/1507/pull-ci-openshift-assisted-test-infra-master-e2e-metal-single-node-live-iso/1500835058892673024

why not? it has the tar: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_assisted-test-infra/1507/pull-ci-openshift-assisted-test-infra-master-e2e-metal-single-node-live-iso/1500835058892673024/artifacts/e2e-metal-single-node-live-iso/baremetalds-sno-gather/artifacts/build/ibip/

be aware ``post-tests-must-gather`` is code from another repo and will require different handling

eliorerz (approved):
LGTM

osherdp:
ok, I'll fix ``e2e-metal-assisted-kube-api-net-suite`` in another go. either way before this fix nothing has been gathered at all
/unhold

PR #1508, opened 2022-03-07
MGMT-9177: Make OVN network type for SNO

empovit:
OpenShiftSDN network type is not supported with SNO,
so use OVNKubernetes for SNO targets.


eliorerz (approved):
LGTM

PR #1509, opened 2022-03-07
NO-ISSUE: Fix boolean var check not adhering to the default true|false in Makefile rule

mresvanis:
This PR fixes the boolean variable `ENABLE_KUBE_API` check in the `deploy_nodes_with_install` Makefile rule. 

The default value of `ENABLE_KUBE_API` is ["false"](https://github.com/openshift/assisted-test-infra/blob/master/Makefile#L98), but we were checking for ["no"](https://github.com/openshift/assisted-test-infra/blob/master/Makefile#L236).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -233,7 +233,7 @@ install_cluster:
 #########
 
 deploy_nodes_with_install: start_load_balancer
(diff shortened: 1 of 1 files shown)


eliorerz (approved):
LGTM

PR #1510, opened 2022-03-07
NO-ISSUE: add ADDITIONAL_NTP_SOURCE env var to skipper

mresvanis:
This PR adds the `ADDITIONAL_NTP_SOURCE environment variable to the variables that skipper passes on to the containers it spins up. 

> The `ADDITIONAL_NTP_SOURCE` variable is used to configure user defined [NTP sources](https://github.com/openshift/assisted-service/blob/master/swagger.yaml#L6976) in the spawned cluster.


eliorerz (approved):
LGTM

PR #1518, opened 2022-03-09
MGMT-9549:Trigger for remote deploy HW specs

paul-maidment:
It has been observed that while testing against production, there is an
issue with disk validation on workers. It has been found that setting
the disk size of a worker to 128GB resolves the issue. This trigger has
been added to ensure that this happens.

### DIFF
--- a/src/tests/global_variables/default_variables.py
+++ b/src/tests/global_variables/default_variables.py
@@ -1,4 +1,5 @@
 from contextlib import suppress
+from dataclasses import dataclass
 from typing import Any, ClassVar
 
 from assisted_test_infra.test_infra.helper_classes.config.base_config import Triggerable
@@ -8,6 +9,7 @@ from tests.global_variables.env_variables_defaults import _EnvVariables
 from triggers import Trigger, get_default_triggers
 
 
+@dataclass(frozen=True)
 class DefaultVariables(_EnvVariables, Triggerable):
     __instance: ClassVar = None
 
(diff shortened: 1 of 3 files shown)


src/triggers/default_triggers.py
```diff
@@ -8,6 +8,18 @@
 
 _default_triggers = frozendict(
     {
+        "production": Trigger(
+            condition=("service_base_url", consts.RemoteEnvironment.PRODUCTION),
+            worker_disk=128849018880
```
osherdp, line 13:
```suggestion
            worker_disk=120 * 2 ** 30  # 120 GB
```
or any other option that will make the size in use more clear

osherdp (changes requested):
great idea!

src/triggers/default_triggers.py
eliorerz: And please keep that value as const

src/tests/global_variables/default_variables.py
```diff
@@ -58,4 +60,4 @@ def get_api_client(self, offline_token=None, **kwargs) -> InventoryClient:
         if not url:
             url = utils.get_local_assisted_service_url(self.namespace, "assisted-service", self.deploy_target)
 
-        return ClientFactory.create_client(url, offline_token, **kwargs)
+        return ClientFactory.create_client(url, offline_token, **kwargs)
```
osherdp, line 63: can you add blank lines at the end of files?

osherdp (approved):

osherdp (approved):

osherdp:
@paul-maidment I guess it's no longer WIP, right?

osherdp (approved):

osherdp:
should be fixed by now

PR #1530, opened 2022-03-11
WIP: NO-ISSUE: ISO download rearranged to make it a little  more reusable.

paul-maidment:
The existing ISO download implementation does not support retry or resume. Much better to have this use curl for the underlying implementation. Also adjusted bootstrap in place to use a more abstract way of handling ISO download.


src/assisted_test_infra/test_infra/utils/utils.py
```diff
@@ -580,9 +580,9 @@ def get_assisted_controller_status(kubeconfig):
 
 
 def download_iso(image_url, image_path):
-    with requests.get(image_url, stream=True, verify=False) as image, open(image_path, "wb") as out:
-        for chunk in image.iter_content(chunk_size=1024):
-            out.write(chunk)
+    run_command(
+        f"curl --location {image_url} --retry 10 --retry-connrefused -o {image_path} --continue-at -"
+    )
```
eliorerz, line 585:
```suggestion
@retry(...)
def download_iso(image_url, image_path):
    with requests.get(image_url, stream=True, verify=False) as image, open(image_path, "wb") as out:
        for chunk in image.iter_content(chunk_size=1024):
            out.write(chunk)
    )
```

src/assisted_test_infra/test_infra/utils/utils.py
```diff
@@ -580,9 +580,9 @@ def get_assisted_controller_status(kubeconfig):
 
 
 def download_iso(image_url, image_path):
```
eliorerz, line 582: I'm not sure we need that method.
On the REST flow we are using [download_file](https://github.com/openshift/assisted-test-infra/blob/3df9de413412d2436b313a696827804ae892f756/src/assisted_test_infra/test_infra/utils/utils.py#L38) method.
Maybe we want to consider to try to use it here instead of duplicate the logic

paul-maidment: Fair point, I will look into that.

src/assisted_test_infra/test_infra/utils/utils.py
paul-maidment: I considered doing it this way but we lose "resume" functionality offered by curl.

src/tests/base_kubeapi_test.py
```diff
@@ -125,7 +125,7 @@ def download_iso_from_infra_env(cls, infra_env: InfraEnv, iso_download_path: str
         log.info("getting iso download url")
         iso_download_url = infra_env.get_iso_download_url()
         log.info("downloading iso from url=%s", iso_download_url)
-        utils.download_iso(iso_download_url, iso_download_path)
+        utils.download_file(iso_download_path, iso_download_url)
```
osherdp, line 128: how does it work?
I see ``download_file`` has a ``verify_ssl`` parameter

src/tests/test_bootstrap_in_place.py
```diff
@@ -84,9 +84,7 @@ def download_live_image(self, download_path: str, rhcos_url: str):
             return
 
         log.info("Downloading iso to %s", download_path)
-        utils.run_command(
-            f"curl --location {rhcos_url} --retry 10 --retry-connrefused -o {download_path} --continue-at -"
-        )
+        utils.download_file(download_path, rhcos_url)
```
osherdp, line 87: same question as above

osherdp (changes requested):

src/tests/test_bootstrap_in_place.py
```diff
@@ -84,9 +84,7 @@ def download_live_image(self, download_path: str, rhcos_url: str):
             return
 
         log.info("Downloading iso to %s", download_path)
-        utils.run_command(
-            f"curl --location {rhcos_url} --retry 10 --retry-connrefused -o {download_path} --continue-at -"
-        )
+        utils.download_file(download_path, rhcos_url, True)
```
osherdp, line 87:
```suggestion
        utils.download_file(download_path, rhcos_url, verify_ssl=True)
```

osherdp (changes requested):

PR #1534, opened 2022-03-14
Use pypi's assisted-service-client

celebdor:
Now that the assisted-service-client is published in the Python Package
Index, it makes sense to use that instead of copying it from the
assisted-service image.

Additionally, this should make it easier once we branch out to get the
matching assisted-service-client for specific branches rather than the
latest upstream. Though the assisted-service-client should be backwards
compatible, so it is not a big deal


celebdor:
It seems this is being used by developers to get their locally build swagger client changes. A different approach should be taken

PR #1539, opened 2022-03-14
NO-ISSUE: Increase libvirt wait for machine IP timeout from 360 to 600

omertuc:
See commit message


eliorerz (approved):
LGTM

eliorerz:
@omertuc I'm not sure that what is casing the `Timeout of 360 seconds expired waiting for Waiting for Ips` error but lets give it a try

omertuc:
> @omertuc I'm not sure that what is casing the `Timeout of 360 seconds expired waiting for Waiting for Ips` error but lets give it a try

If you look at sosreport libvirt logs you can see that the machine does have an IP, leading me to believe it's a matter of just needing more time.

This test is different that other tests because it's using a full OCP cluster as hub and installing another cluster as spoke, meaning there are a lot more VMs, so maybe things are a bit slower and we hit this timeout

omertuc:
This did not help...

PR #1558, opened 2022-03-24
MGMT-7105: VSphere CI job - supporting both upper and lower case vsph…

michaellevy101:
…ere params.

Create vsphere machines on the CI dedicated folder


skipper.env
```diff
@@ -89,12 +91,19 @@ DEBUG_SERVICE_PORT
...
 VSPHERE_NETWORK
+vsphere_network
 VSPHERE_VCENTER
+vsphere_vcenter
 VSPHERE_DATACENTER
+vsphere_datacenter
 VSPHERE_DATASTORE
+vsphere_datastore
 VSPHERE_PASSWORD
+vsphere_password
```
eliorerz, line 106: Why in lower case and why it's duplicated?

skipper.env
```diff
@@ -1,3 +1,5 @@
+BUILD_ID
+build_id
```
eliorerz, line 2: Why in lower case and why it's duplicated?

src/assisted_test_infra/test_infra/helper_classes/config/controller_config.py
```diff
@@ -10,6 +10,9 @@
 
 @dataclass
 class BaseNodeConfig(BaseConfig, ABC):
+    # CI build id
+    build_id: str = None
```
eliorerz, line 14: What is `build-id` and why do we need it for?

eliorerz (changes requested):

skipper.env
michaellevy101: The CI jobs use lowercase parameters while test-infra uses uppercase. This makes both of them work.

skipper.env
michaellevy101: The CI jobs use lowercase parameters while test-infra uses uppercase. This makes both of them work.

src/assisted_test_infra/test_infra/helper_classes/config/controller_config.py
michaellevy101: to create the vsphere machines under a folder which contains this build_id

skipper.env
eliorerz: I think that you should export then in upper case. 
@osherdp WDYT?

src/assisted_test_infra/test_infra/helper_classes/config/controller_config.py
eliorerz: If this is Vsphere only variable, please pass it to Vsphere on dedicated configuration or on some other way

skipper.env
osherdp: probably, if it's possible. there's also a chance it depends on other components
@michaellevy101 is it possible?

src/assisted_test_infra/test_infra/helper_classes/config/controller_config.py
michaellevy101: It isn't. It's the CI build-id. I use it now for vsphere but it can be used elsewhere

skipper.env
michaellevy101: yes, It's. but i think supporting they both is simpler

src/assisted_test_infra/test_infra/helper_classes/config/controller_config.py
osherdp: I think it's just confusing to duplicate the values with different letter case
Any easy action that can be made to remove confusion is a good idea

eliorerz, line 14: From what I'm seeing we don't need this variable and if we do, BaseNodeConfig is definitely not where it should be.
BuildID is specific for the CI and it's more of a global variable.

src/tests/global_variables/env_variables_defaults.py
```diff
@@ -96,6 +96,8 @@ class _EnvVariables(DataPool, ABC):
         ["HOLD_INSTALLATION"], loader=lambda x: bool(strtobool(x)), default=env_defaults.DEFAULT_HOLD_INSTALLATION
     )
 
+    build_id: EnvVar = EnvVar(["BUILD_ID"])
```
eliorerz, line 99:
```suggestion
    ci_build_id: EnvVar = EnvVar(["BUILD_ID"])
```

src/assisted_test_infra/test_infra/controllers/node_controllers/vsphere_controller.py
```diff
@@ -28,6 +28,8 @@ def prepare_nodes(self):
         config = {**self._config.get_all(), **self._entity_config.get_all(), "cluster_name": self.cluster_name}
         # The ISO file isn't available now until preparing for installation
         del config["iso_download_path"]
+        folder_name = "assisted-test-infra-ci" if self._config.build_id is not None else "assisted-test-infra"
+        self._create_folder(folder_name)
```
eliorerz, line 32: Why not to use the same mechanism as we do on `TerraformController` (`ClusterName.suffix`)?

src/assisted_test_infra/test_infra/helper_classes/config/controller_config.py
michaellevy101: do you think that the controller should have access to the global parameters? It should have access to the controller config

src/tests/global_variables/env_variables_defaults.py
michaellevy101: It could be any build-id not only ci. it's the id of the enclosing build 
anyway it's not critical to me. are you sure?

src/assisted_test_infra/test_infra/controllers/node_controllers/vsphere_controller.py
michaellevy101: To make it easier to manage, we separated the test-infra clusters from the CI clusters. Each cluster was placed inside a folder with its build-id.
The mechanism is new, since our existing virtual machines ran on a bare-metal host. Here, the cluster is outside the bare metal host.

src/assisted_test_infra/test_infra/controllers/node_controllers/vsphere_controller.py
```diff
@@ -182,30 +184,55 @@ def __get_vms(self) -> List[Dict[str, Any]]:
...
+    def __new_connection(self):
+        return SmartConnect(
+            host=self._config.vsphere_vcenter,
+            user=self._config.vsphere_username,
+            pwd=self._config.vsphere_password,
+            disableSslCertValidation=True,
+        )
+
+    @staticmethod
+    def __search_for_obj(content, vim_type, name):
```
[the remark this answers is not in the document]
michaellevy101: the type hint is missing from the third party library.

src/assisted_test_infra/test_infra/controllers/node_controllers/vsphere_controller.py
michaellevy101: It makes sense to group everything in one place since the CI (test-infra) machine is created within the same folder.

eliorerz (approved):
LGTM

PR #1578, opened 2022-04-01
MGMT-9350: Add iPXE boot flow

CrystalChun:
[MGMT-9350](https://issues.redhat.com/browse/MGMT-9350)
Adds an end-to-end test of the iPXE boot flow
to test infra.

/cc @eliorerz 
/cc @filanov


src/assisted_test_infra/test_infra/helper_classes/cluster.py
```diff
@@ -25,6 +26,7 @@
 from assisted_test_infra.test_infra.utils import logs_utils, network_utils, operators_utils
 from assisted_test_infra.test_infra.utils.waiting import wait_till_all_hosts_are_in_status
 from service_client import InventoryClient, log
+from tests.config import TerraformConfig
```
eliorerz, line 29: Do not import `tests` package into `assisted_test_infra` package, you can use the abstract base configs object instead.

eliorerz: But something is not clear to me here, why do we need to import the controller configurations in `cluster.py`? The cluster object have nodes object that managing the platform. 
Cluster supposed to not be depend on specific platform and therefore we don't need to import any controller configurations to `cluster.py`

scripts/local_ipxe_server.py
```diff
@@ -0,0 +1,21 @@
...
+args = parser.parse_args()
+
+# Make sure the server is hosting the iPXE scripts directory
+dirpath=args.dir
+os.chdir(dirpath)
+
+# Create server object
+server_object = HTTPServer(server_address=(args.ip, args.port), RequestHandlerClass=CGIHTTPRequestHandler)
+# Start the web server
+server_object.serve_forever()
```
eliorerz, line 21: Create an ipxe controller, make it run inside container (the same as we did on the proxy controller).
Scripts is more for external test preparation, e.g. create-environment, deploy-assisted-service and more

skipper.env
```diff
@@ -121,4 +121,9 @@ LOGGER_NAME
...
 ADDITIONAL_NTP_SOURCE
 ENABLE_KUBE_API
-SPOKE_NAMESPACE
\ No newline at end of file
+SPOKE_NAMESPACE
+IPXE_BOOT
+LOCAL_IPXE_SERVER_URL
+LOCAL_IPXE_SERVER_PORT
+LOCAL_IPXE_ARTIFACTS_PATH
+LOCAL_IPXE_SCRIPT_PATH
```
eliorerz, line 129: This is not needed if we are creating an ipxe controller

src/assisted_test_infra/test_infra/controllers/node_controllers/libvirt_controller.py
```diff
@@ -590,6 +590,17 @@ def set_ram_kib(self, node_name, ram_kib):
             raise Exception(f"Failed to set memory for node: {node_name}")
         log.info(f"Successfully set memory to {ram_kib} for node: {node_name}")
 
+    def set_ipxe_url(self, network_name, ipxe_url):
```
eliorerz, line 593: Add type hint annotations

src/assisted_test_infra/test_infra/helper_classes/cluster.py
```diff
@@ -17,6 +17,7 @@
 from assisted_test_infra.test_infra import BaseClusterConfig, BaseInfraEnvConfig, ClusterName, exceptions, utils
 from assisted_test_infra.test_infra.controllers.load_balancer_controller import LoadBalancerController
 from assisted_test_infra.test_infra.controllers.node_controllers import Node
+from assisted_test_infra.test_infra.controllers.node_controllers.libvirt_controller import LibvirtController
```
eliorerz, line 20: No need to import specific controller in `cluster.py`. See my next comment.

src/assisted_test_infra/test_infra/helper_classes/cluster.py
```diff
@@ -801,10 +803,46 @@ def download_image(self, iso_download_path: str = None) -> Path:
...
+    def run_local_ipxe_server(self):
+        ip = re.search('\d+\.\d+\.\d+\.\d+',self._config.local_ipxe_server_url).group(0)
+        log.info("the ip %s", ip)
+        run_ipxe_server_cmd = f'''python3 {self._config.local_ipxe_server_script_path} \
+            --dir {self._config.local_ipxe_server_artifacts_path} \
+            --ip {ip} \
+            --port {self._config.local_ipxe_server_port}'''
+        log.info("the cmd %s", run_ipxe_server_cmd)
+        utils.run_command(run_ipxe_server_cmd, run_in_background=True)
+
```
eliorerz, line 829: Also here, create an ipxe controller as I mentioned before, and put those methods there.

src/assisted_test_infra/test_infra/helper_classes/cluster.py
```diff
@@ -801,10 +803,46 @@ def download_image(self, iso_download_path: str = None) -> Path:
...
+        local_ipxe_script_file = f"{self._config.local_ipxe_server_artifacts_path}/{self._config.cluster_name}"
+
+        if not os.path.exists(local_ipxe_script_file):
+            log.info("Local iPXE server path doesn't exist, creating the folder: %s", local_ipxe_script_file)
+            utils.recreate_folder(os.path.dirname(local_ipxe_script_file), force_recreate=False)
+        utils.download_file(ipxe_download_url, local_ipxe_script_file, False)
+
+    # Runs a local python server that hosts the iPXE scripts 
+    def run_local_ipxe_server(self):
+        ip = re.search('\d+\.\d+\.\d+\.\d+',self._config.local_ipxe_server_url).group(0)
```
eliorerz, line 821:
```suggestion
    def run_local_ipxe_server(self):
        """Runs a local python server that hosts the iPXE scripts"""
        ip = re.search('\d+\.\d+\.\d+\.\d+',self._config.local_ipxe_server_url).group(0)
```

src/assisted_test_infra/test_infra/helper_classes/cluster.py
```diff
@@ -801,10 +803,46 @@ def download_image(self, iso_download_path: str = None) -> Path:
...
+        log.info("the ip %s", ip)
```
eliorerz, line 822:
```suggestion
        log.info(f"the ip {ip}")
```

src/assisted_test_infra/test_infra/helper_classes/cluster.py
```diff
@@ -801,10 +803,46 @@ def download_image(self, iso_download_path: str = None) -> Path:
...
+            infra_env = self.generate_infra_env()
+            
+            self.prepare_local_ipxe_server(infra_env.id)
+            self.run_local_ipxe_server()
+
+            local_ipxe_server_url = f"{self._config.local_ipxe_server_url}:{self._config.local_ipxe_server_port}/{self._config.cluster_name}"
+            network_name = self.nodes.get_cluster_network()
+            
+            libvirt_controller = LibvirtController(config=TerraformConfig(), entity_config=self._config)
+            libvirt_controller.set_ipxe_url(network_name=network_name, ipxe_url=local_ipxe_server_url)
```
eliorerz, line 843: Also here, create an ipxe controller as I mentioned before, and put create a method from this code.
Note that we are creating a controller only in our test preparation (`tests/base_test.py` fixtures), I think that this method needs to run as a fixture before test starts (maybe after cluster creation)

eliorerz (changes requested):

eliorerz:
Run make reformat to fix most of the `lint` issues.
You can check before submitting a PR using `make flake8`

README.md
```diff
@@ -44,6 +44,7 @@ This project deploys the OpenShift Assisted Installer in Minikube and spawns lib
   - [On-prem](#on-prem)
   - [Run operator](#run-operator)
   - [Cluster-API-provider-agent](#Cluster-API-provider-agent)
+  - [Test iPXE boot flow](#test-ipxe-boot-flow)
```
osherdp, line 47: Big like for documenting usage! :+1:

src/assisted_test_infra/test_infra/controllers/node_controllers/libvirt_controller.py
```diff
@@ -590,6 +590,17 @@ def set_ram_kib(self, node_name, ram_kib):
...
+        network = self.get_network_by_name(network_name)
+        current_xml = network.XMLDesc(0)
+        xml = minidom.parseString(current_xml.encode("utf-8"))
+        dhcpElement = xml.getElementsByTagName("dhcp")[0]
+        bootp = xml.createElement("bootp")
+        bootp.setAttribute("file", ipxe_url)
+        dhcpElement.appendChild(bootp)
+        self.destroy_network(network)
+        self.create_network(xml.toprettyxml())
```
osherdp, line 602: ``network.update(...)`` doesn't work?

src/assisted_test_infra/test_infra/helper_classes/cluster.py
```diff
@@ -801,10 +803,46 @@ def download_image(self, iso_download_path: str = None) -> Path:
             )
         return self._infra_env.download_image(iso_download_path)
 
+    # Downloads the iPXE script from assisted service to a local file
```
osherdp, line 806: add it as a docstring. e.g.:
```
def prepare_local_ipxe_server(...):
    """Downloads the iPXE script from..."""
    assisted_service_url = ...
```

src/assisted_test_infra/test_infra/helper_classes/cluster.py
```diff
@@ -801,10 +803,46 @@ def download_image(self, iso_download_path: str = None) -> Path:
...
+    def prepare_local_ipxe_server(self, infra_env_id):
+        assisted_service_url = utils.get_assisted_service_url_by_configmap("assisted-service-config", "assisted-installer")
+        remote_ipxe_path = f"/api/assisted-install/v2/infra-envs/{infra_env_id}/downloads/files?file_name=ipxe-script"
```
osherdp, line 809: we decided a while ago doing all communication with the service via ``assisted_service_api.py`` script, and using the automatically generated one
That's mainly to have a single point for abstracting-away this kind of communication, and I think this method (or maybe part of it) should reside in the "adapter"

terraform_files/baremetal/main.tf
```diff
@@ -83,7 +83,8 @@ module "masters" {
   cpu_mode          = var.master_cpu_mode
   cluster_domain    = var.cluster_domain
   vtpm2             = var.master_vtpm2
-
+  boot_devices      = var.master_boot_devices
```
osherdp, line 86: all edits have been made to the ``baremetal`` template, meaning not supporting the ``none`` (platform) and ``infra-env`` (all sorts of late-binding) templates
it's fine by me (and I think it will be easier to first concentrate on the "regular" case), just wanted to make sure we're all on the same page

src/assisted_test_infra/test_infra/utils/utils.py
```diff
@@ -272,6 +274,13 @@ def get_local_assisted_service_url(namespace, service, deploy_target):
 
         raise RuntimeError(f"The parsed url {url} to service {service} in {namespace} namespace was not reachable.]")
 
+def get_assisted_service_url_by_configmap(configmap, namespace):
```
osherdp, line 277: what's the difference from ``get_local_assisted_service_url``?

osherdp (changes requested):

terraform_files/baremetal/main.tf
```diff
@@ -116,7 +117,8 @@ module "workers" {
   cpu_mode          = var.worker_cpu_mode
   cluster_domain    = var.cluster_domain
   vtpm2             = var.worker_vtpm2
-
+  boot_devices      = var.master_boot_devices
```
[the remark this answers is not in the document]
CrystalChun: Yes! Thanks for the catch!

CrystalChun:
Still making and testing changes based on review comments, will re-request reviews once ready!

src/triggers/default_triggers.py
```diff
@@ -57,6 +57,12 @@
         "lso_operator": OlmOperatorsTrigger(condition="lso"),
         "cnv_operator": OlmOperatorsTrigger(condition="cnv"),
         "odf_operator": OlmOperatorsTrigger(condition="odf"),
+        "ipxe_boot": Trigger(
+            condition=("ipxe_boot", True),
+            download_image=False,
+            master_boot_devices=["hd", "network"],
+            worker_boot_devices=["network"],
```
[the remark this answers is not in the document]
CrystalChun: Thanks for the catch! I missed this due to [this](https://github.com/openshift/assisted-test-infra/pull/1578#discussion_r841205824)

skipper.env
CrystalChun: Removed them, thank you!

src/assisted_test_infra/test_infra/helper_classes/cluster.py
CrystalChun: Moved some of the logic to the `assisted_service_api.py` file under the `download_and_save_infra_env_file` function. How does it look?

src/assisted_test_infra/test_infra/utils/utils.py
CrystalChun: Thanks for the catch! I didn't look closely at that function. Removed the configmap function!

scripts/local_ipxe_server.py
CrystalChun: Created an ipxe controller! How does it look?

src/assisted_test_infra/test_infra/controllers/node_controllers/libvirt_controller.py
CrystalChun: I didn't realize that function existed! All the examples usually showed create and destroy. I found the function
```
    def update(self, command, section, parentIndex, xml, flags=0):
        """Update the definition of an existing network, either its live
        running state, its persistent configuration, or both. """
        ret = libvirtmod.virNetworkUpdate(self._o, command, section, parentIndex, xml, flags)
        if ret == -1:
            raise libvirtError('virNetworkUpdate() failed')
        return ret
```
Are you familiar with using it? I'm not sure what to pass in for the `command`, `section`, and `parentIndex` parameters 😅

osherdp: In fact I'm not so sure myself
Now seeing this function, if it's that undocumented and not trivial to use then we'd better off with the existing code

src/assisted_test_infra/test_infra/helper_classes/cluster.py
CrystalChun: Gotcha! This was my attempt at trying to understand and make the code "work", thank you for the explanation! I've removed this import from `cluster.py`

src/assisted_test_infra/test_infra/helper_classes/cluster.py
CrystalChun: Thanks for showing me the syntax and proper way to document!

src/assisted_test_infra/test_infra/helper_classes/cluster.py
CrystalChun: Thanks for showing me the syntax and proper way to document!

src/assisted_test_infra/test_infra/helper_classes/cluster.py
CrystalChun: Thanks for showing me the syntax!

src/assisted_test_infra/test_infra/helper_classes/cluster.py
CrystalChun: Removed the import!

CrystalChun:
@eliorerz @osherdp Ready for review!

scripts/local_ipxe_server.py
eliorerz: This is the proxy controller https://github.com/openshift/assisted-test-infra/blob/master/src/assisted_test_infra/test_infra/controllers/proxy_controller/proxy_controller.py
And this the fixture that create the controller https://github.com/openshift/assisted-test-infra/blob/master/src/tests/base_test.py#L665-#L679

src/assisted_test_infra/test_infra/controllers/ipxe_controller/ipxe_controller.py
```diff
@@ -0,0 +1,58 @@
...
+        self.port = port
+        self.dir = os.path.dirname(os.path.realpath(__file__))
+        self.ipxe_scripts_folder = f"{self.dir}/server/ipxe_scripts"
+
+    def remove(self):
+        log.info(f"Removing iPXE Server {self.name}")
+        utils.remove_running_container(container_name=self.name)
+        self._remove_ipxe_scripts_folder()
+
+    def start(self, api_client, infra_env_id, cluster_name):
```
eliorerz, line 27: Maybe `api_client` can be part of the class properties

src/assisted_test_infra/test_infra/controllers/ipxe_controller/ipxe_controller.py
```diff
@@ -0,0 +1,58 @@
...
+            f"--publish {self.port}:{self.port}",
+        ]
+        utils.run_container(container_name=self.name, image=self.name, flags=run_flags)
+
+    def build_server_image(self):
+        log.info(f"Creating Image for iPXE Server {self.name}")
+        build_flags = f"--build-arg SERVER_IP={self.ip} --build-arg SERVER_PORT={self.port}"
+        utils.run_command(f"podman {consts.PODMAN_FLAGS} build {self.dir}/server -t {self.name} {build_flags}")
+
+    def download_ipxe_script(self, api_client, infra_env_id, cluster_name):
```
eliorerz, line 47: Same here. Add type hint annotations.

eliorerz, line 47: If not used elsewhere this method can be set as protected/private

src/assisted_test_infra/test_infra/controllers/ipxe_controller/server/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM python:3
```
eliorerz, line 1: Use more slim image and specific python version.
e.g 
```Dockerfile
FROM python:3.9.12-alpine3.15
```

```
python:3                 ~ 919MB
python:3.9.12-alpine3.15 ~ 47MB
```

src/triggers/default_triggers.py
```diff
@@ -57,6 +57,12 @@
         "lso_operator": OlmOperatorsTrigger(condition="lso"),
         "cnv_operator": OlmOperatorsTrigger(condition="cnv"),
         "odf_operator": OlmOperatorsTrigger(condition="odf"),
+        "ipxe_boot": Trigger(
+            condition=("ipxe_boot", True),
+            download_image=False,
+            master_boot_devices=["hd", "network"],
+            worker_boot_devices=["hd", "network"],
+        ),
```
eliorerz, line 65: Signature will be different when https://github.com/openshift/assisted-test-infra/pull/1523 will merge
```suggestion
        "ipxe_boot": Trigger(
            condition=lambda config: config.ipxe_boot == True,
            download_image=False,
            master_boot_devices=["hd", "network"],
            worker_boot_devices=["hd", "network"],
        ),
```

src/assisted_test_infra/test_infra/controllers/ipxe_controller/server/Dockerfile
eliorerz, line 1: @osherdp Not sure if this is the right path for this file, WDYT?

eliorerz (changes requested):

src/assisted_test_infra/test_infra/controllers/ipxe_controller/server/Dockerfile
osherdp: might cause problems with rate-limiters of docker-hub
meaning we should strive to using quay.io, otherwise CI will get blocked because of its high usage of docker-hub

CrystalChun: I couldn't find `python:3.9.12-alpine3.15` in a quay repository. 

Should I tag and push the docker-hub image into my own repo and use that? The Dockerfile would look like this:
```
FROM quay.io/cchun/python:3.9.12-alpine3.15
```

Is there a better or more preferred way to do this?

osherdp: maybe there is another way to do that, but I think that will suffice 
go ahead and save it in your repo (just make sure you're toggling it public)

CrystalChun: Added it to my repo and made the change to the Dockerfile. Thanks!

src/triggers/default_triggers.py
CrystalChun: Made the change, but used `is` instead of `==`

src/assisted_test_infra/test_infra/controllers/ipxe_controller/ipxe_controller.py
CrystalChun: Modified to private, thanks!

src/assisted_test_infra/test_infra/controllers/ipxe_controller/ipxe_controller.py
CrystalChun: Good idea, made it part of the class property. Thanks!

src/assisted_test_infra/test_infra/controllers/ipxe_controller/ipxe_controller.py
```diff
@@ -0,0 +1,58 @@
...
+from assisted_test_infra.test_infra import utils
+from service_client import log
+
+
+class IPXEController:
+    def __init__(
+        self,
+        name=None,
+        port=consts.DEFAULT_IPXE_SERVER_PORT,
+        ip=consts.DEFAULT_IPXE_SERVER_IP,
```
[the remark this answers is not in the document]
CrystalChun: Added, thanks!

src/assisted_test_infra/test_infra/controllers/ipxe_controller/ipxe_controller.py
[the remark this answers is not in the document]
CrystalChun: Added, thanks!

src/assisted_test_infra/test_infra/controllers/ipxe_controller/ipxe_controller.py
CrystalChun: Added, thanks!

scripts/local_ipxe_server.py
CrystalChun: Awesome! Thank you for the pointers :D

scripts/install_environment.sh
```diff
@@ -256,6 +256,11 @@ function additional_configs() {
     sudo sed -ir 's/net.ipv6.conf.all.disable_ipv6[[:blank:]]*=[[:blank:]]*1/net.ipv6.conf.all.disable_ipv6 = 0/g' /etc/sysctl.conf
     sudo sed -i -e '/net.core.somaxconn/d' -e '$a net.core.somaxconn = 2000' /etc/sysctl.conf
     sudo sysctl --load
+
+    if [ $IPXE_BOOT = "true" ]; then
```
eliorerz, line 260: ```IPXE_BOOT: unbound variable```

CrystalChun: Fixed, thank you!

osherdp: I think it needs to be ``${IPXE_BOOT:-false}`` to handle the default value

osherdp: I'm also okay with opening that port unconditionally, so whichever of those options is totally okay

CrystalChun: I think the syntax `if [ ${IPXE_BOOT-false} = "true" ]` works too, setting `IPXE_BOOT` to false if it's not set to anything. Is it preferable to use `${IPXE_BOOT:-false}` instead?

CrystalChun:
@eliorerz @osherdp ready for re-review!

src/assisted_test_infra/test_infra/controllers/ipxe_controller/ipxe_controller.py
```diff
@@ -0,0 +1,58 @@
...
+    ):
+        self.name = name
+        self.ip = ip
+        self.port = port
+        self.dir = os.path.dirname(os.path.realpath(__file__))
+        self.ipxe_scripts_folder = f"{self.dir}/server/ipxe_scripts"
```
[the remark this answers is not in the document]
CrystalChun: Made protected, thanks!

README.md
CrystalChun: Thanks :D

src/assisted_test_infra/test_infra/helper_classes/cluster.py
CrystalChun: Created the controller!

src/assisted_test_infra/test_infra/helper_classes/cluster.py
CrystalChun: Created the controller!

scripts/install_environment.sh
osherdp: Not critical
``${VAR:-value}`` is able to cope with empty values as well, but that's not such an important use-case

scripts/install_environment.sh
```diff
@@ -256,6 +256,11 @@ function additional_configs() {
     sudo sed -ir 's/net.ipv6.conf.all.disable_ipv6[[:blank:]]*=[[:blank:]]*1/net.ipv6.conf.all.disable_ipv6 = 0/g' /etc/sysctl.conf
     sudo sed -i -e '/net.core.somaxconn/d' -e '$a net.core.somaxconn = 2000' /etc/sysctl.conf
     sudo sysctl --load
+
+    if [ ${IPXE_BOOT-false} = "true" ]; then
+        echo "Opening port 8500 for iPXE boot."
+        sudo firewall-cmd --zone=libvirt --add-port=8500
```
osherdp, line 262: In my env it says:
```
# sudo firewall-cmd --zone=libvirt --add-port=8500
Error: INVALID_PORT: bad port (most likely missing protocol), correct syntax is portid[-portid]/protocol
```

But ``8500/tcp`` works. Maybe some operating systems are handling it differently 
I guess CI doesn't catch it because ``IPXE_BOOT`` is not enabled by default

osherdp (changes requested):

scripts/install_environment.sh
CrystalChun: Thank you for catching this! Modified it to the correct syntax

scripts/install_environment.sh
CrystalChun: Modified it! Thanks!!

eliorerz (approved):
LGTM

osherdp:
so... does it work?
I'm not sure how we can make sure those libvirt hosts were using the ipxe server. The sos-reports are not able to retrieve data from back when the nodes were provisioned, because we've had to restart them in the installation process

CrystalChun:
@osherdp I think so, we can verify that the iPXE scripts were being queried in the python iPXE server. 

We can also add logging into the image service hosting the boot artifacts to show that they're being queried too. https://github.com/openshift/assisted-image-service/blob/main/internal/handlers/boot_artifacts.go WDYT?

osherdp:
Sounds good
Either way, not a blocker here, so feel free to unhold this PR

CrystalChun:
Good questions and thank you both for your reviews!! I'll unhold this now

/unhold

PR #1581, opened 2022-04-03
NO-ISSUE - Fix fetching pull-request when running `bring_assisted_service` twice

eliorerz:
Allow git fetch to update the head which corresponds to the given PULL_NUMBER. Adding the `--update-head-ok` flag disables the check.

/cc @osherdp


osherdp (approved):

osherdp:
What's the ``git checkout .`` for?

eliorerz:
> What's the `git checkout .` for?

test-infra sometimes alter files (e.g. configmap) inside assisted-service directory.
Running `git checkout .` will ensure that we discard all the changes in the local cloned assisted-service.
Now that I'm writing it, I'm not sure that it is needed here. WDYT?

osherdp (approved):

PR #1602, opened 2022-04-07
NO-ISSUE: [release-ocm-2.4]: Change catalog image to one that's actually being updated

omertuc:
See commit message


osherdp:
Don't we set INDEX_IMAGE in prow somewhere?
It's been a while since I've seen this code, but either way thought to ask

PR #1603, opened 2022-04-07
MGMT-9909: Use centos stream 9 in Dockerfile

adriengentil:
Use centos stream 9 in infra test Dockerfile, it has the benefits to
bring an updated version of dracut that stop polluting the logs when it
tries to build the initramfs from inside a container.

### DIFF
--- a/Dockerfile.assisted-test-infra
+++ b/Dockerfile.assisted-test-infra
@@ -1,23 +1,25 @@
 FROM quay.io/edge-infrastructure/assisted-service:latest AS service
 
-FROM quay.io/centos/centos:stream8
+FROM quay.io/centos/centos:stream9
 
-RUN dnf -y install \
+# CRB repo is required for libvirt-devel
+RUN dnf -y install --enablerepo=crb \
   make \
   gcc \
   unzip \
   wget \
-  curl \
+  curl-minimal \
   git \
   podman \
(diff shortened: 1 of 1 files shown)


osherdp:
Maybe just tpmv2, to make sure installed version of swtpm is a good one

Dockerfile.assisted-test-infra
```diff
@@ -1,23 +1,25 @@
...
-  curl \
+  curl-minimal \
   git \
   podman \
   httpd-tools \
   jq \
   nss_wrapper \
-  python39 \
-  python39-devel \
+  python3 \
```
osherdp, line 17: Version there is by default 3.9?

osherdp (approved):

Dockerfile.assisted-test-infra
adriengentil: yes, python38 is not there anymore

Dockerfile.assisted-test-infra
```diff
@@ -1,23 +1,25 @@
...
+  python3-devel \
   libvirt-client \
   libvirt-devel \
   libguestfs-tools \
+  libxslt \
```
adriengentil, line 22: This one is for terraform, it was pulled by other dependencies before

PR #1604, opened 2022-04-07
NO-ISSUE: lint all directories in codebase

osherdp:
Mainly a refactor. It contains two types of changes:
 * Removing the ansible-bm-install/ directory which seems unused by QE or any other external usage from what I can see.
 * Make flake8 and black pass for all code in this repository (except any file/dir that has been explicitly been ignored).


Makefile
```diff
@@ -329,14 +329,14 @@ pre-commit:
 	pre-commit run --files ./src/assisted_test_infra/test_infra/* ./src/tests/*
 
 _reformat:
-	black $(LINT_CODE_STYLING_DIRS) --line-length=120
-	isort $(LINT_CODE_STYLING_DIRS) --profile=black --line-length=120
+	black . --line-length=120
```
adriengentil, line 332: should the lint include these new targets and fail if changes are detected?

eliorerz (approved):
LGTM

Makefile
osherdp: That's a good question
Checking

osherdp: OK, so it seems like files in ``scripts/`` dir are not being checked by ``flake8``
So doing ``make reformat`` will fix all lint errors (which is a required behavior), but pep8/flake8/black are not enforced for ``scripts/`` dir (which is a desired behavior)
I'll unhold this PR and check how to include this dir as well (maybe ``flake8`` only considers packages defined in ``pyproject.toml``? I still unsure about the behavior here)

PR #1442, opened 2022-02-06
Refactor makefile targets

osherdp:
Removing most of the redundant makefile targets, and updating documentation to reflect current usage of assisted-test-infra.

PR #1606, opened 2022-04-08
Handle forward-compatibility with 'setup' target

osherdp:
Back in https://github.com/openshift/assisted-test-infra/pull/1442, we've had changed our makefile targets to remove some of
the redundancy. Most important one to handle is the ``create_full_environment`` -> ``setup`` change, because ``create_full_environment`` is heavily referenced in CI.

This PR introduce ``setup`` target so that we can use it in cross-branches places, e.g. assisted-baremetal-setup step.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,7 +155,7 @@ endif
 # General #
 ###########
 
-all: create_full_environment run_full_flow_with_install
+all: setup run_full_flow_with_install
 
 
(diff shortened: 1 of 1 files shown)


osherdp:
/retest
Flakes. We should figure out how to do image replacement inside the packet server

eliorerz (approved):
LGTM

eliorerz:
FYI @lalon4 @nshidlin

PR #1608, opened 2022-04-10
Support passing vips IPs as env variables

michaellevy101:
Allow configuring the vips from outside by env variables.
For example having static reserved vips in the vsphere CI env 
(Using DHCP allocation requires downgrading the vsphere security so the only way is using static ip as Openshift CI does)


eliorerz:
Please add description why do we need it for

src/assisted_test_infra/test_infra/helper_classes/cluster.py
```diff
@@ -317,8 +317,8 @@ def set_network_params(self, controller=None):
         else:
             log.info("Assigning VIPs statically")
             access_vips = controller.get_ingress_and_api_vips()
-            api_vip = access_vips["api_vip"]
-            ingress_vip = access_vips["ingress_vip"]
+            api_vip = access_vips["api_vip"] if access_vips else None
+            ingress_vip = access_vips["ingress_vip"] if access_vips else None
```
eliorerz, line 321: Why getting api vip and ingress vip as env variables and not using them?

eliorerz: You might want to return those value from `get_ingress_and_api_vips` (Or that I'm missing something)?

src/tests/global_variables/env_variables_defaults.py
```diff
@@ -97,6 +97,9 @@ class _EnvVariables(DataPool, ABC):
         ["HOLD_INSTALLATION"], loader=lambda x: bool(strtobool(x)), default=env_defaults.DEFAULT_HOLD_INSTALLATION
     )
 
+    api_vip: EnvVar = EnvVar(["API_VIP"], default=None)
+    ingress_vip: EnvVar = EnvVar(["INGRESS_VIP"], default=None)
```
eliorerz, line 101: `default` is None by default

eliorerz (changes requested):

src/assisted_test_infra/test_infra/helper_classes/cluster.py
michaellevy101: Underlying platform is managed by the controller. While the controller can provide the VIP's IPs, if they are provided as env variables that are unrelated to the underline platform, the controller should not return them.
Instead of duplicating everything at the controller level, we should try to implement it in the most generic way possible. 
We have 4 major cases here:
1. DHCP
2. Static 
3. Taking free IP
4. Custom - by controller
All controllers should share the first three cases

michaellevy101: the logic here is: If the controller provides vips then use them else take it from the config (env variable)

michaellevy101:
> Please add description why do we need it for

isn't a

src/assisted_test_infra/test_infra/helper_classes/cluster.py
michaellevy101: I used env variables because sometimes the VIPs allocation has different logic between Envs like CI and Local and cannot be determine by the controller especially for real virtualization solutions

eliorerz: > I used env variables because sometimes the VIPs allocation has different logic between Envs like CI and Local and cannot be determine by the controller especially for real virtualization solutions

But you are not using those in your code. You added api and ingress vips to the variables defaults but they are not being used elsewhere in the code

src/tests/global_variables/env_variables_defaults.py
```diff
@@ -97,6 +97,9 @@ class _EnvVariables(DataPool, ABC):
         ["HOLD_INSTALLATION"], loader=lambda x: bool(strtobool(x)), default=env_defaults.DEFAULT_HOLD_INSTALLATION
     )
 
+    api_vip: EnvVar = EnvVar(["API_VIP"])
+    ingress_vip: EnvVar = EnvVar(["INGRESS_VIP"])
```
eliorerz, line 101: Add `API_VIP` and `INGRESS_VIP` to `skipper.env`

src/assisted_test_infra/test_infra/helper_classes/cluster.py
michaellevy101: please see:
https://github.com/openshift/assisted-test-infra/blob/8fdf16a0b255f2104de00f254f9bbb4ed8fbaaea/src/assisted_test_infra/test_infra/helper_classes/cluster.py#L406

eliorerz (approved):
LGTM

PR #1610, opened 2022-04-12
Gather release-image.service log

osherdp:
In addition to bootkube, kubelet and crio, the release-image service is the first one to start by pulling the relevant OCP image.
This is important to gather especially on IPv6 tests (where IPv6 connectivity depends on proxy or an internal registry and might not work properly).

/cc @eliorerz @eranco74 
/hold


eranco74 (commented):
/lgtm

PR #1611, opened 2022-04-12
MGMT-9840 - assisted-test-infra doesn't rebase assisted-service in PRs

eliorerz:
We have the following scenario:

- A commit is pushed to master that updates the 'deploy/' directory (e.g., RBAC) as well as updates code that relies on those changes.
- A PR is submitted that isn't rebased on the above change

The PR won't pass tests because:

- When building the service image, the repo is rebased and therefore includes the code changes
- Assisted-test-infra doesn't rebase and therefore the 'deploy/' changes won't be applied
- The test is using rebased code without rebased deploy, and therefore the service will fail

/cc @osherdp

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,7 @@ OPENSHIFT_CI := $(or ${OPENSHIFT_CI}, "false")
 JOB_TYPE := $(or ${JOB_TYPE}, "")
 REPO_NAME := $(or ${REPO_NAME}, "")
 PULL_NUMBER := $(or ${PULL_NUMBER}, "")
+PULL_BASE_REF := $(or ${PULL_BASE_REF}, "")
 
 # lint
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -285,6 +285,7 @@ ifeq ($(shell [[ $(OPENSHIFT_CI) == "true" && $(REPO_NAME) == "assisted-service"
 	@cd assisted-service && \
 	git fetch origin pull/$(PULL_NUMBER)/head:assisted-service-pr-$(PULL_NUMBER) && \
 	git checkout assisted-service-pr-$(PULL_NUMBER)
+	git rebase origin/$(PULL_BASE_REF)
```
osherdp, line 288: You'll want ``PULL_BASE_REF := $(or ${PULL_BASE_REF}, "")`` at the top of the makefile (like with ``PULL_NUMBER`` and others)
In addition, adding it here: https://github.com/openshift/release/blob/9eb742d4427e133e7c62a8170efb6acf0e64ace1/ci-operator/step-registry/assisted/baremetal/setup/assisted-baremetal-setup-commands.sh#L103

osherdp (approved):

PR #1612, opened 2022-04-13
NO-ISSUE - Add WORKERS_COUNT and MASTERS_COUNT to skipper.env

eliorerz:
/cc @osherdp


osherdp:
It should replace NUM_MASTERS and NUM_WORKERS?
If so, why we those new vars are not documented?

osherdp (approved):

PR #1613, opened 2022-04-13
MGMT-9849: convert test_kube_api to multi-node

slaviered:


src/tests/test_kube_api.py
```diff
@@ -84,7 +84,6 @@ def kube_api_test(
 
         # TODO resolve it from the service if the node controller doesn't have this information
         #  (please see cluster.get_primary_machine_cidr())
```
osherdp, line 86:
```suggestion
```

src/tests/test_kube_api.py
```diff
@@ -127,27 +126,21 @@ def kube_api_test(
         infra_env.create(
             cluster_deployment, secret, proxy, ignition_config_override, ssh_pub_key=cluster_config.ssh_public_key
         )
+
         agents = self.start_nodes(nodes, infra_env, cluster_config)
 
-        log.info("Waiting for agent status verification")
-        Agent.wait_for_agents_to_install(agents)
+        if len(nodes) == 1:
+            self._set_agent_cluster_install_machine_cidr(agent_cluster_install, nodes)
```
osherdp, line 133: In order to simulate Git-Ops as good as we can, we should not patch k8s objects but rather have all objects set only once
Can we do that throughout the whole test? (also applies to ``set_api_vip`` and ``set_ingress_vip`` at the bottom)
Otherwise we might evade some pretty problematic bugs

osherdp (changes requested):

src/tests/test_kube_api.py
slaviered: @ori-amizur can we change the flow of this test so the VIP will be calculated before creating the aci? Does it requires staring the nodes?

ori-amizur: No. You can set the VIPs in the ACI spec

eliorerz:
```2022-04-13 13:55:13,282 - root - DEBUG - 139691424477440 - DefaultVariables - Trigger set `workers_count` to `0`, Condition: ['lambda config: config.masters_count == 1']
```
Test is still setting MASTERS_COUNT to 1, probably because of https://github.com/openshift/release/blob/d3486803d4390caf967e7430a1f40b2a51760f54/ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml#L390.
Please add a test for SNO and change the current kube-api test to run in multi-node

eliorerz (changes requested):
/hold until making sure that the kube-api test is running in multi-node mode

src/tests/test_kube_api.py
```diff
@@ -127,27 +126,29 @@ def kube_api_test(
...
+        else:
+            # for multi node allocate 2 address at a safe distance from the beginning
+            # of the available address block to allow enough addresses for workers
+            access_vips = nodes.controller.get_ingress_and_api_vips()
+            api_vip = access_vips["api_vip"]
+            ingress_vip = access_vips["ingress_vip"]
+            # patch the aci with the vips. The cidr will be derived from the range
+            agent_cluster_install.set_api_vip(api_vip)
+            agent_cluster_install.set_ingress_vip(ingress_vip)
+            nodes.controller.set_dns(api_vip=api_vip, ingress_vip=ingress_vip)
```
eliorerz, line 148:
```suggestion
        if len(nodes) == 1:
            # for single node set the cidr and take the actual ip from the host
            # the vips is the ip of the host
            self._set_agent_cluster_install_machine_cidr(agent_cluster_install, nodes)
            self.set_single_node_ip(cluster_deployment, nodes)  # patch the aci with the cidr. no need to specify vips
            api_vip = ingress_vip = get_ip_for_single_node(cluster_deployment, nodes.is_ipv4)
        else:
            # for multi node allocate 2 address at a safe distance from the beginning
            # of the available address block to allow enough addresses for workers
            access_vips = nodes.controller.get_ingress_and_api_vips()
            api_vip = access_vips["api_vip"]
            ingress_vip = access_vips["ingress_vip"]
            # patch the aci with the vips. The cidr will be derived from the range
            agent_cluster_install.set_api_vip(api_vip)
            agent_cluster_install.set_ingress_vip(ingress_vip)

        nodes.controller.set_dns(api_vip=api_vip, ingress_vip=ingress_vip)
```

src/tests/base_kubeapi_test.py
```diff
@@ -150,12 +151,35 @@ def start_nodes(cls, nodes: Nodes, infra_env: InfraEnv, entity_config: BaseEntit
...
+        node_list = nodes.controller.list_nodes()
         for agent in agents:
             agent.approve()
-            cls.set_agent_hostname(nodes[0], agent, entity_config.is_ipv4)  # Currently only supports single node
+            cls._set_host_name_from_node(node_list, agent, entity_config.is_ipv4)
 
         return agents
 
+    @classmethod
+    def _set_host_name_from_node(cls, nodes: List[Node], agent: Agent, is_ipv4: bool) -> None:
```
eliorerz, line 162: Maybe need some documentation here

eliorerz (changes requested):

src/tests/base_kubeapi_test.py
```diff
@@ -150,12 +151,42 @@ def start_nodes(cls, nodes: Nodes, infra_env: InfraEnv, entity_config: BaseEntit
...
+    """
+    Use the MAC address that is listed in the virt node object to find the interface entry
+    in the host's inventory and take the host name from there
+    The setting of the hostname is required for IPv6 only, because the nodes are booted with
+    hostname equal to localhost which is neither unique not legal name for AI host
+    """
+
+    @classmethod
+    def _set_host_name_from_node(cls, nodes: List[Node], agent: Agent, is_ipv4: bool) -> None:
+        if is_ipv4:
```
eliorerz, line 170:
```suggestion
    @classmethod
    def _set_host_name_from_node(cls, nodes: List[Node], agent: Agent, is_ipv4: bool) -> None:
        """
    Use the MAC address that is listed in the virt node object to find the interface entry
    in the host's inventory and take the host name from there
    The setting of the hostname is required for IPv6 only, because the nodes are booted with
    hostname equal to localhost which is neither unique not legal name for AI host
    """

        if is_ipv4:
```

eliorerz (changes requested):

eliorerz (approved):
LGTM

PR #1616, opened 2022-04-20
Fix errors when test-infra tries to access data/

osherdp:
For some targets, we see those kind of errors:
```
make[1]: Entering directory
'/mnt/nvme/assisted-test-infra/assisted-service'
/bin/sh: /mnt/nvme/assisted-test-infra/data/default_must_gather_versions.json: No such file or directory
cat: /mnt/nvme/assisted-test-infra/data/default_hw_requirements.json: No such file or directory
/bin/sh: /mnt/nvme/assisted-test-infra/data/default_release_images.json: No such file or directory
/bin/sh: /mnt/nvme/assisted-test-infra/data/default_os_images.json: No such file or directory
```

It turns out that assisted-service uses ``MAKEFILE_LIST`` in order to determine the relevant Makefile root-dir, taking into account cases like ``make -f <file>`` and ``include`` statements in makefiles. Outcome of that actually makes it take the path of assisted-test-infra instead.

This change will set root-dir from outside, as test-infra is totally aware of the location of the assisted-service root-dir.

/cc @osherdp 
/hold

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -150,7 +150,7 @@ delete_minikube:
 ####################
 
 destroy_onprem:
-	make -C assisted-service/ clean-onprem || true
+	ROOT_DIR=$(realpath assisted-service/) make -C assisted-service/ clean-onprem || true
 
 ####################
 # Load balancer    #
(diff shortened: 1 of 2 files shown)


eliorerz (approved):
LGTM

PR #1617, opened 2022-04-20
MGMT-9840 - Move bring_assisted_service Makefile target to bash script

eliorerz:
On assisted-service, when opening a new PR, on some cases (e.g. change on  the `deploy/` directory) the test is using rebased code without rebased all the service code, and therefore the service will fail.
This PR moves the `bring_assisted_service` target code from Makefile to a new bsah script and rebase assisted-service with `PULL_BASE_REF` after checking out the PR branch

### DIFF
--- a/scripts/bring_assisted_service.sh
+++ b/scripts/bring_assisted_service.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+set -o xtrace
+
+export SERVICE_REPO=${SERVICE_REPO:-https://github.com/openshift/assisted-service}
+export SERVICE_BRANCH=${SERVICE_BRANCH:-master}
+export SERVICE_BASE_REF=${SERVICE_BASE_REF:-master}
+export OPENSHIFT_CI=${OPENSHIFT_CI:-false}
+export REPO_NAME=${REPO_NAME:-assisted-service}
+export JOB_TYPE=${JOB_TYPE:-}
+export PULL_BASE_REF=${PULL_BASE_REF:-master}
+
+if ! [[ -d "assisted-service" ]]; then
+  echo "Can't find assisted-service source locally, cloning ${SERVICE_REPO}"
+  git clone "${SERVICE_REPO}"
+fi
+
+service_active_branch=$(cd assisted-service/ && git rev-parse --abbrev-ref HEAD)
+pr_branch_name=assisted-service-pr-${PULL_NUMBER}
+
+if [[ "${OPENSHIFT_CI}" == "true" && "${REPO_NAME}" == "assisted-service" && "${JOB_TYPE}" == "presubmit" ]]; then
+  if [[ "${service_active_branch}" == "${pr_branch_name}" ]]; then
+    # Assisted-Service source code is already updated and rebased with PULL_BASE_REF.
+    # git fetch cannot be called twice after rebase so if the PR branch already exist the assumption is that it was
+    # already fetched and rebased on another call of this target
+    echo "Nothing to update. ${REPO_NAME} is already updated, branch: ${pr_branch_name}"
+    exit 0
+  fi
+
+  echo "Running in assisted-service pull request"
+  cd assisted-service
+  git fetch -v origin "pull/${PULL_NUMBER}/head:${pr_branch_name}"
+  git checkout "${pr_branch_name}"
+  git rebase -v "origin/${PULL_BASE_REF}"
+else
+  cd assisted-service
+  git fetch --force origin "${SERVICE_BASE_REF}:FETCH_BASE" "${SERVICE_BRANCH}"
+  git reset --hard FETCH_HEAD
+  git rebase FETCH_BASE
+fi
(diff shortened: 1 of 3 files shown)


scripts/bring_assisted_service.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
+export SERVICE_REPO=${SERVICE_REPO:-}
+export OPENSHIFT_CI=${OPENSHIFT_CI:-flase}
```
osherdp, line 4:
```suggestion
export OPENSHIFT_CI=${OPENSHIFT_CI:-false}
```

scripts/bring_assisted_service.sh
```diff
@@ -0,0 +1,35 @@
...
+export REPO_NAME=${REPO_NAME:-assisted-service}
+export JOB_TYPE=${JOB_TYPE:-}
+export PULL_BASE_REF=${PULL_BASE_REF:-master}
+
```
osherdp, line 8: delete one blank line

scripts/bring_assisted_service.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
+export SERVICE_REPO=${SERVICE_REPO:-}
```
osherdp, line 3: this one also deserves a default, no?

scripts/bring_assisted_service.sh
```diff
@@ -0,0 +1,35 @@
...
+if [[ "${OPENSHIFT_CI}" = "true" && "${REPO_NAME}" = "assisted-service" && "${JOB_TYPE}" = "presubmit" ]]; then
+  if [[ "${service_active_branch}" = "${pr_branch_name}" ]]; then
+    echo "Nothing to update. ${REPO_NAME} is already updated, branch: ${pr_branch_name}"
+    exit 0
+  fi
+
+  echo "Running in assisted-service pull request"
+	cd assisted-service && \
+	git fetch -v origin "pull/${PULL_NUMBER}/head:${pr_branch_name}" && \
+	git checkout "${pr_branch_name}" && \
```
osherdp, line 28: some weird indentation here

osherdp: and instead of ``&&`` we can make each line fail inherently with ``set``

osherdp (changes requested):

osherdp:
Can you set some description to the PR?

scripts/bring_assisted_service.sh
```diff
@@ -0,0 +1,35 @@
...
+  echo "Can't find assisted-service source locally, cloning ${SERVICE_REPO}"
+  git clone "${SERVICE_REPO}"
+fi
+
+service_active_branch=$(cd assisted-service/ && git rev-parse --abbrev-ref HEAD)
+pr_branch_name=assisted-service-pr-${PULL_NUMBER}
+
+if [[ "${OPENSHIFT_CI}" == "true" && "${REPO_NAME}" == "assisted-service" && "${JOB_TYPE}" == "presubmit" ]]; then
+  if [[ "${service_active_branch}" == "${pr_branch_name}" ]]; then
+    echo "Nothing to update. ${REPO_NAME} is already updated, branch: ${pr_branch_name}"
```
eranco74, line 21: Why do we know there is no need to rebase here?

eliorerz: It fails when we call it twice (on different places on the code). If the branch is already exists it means that we already fetched, checked-out and rebsaed the code

eranco74: I guess it fails on the `checkout` command, you can mitigate it by adding `-B` to the command (it will reset the branch in case it already exists).

If you decide to keep it this way please add a comment on why we do this check and that what it means

eliorerz: It's failing when trying to run the command 
```bash
git fetch -v origin pull/3678/head:assisted-service-pr-3678
--
fatal: Refusing to fetch into current branch refs/heads/assisted-service-pr-3678 of non-bare repository
```

eliorerz: After the rebase it fails even with this flag.
With the flag there is another failure:

```
From https://github.com/openshift/assisted-service
 ! [rejected]          refs/pull/3678/head -> assisted-service-pr-3678  (non-fast-forward)
```

eliorerz: And from the documentation the recommendation is not to use the `--update-head-ok` flag

```
--update-head-ok
By default git fetch refuses to update the head which corresponds to the current branch. This flag disables the check. This is purely for the internal use for git pull to communicate with git fetch, and unless you are implementing your own Porcelain you are not supposed to use it.
```

scripts/bring_assisted_service.sh
eliorerz: Done

scripts/bring_assisted_service.sh
eliorerz: I documented it

osherdp (approved):

PR #1618, opened 2022-04-21
NO-ISSUE: Remove cluster ID suffix from cluster_event junit tests

vrutkovs:
This ensures "cluster_events" test name is static, so that test name would be persistent across runs.

### DIFF
--- a/src/service_client/assisted_service_api.py
+++ b/src/service_client/assisted_service_api.py
@@ -36,7 +36,9 @@ class InventoryClient(object):
         self.operators = api.OperatorsApi(api_client=self.api)
         self.manifest = api.ManifestsApi(api_client=self.api)
 
-        fmt = CaseFormatKeys(case_name="cluster_id", severity_key="severity", case_timestamp="event_time")
+        fmt = CaseFormatKeys(
+            case_name="cluster-event-test", static_case_name=True, severity_key="severity", case_timestamp="event_time"
+        )
         self._events_junit_exporter = JsonJunitExporter(fmt)
 
     def get_host(self, configs: Configuration) -> str:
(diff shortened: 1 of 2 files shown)


src/service_client/assisted_service_api.py
```diff
@@ -328,7 +328,7 @@ def download_cluster_events(self, cluster_id: str, output_file: str, categories=
         with open(output_file, "wb") as _file:
             events = self.get_events(cluster_id, categories=categories)
             _file.write(json.dumps(events, indent=4).encode())
-            self._events_junit_exporter.collect(events, suite_name="cluster_events", xml_suffix=cluster_id)
```
eliorerz, line 331: The `xml_suffix` is just setting the xml file name and not the test suite/case name

eliorerz: When we have multiple clusters on single test we don't want to override the report

eliorerz: To fix that issue we might need to change the format https://github.com/openshift/assisted-test-infra/blob/d005d3271ad2aa38ae3ea71f08a82fd8deec0886/src/service_client/assisted_service_api.py#L39

eliorerz:
I updated the option to set static test case name on https://github.com/eliorerz/junit-report/blob/591ac478ee531ceea9d87db220d4784ea4e8deeb/src/junit_report/json_junit_exporter.py#L19
e.g. 
```python
        fmt = CaseFormatKeys(case_name="clsuter-event-test", severity_key="severity", case_timestamp="event_time",  static_case_name=True)
```
Just update junit-report to `junit-report-0.2.6 `

vrutkovs:
/test e2e-metal-assisted-none

/hold cancel

Test name is static now

eliorerz (approved):
LGTM

eliorerz (approved):

PR #1620, opened 2022-04-21
Adding a cluster method to set encryption params

lalon4:
To support a scenario where disk encryption turned on while not set on the node's BIOS, and for other cases, added a method to set  disk_encryption params
Corresponding e2e test: https://gitlab.cee.redhat.com/ocp-edge-qe/kni-assisted-installer-auto/-/merge_requests/763/diffs


eranco74 (commented):
/lgtm

PR #1630, opened 2022-04-26
Remove create_full_environment target

osherdp:
After https://github.com/openshift/release/pull/27909 we should no longer have the old target.
/cc @eliorerz


eliorerz:
It seems that `kube-api-net-suite` is still using `create_full_environment`

osherdp:
> It seems that `kube-api-net-suite` is still using `create_full_environment`

right
I'll first change those occurrences

eliorerz (approved):

PR #1632, opened 2022-04-26
NO-ISSUE - Assert if no PULL_SECRET provided

eliorerz:
Fail the test if missing pull secret 

@osherdp


osherdp (changes requested):
Code is good, but you have a type in the commit message "secrete"

osherdp (approved):

PR #1645, opened 2022-05-03
MGMT-10223: capi test should update nodepool replicas instead of nodeCount

eranco74:
This is required foloowing this change in hypershift:
https://github.com/openshift/hypershift/pull/1205


eranco74:
The test is failing on post installation conformance tests, such failures are tracked in the  it has nothing to do with this PR

PR #1646, opened 2022-05-05
MGMT-9179 - Single node bootstrap-in-place flow should use OVNKubernetes

omertuc:
In recent versions of OCP `OVNKubernetes` replaced `OpenShiftSDN` as the
default network type for Single Node clusters (currently just in
Assisted Installer's default configuration, in the future it would be
the default for the regular installer / OCP documentation). Later SNO
OpenShiftSDN would be completely forbidden.

We want our BIP flow in this repo to also use OVNKubernetes so we'll
be testing the default network type customers are likely to use in the
future and not the older default OpenShiftSDN.


eranco74 (commented):
/lgtm

PR #1647, opened 2022-05-08
MGMT-9682: fix disk sizes used in various places

osherdp:
It seems like lots of places are using GiB (Gibibyte == 2 ** 30) whereas requirements in OCP documentation are mentioned in GB (Gigabyte == 10 ** 9) quantities.

This only seems to be applied to the usage rather than the restrictions (that use the proper quantities).

Connected to [MGMT-9682](https://issues.redhat.com//browse/MGMT-9682) effort.


src/consts/consts.py
```diff
@@ -144,7 +144,7 @@ def all(cls):
 )
 
 # DISK SIZES
-DISK_SIZE_120GB = 120 * 2**30
+DISK_SIZE_120GB = 120 * 10**9
```
eliorerz, line 147: Maybe declare a 1GB unit const and use it across the project?

osherdp: sure, done

osherdp:
/test e2e-metal-assisted-cnv e2e-metal-assisted-odf e2e-metal-assisted-cnv
known to cause some troubles, but trying out anyway

osherdp:
Actually if it passed preflight validations of the assisted-service then this change should be okay
/unhold

PR #1661, opened 2022-05-17
[release-ocm-2.4] revert libvirt downgrade on Rocky Linux

osherdp:
On Rocky (and some RHEL releases) we used to need to install a downgraded libvirt version, because of a malformed version which is the latest available in the relevant registries. Seems like they've updated the registries, so we no longer need to do that.

Backports #1660


osherdp:
/unhold
late-binding job failure is a known issue

PR #1662, opened 2022-05-17
Fix swtpm installation to always happen

osherdp:
Back in openshift/assisted-test-infra#1654 we changed behavior to install swtpm packages only when disk-encryption is defined. This change reverts this change as swtpm packages are now residing at Rocky Linux
registries.

Having the packages in Rocky Linux registries also mean we no longer need EPEL, and it's really simplifying our code and dependencies.

/cc @osherdp 
/hold


osherdp:
/cc @adriengentil @eliorerz @nmagnezi 
Just so you know, the swtpm version in Rocky Linux registries is 0.7.0 meaning we still have stefanberger/swtpm/issues/644
To get rid of the flakiness we'll need to find out how 0.7.1 can be added to the Rocky Linux registries

adriengentil:
> /cc @adriengentil @eliorerz @nmagnezi Just so you know, the swtpm version in Rocky Linux registries is 0.7.0 meaning we still have [stefanberger/swtpm/issues/644](https://github.com/stefanberger/swtpm/issues/644) To get rid of the flakiness we'll need to find out how 0.7.1 can be added to the Rocky Linux registries

the version of swtpm we had in the epel was the 0.7.1?

osherdp:
> > /cc @adriengentil @eliorerz @nmagnezi Just so you know, the swtpm version in Rocky Linux registries is 0.7.0 meaning we still have [stefanberger/swtpm/issues/644](https://github.com/stefanberger/swtpm/issues/644) To get rid of the flakiness we'll need to find out how 0.7.1 can be added to the Rocky Linux registries
> 
> the version of swtpm we had in the epel was the 0.7.1?

No, it was 0.5.3 iirc
Which also had the fix backported

PR #1663, opened 2022-05-18
Always pulling referenced service image

osherdp:
Seems like since openshift/assisted-test-infra#1595 we no longer trying to pull the assisted-service base image in assisted-test-infra Dockerfile. This is true anytime someone does ``make image_build``
without doing ``make setup`` before.

To make it more convenient, I moved this part to the former target which either way seems like the right place for it.


osherdp:
/cc @eliorerz @lalon4  
can you review?

lalon4:
/lgtm
thanks!

PR #1678, opened 2022-06-08
NO-ISSUE: Fixing capi test by adding label worker to minikube

tsorya:
NO-ISSUE: Fixing capi test by adding label worker to minikube
In 4.11 ovn is a default network for hypershift cluster and it requires
worker label to run
NO-ISSUE: small improvements to case deploy scripts to cleanup
hypershift image locally before pulling

### DIFF
--- a/scripts/setup_capi_env.sh
+++ b/scripts/setup_capi_env.sh
@@ -19,11 +19,16 @@ function waitForPodsReadyStatus(){
 deploy_hypershift() {
   kubectl apply -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.51.1/bundle.yaml || true
   HYPERSHIFT_BIN_DIR=$BASE_DIR/hypershift/bin
+
+  if [[ "$(${CONTAINER_COMMAND} images -q $HYPERSHIFT_IMAGE 2> /dev/null)" != "" ]]; then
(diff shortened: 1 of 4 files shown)


scripts/setup_capi_env.sh
```diff
@@ -36,3 +41,5 @@ echo "Deploying HyperShift"
 deploy_hypershift
 echo "Alow route to minikube network - required for hosts to pull ignition"
 iptables -D LIBVIRT_FWI -o virbr1 -j REJECT --reject-with icmp-port-unreachable || true
+echo "Set worker label for minikube"
+kubectl get nodes minikube && kubectl label node minikube node-role.kubernetes.io/worker="" --overwrit
```
osherdp, line 45: Shouldn't it be "overwrite"?

osherdp (changes requested):

scripts/setup_capi_env.sh
tsorya: oops, copy pasted

osherdp:
/hold
Because otherwise nothing will wait for your optional test to finish

tsorya:
now there is another issue, container is stuck in creating

eranco74 (commented):
/lgtm

PR #1679, opened 2022-06-09
MGMT-10775: Trying some stuff

mkowalski:
/hold

For a sample libvirt network the configuration looks as follows

```
  <ip address='192.168.122.1' netmask='255.255.255.0'>
    <dhcp>
      <range start='192.168.122.100' end='192.168.122.254'/>
    </dhcp>
  </ip>
```

I want to see if this is prone to collide when we want to manually reserve 2 IPs in the block, but at the same time there are VM resources created which also consume IP addresses from that pool.


mkowalski:
The cluster payload after the job succeeded looks

```
"api_vip": "192.168.127.50",
"ingress_vip": "192.168.127.51",
```

so seems like this is a potential workaround for the libvirt defaults. A discussion is needed whether we want to carry it or no.

mkowalski:
/close

We'd better used the XML trick to modify the DHCP range in the libvirt's network

PR #1682, opened 2022-06-13
NO-ISSUE: added 20 minutes WAIT_FOR_HYPERSHIFT_CONTROL_PLANE_TIMEOUT

eranco74:
This is required since recently hypershift is installing ocp 4.11 which
uses OVNKubernetes, this seems to cause a delay when starting the
control plane pods


osherdp (approved):

gamli75 (approved):

eranco74:
> Don't forget to backport to 2.5

2.5 shouldn't need this since it should be using OCP 4.10 release image.
We suspect that the timeout is caused by OVNKuberenetes (default network in OCP 4.11)

PR #1683, opened 2022-06-13
MGMT-9914: Integration of the ODF LVM operator with assisted installer

dvachhan:
This PR depends on [openshift/assisted-service#3921](https://github.com/openshift/assisted-service/pull/3921)


osherdp:
/uncc
@dvachhan please cc me again when you want me to review :)

osherdp:
@dvachhan can you add a job so that we can see how it works?
for example: https://github.com/openshift/release/blob/b21f28ae0b8d8a04b296c0c973fab00280a07ffa/ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml#L260-L267

src/consts/olm_operators.py
```diff
@@ -79,6 +80,11 @@ def values(cls) -> dict:
                 worker_count=4,
             ),
             OperatorType.LSO: cls.get_resource_dict(),
+            OperatorType.LVM: cls.get_resource_dict(
+                master_memory=1200,
```
osherdp, line 84: just so I understand, those are additional resources that are required for the operator right?

osherdp (changes requested):

gamli75 (approved):

adriengentil (approved):

PR #1693, opened 2022-06-15
Change the default disk size from 120GB to 100GB

osherdp:
Since all product dependant places have been handled for the various changes in MGMT-9682, we can change the actual usage of it in our tests.


src/consts/resources.py
```diff
@@ -4,7 +4,7 @@
 DEFAULT_MASTER_MEMORY: int = 16984
 DEFAULT_MTU: int = 1500
 DEFAULT_WORKER_DISK: int = 20 * GB
-DEFAULT_MASTER_DISK: int = 120 * GB
+DEFAULT_MASTER_DISK: int = 100 * GB
```
eliorerz, line 7:
```suggestion
DEFAULT_MASTER_DISK: int = DISK_SIZE_100GB
```

eliorerz (changes requested):

src/consts/resources.py
osherdp: Done!

eliorerz (approved):
LGTM

PR #1695, opened 2022-06-16
MGMT-10830 - Create base class for test-infra containerized controllers

eliorerz:
Join base logic for test-infra containerized controllers
After adding tang support to test-infra we have 3 duplicated code/behaviors on the containerized controller: ipxe_controller, tang_controller, and on proxy_controller. This PR move the base controller behavior under new base class.

/cc @osherdp

### DIFF
--- a/src/assisted_test_infra/test_infra/controllers/containerized_controller.py
+++ b/src/assisted_test_infra/test_infra/controllers/containerized_controller.py
@@ -1,0 +1,46 @@
+from abc import ABC
+from typing import List, Optional
+
+from assisted_test_infra.test_infra.utils import utils
+from service_client import log
+
+
+class ContainerizedController(ABC):
+    def __init__(self, name: str, port: int, image: str, extra_flags: Optional[List[str]] = None) -> None:
+        self._name = name
+        self._port = port
+        self._image = image
+        self._extra_flags = [] if not extra_flags else extra_flags
+        self._is_running = False
+
+    def run(self, **kwargs):
+        if self._is_running:
+            log.info(f"{self.__class__.__name__} server is already running...")
+            return
+        log.info(f"Running {self.__class__.__name__} Server {self._name}")
+
+        self._on_container_start(**kwargs)
+        base_run_flags = [
+            "-d",
+            "--restart=always",
+            "--network=host",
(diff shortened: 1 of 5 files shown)


eliorerz:
/hold
Wait for `e2e-metal-assisted-ipxe`, `e2e-metal-assisted-tang` and  `e2e-metal-assisted-ipv6` to pass

src/assisted_test_infra/test_infra/controllers/containerized_controller.py
```diff
@@ -0,0 +1,41 @@
...
+from assisted_test_infra.test_infra.utils import utils
+from service_client import log
+
+
+class ContainerizedController(ABC):
+    def __init__(self, name: str, port: int, image: str) -> None:
+        self._name = name
+        self._port = port
+        self._image = image
+        self._extra_flags = []
```
osherdp, line 12: maybe enable setting ``_extra_flags`` from ctor parameters as well

src/assisted_test_infra/test_infra/controllers/proxy_controller/proxy_controller.py
```diff
@@ -6,82 +6,71 @@
...
             self.authenticated = authenticated
-            self.image = "quay.io/sameersbn/squid"
-            self.dir = dir
+            self.image = ""
             self._is_ipv6 = is_ipv6
             self._set_server_address(host_ip)
             self._create_conf_from_template(denied_port=denied_port)
             self._create_user_file_for_auth()
-            self._run_proxy_server()
+            self.run()
```
osherdp, line 40: kind of weird to do the ``run`` from the c'tor. can we move it from this scope?

osherdp (changes requested):

src/assisted_test_infra/test_infra/controllers/proxy_controller/proxy_controller.py
eliorerz: I didn't want to change its behavior , you think it's ok to do it inside this PR?

osherdp: I think so. it's just moving one line

osherdp (approved):

PR #1696, opened 2022-06-19
NO-ISSUE: it takes time to hypershift to create namespace and we should add secret only after it was created

tsorya:
NO-ISSUE: it takes time to hypershift to create namespace and we should add secret only after it was created


eranco74:
I think you should explicitly wait for the namespace to get created but let's see if this works

tsorya:
> I think you should explicitly wait for the namespace to get created but let's see if this works

there is no need for it

PR #1699, opened 2022-06-20
move location of create_full_environment.sh

osherdp:
No need to have it in root dir, so moving it to its natural place.

/cc @eliorerz 
/hold


eliorerz (approved):
LGTM

PR #1703, opened 2022-06-21
NO-ISSUE - Use the default node controller in day2 test

eliorerz:
Currently we are creating a new controller configuration on `day2cluster.start_install_and_wait_for_installed` that cause the configuration not to be triggered on test start. 
This PR using the already exists controller instead of creating a new one 

/cc @omertuc

### DIFF
--- a/src/assisted_test_infra/test_infra/helper_classes/day2_cluster.py
+++ b/src/assisted_test_infra/test_infra/helper_classes/day2_cluster.py
@@ -11,13 +11,12 @@ import waiting
 
 import consts
 from assisted_test_infra.test_infra import utils
-from assisted_test_infra.test_infra.controllers.node_controllers.libvirt_controller import LibvirtController
+from assisted_test_infra.test_infra.controllers import NodeController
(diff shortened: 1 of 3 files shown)


src/assisted_test_infra/test_infra/helper_classes/day2_cluster.py
```diff
@@ -88,9 +87,6 @@ def start_install_and_wait_for_installed(self):
         tfvars = utils.get_tfvars(self.config.tf_folder)
         tf_network_name = tfvars["libvirt_network_name"]
 
-        config = TerraformConfig()
-        config.nodes_count = num_nodes_to_wait
```
omertuc, line 92: What about this

eliorerz: It's not doing anything (deprecated)
https://github.com/openshift/assisted-test-infra/blob/master/src/assisted_test_infra/test_infra/helper_classes/config/controller_config.py#L48-#L54

PR #1704, opened 2022-06-21
Add separate log file for each test worker during parallel pytest runs

mchernik:
Signed-off-by: Mykhailo Chernyk <mchernik@users.noreply.github.com>

### DIFF
--- a/src/tests/conftest.py
+++ b/src/tests/conftest.py
@@ -1,3 +1,4 @@
+from logging import getLogger, FileHandler
 from typing import List
 
 import pytest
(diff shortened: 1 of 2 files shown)


lalon4:
Thanks Misha! need to make sure we will collect those loggers (to s3). Also those files will be replacing the current test_infra.log?

src/tests/conftest.py
```diff
@@ -1,16 +1,29 @@
...
 from service_client.client_validator import verify_client_version
 from tests.config import global_variables
 
 assert global_variables.pull_secret is not None, "Missing pull secret"
 
+@pytest.fixture(scope="session", autouse=True)
+def setup_worker_log(worker_id):
+    def configure_logger(logger):
+        fmt = SensitiveFormatter("%(asctime)s - %(process)d - %(name)s - %(levelname)s - %(message)s")
+        fh = FileHandler("test_infra_{worker_id}.log")
```
tsorya, line 19: what is worker id?

tsorya: Why we can't use test name for file name? I just not sure this worker_id somehow will help. We will have 100 log files in regression, how we should know wich one to open? 
Lets create logs folder and add it to .gitignore

mchernik:
> Thanks Misha! need to make sure we will collect those loggers (to s3). Also those files will be replacing the current test_infra.log?

There will be the same test_infra.log and a subset log file per worker.

src/tests/conftest.py
mchernik: To my understanding, the biggest problem was that logs from different parallel executors were mixed in single file, so it was a problem to track a source of a problem. Currently there would be single continuous log per executor. 

Not sure if per-test logging is possible, taking into account that fixtures can be run before/after test. I'll look into it.

PR #1705, opened 2022-06-21
Provision scsi disks for libvirt domains

trewest:
Changes:

- Changes terraform provisioned disks to scsi
- Change libvirt_domain machine type to q35
- Renames `consolemodel.xsl` to `console_cdrom_model.xsl`
- Adds xsl to update cdrom bus to sata rather than ide


terraform_files/baremetal/console_cdrom_model.xsl
```diff
@@ -15,4 +15,10 @@
         <log file="/var/log/libvirt/qemu/{/domain/name}-console.log" append="on"/>
       </xsl:copy>
   </xsl:template>
+
+  <xsl:template match="/domain/devices/disk[@device='cdrom']/target/@bus">
+  <xsl:attribute name="bus">
```
osherdp, line 20: Any chance you can fix indentation?
Thanks!

terraform_files/baremetal_host/main.tf
```diff
@@ -68,8 +69,10 @@ resource "libvirt_domain" "host" {
     }
   }
 
+  machine = "q35"
+
   xml {
-    xslt = file("consolemodel.xsl")
+    xslt = file("console_cdrom_model.xsl")
```
osherdp, line 75:
```suggestion
    xslt = file("../baremetal_host/console_cdrom_model.xsl")
```
then those other files under ``terraform_files/baremetal``, ``terraform_files/none`` & ``terraform_files/baremetal_infra_env`` can safely be deleted and we can save ourselves the need to update it everywhere

terraform_files/baremetal_host/main.tf
```diff
@@ -68,8 +69,10 @@ resource "libvirt_domain" "host" {
     }
   }
 
+  machine = "q35"
```
osherdp, line 72: I think we should leave default behavior as-is and just allow templating this stuff if requested
You can never know what other environments are supporting and what's not, so even deciding to transition should be done gradually

osherdp (changes requested):

terraform_files/baremetal_host/main.tf
trewest: done

terraform_files/baremetal_host/main.tf
osherdp: the same way we're templating ``vcpu``, ``memory`` and ``name`` in this file (to name a few). Those values are being set based on user input or other calculations

osherdp: For more info, please read: https://github.com/openshift/assisted-test-infra/blob/master/GUIDE.md#adding-a-new-e2e-flow

trewest: okay, I see what you are saying now

trewest:
Relies on some commits provided by: https://github.com/openshift/assisted-test-infra/pull/1709

PR #1706, opened 2022-06-22
NO-ISSUE - Fix unbound `SUBSYSTEM_LOCAL_REGISTRY` variable when `DEBUG_SERVICE=true`

eliorerz:
`SUBSYSTEM_LOCAL_REGISTRY` was changed to `SUBSYSTEM_LOCAL_REGISTRY` and wasn't changed on `deploy_assisted_service.sh`

/cc @osherdp @omertuc


scripts/deploy_assisted_service.sh
```diff
@@ -7,6 +7,7 @@ set -o xtrace
 export SERVICE_NAME=assisted-service
 export SERVICE_URL=$(get_main_ip)
 export AUTH_TYPE=${AUTH_TYPE:-none}
+export LOCAL_ASSISTED_ORG=${LOCAL_ASSISTED_ORG:-none}
```
eliorerz, line 10: Ignore that . will be removed

PR #1708, opened 2022-06-22
NO-ISSUE: Remove Fedora support, add warning about this repo

omertuc:
Users are not recommended to run the code in this repo on their local
machine, as it makes some surprising opinionated configuration changes
that users might not expect (installing packages, configuring libvirt,
etc). We should give a clear warning about that in the README.

Also since we don't support Fedora, our `create_full_environment.sh`
should be modified to reflect that and block users trying to install
on Fedora. If we detect an OS we don't support, on top of saying we
don't support it, we give the same warning as in the README.

### DIFF
--- a/scripts/create_full_environment.sh
+++ b/scripts/create_full_environment.sh
@@ -8,8 +8,9 @@ function error() {
 
 # Check OS
 OS=$(awk -F= '/^ID=/ { print $2 }' /etc/os-release | tr -d '"')
-if [[ ! ${OS} =~ ^(centos)$ ]] && [[ ! ${OS} =~ ^(rhel)$ ]] && [[ ! ${OS} =~ ^(fedora)$ ]] && [[ ! ${OS} =~ ^(rocky)$ ]]; then
-    error "\"${OS}\" is an unsupported OS. We support only CentOS, RHEL, Fedora or Rocky."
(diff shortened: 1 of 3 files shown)


scripts/create_full_environment.sh
```diff
@@ -8,8 +8,9 @@ function error() {
 
 # Check OS
 OS=$(awk -F= '/^ID=/ { print $2 }' /etc/os-release | tr -d '"')
-if [[ ! ${OS} =~ ^(centos)$ ]] && [[ ! ${OS} =~ ^(rhel)$ ]] && [[ ! ${OS} =~ ^(fedora)$ ]] && [[ ! ${OS} =~ ^(rocky)$ ]]; then
-    error "\"${OS}\" is an unsupported OS. We support only CentOS, RHEL, Fedora or Rocky."
+if [[ ! ${OS} =~ ^(centos)$ ]] && [[ ! ${OS} =~ ^(rhel)$ ]] && [[ ! ${OS} =~ ^(rocky)$ ]]; then
+    error "\"${OS}\" is an unsupported OS. We support only CentOS, RHEL or Rocky."
```
rccrdpccl, line 12: Shall we add Rocky to the README then?

osherdp (approved):

PR #1709, opened 2022-06-22
MGMT-10913: Use our fork of dmacvicar/terraform-provider-libvirt

omertuc:
https://github.com/dmacvicar/terraform-provider-libvirt is recently
unmaintained and doesn't seem to be merging pull requests.

There are some changes we wanted to make in the project but couldn't
because the maintainer haven't merged our pull requests for a while now,
so we decided we should [fork it](https://github.com/openshift-assisted/terraform-provider-libvirt) instead for now.

Changes include:
- Race condition when creating two libvirt networks at the same time
- Some missing ARM support (@tsorya can elaborate)
- Like three other bugs / issues mentioned in our code if you grep for dmacvicar

This pull request makes it so that when running tests, the fork will be
cloned, built, and installed. All terraform files have been modified
to use the fork.

We always use pull and build the latest version of our fork (master
branch).

### DIFF
--- a/Makefile.terraform_fork
+++ b/Makefile.terraform_fork
@@ -1,0 +1,43 @@
+TF_LIBVIRT_PROJECT = terraform-provider-libvirt
+TF_LIBVIRT_BINARY = terraform-provider-libvirt
+TF_LIBVIRT_FORK_URL = https://github.com/openshift-assisted/$(TF_LIBVIRT_PROJECT)
+TF_LIBVIRT_FORK_DIR = /opt/$(TF_LIBVIRT_PROJECT)
+TF_LIBVIRT_LOCAL_REGISTRY = /usr/share/terraform/plugins/assisted-test-infra-registry.local/openshift-assisted/libvirt/9.9.9/linux_amd64
+
+# Allows users to specify the git revision that will be switched to before building the fork
+# Only git commit hashes are allowed. It must be an ancestor of the main branch.
+TERRAFORM_PROVIDER_LIBVIRT_FORK_GIT_REVISION := $(or ${TERRAFORM_PROVIDER_LIBVIRT_FORK_GIT_REVISION},"")
+
+.PHONY: install
+install: $(TF_LIBVIRT_LOCAL_REGISTRY)/$(TF_LIBVIRT_BINARY)
+
+# A folder containing a clone of our fork of the libvirt terraform provider
+$(TF_LIBVIRT_FORK_DIR):
+	git clone $(TF_LIBVIRT_FORK_URL) $@ 
+
+# A folder acting as local registry of libvirt providers
+# (See https://www.terraform.io/language/providers/requirements#in-house-providers)
+$(TF_LIBVIRT_LOCAL_REGISTRY):
+	mkdir -p $@
+
+# .PHONY because we can't know whether it's stale or not, we always want to git pull
+# and run the build, the go toolchain will decide if anything needs rebuilding in practice,
+# we should not leave it up to make.
+.PHONY: $(TF_LIBVIRT_FORK_DIR)/$(TF_LIBVIRT_BINARY)
+
+# Create the actual binary by building. Requires Go.
+$(TF_LIBVIRT_FORK_DIR)/$(TF_LIBVIRT_BINARY): $(TF_LIBVIRT_FORK_DIR)
+	git -C $(TF_LIBVIRT_FORK_DIR) fetch origin main
+	if [ "$(TERRAFORM_PROVIDER_LIBVIRT_FORK_GIT_REVISION)" = "" ]; then\
+		# User didn't ask for any particular revision, so use the tip of the main branch\
+		git -C $(TF_LIBVIRT_FORK_DIR) switch --detach origin/main;\
+	else \
+		# User asked for a particular revision, switch to it before building\
+		git -C $(TF_LIBVIRT_FORK_DIR) switch --detach "$(TERRAFORM_PROVIDER_LIBVIRT_FORK_GIT_REVISION)";\
+	fi;
+	make -C $(TF_LIBVIRT_FORK_DIR)
+
+# Copy the build binary into our local "terraform registry"
+$(TF_LIBVIRT_LOCAL_REGISTRY)/$(TF_LIBVIRT_BINARY): $(TF_LIBVIRT_FORK_DIR)/$(TF_LIBVIRT_BINARY) $(TF_LIBVIRT_LOCAL_REGISTRY) 
+	cp -r $< $@
+
--- a/terraform_files/baremetal_infra_env/main.tf
+++ b/terraform_files/baremetal_infra_env/main.tf
@@ -1,8 +1,8 @@
 terraform {
   required_providers {
     libvirt = {
-      source = "dmacvicar/libvirt"
-      version = "0.6.12"
+      source = "assisted-test-infra-registry.local/openshift-assisted/libvirt"
+      version = "9.9.9"
     }
   }
 }
--- a/Dockerfile.assisted-test-infra
+++ b/Dockerfile.assisted-test-infra
@@ -37,6 +37,11 @@ ENV GOPATH=/go
 ENV GOCACHE=/go/.cache
 ENV PATH=$PATH:/usr/local/go/bin:/go/bin
 
+COPY Makefile.terraform_fork .
+
+ENV TERRAFORM_PROVIDER_LIBVIRT_FORK_GIT_REVISION=734281d20f9c50da411cf2d7db1aa4303746b3f2
+RUN make -f Makefile.terraform_fork install
+
 COPY . .
 
 RUN chgrp -R 0 /home/assisted-test-infra && \
--- a/skipper.yaml
+++ b/skipper.yaml
@@ -1,6 +1,5 @@
 registry: quay.io
-build-container-image: test-infra
-build-container-tag: latest
+build-container-image: assisted-test-infra
 
 volumes:
   # programs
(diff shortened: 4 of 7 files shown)


Makefile
```diff
@@ -337,7 +341,7 @@ test_parallel:
 	skipper make $(SKIPPER_PARAMS) _test_parallel
 	scripts/assisted_deployment.sh set_all_vips_dns
 
-_test_setup:
+_test_setup: $(TF_LIBVIRT_LOCAL_REGISTRY)/$(TF_LIBVIRT_BINARY)
```
tsorya, line 344: why not building and adding it as part of build image? i think it will be more cleaner

Makefile.terraform_fork
```diff
@@ -0,0 +1,33 @@
+# Needed for terraform to pick up our locally compiled fork of the terraform
+# libvirt provider
+PATH := ${PATH}:$(ROOT_DIR)/terraform-provider-libvirt
+
+TF_LIBVIRT_PROJECT = terraform-provider-libvirt
+TF_LIBVIRT_BINARY = terraform-provider-libvirt
+TF_LIBVIRT_FORK = https://github.com/openshift-assisted/$(TF_LIBVIRT)
```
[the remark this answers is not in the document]
omertuc: Yes :thinking: how did this work for me :man_shrugging:

Makefile.terraform_fork
```diff
@@ -0,0 +1,33 @@
...
+TF_LIBVIRT_LOCAL_REGISTRY = /usr/share/terraform/plugins/assisted-test-infra-registry.local/openshift-assisted/libvirt/9.9.9/linux_amd64
+
+# A folder containing our fork of the libvirt terraform provider
+$(TF_LIBVIRT_PROJECT):
+	git clone  $@
```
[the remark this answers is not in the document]
omertuc: Right, now sure how that happened

Makefile
omertuc: Because the build image will not automatically get updated when we push to the fork

omertuc: But it could be resolved, I'll try something

tsorya: I don't think we need to update build image each time we push to fork (it will happened once in a month maybe). In ci we are building image each time so there is no problem with it and in dev envs we will build image only when will need something from fork

osherdp: Can I suggest something else?
We can store this provider in the terraform registry, saving our time building it and leaving test-infra a bit cleaner 
Any major downside for that suggestion?

tsorya: Nope, sounds great

osherdp: Ack, on it

terraform_files/baremetal_infra_env/main.tf
```diff
@@ -1,8 +1,8 @@
 terraform {
   required_providers {
     libvirt = {
-      source = "dmacvicar/libvirt"
-      version = "0.6.12"
+      source = "assisted-test-infra-registry.local/openshift-assisted/libvirt"
```
osherdp, line 4:
```suggestion
      source = "openshift-assisted/libvirt"
```

Makefile
osherdp: Done! Added usage instructions as a suggestion comment

osherdp (changes requested):

terraform_files/baremetal_infra_env/main.tf
omertuc: ?

How would that work?

Makefile
omertuc: We'd also have to do "proper" versioning rather than 9.9.9 which is yet even more complexity to the CI of the fork

omertuc: Also publishing manually is out of the question, it requires tribal knowledge that will be really hard to figure out for anyone in the future.

omertuc: Where will we keep the credentials for the terraform registry account?

osherdp: Wow, lots of questions 😂
We currently have automatic publishing on every new release. It should be very easy managing that

omertuc: >We currently have automatic publishing on every new release

Where?

osherdp: openshift-assisted/terraform-provider-libvirt
Take a look at github actions for that repo

omertuc: Can't find the part that publishes to the terraform registry in those actions

omertuc: Also I'd hate having to rely on github actions. And I'm still not sure who's gonna own the terraform account and create new tokens for it whenever we need to

omertuc: >saving our time building 

It's a very fast build, and it only happens when the skipper dockerfile is rebuilt, it doesn't happen every time

> and leaving test-infra a bit cleaner

It's leaving test-infra cleaner but requires a lot of maintenance, knowledge and work behind the scenes that I'm not sure is worth it just to get rid of a simple Makefile

osherdp: @omertuc all this setup is already been set up by me, with tokens & terraform account and everything else. You can try it out with new tags+releases and see how it goes it's very simple

As always, it's possible for something to get broken, but I think it's not very probable. what do you think about giving it a go and trying it out? If something with github-actions will get broken then it will block us from publishing a new provider version, it will not affect stability in test-infra directly

omertuc: >Can't find the part that publishes to the terraform registry in those actions

You didn't address this ^

>set up by me

This is the part of the problem, only you know how it works, only you know how to login to that account and with which email to do it. 

With my solution we're cloning a public repo and building it with a makefile - there's no secret knowledge. With your solution, there's an entire process going on behind the scenes that is:

1. hard to discover (one would have to *know* it's done by github actions, it's not obvious)
2. full of secret knowledge - I still don't know where is the part that publishes to the terraform registry - I didn't see it in neither the github actions or the goreleaser yaml. I still don't know which email is used for the terraform account. I still don't know which password is used to login to the terraform account. All I have is a magical token that may stop working one day with no way to create a new one.

>it's possible for something to get broken, but I think it's not very probable

I'm not afraid github actions will break. I'm afraid of having another CI system people in this team would need to learn and be familiar with in order to maintain.

osherdp: @omertuc, same argument can be said for that new fork which contain commits that basically only you know what they are doing.
I think not knowing how publishing goes is a smaller knowledge debt than that whole code-base of the provider that we're going to maintain. Don't you agree about that point?

If we're going to "trust' code that was there in the provider then we should also trust the existing automation for publishing the provider (because it's 100% the same)

osherdp: https://github.com/openshift-assisted/terraform-provider-libvirt/actions/workflows/release.yml

omertuc: >same argument can be said for that new fork which contains commits that basically only you know what they are doing.

That's irrelevant as we need those commits, that's the entire point of the fork. The commits are gonna be there with or without using the terraform registry vs a local one. My commits come with a detailed technical explanation of the problem they solve and how they solve it, there's no hidden knowledge (e.g. password to the terraform account).

> I think not knowing how publishing goes is a smaller knowledge debt than that whole code-base of the provider that we're going to maintain. Don't you agree about that point?

Still irrelevant, we've decided that we have to maintain the fork, it has to be done. There's no way around that. There is a way around avoiding publishing to the online terraform registry, and that way is a simple Makefile. The entire process can be easily followed - Clone, make, install to local registry. With the online terraform registry - once I reach the registry, I have no way to know what's behind it. Who's publishing to it. How they're doing it, when they're doing it. How can I change that process to release a new version. To you it's obvious that's you're using github actions, but it's not obvious to someone just seeing `openshift-assisted/libvirt` in a tf file. 

>trust the existing automation for publishing the provider (because it's 100% the same)

I 100% trust it, but I have no idea how to operate it. I literally can't - I don't know which email you used to sign up to terraform. I don't know which password you used to sign up for terraform. If the token expires, we're literally locked out of the account - I have no way of finding that information if I wanted to. I would understand if it had a lot of value but it barely gives me any value and requires me to be dependent on:

- A new CI system, GitHub actions.
- A collection of secrets, some of them in GitHub (token), some of them only @osherdp knows (email and password for the account used to generate the token)
- The Terraform Registry
- A need to maintain version numbers and releases to my fork, which are different from the version numbers used by the original repo, which is confusing. With the makefile I just use the hash of the git commit I want to depend on and that's it.

>Can't find the part that publishes to the terraform registry in those actions

You still haven't answered this. Please point me at the lines in the code responsible for pushing the artifacts to the terraform registry, I literally can't find them. I see it published to the release page https://github.com/openshift-assisted/terraform-provider-libvirt/releases but where does it publish to the terraform registry?

osherdp: Answering your question, it is this github-action step that publishes to the terraform registry: https://github.com/openshift-assisted/terraform-provider-libvirt/blob/main/.github/workflows/release.yml#L41-L49

As for the other considerations, I guess we can't convince each other because each one has something else more prioritized in his trade-off calculations (again, it's all good. we should sometimes have different opinions)
I'll suggest letting other people speak their minds. @tsorya @eliorerz should we `` go build`` the new provider in test-infra or have it published with github-actions where the code base is located?

omertuc: >Answering your question, it is this github-action step that publishes to the terraform registry: https://github.com/openshift-assisted/terraform-provider-libvirt/blob/main/.github/workflows/release.yml#L41-L49

All I see in the lines you linked is that it's triggering the go-releaser config https://github.com/openshift-assisted/terraform-provider-libvirt/blob/a59c51044c8a75d4bd90487f4db19e795eff5204/.goreleaser.yml#L1-L53

which doesn't seem to be publishing to the terraform registry.

osherdp: That's the only relevant **line of code** for that process. There is a github webhook for release events that's taking it from there. Take a look at https://www.terraform.io/registry/providers/publishing#webhooks

omertuc: Ack, thanks

PR #1711, opened 2022-06-23
Fixing debugging the service from test-infra by setting DEBUG_SERVICE…

michaellevy101:
The image doesn't pushed to the local registry


eliorerz (approved):
LGTM

PR #1718, opened 2022-06-26
NO-ISSUE: Use ASSISTED_OPENSHIFT_INSTALL_RELEASE_IMAGE for hypershift release image

eranco74:
I hope this will get the periodic job working again

### DIFF
--- a/src/tests/test_kube_api.py
+++ b/src/tests/test_kube_api.py
@@ -198,7 +198,7 @@ class TestKubeAPI(BaseKubeAPI):
                     agent_namespace=spoke_namespace,
                     provider_image=os.environ.get("PROVIDER_IMAGE", ""),
                     hypershift_cpo_image=os.environ.get("HYPERSHIFT_IMAGE", ""),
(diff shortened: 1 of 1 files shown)


src/tests/test_kube_api.py
```diff
@@ -198,7 +198,7 @@ def capi_test(
                     agent_namespace=spoke_namespace,
                     provider_image=os.environ.get("PROVIDER_IMAGE", ""),
                     hypershift_cpo_image=os.environ.get("HYPERSHIFT_IMAGE", ""),
-                    release_image=os.environ.get("OPENSHIFT_INSTALL_RELEASE_IMAGE", ""),
+                    release_image=os.environ.get("ASSISTED_OPENSHIFT_INSTALL_RELEASE_IMAGE", ""),
```
osherdp, line 201: Are you sure it's being defined for test-infra based jobs?

osherdp (changes requested):

src/tests/test_kube_api.py
eranco74: Hmmm, you are right, it seems that it's being passed only for operator workflows.

PR #1729, opened 2022-06-30
MGMT-9909: Use centos stream 9 in Dockerfile

adriengentil:
Use centos stream 9 in infra test Dockerfile, it has the benefits to
bring an updated version of dracut that stop polluting the logs when it
tries to build the initramfs from inside a container.


adriengentil:
/retitle MGMT-9909: Use centos stream 9 in Dockerfile

osherdp (approved):

PR #1731, opened 2022-07-04
additional parameter(timeout) for method wait_for_specific_host_status  and new host event parameter DISCONNECTED

eliajahshan:
@eliorerz @lalon4
Function wait_for_specific_host_status , will have a new parameter (timeout) 
new Parameter in host events : DISCONNECTED

change is needed to support kni test , which validate that cluster installation is timeout after 3 min when a host which already reached the state "write to disk" is now disconnected for 3 min+


eliorerz:
Need to fix linter first.
Use `make reformat` that might fix the issue

eliorerz (approved):
LGTM

PR #1732, opened 2022-07-04
MGMT-10946: fix memory usage based on OCP requirements

osherdp:
* Update all SNO installations (defined in assisted-test-infra at least) to use exactly 16GiB of RAM as official requirement dictates. It should apply to e2e-metal-assisted-single-node job and for kube-api jobs.
* Update other memory definitions of master and worker nodes to be actual multiples of gigibytes.


src/consts/resources.py
```diff
@@ -1,12 +1,12 @@
 from consts import DISK_SIZE_100GB, GB
 
-DEFAULT_WORKER_MEMORY: int = 8892
-DEFAULT_MASTER_MEMORY: int = 16984
+DEFAULT_WORKER_MEMORY: int = 8 * 1024  # mebibytes
```
eliorerz, line 3: Why not declare `MiB = 1024` as same as we did on `GB`?

eliorerz: That will make all the comments redundant

eliorerz (changes requested):

src/consts/resources.py
osherdp: ``MiB = 1024`` seems incorrect, because 1024 is actually K

osherdp: I initially did ``(8 * GiB) // MiB`` or something alike, not sure what's more clear of the two options

eliorerz: Maybe something like `DEFAULT_WORKER_MEMORY: int = 3 * MiB_UNITS` where `MiB_UNITS=1024`?

osherdp: Done!

osherdp:
AFAIK conformance tests failure in [e2e-metal-single-node-with-worker-live-iso](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_assisted-test-infra/1732/pull-ci-openshift-assisted-test-infra-master-e2e-metal-single-node-with-worker-live-iso/1545038081113985024) are known issue

eliorerz (approved):
LGTM

osherdp:
/test e2e-metal-assisted-odf
seems like the existing flake we know about. retesting

osherdp:
known issue

PR #1738, opened 2022-07-05
NO-ISSUE: allow cross network traffic

adriengentil:
The rootfs is now downloaded from the assisted-image-service instead of
internet.

As a consequence, when we try to setup a cluster located out of the hub
cluster network, the machine that boots the assisted ISO fails to
retrieve the rootfs because libvirt denies the traffic between guest
networks.

This change sets up a hook in libvirt that drops the rules denying
forwarded traffic between guest networks.

Here is a job failing because of this issue:
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-assisted-service-master-edge-e2e-metal-assisted-kube-api-late-binding-single-node-periodic/1544506948760637440

### DIFF
--- a/scripts/install_environment.sh
+++ b/scripts/install_environment.sh
@@ -37,6 +37,8 @@ function install_libvirt() {
     sudo sed -i -e 's/#tcp_port/tcp_port/g' /etc/libvirt/libvirtd.conf
     sudo sed -i -e 's/#security_driver = "selinux"/security_driver = "none"/g' /etc/libvirt/qemu.conf
 
+    allow_libvirt_cross_network_traffic
+
     if ! version_is_greater "$current_version" "$minimum_version"; then
         add_libvirt_listen_flag
     else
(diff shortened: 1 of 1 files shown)


scripts/install_environment.sh
```diff
@@ -87,6 +89,19 @@ function is_libvirtd_tcp_socket_enabled_and_running() {
...
 
+function allow_libvirt_cross_network_traffic() {
+    # Flush LIBVIRT_FWI chain each time the network configuration is being updated
+    echo "Installing libvirt network hook to allow cross network traffic"
+    hook_dir="/etc/libvirt/hooks/network.d"
+    hook_filename="${hook_dir}/allow-cross-network-traffic.sh"
+    sudo mkdir -p "${hook_dir}"
+    sudo tee "${hook_filename}" << EOF
+#!/usr/bin/env sh
+iptables -F LIBVIRT_FWI || true
```
omertuc, line 100: Long form preferred in scripts

```suggestion
iptables --flush LIBVIRT_FWI || true
```

omertuc, line 100: Could this [simpler](https://github.com/openshift/assisted-service/blob/3cf458c0af48b42940efaf19e2a2351ae79fc60f/deploy/operator/ztp/none_platform_utils.sh#L71-L76) code be achieving the same thing as you're trying to do?

adriengentil: libvirt set its own rules out of firewalld abstraction.
The `libvirt` zone in firewalld is mostly used to control guest to host traffic (and allows traffic forwarding, I guess to prevent clashes with the rules libvirt sets).

Here, flushing the chain `LIBVIRT_FWI`, managed by libvirt, allows the traffic between guest networks.

https://libvirt.org/firewall.html#firewalld-and-the-virtual-network-driver

omertuc:
WIP?

scripts/install_environment.sh
```diff
@@ -87,6 +89,19 @@ function is_libvirtd_tcp_socket_enabled_and_running() {
     fi
 }
 
+function allow_libvirt_cross_network_traffic() {
+    # Flush LIBVIRT_FWI chain each time the network configuration is being updated
```
omertuc, line 93:
```suggestion
    # Flush LIBVIRT_FWI chain each time the network configuration is being updated
    # We do this because .... needs to ... for ....
```

omertuc:
Can you link in the commit message to a prow job that is failing because of this issue?

carbonin:
FYI @CrystalChun in case you hadn't seen this

carbonin:
@adriengentil what would be involved in getting dns to work cross network also?

I was trying to use test-infra to get an environment where assisted-service could access the kubernetes API endpoint on an installed cluster, but ended up with dns resolution issues.

I assume this doesn't solve that outright, but could you point me in the right direction?
This would help a lot with https://issues.redhat.com/browse/MGMT-11143 - we wouldn't have to use the devscripts setup just to get routing to the spoke.

adriengentil:
> FYI @CrystalChun in case you hadn't seen this

This PR is nothing without https://github.com/openshift/assisted-test-infra/pull/1747 :)

PR #1739, opened 2022-07-05
Delete tfstate file during destroy nodes

lalon4:
Since recently one of our e2e test is failing (test_unique_vip_hostname). 
The test create cluster and nodes, and than delete created cluster and nodes, and create another set of cluster and nodes. Issue is that lately it is failing to create the second set of nodes - tf apply get an error: 
"Storage pool not found: no storage pool with matching name".
Turns out that whenever we clear the nodes we use libvirt cleanup (instead of tf) so tf state assume disks are already exists, while in reality they were cleaned by delete all virsh resources.
This PR clean tf state file to avoid this failure.

### DIFF
--- a/src/assisted_test_infra/test_infra/controllers/node_controllers/terraform_controller.py
+++ b/src/assisted_test_infra/test_infra/controllers/node_controllers/terraform_controller.py
@@ -309,6 +309,11 @@ class TerraformController(LibvirtController):
         self._delete_virsh_resources(
             self._entity_name.get(), self.params.libvirt_network_name, self.params.libvirt_secondary_network_name
         )
+        tfstate_path = f"{self.tf_folder}/{self.tf.STATE_FILE}"
+        if os.path.exists(tfstate_path):
+            log.info(f"Deleting tf state file: {tfstate_path}")
+            os.remove(tfstate_path)
+
         if delete_tf_folder:
             log.info("Deleting %s", self.tf_folder)
             shutil.rmtree(self.tf_folder)


src/assisted_test_infra/test_infra/controllers/node_controllers/terraform_controller.py
```diff
@@ -309,6 +309,11 @@ def destroy_all_nodes(self, delete_tf_folder=False):
         self._delete_virsh_resources(
             self._entity_name.get(), self.params.libvirt_network_name, self.params.libvirt_secondary_network_name
         )
+        tfstate_path = f"{self.tf_folder}/terraform.tfstate"
```
osherdp, line 312: 5 lines above we're doing "terraform destroy"
Why doesn't it update the state (and delete the storage pool)?

osherdp (changes requested):

src/assisted_test_infra/test_infra/controllers/node_controllers/terraform_controller.py
lalon4: thats a good question, we also have _delete_virsh_resources right after, explicitly to delete all resources, so i am not sure what tf destroy do

eliorerz, line 312:
```suggestion
        tfstate_path = f"{self.tf_folder}/{self.STATE_FILE}"
```

eliorerz (changes requested):

src/assisted_test_infra/test_infra/controllers/node_controllers/terraform_controller.py
osherdp: Let's first understand what's going on there, maybe the ``_delete_virsh_resources`` action can be a redundant one (it's still called terraform-controller so terraform makes sense to be the main access point)

lalon4: Your right, and yet, even if its redundant, it require more investigations and understanding why we even use the virsh cleanup. We can open a task for investigating / fixing this - but since its a regression (that probably related to newer tf version as i haven't found why it suddenly stopped working), i would like to have this workaround, shouldn't impact any existing functionality. WDYT?

lalon4: https://issues.redhat.com/browse/MGMT-11072

osherdp (approved):

src/assisted_test_infra/test_infra/controllers/node_controllers/terraform_controller.py
lalon4: Sure thanks Osher, although i must admit i invested 2 days on that failure, i involved both @eliorerz and @tsorya trying to understand what was changed, so from my POV and based on the conversation with Igal we thought that it would be the best solution. As for the redundant part - i dont really think its redundant as i mentioned on the ticket, we do virsh actions during tests so by definition tf state wont be up to date. So its not a habit, but rather i got to wrong conclusion, anyway this is the reason we have reviews for.

osherdp: Just to make sure I'm understood: I have no problem with the process that was made before this PR got created. It's okay to make mistakes and/or not to pay attention to some small details as nobody's perfect 

What I want us to get better at is situations where we understand something is weird, and yet feel the urge to merge workarounds. Please let me know we're on the same page about that and that we'll gonna minimize those occasions 🙏

PR #1746, opened 2022-07-11
NO-ISSUE: print ipatbles rules in case setup_capi_env failed to update iptables LIBVIRT_FWI chain

eranco74:


eranco74:
/close
should get fixed by https://github.com/openshift/assisted-test-infra/pull/1747

PR #1748, opened 2022-07-11
MGMT-11119: CI workaround: Get swtpm from source 

nmagnezi:
This is a workaround for MGMT-8707, until we get the fixed RPMs to land in CentOS / RHEL 8.x

### DIFF
--- a/scripts/install_environment.sh
+++ b/scripts/install_environment.sh
@@ -21,9 +21,45 @@ function install_libvirt() {
         libvirt-devel \
         libvirt-daemon-kvm \
         qemu-kvm \
-        libgcrypt \
-        swtpm \
-        swtpm-tools
+        libgcrypt
+
+    # TODO: get swtpm RPMs from EPEL as soon as 0.7.1 or later lands there.
+    # based on: https://github.com/stefanberger/swtpm/wiki#compile-and-install-on-linux
+
+    echo "Enable EPEL"
+    sudo dnf install -y  epel-release
+
+    echo "Install swtpm dependencies"
+    sudo dnf install -y \
+        libtasn1-devel \
+        expect \
+        socat \
+        python3-twisted \
+        fuse-devel \
+        glib2-devel \
+        gnutls-devel \
+        gnutls-utils \
+        gnutls \
+        json-glib-devel \
+        autoconf \
+        automake \
+        libtool \
+        openssl-devel \
+        libtpms \
+        libtpms-devel \
+        libseccomp-devel \
+        net-tools \
+        iproute \
+        git \
+        make
+
+    echo "Install swtpm v0.7.1 from source"
+    git clone --depth 1 --branch v0.7.1 https://github.com/stefanberger/swtpm.git
+    pushd swtpm
+    ./autogen.sh --with-openssl --prefix=/usr
+    make -j4
+    make install
+    popd
 
     sudo systemctl enable libvirtd
 


scripts/install_environment.sh
```diff
@@ -16,14 +16,15 @@ function install_libvirt() {
...
+    # TODO: get swtpm RPMs from EPEL as soon as 0.7.1 or later lands there.
     sudo dnf install -y \
         libvirt \
         libvirt-devel \
         libvirt-daemon-kvm \
         qemu-kvm \
         libgcrypt \
-        swtpm \
-        swtpm-tools
+        https://download-ib01.fedoraproject.org/pub/fedora/linux/updates/35/Everything/x86_64/Packages/s/swtpm-0.7.2-1.20220307git21c90c1.fc35.x86_64.rpm \
```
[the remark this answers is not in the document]
nmagnezi: Can do, but how does curl makes a difference here?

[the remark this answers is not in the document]
osherdp: curl with retry mechanism can make a difference
Or just curl --retry 5 (although it doesn't retry on non-transient errors)

nmagnezi:
this turned out to be an endless dependency chain..
I'll try a different approach

scripts/install_environment.sh
```diff
@@ -21,9 +21,37 @@ function install_libvirt() {
...
         libvirt-daemon-kvm \
         qemu-kvm \
-        libgcrypt \
-        swtpm \
-        swtpm-tools
+        libgcrypt
+
+    # TODO: get swtpm RPMs from EPEL as soon as 0.7.1 or later lands there.
+    # based on: https://github.com/stefanberger/swtpm/wiki#compile-and-install-on-linux
+   dnf install -y \
```
osherdp, line 28: interesting
if it doesn't add too much time I wouldn't mind going that direction

nmagnezi: worked for me locally but seems to fail here. I suspect it might be missing EPEL. looking into it.

scripts/install_environment.sh
```diff
@@ -21,9 +21,43 @@ function install_libvirt() {
...
+        net-tools \
+        iproute \
+        git \
+        make
+
+    echo "Install swtpm v0.7.1 from source"
+    git clone --depth 1 --branch v0.7.1 https://github.com/stefanberger/swtpm.git
+    pushd swtpm
+    ./autogen.sh --with-openssl --prefix=/usr
+    popd
```
osherdp, line 60: how much time did it add this whole installation process?
because we might consider doing it specifically for tpm flow

osherdp (changes requested):

scripts/install_environment.sh
nmagnezi: checking

scripts/install_environment.sh
```diff
@@ -21,9 +21,43 @@ function install_libvirt() {
...
+        libseccomp-devel \
+        net-tools \
+        iproute \
+        git \
+        make
+
+    echo "Install swtpm v0.7.1 from source"
+    git clone --depth 1 --branch v0.7.1 https://github.com/stefanberger/swtpm.git
+    pushd swtpm
+    ./autogen.sh --with-openssl --prefix=/usr
```
adriengentil, line 59: don't we need `make` + `make install` as described in the doc?

also, I see that `swtmp` packages are still installed in the build log: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_assisted-test-infra/1748/pull-ci-openshift-assisted-test-infra-master-e2e-metal-assisted/1547106888800276480/artifacts/e2e-metal-assisted/assisted-baremetal-setup/build-log.txt

osherdp: how is that possible?

nmagnezi: Added those.
as for why the RPMs are still installed - I'm not sure. I don't see any other places in the code that mention those RPMs.
@osherdp @eliorerz , do you know?

scripts/install_environment.sh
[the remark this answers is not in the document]
nmagnezi: Closing this one, as I moved to install from source.

scripts/install_environment.sh
nmagnezi: Indeed, EPEL was the issue here.

scripts/install_environment.sh
nmagnezi: based on the job history, [job duration for successful runs](https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-assisted-test-infra-master-e2e-metal-assisted-tpmv2) was not increased.

scripts/install_environment.sh
nmagnezi: So as @adriengentil mentioned over Slack, swtpm gets installed as a dependency of [libvirt-daemon-driver-qemu](https://pkgs.org/download/libvirt-daemon-driver-qemu).

I have tested `install_environment.sh` locally on a RockyLinux 8.6 container, and the install from source procedure seems to override the older swtpm binaries:


**check done after swtpm-0.7.0 got installed via RPM**
```sh
[root@5091c205cc54 /]# swtpm --version
TPM emulator version 0.7.0, Copyright (c) 2014-2021 IBM Corp.
[root@5091c205cc54 /]# which swtpm
/usr/bin/swtpm


[root@5091c205cc54 /]# /usr/bin/swtpm_ioctl --version
TPM emulator control tool version 0.7.0, Copyright (c) 2015 IBM Corp.
```

**a check done after swtpm-0.7.1 got installed from source**

```sh
[root@5091c205cc54 /]# /usr/bin/swtpm_ioctl --version
TPM emulator control tool version 0.7.1, Copyright (c) 2015 IBM Corp.
[root@5091c205cc54 /]# swtpm --version
TPM emulator version 0.7.1, Copyright (c) 2014-2021 IBM Corp
```

can we continue with that approach?

scripts/install_environment.sh
nmagnezi: the added 'time dnf install' commands (enable EPEL + install RPM deps) did not show the cmd duration; however, the swtpm install commands did:

swtpm install commands:
real	0m1.163s
real	0m8.621s
real	0m5.370s
real	0m0.607s

15.758 seconds in total.

osherdp (approved):

PR #1756, opened 2022-07-14
Allow usage of AlmaLinux OS in assisted-test-infra

osherdp:
This allows AlmaLinux as a backup plan so that we'll be able to use it if Rocky fails us as a CI candidate.
I tested full installation with AlmaLinux 8 via Equinix, it passed successfully.

### DIFF
--- a/GUIDE.md
+++ b/GUIDE.md
@@ -2,7 +2,7 @@
 
 ## Prerequisites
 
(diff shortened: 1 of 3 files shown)


eliorerz:
Have you tried running the base flow (test_install) with `AlmaLinux` instead of Rocky?

osherdp:
Yes, see description 😄

eliorerz (approved):
LGTM

PR #1759, opened 2022-07-17
MGMT-11319 - Remove redundant `destroy_nodes` when preparing nodes

eliorerz:
Currently we are running the "nodes  cleanup" on every test regardless the nodes status.
Each cluster name is unique so as the terraform configuration folder so there shouldn't be a case where a cluster is created with the same name.
Furthermore, we should not prepare the nodes twice in the same test. If for some reason we want to do it, we can always do it manually, e.g.

```python
...
nodes.prepare_nodes()
...
nodes.destroy_all_nodes()
nodes.prepare_nodes()
...
```

/cc @osherdp

### DIFF
--- a/src/tests/test_day2.py
+++ b/src/tests/test_day2.py
@@ -11,6 +11,8 @@ class TestDay2(BaseTest):
     @JunitTestSuite()
     def test_deploy_day2_nodes_cloud(self, cluster, day2_cluster, controller):
         if not global_variables.cluster_id:
+            cluster.nodes.destroy_all_nodes()
+            cluster.nodes.prepare_nodes()
             cluster.prepare_for_installation()
             cluster.start_install_and_wait_for_installed()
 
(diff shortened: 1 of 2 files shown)


osherdp (approved):

eliorerz:
See me surprised that the day2 test is failing :|

osherdp (approved):

lalon4:
@eliajahshan can we lgtm? do we need any update on kni?

src/tests/test_day2.py
```diff
@@ -11,6 +11,8 @@ class TestDay2(BaseTest):
     @JunitTestSuite()
     def test_deploy_day2_nodes_cloud(self, cluster, day2_cluster, controller):
         if not global_variables.cluster_id:
+            cluster.nodes.destroy_all_nodes()
```
lalon4, line 14: @eliorerz why we need the destroy here?

eliorerz: Because this flow isn't using the global resources.  It will be fixed as part of the day2 pr

PR #1761, opened 2022-07-18
NO-ISSUE - Allow KeyboardInterrupt during download logs

eliorerz:
Allow the user to skip on the download logs part using KeyboardInterrupt signal.
At the current situation, the test is to trying to download the logs 3 times with 300 seconds sleep between them. Pressing sending KeyboardInterrupt signal is not always working (on sleep) and the test is exit without a proper cleanup.
Also removing the unused `pull_secret` field from `download_logs` function.

/cc @osherdp

### DIFF
--- a/src/assisted_test_infra/download_logs/__main__.py
+++ b/src/assisted_test_infra/download_logs/__main__.py
@@ -56,7 +56,6 @@ def main():
             args.dest,
             args.must_gather,
             args.update_by_events,
-            pull_secret=args.pull_secret,
         )
     else:
         clusters = get_clusters(client, args.download_all)
@@ -67,9 +66,7 @@ def main():
 
         for cluster in clusters:
             if args.download_all or should_download_logs(cluster):
-                download_cluster_logs(
-                    client, cluster, args.dest, args.must_gather, args.update_by_events, pull_secret=args.pull_secret
-                )
+                download_cluster_logs(client, cluster, args.dest, args.must_gather, args.update_by_events)
 
         log.info("Cluster installation statuses: %s", dict(Counter(cluster["status"] for cluster in clusters).items()))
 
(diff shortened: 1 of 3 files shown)


src/assisted_test_infra/download_logs/__main__.py
```diff
@@ -56,7 +56,6 @@ def main():
             args.dest,
             args.must_gather,
             args.update_by_events,
-            pull_secret=args.pull_secret,
```
osherdp, line 59: This field is unused?
If so, can you include explanation for those changes in PR description?

src/assisted_test_infra/download_logs/download_logs.py
```diff
@@ -200,7 +198,8 @@ def download_logs(
                     # Skip sleeping on last retry
                     if i < MAX_RETRIES - 1:
                         log.info(f"Going to retry in {retry_interval} seconds")
-                        time.sleep(retry_interval)
+
+                    [time.sleep(1) for _ in range(retry_interval)]  # allow Ctrl+C when running locally
```
osherdp, line 202: when not using value of the list-comprehension, I prefer:
```suggestion
                    # allow Ctrl+C when running locally
                    for _ in range(retry_interval):
                        time.sleep(1)
```

osherdp, line 202: why do we need this?
if I'm doing ``python -c "import time; time.sleep(20)"``` I have no problem killing it

osherdp (changes requested):

src/assisted_test_infra/download_logs/__main__.py
eliorerz: It's unused. During this change I noticed that the pull_secret is grayed out so I removed it.
I'll add it to the PR description.

src/assisted_test_infra/download_logs/download_logs.py
eliorerz: My mistake, the KeyboardInterrupt does kill the sleep program (but I don't think that the waiting package)

osherdp (approved):

PR #1764, opened 2022-07-19
NO-ISSUE - Prevent from using the same assisted-service image while using DEBUG_SERVICE flag

eliorerz:
/cc @osherdp @michaellevy101

### DIFF
--- a/scripts/deploy_assisted_service.sh
+++ b/scripts/deploy_assisted_service.sh
@@ -126,6 +126,13 @@ else
         DEBUG_DEPLOY_AI_PARAMS="REPLICAS_COUNT=1"
         # Override the SERVICE environment variable with the local registry debug image
         export SERVICE="${SUBSYSTEM_LOCAL_REGISTRY}/assisted-service:latest"
+
+        # Force removing the debugable image before re-deploying the service
+        # Change replicas to 0 so when applying the deployment it will identify the changes and allow deleting the
+        # image if it's already in use in the current pods
+        kubectl scale -n assisted-installer --replicas=0 deployment assisted-service || true
+        kubectl wait --for=delete -n assisted-installer pod $(kubectl -n assisted-installer get pods | grep assisted-service | awk '{print $1}')
+        minikube image rm "${SERVICE}" || true
     fi
 
     skipper run src/update_assisted_service_cm.py


scripts/deploy_assisted_service.sh
```diff
@@ -126,6 +126,13 @@ else
         DEBUG_DEPLOY_AI_PARAMS="REPLICAS_COUNT=1"
         # Override the SERVICE environment variable with the local registry debug image
         export SERVICE="${SUBSYSTEM_LOCAL_REGISTRY}/assisted-service:latest"
+
+        # Force removing the debugable image before re-deploying the service
+        # Change replicas to 0 so when applying the deployment it will identify the changes and allow deleting the
+        # image if it's already in use in the current pods
+        kubectl scale -n assisted-installer --replicas=0 deployment assisted-service || true
```
osherdp, line 133: in which case would you like to accept failure of this operation?

scripts/deploy_assisted_service.sh
```diff
@@ -126,6 +126,13 @@ else
...
+        sleep 10
+        minikube image rm "${SERVICE}" || true
```
osherdp, line 135: same question as above

scripts/deploy_assisted_service.sh
```diff
@@ -126,6 +126,13 @@ else
         DEBUG_DEPLOY_AI_PARAMS="REPLICAS_COUNT=1"
         # Override the SERVICE environment variable with the local registry debug image
         export SERVICE="${SUBSYSTEM_LOCAL_REGISTRY}/assisted-service:latest"
+
+        # Force removing the debugable image before re-deploying the service
+        # Change replicas to 0 so when applying the deployment it will identify the changes and allow deleting the
+        # image if it's already in use in the current pods
+        kubectl scale -n assisted-installer --replicas=0 deployment assisted-service || true
+        sleep 10
```
osherdp, line 134: can we leverage ``kubectl wait --for=delete`` instead of sleeping?

osherdp (changes requested):

scripts/deploy_assisted_service.sh
eliorerz: Never. it will probably fail only if this is the first deployment, so in this case we want to ignore this failure

scripts/deploy_assisted_service.sh
eliorerz: Same answer as above, if the image isn't exist we want to ignore the failure

scripts/deploy_assisted_service.sh
eliorerz: Good idea. Thanks :)

osherdp (approved):

PR #1766, opened 2022-07-22
WIP NO-ISSUE: Define a fully dynamic IP range

adriengentil:
In order to avoid IP collisions, start the dhcp range from `.128` to
`.254`. Statically assigned IPs usually use the lower range:

- `.100` and `.101` addresses are sometime used for the API and ingress
  VIPs
- we statically map the mac addresses of the master/worker VMs from
  `.10`


adriengentil:
For the context this PR tries to solves the issue that happened in this job: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-assisted-test-infra-release-ocm-2.5-e2e-metal-assisted-day2-periodic/1549890537048248320

The DHCP server assigned `192.168.127.15` to the API VIP, the DNS to api was configured to points to this IP.
Then, we spawned a worker, it took `192.168.127.15` because we configured the DHCP server to do so. As a consequence, the DHCP server updated the API VIP to `192.168.127.17` but the DNS was not updated which led to a failure of the job (the worker was not able to contact the API server).

/cc @eliorerz @osherdp

adriengentil:
BTW, it looks like the DNS to the ingress VIP in day2 job is not right, it is set to `192.168.127.101` instead of `192.168.127.226`. But it doesn't seem to make the job fail.

terraform_files/baremetal/limit_ip_dhcp_range.xsl
```diff
@@ -12,11 +12,11 @@
...
-        <!-- Transform end range of e.g. "192.168.122.254" to "192.168.122.90" -->
-        <xsl:value-of select="concat(substring-before(@end,'.254'),'.90')" />
+      <xsl:attribute name="start">
+        <!-- Transform start range of e.g. "192.168.122.2" to "192.168.122.128" -->
+        <xsl:value-of select="concat(substring-before(@end,'.2'),'.128')" />
       </xsl:attribute>
-      <xsl:apply-templates select="@*[not(local-name()='end')]|node()"/>
+      <xsl:apply-templates select="@*[not(local-name()='start')]|node()"/>
     </xsl:copy>
   </xsl:template>
```
osherdp, line 22: similar handling should be apply to the ipv6 case I assume (down below)

terraform_files/baremetal_infra_env/limit_ip_dhcp_range.xsl
```diff
@@ -12,11 +12,11 @@
   <xsl:template match="/network/ip[@family='ipv4']/dhcp/range">
```
osherdp, line 12: unless I'm missing something, I see no reason not to apply it to the none-platform template as well right?

osherdp: btw if the answer above is true we can have just 1 instance of that ``xsl`` file, place it under ``terraform_files`` and reference it like ``xslt = file("../limit_ip_dhcp_range.xsl")``

osherdp:
> For the context this PR tries to solves the issue that happened in this job: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-assisted-test-infra-release-ocm-2.5-e2e-metal-assisted-day2-periodic/1549890537048248320
> 
> The DHCP server assigned `192.168.127.15` to the API VIP, the DNS to api was configured to points to this IP. Then, we spawned a worker, it took `192.168.127.15` because we configured the DHCP server to do so. As a consequence, the DHCP server updated the API VIP to `192.168.127.17` but the DNS was not updated which led to a failure of the job (the worker was not able to contact the API server).
> 
> /cc @eliorerz @osherdp

So the domain server has not really been reserving that IP for just one entity?
Doesn't it mean we might have it on other places? (the domain server returning an IP to a master node and then returning the same one to a worker node)

osherdp (changes requested):

adriengentil:
> > For the context this PR tries to solves the issue that happened in this job: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-assisted-test-infra-release-ocm-2.5-e2e-metal-assisted-day2-periodic/1549890537048248320
> > The DHCP server assigned `192.168.127.15` to the API VIP, the DNS to api was configured to points to this IP. Then, we spawned a worker, it took `192.168.127.15` because we configured the DHCP server to do so. As a consequence, the DHCP server updated the API VIP to `192.168.127.17` but the DNS was not updated which led to a failure of the job (the worker was not able to contact the API server).
> > /cc @eliorerz @osherdp
> 
> So the domain server has not really been reserving that IP for just one entity? Doesn't it mean we might have it on other places? (the domain server returning an IP to a master node and then returning the same one to a worker node)

I guess we don't know the MAC address of the API VIP in order to reserve it when we create the nodes, it's something random, no?

terraform_files/baremetal/limit_ip_dhcp_range.xsl
adriengentil: Ok, I'll have a look, I didn't know if was working in the same way

adriengentil:
> Right, it's unknown Something (bootstrap's agent maybe? I'm not sure) requests API/ingress IPs from DHCP, and it's very possible that the lease time is too short and those requests are becoming invalidated

lease time is 1h by default. Otherwise, we should not rely on dhcp for the VIPs, and always set them to 192.168.127.100/192.168.127.101, is using DHCP for the tests is expected?

osherdp:
> > Right, it's unknown Something (bootstrap's agent maybe? I'm not sure) requests API/ingress IPs from DHCP, and it's very possible that the lease time is too short and those requests are becoming invalidated
> 
> lease time is 1h by default. Otherwise, we should not rely on dhcp for the VIPs, and always set them to 192.168.127.100/192.168.127.101, is using DHCP for the tests expected?

We have two modes based on what the user has selected (statically setting 100+101 or asking DHCP to dynamically choose them). It should be configured with the vip_dhcp_allocation field
For some flows, we can only have one mode available (we did not implement dhcp mode for ipv6 clusters)

terraform_files/baremetal/limit_ip_dhcp_range.xsl
adriengentil: We don't do any reservation in IPv6 so it shouldn't be affected

terraform_files/baremetal_infra_env/limit_ip_dhcp_range.xsl
adriengentil: Done

terraform_files/baremetal/limit_ip_dhcp_range.xsl
osherdp: ah right

PR #1767, opened 2022-07-25
Support podman-remote

michaellevy101:
This PR supports podman-remote for test-infra, now the images will be visible to the host container tool.

### DIFF
--- a/scripts/install_environment.sh
+++ b/scripts/install_environment.sh
@@ -3,6 +3,7 @@ set -euo pipefail
 
 export EXTERNAL_PORT=${EXTERNAL_PORT:-y}
 export ADD_USER_TO_SUDO=${ADD_USER_TO_SUDO:-n}
+readonly PODMAN_MINIMUM_VERSION="3.2.0"
 
 function version_is_greater() {
     if [ "$(head -n1 <(printf '%s\n' "$2" "$1" | sort -V))" = "$2" ]; then
@@ -146,20 +147,40 @@ EOF
     chmod +x "${hook_filename}"
 }
 
+function install_podman(){
+    sudo systemctl disable --now podman.socket || true
+    sudo rm -rf /run/user/${UID}/podman
+    sudo rm -rf /run/podman
+    sudo dnf install podman -y
+    sudo systemctl enable --now podman.socket
+    systemctl --user enable --now podman.socket
+    sudo loginctl enable-linger $USER
+}
+
 function install_runtime_container() {
     echo "Installing container runtime package"
-    if ! [ -x "$(command -v docker)" ] && ! [ -x "$(command -v podman)" ]; then
-        sudo dnf install podman -y
-    elif [ -x "$(command -v podman)" ]; then
-        current_version="$(head -n1 <(podman version) | awk '{print $2}')"
-        minimum_version="1.6.4"
-        if ! version_is_greater "$current_version" "$minimum_version"; then
-            sudo dnf install podman-$minimum_version -y
-        fi
-    else
-        echo "docker or podman is already installed"
+    if [ -x "$(command -v docker)" ]; then
+        echo "docker is already installed"
+        return
     fi
 
+    if [ ! -x "$(command -v podman)" ]; then
+        install_podman
+        current_version="$(head -n1 <(podman version) | awk '{print $2}')"
+    else
+        current_version="$(head -n1 <(podman version) | awk '{print $2}')"
+
+        if [! version_is_greater "$current_version" "$PODMAN_MINIMUM_VERSION" || ! systemctl is-active --quiet service ]; then
+            install_podman
+            current_version="$(head -n1 <(podman version) | awk '{print $2}')"
+        else
+            echo "podman is already installed"
+        fi
+    fi
+
+    if ! version_is_greater "$current_version" "$PODMAN_MINIMUM_VERSION"; then
+        echo "podman version ($current_version) is too old and might not work as expected"
+    fi
 }
 
 function install_packages() {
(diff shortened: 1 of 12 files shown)


scripts/run_minikube.sh
```diff
@@ -26,7 +26,7 @@ function init_minikube() {
     for i in {1..5}
     do
         minikube delete
-        minikube start --driver=kvm2 --memory="${MINIKUBE_RAM_MB}" --cpus=4 --force --wait-timeout=15m0s --disk-size="${MINIKUBE_DISK_SIZE}" --addons=registry || true
+        minikube start --driver=kvm2 --listen-address=0.0.0.0 --memory="${MINIKUBE_RAM_MB}" --cpus=4 --force --wait-timeout=15m0s --disk-size="${MINIKUBE_DISK_SIZE}" --addons=registry || true
```
adriengentil, line 29:
```
    --listen-address='':
        IP Address to use to expose ports (docker and podman driver only)
```
should it be cleaned up?

scripts/utils.sh
```diff
@@ -185,4 +185,49 @@ function configure_none_platform_iptables_rules() {
...
 }
 
+function running_from_skipper() {
+   # The SKIPPER_UID environment variable is an indication that we are running on a skipper container.
+   [ -n "${SKIPPER_UID+x}" ]
+}
+
+function get_container_runtime_command() {
+  # if CONTAINER_TOOL is defined skipping
+  if [ ! -n "${CONTAINER_TOOL+x}" ]; then
```
adriengentil, line 195:
```suggestion
  if [ -z "${CONTAINER_TOOL+x}" ]; then
```
looks clearer

scripts/utils.sh
```diff
@@ -185,4 +185,49 @@ function configure_none_platform_iptables_rules() {
...
+    if running_from_skipper; then
+      if [ -z ${CONTAINER_RUNTIME_COMMAND} ]; then
+        echo "CONTAINER_RUNTIME_COMMAND doesn't set on old skipper version -> default to docker. Upgrade your skipper to the latest version"
+      fi
+
+      if [ "${CONTAINER_RUNTIME_COMMAND:-docker}" == "docker" ]; then
```
adriengentil, line 201:
```suggestion
      if [ "${CONTAINER_RUNTIME_COMMAND:-docker}" = "docker" ]; then
```

scripts/install_environment.sh
```diff
@@ -146,15 +147,25 @@ EOF
     chmod +x "${hook_filename}"
 }
 
+function install_podman(){
```
adriengentil, line 150: nit: for consistency, use the same indentation level as in the rest of the file

scripts/utils.sh
```diff
@@ -185,4 +185,49 @@ function configure_none_platform_iptables_rules() {
...
+    exit
+  fi
+
+  if [ "$(get_container_runtime_command)" = "podman-remote" ]; then
+    if podman-remote4 info 2>&1 | grep "server API version is too old" &> /dev/null; then
+      echo "using podman-remote version 3"
+      ln $(which podman-remote3) /tools/podman-remote
+    else
+      echo "using podman-remote version 4"
+      ln $(which podman-remote4) /tools/podman-remote
```
adriengentil, line 228: what is `/tools` directory? it is part of skipper magic?

adriengentil (changes requested):

michaellevy101:
@adriengentil It's not ready yet

scripts/run_minikube.sh
michaellevy101: yes, This PR isn't ready yet

scripts/utils.sh
michaellevy101: no, should be create as part of the image

scripts/install_environment.sh
```diff
@@ -146,20 +147,40 @@ EOF
...
-    if ! [ -x "$(command -v docker)" ] && ! [ -x "$(command -v podman)" ]; then
-        sudo dnf install podman -y
-    elif [ -x "$(command -v podman)" ]; then
-        current_version="$(head -n1 <(podman version) | awk '{print $2}')"
-        minimum_version="1.6.4"
-        if ! version_is_greater "$current_version" "$minimum_version"; then
-            sudo dnf install podman-$minimum_version -y
-        fi
+    if [ -x "$(command -v docker)" ]; then
+      echo "docker is already installed"
```
osherdp, line 163: fix indentation to be 4 spaces like the rest of this file

scripts/install_environment.sh
```diff
@@ -146,20 +147,40 @@ EOF
...
+      return
+    fi
+
+    if [ ! -x "$(command -v podman)" ]; then
+      install_podman
```
osherdp, line 168: if on some environment (and I think it's most local environments) we have podman installed but without the podman socket, then the podman socket will not be configured
maybe we should check if there's an active socket and based on that do or don't do the socket configuration

skipper.yaml
```diff
@@ -36,8 +36,14 @@ volumes:
   - /dev/:/dev
   - /run/udev:/run/udev
 
+  # podman - sharing the podman.socket between the host and the skipper container
+  - $XDG_RUNTIME_DIR/podman/podman.sock:/run/podman/podman.sock
```
osherdp, line 40: in assisted-service we have ``/run/podman/podman.sock:/run/podman/podman.sock``. what is it different?

osherdp (changes requested):

src/consts/consts.py
```diff
@@ -125,7 +125,6 @@ def all(cls):
 WORKER_TF_CPU_MODE = HOST_PASSTHROUGH_CPU_MODE
 NUMBER_OF_MASTERS = 3
 NAMESPACE_POOL_SIZE = 15
-PODMAN_FLAGS = "--cgroup-manager=cgroupfs --storage-driver=vfs --events-backend=file"
```
eliorerz, line 128: I just saw `--cgroup-manager=cgroupfs --storage-driver=vfs --events-backend=file` is hard-coded in https://github.com/openshift/assisted-test-infra/blob/c5d5a2867a6f552d250e39e0b3513c8fdfd33750/src/assisted_test_infra/test_infra/controllers/tang_controller/tang_controller.py#L38

skipper.yaml
michaellevy101: It depends on the podman-socket installation.

eliorerz (approved):
LGTM
/hold

PR #1772, opened 2022-07-27
NO-ISSUE - Change test-infra image to assisted-test-infra

eliorerz:
Change the test-infra image to be the same as in the dockerfile: Dockerfile.assisted-test-infra

/cc @osherdp


Makefile
```diff
@@ -133,6 +133,7 @@ image_build:
 	scripts/pull_dockerfile_images.sh
 	sed 's/^FROM .*assisted-service.*:latest/FROM $(subst /,\/,${SERVICE})/' Dockerfile.assisted-test-infra | \
 	 $(CONTAINER_COMMAND) build --network=host -t $(IMAGE_NAME) -f- .
+# 	$(CONTAINER_COMMAND) tag $(IMAGE_NAME) test-infra:latest  # For backwards computability
```
osherdp, line 136: backwards compatibility in local envs or something else?

osherdp (changes requested):

Makefile
eliorerz: I need to uncomment this line, it will tag the `assisted-test-infra:latest`  to `test-infra:latest`.
I just wanted to run the test using the new image name

eliorerz: But basically local envs

.flake8
```diff
@@ -17,3 +17,5 @@ exclude =
     src/consts/__init__.py
     src/assisted_test_infra/test_infra/utils/__init__.py
     .git
+    assisted-service/
+    swtpm/
```
osherdp, line 21: this can be removed now

Makefile
```diff
@@ -133,6 +133,7 @@ image_build:
 	scripts/pull_dockerfile_images.sh
 	sed 's/^FROM .*assisted-service.*:latest/FROM $(subst /,\/,${SERVICE})/' Dockerfile.assisted-test-infra | \
 	 $(CONTAINER_COMMAND) build --network=host -t $(IMAGE_NAME) -f- .
```
osherdp, line 135: A bit nitty, but can you change it to 2*tab instead of tab+space?
At first I wasn't aware it's a continuation of the previous line

osherdp (changes requested):

osherdp (approved):
please drop a message for qe as a FYI in case something breaks/changes

PR #1774, opened 2022-07-28
Fix EPEL installation for RHEL and CentOS

osherdp:
Existing code is good for Rocky Linux and Alma Linux, but for RHEL we'll need to install it differently.

/cc @nmagnezi @eranco74


eranco74 (commented):
/lgtm

PR #1776, opened 2022-07-28
Fix podman verification code

osherdp:
Fixing a couple of issues:
* Bash syntax was completely off. I get locally:
```
./scripts/install_environment.sh: line 173: [!: command not found
Failed to retrieve unit state: Unit name ].service is neither a valid
invocation ID nor unit name
```
* Getting the podman version does not work. ``head -n1 <(podman version) | awk '{print $2}'`` returns ``Podman`` for me...
* I simplified the branching a bit to make code look better.

/cc @michaellevy101 @eliorerz 
/hold
(not completely tested yet)

### DIFF
--- a/scripts/install_environment.sh
+++ b/scripts/install_environment.sh
@@ -179,22 +179,22 @@ function install_runtime_container() {
         return
     fi
 
-    if [ ! -x "$(command -v podman)" ]; then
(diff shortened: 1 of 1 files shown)


eliorerz (approved):
LGTM

PR #1783, opened 2022-08-03
install podman from 4.0 stream on RHEL 8 machines

osherdp:
Handling several issues related to a QE environment that had an outdated (v2) version of podman:
* When installing podman, use ``--best`` to allow it to upgrade the package if that's possible.
* For RHEL 8, enable ``container-tools:4.0`` in order to be able to install later podman versions than v2.x.

### DIFF
--- a/scripts/install_environment.sh
+++ b/scripts/install_environment.sh
@@ -13,8 +13,28 @@ function version_is_greater() {
     false
 }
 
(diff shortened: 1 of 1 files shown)


osherdp:
/hold
waiting for the tests

eliorerz (approved):
LGTM

PR #1784, opened 2022-08-03
MGMT-11491: kubeAPI flow should work with no default OCP version

nmagnezi:
As detailed in Jira, the lack of a default OCP version fails this flow.


eliorerz (approved):
LGTM

PR #1785, opened 2022-08-04
MGMT-11449: govc installation from GitHub gets rate-limiter failures

adriengentil:
Install govc from the container registry instead of relying on github.


adriengentil:
/hold

until the mapping in openshift/release is done.

Dockerfile.assisted-test-infra
```diff
@@ -38,6 +37,8 @@ RUN curl --retry 5 -L https://github.com/containers/podman/releases/download/v3.
     mv podman-remote-static /tools/podman-remote4 && \
     rm -f /tmp/podman-remote4.tar.gz
```
adriengentil, line 38: :point_up: we will likely get the same issue with `podman-remote`

osherdp (approved):

adriengentil:
/hold
oups, I want to wait for a build

PR #1786, opened 2022-08-08
MGMT-11415: add platform support to kubeapi test

danielerez:
In order to support vSphere platform on KubeAPI test introduced the following changes:
* Added 'platform_type' to ACI creation on kube_api_test
* VSphereController: get_ingress_and_api_vips func should return api_vip/ingress_vip params as dhcp is not supported using kube-api.
* Ensured _set_host_name_from_node logic is performed for both ipv4 and ipv6 (as setting the hostname is always required on vSphere).
* Removed unused set_agents_hostnames and set_agent_hostname funcs.

Note: pending on https://github.com/openshift/assisted-service/pull/4219

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,4 @@
 storage_pool
 **/*.iso
+**/minikube*
 
--- a/src/assisted_test_infra/test_infra/controllers/node_controllers/vsphere_controller.py
+++ b/src/assisted_test_infra/test_infra/controllers/node_controllers/vsphere_controller.py
@@ -1,3 +1,4 @@
+import os
 import time
 from builtins import list
 from typing import Any, Callable, Dict, List, Tuple
@@ -25,6 +26,10 @@ class VSphereController(NodeController):
         self._tf = terraform_utils.TerraformUtils(working_dir=folder, terraform_init=False)
 
     def prepare_nodes(self):
+        if not os.path.exists(self._entity_config.iso_download_path):
+            utils.recreate_folder(os.path.dirname(self._entity_config.iso_download_path), force_recreate=False)
+            # if file not exist lets create dummy
+            utils.touch(self._entity_config.iso_download_path)
         config = {**self._config.get_all(), **self._entity_config.get_all(), "cluster_name": self.cluster_name}
         # The ISO file isn't available now until preparing for installation
         del config["iso_download_path"]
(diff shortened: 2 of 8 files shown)


.dockerignore
```diff
@@ -1,3 +1,4 @@
 storage_pool
 **/*.iso
+**/minikube*
```
osherdp, line 3: maybe change to 
```suggestion
minikube*
```
I wouldn't want any ``minikube*.py`` module we might want to add to be ignored

src/consts/consts.py
```diff
@@ -294,6 +294,16 @@ class Platforms:
     VSPHERE = "vsphere"
 
 
+class PlatformTypes:
```
osherdp, line 297:
```suggestion
class PlatformType:
```

I know some of those classes in this file are following either singular or plural, I think the former is more readable and ultimately we should move other classes to this form

osherdp (changes requested):

eliorerz (changes requested):

src/assisted_test_infra/test_infra/utils/kubeapi_utils.py
```diff
@@ -106,6 +86,21 @@ def get_hostname_for_agent(agent, nodes_details):
                 return node_metadata["name"]
 
 
+def get_platform_type(platform):
```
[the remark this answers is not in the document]
danielerez: sure, done

src/consts/consts.py
danielerez: The problem is the the service expects the new format (i.e. VSphere), so platform.lower() won't work when passing it to the service. That's why we need the conversion method between the formats. Can we do it nicer?

.dockerignore
danielerez: we're adding the entire minikube_home folder, so it should already be ignored (/home/assisted-test-infra/minikube_home/.minikube/machines/minikube/minikube.rawdisk)

src/assisted_test_infra/test_infra/controllers/node_controllers/vsphere_controller.py
```diff
@@ -21,10 +22,16 @@ class VSphereController(NodeController):
     def __init__(self, config: BaseVSphereConfig, cluster_config: BaseClusterConfig):
         super().__init__(config, cluster_config)
         self.cluster_name = cluster_config.cluster_name.get()
+        self.api_vip = cluster_config.api_vip
+        self.ingress_vip = cluster_config.ingress_vip
```
eliorerz, line 26: Missed that last review. please remove it and return it using `get_ingress_and_api_vips`

danielerez: sure, changed.

osherdp (approved):

PR #1797, opened 2022-08-17
MGMT-11337: Define wildcard DNS for ingress DNSes

osherdp:
The job e2e-metal-assisted-day2-single-node currently fails with the following error:

```
2022-07-22 05:43:31,208 - root - INFO - 139970722252608 - Asked hosts to
be in one of the statuses from ['known'] and currently hosts statuses
are [(1, 'eecb5396-0b83-422b-ae31-7418fee62663',
'test-infra-cluster-c717123a-worker-0', 'worker', 'insufficient', "Host
cannot be installed due to following failing validation(s): Couldn't
resolve domain name *.apps.test-infra-cluster-c717123a.redhat.com on the
host. To continue installation, create the necessary DNS entries to
resolve this domain name to your cluster's application ingress IP
address")]
```

assisted-test-infra does not currently create the wildcard DNS for the cluster because the terraform provider did not support that up until now.

Since the provider now supports the option, we should now be able to create the wildcard DNS using custom dnsmasq options.

In this change, I did not try to overcome the following problems I find in the current design:
* Having an actual wildcard DNS matching in the infra. It requires handling something that I yet to overcome which is related to making the libvirt dnsmasq server an authorized DNS server. Maybe a further refactoring PR will be able to actually remove the redundant DNS entries *.apps...
* We have some other duplications I want to remove (e.g. https://github.com/openshift/assisted-test-infra/blob/708d79aa066fface6f52dc274d1c6e2d7dcfe193/src/assisted_test_infra/test_infra/controllers/node_controllers/terraform_controller.py#L165). We'll have to handle it some other time because I wouldn't want this PR to get any bigger.
* Remove the need for ``single_node_ip`` variable. We can do just fine with ``api/ingress_vip``.

### DIFF
--- a/src/assisted_test_infra/test_infra/controllers/node_controllers/libvirt_controller.py
+++ b/src/assisted_test_infra/test_infra/controllers/node_controllers/libvirt_controller.py
@@ -607,7 +607,7 @@ class LibvirtController(NodeController, ABC):
     def format_node_disk(self, node_name: str, disk_index: int = 0) -> None:
         raise NotImplementedError
 
-    def get_ingress_and_api_vips(self) -> dict:
+    def get_ingress_and_api_vips(self, is_highly_available: bool = False) -> dict:
         raise NotImplementedError
 
     def get_cluster_network(self) -> str:
--- a/src/assisted_test_infra/test_infra/controllers/node_controllers/node_controller.py
+++ b/src/assisted_test_infra/test_infra/controllers/node_controllers/node_controller.py
@@ -102,7 +102,7 @@ class NodeController(ABC):
         pass
 
     @abstractmethod
-    def get_ingress_and_api_vips(self) -> dict:
+    def get_ingress_and_api_vips(self, is_highly_available: bool = False) -> dict:
         pass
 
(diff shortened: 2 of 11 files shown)


adriengentil:
/lgtm
looks like the kube-api-net-suite fails for an unrelated reason

eliorerz:
@lalon4 This PR breaks 3 tests on `kni-assisted-installer-auto` due to signature change of `get_ingress_and_api_vips`:
- `test_static_ips`
- `test_cancel_and_retest`
-  `test_scaleup`

osherdp:
> @lalon4 This PR breaks 3 tests on `kni-assisted-installer-auto` due to signature change of `get_ingress_and_api_vips`:
> 
> * `test_static_ips`
> * `test_cancel_and_retest`
> * `test_scaleup`

it has a default, so it's not necessarily break them, right?
what needs to make sure is that those tests are not installing SNO (and in that case need to have it fixed)

osherdp:
hmm, it has the wrong default value so I'll have to change it either way

PR #1798, opened 2022-08-18
Set hypershift operator image only when the HYPERSHIFT_OPERATOR_IMAGE env was provided

eranco74:
In case we override the hypershift-operator-image to the latest hypershift image (might happen because the setup_capi.sh script export a similar env) we might end up with a mismatch between the release we are installing and the hypershift control plane operator image

### DIFF
--- a/src/tests/test_kube_api.py
+++ b/src/tests/test_kube_api.py
@@ -209,7 +209,7 @@ class TestKubeAPI(BaseKubeAPI):
                     pull_secret_file=ps,
                     agent_namespace=spoke_namespace,
                     provider_image=os.environ.get("PROVIDER_IMAGE", ""),
-                    hypershift_cpo_image=os.environ.get("HYPERSHIFT_IMAGE", ""),
+                    hypershift_cpo_image=os.environ.get("HYPERSHIFT_OPERATOR_IMAGE", ""),
                     release_image=os.environ.get("OPENSHIFT_INSTALL_RELEASE_IMAGE", ""),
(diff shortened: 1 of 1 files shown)


eranco74:
kube-api-net-suite failed due to timeout waiting for agent to be installed
[2022-08-21 11:39:16] INFO     root:agent.py:201 Waiting for agents to have the condition 'Installed' = 'True' and currently agent conditions are {'a98237f9-9434-4fb8-9ddb-14fc93da2e4a': {'SpecSynced': 'True', 'Connected': 'True', 'RequirementsMet': 'True', 'Validated': 'True', 'Installed': 'False', 'Bound': 'True'}}

The host installation started at:
```
    "event_time": "2022-08-21T10:40:58.182Z",
    "host_id": "a98237f9-9434-4fb8-9ddb-14fc93da2e4a",
    "infra_env_id": "71984154-9ff1-4996-82e1-c57c6f03bdb9",
    "message": "Host test-infra-cluster-3589f426-master-0: updated status from installing to installing-in-progress (Starting installation)",
```    
The host was rebooted at:
```
    "event_time": "2022-08-21T10:44:56.222Z",
    "host_id": "a98237f9-9434-4fb8-9ddb-14fc93da2e4a",
    "infra_env_id": "71984154-9ff1-4996-82e1-c57c6f03bdb9",
    "message": "Host: test-infra-cluster-3589f426-master-0, reached installation stage Rebooting",
```
And just never reported back

PR #1807, opened 2022-08-24
Fix default value for getting vips

osherdp:
Back in https://github.com/openshift/assisted-test-infra/pull/1797 I accidentally used the wrong non-compatible default value for those functions (we'd like them to behave like previously defined).


osherdp:
/hold
testing
/cc @eliorerz @lalon4

eliorerz (changes requested):
Please change this also for `NutanixController`
https://github.com/openshift/assisted-test-infra/blob/master/src/assisted_test_infra/test_infra/controllers/node_controllers/nutanix_controller.py#L127

eliorerz (approved):
LGTM

PR #1810, opened 2022-08-28
NO-ISSUE - Install openvpn as dependency for nutanix CI

eliorerz:
For Nutanix CI workflow we need the ability to connect to `nutanix-lts-dev` environment.
For that matter we need to connect to the environment VPN using openvpn.

/cc @osherdp


osherdp:
Can you add a bit more details?

osherdp (approved):

PR #1811, opened 2022-08-28
MGMT-11852 - Update platform to none on sno installation

eliorerz:
After https://github.com/openshift/assisted-service/pull/4158 SNO cluster must be none platform when enabling user-managed-networking 

/cc @osherdp

### DIFF
--- a/src/triggers/default_triggers.py
+++ b/src/triggers/default_triggers.py
@@ -28,6 +28,7 @@ _default_triggers = frozendict(
             workers_count=0,
             high_availability_mode=consts.HighAvailabilityMode.NONE,
             user_managed_networking=True,
+            platform=consts.Platforms.NONE,
             vip_dhcp_allocation=False,
             master_memory=resources.DEFAULT_MASTER_SNO_MEMORY,
             master_vcpu=resources.DEFAULT_MASTER_SNO_CPU,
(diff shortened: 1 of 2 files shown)


src/triggers/default_triggers.py
```diff
@@ -24,6 +24,7 @@
             workers_count=0,
             high_availability_mode=consts.HighAvailabilityMode.NONE,
             user_managed_networking=True,
+            platform=consts.Platforms.NONE,
```
osherdp, line 27: hmm, but it means we changed the API
if we have a customer using REST API (or maybe even UI) it means he needs to add this platform

Is it possible making assisted-service infer none-platform when on SNO? (I guess like before)

osherdp (changes requested):

osherdp (approved):

osherdp (changes requested):
You'll have to remove this: https://github.com/openshift/assisted-test-infra/blob/593f07f9266df78b64978da865131cce2a01ea56/Makefile#L102

osherdp (approved):

src/triggers/default_triggers.py
eliorerz: No, because we are setting `baremetal` as default platform. Baremetal and UMN enabled are not a valid input anymore.
Maybe setting platform to python None (null) will be better than changing the platform to none (although that in the service it will be none anyway)

osherdp:
/retest
just equinix running out of machines

osherdp (approved):

PR #1815, opened 2022-09-01
NO-ISSUE: Add missing sudo in install_environment.sh

jhernand:
Currently the `make install` command that installs the _swtpm_ package
and the `chmod` command that sets execution permission for the _libvirt_
cross network traffic hook are missing the `sudo` prefix, so they don't
work when using a non-root user. This patch fixes that.


jhernand:
@osherdp @eliorerz I found this issue while trying to run the tests in my environment with a non-root user, please take a look.

osherdp (approved):

PR #1817, opened 2022-09-05
NO-ISSUE - Use baremetal platform on SNO

eliorerz:
/cc @tsorya

### DIFF
--- a/src/assisted_test_infra/test_infra/helper_classes/cluster.py
+++ b/src/assisted_test_infra/test_infra/helper_classes/cluster.py
@@ -111,6 +111,9 @@ class Cluster(Entity):
             tang_servers=self._config.tang_servers,
         )
 
+        platform = (
+            models.Platform(type=consts.Platforms.NONE) if self.is_sno else models.Platform(type=self._config.platform)
+        )
         cluster = self.api_client.create_cluster(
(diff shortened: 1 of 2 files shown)


tsorya:
failure is relevant to the change 
HTTP response body: {"code":400,"message":"parsing newClusterParams body from \"\" failed, because json: cannot unmarshal string into Go struct field ClusterCreateParams.platform of type models.Platform"}

romfreiman:
why baremetal platfrom? what am i missing? sno is none

eliorerz:
> failure is relevant to the change HTTP response body: {"code":400,"message":"parsing newClusterParams body from "" failed, because json: cannot unmarshal string into Go struct field ClusterCreateParams.platform of type models.Platform"}

Yeah, I missed something there ...

tsorya:
It will set none but in test-infra sno should be treated as baremetal and use baremetal teraform. At least till we will not refactor it

eliorerz:
> why baremetal platfrom? what am i missing? sno is none

Only on test-infra, the cluster is created with `platform_type=none`.
The flow in test infra requires baremetal platfrom for the right terraform configuration.
@tsorya will probably could explain it better but we will deferentially need to change the flow in test infra so we will support SNO  with none platform

PR #1824, opened 2022-09-06
NO-ISSUE - Prevent tests from being skipped when test parameterized is empty

eliorerz:
pytest tests are skipped when parameterized is empty - e.g. `openshift_version` on `test_install`.
This PR force the test to fail when pytest test is skipped

/cc @osherdp

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -331,7 +331,7 @@ test:
 	skipper make $(SKIPPER_PARAMS) _test
 
 _test: $(REPORTS) _test_setup
-	JUNIT_REPORT_DIR=$(REPORTS) python3 ${DEBUG_FLAGS} -m pytest $(or ${TEST},src/tests) -k $(or ${TEST_FUNC},'') -m $(or ${TEST_MARKER},'') --verbose -s --junit-xml=$(PYTEST_JUNIT_FILE)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -331,7 +331,7 @@ test:
 	skipper make $(SKIPPER_PARAMS) _test
 
 _test: $(REPORTS) _test_setup
-	JUNIT_REPORT_DIR=$(REPORTS) python3 ${DEBUG_FLAGS} -m pytest $(or ${TEST},src/tests) -k $(or ${TEST_FUNC},'') -m $(or ${TEST_MARKER},'') --verbose -s --junit-xml=$(PYTEST_JUNIT_FILE)
+	JUNIT_REPORT_DIR=$(REPORTS) python3 ${DEBUG_FLAGS} -m pytest --error-for-skips $(or ${TEST},src/tests) -k $(or ${TEST_FUNC},'') -m $(or ${TEST_MARKER},'') --verbose -s --junit-xml=$(PYTEST_JUNIT_FILE)
```
osherdp, line 334: hmm, it's problematic no?
because ``-k/-m`` will also be considered as a skip no?

osherdp (changes requested):

Makefile
eliorerz: Oh.. I get it now, I don't think it's considered as skip, I think that skip it's when the test is "running" but it's skipped, the - k for example is running only the specified tests (starts with TEST_FUNC), so when running a specific test (e.g. test_install) it will run just those tests, and if one of those tests is marked as skipped than it will fail

eliorerz: I also tested it locally

osherdp (approved):

PR #1827, opened 2022-09-07
NO-ISSUE - Upgrade to python3.10

eliorerz:
Used in a self built image located at https://quay.io/repository/eerez/python-builder
We will do the asme for 3.11 that supposed to be released on October (3.11.0rc1 is already available for installation on `python-builder` repository)

/cc @osherdp

### DIFF
--- a/Dockerfile.assisted-test-infra
+++ b/Dockerfile.assisted-test-infra
@@ -18,13 +18,15 @@ RUN dnf -y install --enablerepo=crb \
   httpd-tools \
   jq \
   nss_wrapper \
-  python3 \
-  python3-devel \
   libvirt-client \
   libvirt-devel \
   libguestfs-tools \
   libxslt \
    && dnf clean all
+
+# Install latest available python version - For more information see https://github.com/eliorerz/tools/tree/master/python_builder
+COPY --from=quay.io/eerez/python-builder:stream9-3.10.7 /python /python-installation
+RUN cd /python-installation && make install && dnf -y install python3-devel && dnf clean all
 
 RUN curl --retry 5 -Lo packer.zip https://releases.hashicorp.com/packer/1.8.0/packer_1.8.0_linux_386.zip && unzip packer.zip -d /usr/bin/ && mv /usr/bin/packer /usr/bin/packer.io && rm -rf packer.zip
 RUN curl --retry 5 -Lo terraform.zip https://releases.hashicorp.com/terraform/1.2.7/terraform_1.2.7_linux_amd64.zip && unzip terraform.zip -d /usr/bin/ && rm -rf terraform.zip
--- a/scripts/override_os_images.py
+++ b/scripts/override_os_images.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 import json
 import os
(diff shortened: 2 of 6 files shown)


Dockerfile.assisted-test-infra
```diff
@@ -18,14 +18,16 @@ RUN dnf -y install --enablerepo=crb \
...
-  python3 \
-  python3-devel \
   libvirt-client \
   libvirt-devel \
   libguestfs-tools \
   libxslt \
    && dnf clean all
 
+# Install latest available python version
+COPY --from=quay.io/eerez/python-builder:stream9-3.10.7 /python /python-installation
```
osherdp, line 28: just something I'm wondering if we can make it happen
is it possible to create ``quay.io/eerez/python-builder:stream9-3.10.7`` (or with some other name) as a pre-included stream9 based image that has python 3.10 installed?
it can save some time building images and be a little more agnostic to how we're getting those python3 dependencies

osherdp: for quite some time we wanted to try the base-images idea, so we might as well go along with it

osherdp (changes requested):

Dockerfile.assisted-test-infra
osherdp: ACK
I can get along with it, but let's just fix some things first:
* Add that URL in a comment, so people will know where to find the source code
* That dir needs some cleanup, for example you have a ``__init__.py`` module with just the string ``FROM``
* Add a small README in that dir to make the cross-connect to the generated image in quay.io

osherdp (changes requested):

Dockerfile.assisted-test-infra
eliorerz: Sure, I just wanted to check first if it's working and get your opinion on the idea. I'll add those fixes.

eliorerz: Updated https://github.com/eliorerz/tools/blob/master/python_builder/README.md

scripts/override_os_images.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/local/bin/python3
```
osherdp, line 1: to the best of my knowledge, this is superior as it's less coupled to specific locations of the python binary (which can change based on installation method) and is equivalent to doing ``python3`` in bash:
```suggestion
#!/usr/bin/env python3
```

osherdp (changes requested):

scripts/deploy_assisted_service.sh
```diff
@@ -135,7 +135,7 @@ else
         minikube image rm "${SERVICE}" || true
     fi
 
-    skipper run src/update_assisted_service_cm.py
+    skipper run python3 src/update_assisted_service_cm.py
```
osherdp, line 138: now this change should be redundant I think

osherdp (changes requested):
sorry, last one I hope

scripts/deploy_assisted_service.sh
eliorerz: On most other places we call python3 explicitly, I think we should move them all to exeute it using `skipper run python3 ...`

osherdp: so why having the shebang line if now using it?

osherdp (approved):

PR #1833, opened 2022-09-08
MGMT-11938 - Remove default platform from test env vars

eliorerz:
- Add `TF_PLATFORM` env var for setting the controller platform. When cluster platform is set, it automatically will set `tf_platform` with the same value. `tf_platform` default value is `baremetal`.
- Align all base node controllers to the same format and fix missing `ABC` inheritance on the class signature.

Note: This PR allow us to create a cluster without actively set the cluster platform, in that case the decision of which platform to use will be decides by the `assisted-service` taking into account all the parameters (e.g. `user-managed-networking`, `high-availability-mode` and more)

/cc @osherdp


eliorerz:
@FYI @lalon4

osherdp:
Looks good two things:
* How should QE change their code (if anything)?
* I guess there's some more paths that are needed to be tested no?
/test e2e-vsphere-assisted 
(and the nutanix one that we don't have as presubmit?)

osherdp:
```
OpenSSL SSL_connect: Connection reset by peer in connection to awscli.amazonaws.com:443 
```
seems pretty rare to happen.
/retest

osherdp (approved):

osherdp:
@eliorerz can you add nutanix presubmit job for test-infra?

eliorerz:
> @eliorerz can you add nutanix presubmit job for test-infra?

Nutanix will not work. We are still waiting for https://github.com/openshift/assisted-service/pull/4165

eliorerz:
> Looks good two things:
> 
> * How should QE change their code (if anything)?
> * I guess there's some more paths that are needed to be tested no?
>   /test e2e-vsphere-assisted
>   (and the nutanix one that we don't have as presubmit?)

1. This fix was made for QE, their code should stay the same.
2. I'll test Nutanix locally, currently we can't test it on CI

PR #1834, opened 2022-09-09
Expose UMN as a setting

vrutkovs:
Use USER_MANAGED_NETWORKING as a setting for UMN, defaults to False.

Later on we'll add optional tests to ensure UMN installs don't regress in CI


eliorerz (approved):
LGTM

eliorerz (approved):

PR #1837, opened 2022-09-11
NO-ISSUE: add describe cluster operators to download_must_gather

eranco74:
Many times the must-gather log is partial or empty, describing the cluster-operators usually provides valubale information about the cluster


src/assisted_test_infra/download_logs/download_logs.py
```diff
@@ -352,6 +352,21 @@ def download_must_gather(kubeconfig: str, dest_dir: str):
...
+    log.info(f"Describing cluster operators using kubeconfig: {kubeconfig}")
+    command = (
+        f"oc --insecure-skip-tls-verify --kubeconfig={kubeconfig} describe co"
+        f" > {must_gather_dir}/cluster_operators.yaml"
+    )
+    try:
+        run_command(command, shell=True, raise_errors=True)
+
+    except RuntimeError as ex:
+        log.warning(f"Failed to run oc describe co: {ex}")
```
osherdp, line 369:
```suggestion
    with SuppressAndLog(RuntimeError):
        run_command(command, shell=True, raise_errors=True)
```

src/assisted_test_infra/download_logs/download_logs.py
```diff
@@ -352,6 +352,21 @@ def download_must_gather(kubeconfig: str, dest_dir: str):
 
     log.debug("Removing must-gather directory %s after we archived it", must_gather_dir)
     shutil.rmtree(must_gather_dir)
+    describe_cluster_operators(kubeconfig, dest_dir)
```
osherdp, line 355: I think it shouldn't be called from ``download_must_gather`` as it's doing something else (especially that it has ``JunitTestCase`` decorator which should be inside of a ``JunitTestCase`` afaik)

osherdp (changes requested):

src/assisted_test_infra/download_logs/download_logs.py
eranco74: Hmm, I think the name is still describing the main thing this function does (getting the co status is informative but just has very little info compared to must-gather logs).
I added an optional parameter (describe_cluster_operators) to the function that indicates that it will optionally run that command as well).
I removed the `JunitTestCase` decorator from the `describe_cluster_operators`

osherdp (approved):

PR #1839, opened 2022-09-12
MGMT-11706: Add test for the upgrade agent feature

jhernand:
This test does the following:

- Prepares a cluster and a set of nodes using the previous version of
  the agent.
- Waits for all the nodes to be int the _ready_ state.
- Restarts the service so that it requires a bad agent image, one that
  that then nodes can't pull.
- Waits till all the nodes move to the _insufficient_ state.
- Restarts the service using the current version of the agent.
- Waits till all the nodes upgrade and move to the _ready_ state.

Related: https://issues.redhat.com/browse/MGMT-11706


src/tests/test_e2e_install.py
```diff
@@ -29,3 +37,86 @@ def test_networking(self, cluster, network_type):
...
+        # Find the namespace:
+        namespace = os.environ.get("NAMESPACE", "assisted-installer")
+
+        # Create the API client:
+        client = kubernetes.dynamic.DynamicClient(
+            kubernetes.client.api_client.ApiClient(
+                configuration=kubernetes.config.load_kube_config(),
+            ),
+        )
+        configmaps_api = client.resources.get(api_version="v1", kind="ConfigMap")
```
osherdp, line 57: if it might affect other tests running in parallel, we might need a way to separate it from other runs and mark it as something that needs to run on its own
is it such a thing?

osherdp (changes requested):
partial review

src/tests/test_e2e_install.py
```diff
@@ -29,3 +37,86 @@ def test_networking(self, cluster, network_type):
     def test_olm_operator(self, cluster, olm_operators):
         cluster.prepare_for_installation()
         cluster.start_install_and_wait_for_installed()
+
+    @JunitTestSuite()
+    def test_upgrade_agent(self, cluster):
+        """
+        This test will prepare the cluster with the default agent image. Once it read to install it will restart the
+        service with an agent image that doesn't exist, so that all the hosts will move to `insufficient`. Then it will
```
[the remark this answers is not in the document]
jhernand: Actually the upgrade mechanism doesn't have the "older" or "newer" concept; it just upgrades from version _a_ to version _b_, where _b_ may be newer or older than _a_.

This test is checking the upgrade from _latest_ to _doesnotexist_, which fails, and then from _doesnotexist_ to _latest_, which works. 

It would be possible to check from an older version to a newer one, but that would require having access to those two versions of the agent images, i.e., the test would need to have two configuration parameters, something like `AGENT_DOCKER_IMAGE_OLD` and `AGENT_DOCKER_IMAGE`. As currently the test uses `:latest` instead of a numbered version I assumed that our jobs don't know anything about agent version numbers either.

Another alternative would be to change the build pipelines so that they publish the current `:latest` image and also a `:previoius` image (which doesn't really need to be the previous version, just a different tag). Then we could write a test to check the upgrade from `:previous` to `:latest`.

To summarize, with the current images that we publish and use in the tests I think that what this test does is reasonable, but I am open to implement one of the above alternatives. Let me know what you think.

src/tests/test_e2e_install.py
jhernand: This will indeed affect other tests that run in parallel if they use the same instance of the service, as it will be restarted with an alternative (and sometimes broken) agent image. How do we do that separation? A new test class? Something else?

osherdp: @eliorerz can you advice?

src/tests/test_e2e_install.py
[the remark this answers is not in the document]
oourfali: I also think that what the test does is sufficient with that regards.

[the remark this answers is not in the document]
eranco74: We can just take some existing image tag and use it for the test.
There is a difference in the upgrade flow if the new required image exists (the agent should successfully pull it and restart) vs the case image doesn't exist where the agents will move to insufficient but the actual agent upgrade flow isn't really tested.

src/tests/test_e2e_install.py
eliorerz: Maybe we can generate a new namespace the same we are doing on the kube_api flow:
https://github.com/openshift/assisted-test-infra/blob/master/src/tests/base_kubeapi_test.py#L39-#L41

osherdp: > A new test class
as a start, we should definitely move it away from ``test_e2e_install.py``, which is meant to have more like 'happy-flow' cases. Maybe it's even enough

As for the namespace idea, is it possible? Doesn't it require installing assisted on that namespace? (which we usually are doing outside of the scope of pytest tests)

src/tests/test_e2e_install.py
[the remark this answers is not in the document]
jhernand: Changed the test so that it finds the previous image (the newest `vX.YZ` tag that has a different digest) and uses it initially. It then waits for nodes to be ready, restarts with a bad image, waits for nodes to be insufficient, and finally restart with the current image and waits for nodes to be ready.

src/tests/test_e2e_install.py
jhernand: Moved the test to a new `TestUpgradeAgent`class.

Doing a separate installation of the service seems complicated to do from the test itself. Can we instead have a separate job that only runs the tests inside this new `TestUpgradeAgent` class?

osherdp: I'm actually wondering if we even want to test this upgrade feature periodically
The reason I'm not sure about that is because it's something that happens in between version upgrades (and not an installation method that happens regularly day-to-day) and mainly important in the presence of incompatibility between agent behavior (which is also quite rare I think)

Not to say the feature is not important (because it is!). If test costs weren't high I wouldn't mind adding much more tests than today

osherdp: as for this PR itself, it's okay as-is I think. just having it in a separate file and making sure not to run it in a bigger suite will be enough

osherdp: and of course making the job optional costs us nothing (forgot to say that before)

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,189 @@
...
+    process = subprocess.Popen(
+        args=["skopeo"] + args,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+    )
+    (stdout, stderr) = process.communicate()
+    if process.returncode != 0:
+        raise Exception(f"failed to run skopeo with args '{args}', finished with code {process.returncode}")
+    if len(stderr) > 0:
+        raise Exception(f"failed to run skopeo with args '{args}', wrote '{stderr}' to the standard error stream")
```
osherdp, line 32: should we really consider it an error? I would expect us to only follow the returncode for that

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,189 @@
...
+    return stdout.decode("utf-8")
```
osherdp, line 33: I think it can be simplified to:
```suggestion
    return subprocess.check_output(["skopeo"] + args).decode()
```

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,189 @@
...
+    Gets the digest of the given image.
+    """
+    result = skopeo(["inspect", "--raw", f"docker://{image}"])
+    result = json.loads(result)
+    return result["config"]["digest"]
+
+
+def repo_tags(repo: str) -> str:
+    """
+    Gets the lit of tags of the given image respository"
```
osherdp, line 55:
```suggestion
    Gets the list of tags of the given image repository"
```

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,189 @@
...
+    Calculates the image repository from the given image reference. For example, if the image is
+    'quay.io/my/image:v1.23' it will return 'quay.io/my/image'.
+    """
+    return re.sub(r":.*$", "", image)
+
+
+def image_digest(image: str) -> str:
+    """
+    Gets the digest of the given image.
+    """
```
osherdp, line 47: nit:
```suggestion
    """Gets the digest of the given image."""
```

osherdp: same elsewhere where it's just a short line

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,189 @@
...
+def repo_tags(repo: str) -> str:
+    """
+    Gets the lit of tags of the given image respository"
+    """
+    result = skopeo(["list-tags", f"docker://{repo}"])
+    result = json.loads(result)
+    return result["Tags"]
+
+
+def parse_tag(tag: str) -> List[int]:
```
osherdp, line 62: we're usually using ``semver`` library for naturally sorting versions. you can take a look here: https://pypi.org/project/semver/

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,189 @@
...
+    else:
+        r = 0
+    return [x, y, z, r]
+
+
+class TestUpgradeAgent(BaseTest):
+    @JunitTestSuite()
+    def test_upgrade_agent(self, cluster):
+        """
+        This test will prepare the cluster with the previouis agent image. Once it is ready to install it will restart
```
osherdp, line 85:
```suggestion
        This test will prepare the cluster with the previous agent image. Once it is ready to install it will restart
```

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,189 @@
...
+        the service with an agent image that doesn't exist, so that all the hosts will move to `insufficient`. Then it
+        will restart the service again with the current image, so that host will upgrade and move to `ready` again.
+        """
+        # Find the namespace:
```
osherdp, line 89: you can remove the comments where it's obvious from the code what it does (and leave the comments where it's more complicated)

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,189 @@
...
+                previous_tag = candidate_tag
+                break
+        assert previous_tag, "Failed to find previous tag for '{current_image}'"
+        previous_image = f"{agent_repo}:{previous_tag}"
+        broken_image = f"{agent_repo}:broken"
+        log.info(f"Previoius image is '{previous_image}'")
+        log.info(f"Broken image is '{broken_image}'")
+        log.info(f"Current image is '{current_image}'")
+
+        def restart_service(agent_image: str) -> None:
```
osherdp, line 131: define it as a class method. it will simplify this test method a bit (``test_upgrade_agent`` is too long right now to get a grasp of the test)

src/tests/test_e2e_install.py
jhernand: That sounds reasonable to me @osherdp. How do we make the job optional?

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,189 @@
...
+            """
+            Checks if the service is already using the given agent image. If it it isn't usisng it then it changes the
```
osherdp, line 133:
```suggestion
            Checks if the service is already using the given agent image. If it isn't using it then it changes the
```

src/tests/test_e2e_install.py
osherdp: Crystal is doing something similar in her new flow: https://github.com/openshift/release/pull/32302
The important bits are ``always_run: false`` and ``optional: true``

And feel free to add it optionally to all relevant component repos (assisted-test-infra, assisted-service, assisted-installer-agent, etc. whatever code that might affect the behavior)

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,189 @@
...
+        r = 0
+    return [x, y, z, r]
+
+
+class TestUpgradeAgent(BaseTest):
+    @JunitTestSuite()
+    def test_upgrade_agent(self, cluster):
+        """
+        This test will prepare the cluster with the previouis agent image. Once it is ready to install it will restart
+        the service with an agent image that doesn't exist, so that all the hosts will move to `insufficient`. Then it
```
osherdp, line 86: > service with an agent image that doesn't exist...

is it still the behavior? I see you've also added upgrade from candidate images and not just a non-existing tag

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,189 @@
...
+
+        # Get the digest and tags of the agent image currently used by the service:
+        service_config = configmaps_api.get(namespace=namespace, name="assisted-service-config")
+        current_image = service_config.data["AGENT_DOCKER_IMAGE"]
+        agent_repo = image_repo(current_image)
+
+        # We expect the repository to contain tags like 'v2.3.1'. We want to find the newest of those tags that is
+        # actually different (different digest) than the current tag used by the service, to make sure that we are using
+        # two different images.
+        all_tags = repo_tags(agent_repo)
```
osherdp, line 109: some part of the following logic can be refactored into a method, to abstract-away some of the more technical things here

src/tests/test_upgrade_agent.py
osherdp, line 131: maybe 
```suggestion
        def load_service_with_agent_image(agent_image: str) -> None:
```

osherdp (changes requested):
have a couple of more comments, but all are just style/refactoring related

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,189 @@
...
+    the command wrote to the standard output stream.
+    """
+    process = subprocess.Popen(
+        args=["skopeo"] + args,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+    )
+    (stdout, stderr) = process.communicate()
+    if process.returncode != 0:
+        raise Exception(f"failed to run skopeo with args '{args}', finished with code {process.returncode}")
```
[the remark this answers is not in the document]
jhernand: Done.

src/tests/test_upgrade_agent.py
jhernand: Well behaved programs should do that, but not all do. Not very relevant in this case. I will do as you suggest.

src/tests/test_upgrade_agent.py
jhernand: Indeed, I forgot this "check_output" method, thanks.

src/tests/test_upgrade_agent.py
jhernand: Done.

src/tests/test_upgrade_agent.py
jhernand: Done.

src/tests/test_upgrade_agent.py
jhernand: Done.

src/tests/test_upgrade_agent.py
jhernand: Done.

src/tests/test_upgrade_agent.py
jhernand: Done.

src/tests/test_upgrade_agent.py
jhernand: Yes, the behavior is first start the server with the previous agent image, then try to upgrade to a broken image, and then try to upgrade to the current image.

src/tests/test_upgrade_agent.py
jhernand: Done.

src/tests/test_upgrade_agent.py
osherdp: wait. do we even have cases of ``v<x>.<y>.<z>-<r>``? Can you share with such an example? (because I don't remember such thing under ``quay.io/edge-infrastructure/*``)

either way, at least the parsing part (without using the compare functions) can be replaced so we'll maintain less code and less regex functions (and I think it's also okay to simplify the code to only consider simple, "no-r" releases)

jhernand: We don't have those cases in `quay.io/edge-infrastructure/*`, but we do in the downstream image repository, for example:

```
registry.redhat.io/rhai-tech-preview/assisted-installer-agent-rhel8:v1.0.0-187
```

That is what we are currently using in production. The fact that the first part is `1.0.0` and the second is large makes me think that we are never updating the first part, only the second. I this scenario if we ignore the second part then we could be potentially selecting a rather old version of the agent, maybe one that doesn't even have support for the upgrade feature. I don't know if we use these tests with that downstream repository. If we do then I think it is worth the extra complexity.

src/tests/test_upgrade_agent.py
jhernand: It can't just be made a class method because it shares the namespace and Kubernetes API clients with the `test_upgrade_agent` method. I will try to move it out anyhow.

src/tests/test_upgrade_agent.py
jhernand: Done.

src/tests/test_upgrade_agent.py
jhernand: Moved it to a separate method.

jhernand:
@osherdp the new version of the patch addresses most of your comments.

oourfali:
@osherdp @eliorerz @eranco74 your re-review will be appreciated.

src/tests/test_upgrade_agent.py
osherdp: But either way we cannot use this code to test staging/integration (and of course production) because all of them are to be managed via app-sre configuration and not by directly changing their configmaps

@lalon4 is it possible for the QE team to use this test logic for some of the use-cases?

src/tests/test_e2e_install.py
osherdp: Can you add those jobs before merging this PR? So we can be assured it's passing successfully right from the start on the CI system

osherdp:
> @osherdp @eliorerz @eranco74 your re-review will be appreciated.

Do no worry. I track all PRs I'm reviewing and I won't leave @oourfali without a review :smiley: 

As for this PR, the code looks good but only those two issues should be handled (trying to summarize everything and provide some rational for the issues):
* IIUC we cannot reuse this code for D/S images/envs, so the regex parsing code is considered redundant IMHO
* We are always testing new code in advance, because you cannot tell how code will react in the CI system (for example, we're using prow's internal registry in CI jobs, meaning you won't see ``quay.io/edge-infrastructure/assisted-service:latest`` but rather something like ``registry.build04.ci.openshift.org/ci-op-r28fztgp/pipeline@sha256:<hash>``. Will it be problematic with current code? probably not, but why risk it). So adding those new jobs in advance and testing them here provides full context and e2e testing instead of providing patches along the way

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,201 @@
...
+        with a non zero exit code. Returns a string containing whatever the command wrote to the standard output stream.
+        """
+        return subprocess.check_output(["skopeo"] + args).decode()
+
+    def _get_image_repo(self, image: str) -> str:
+        """
+        Calculates the image repository from the given image reference. For example, if the image is
+        'quay.io/my/image:v1.23' it will return 'quay.io/my/image'.
+        """
+        return re.sub(r":.*$", "", image)
```
eranco74, line 43: The images reference we have in the CI use digests (instead of tags),  the regex should still work (there's a colon before the digest sha) just note that you will probably get something like this:
registry.build02.ci.openshift.org/ci-op-w23ww3fj/pipeline@sha256
Unsure there's any value in it

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,201 @@
...
+                        },
+                    },
+                },
+            },
+        )
+        log.info(f"Restarted deployment with agent image '{image}'")
+
+    def _get_previous_agent_image(self, current: str) -> str:
+        """
+        We expect the repository to contain tags like 'v2.3.1-187'. This function tries to find the newest of those tags
```
eranco74, line 136: I'm not sure we can make this assumption

eranco74 (commented):
I think that instead of trying to calculate the different agent images during the test, it can get the images as envs (assert the env isn't empty).
The config in the release repo can populate these envs.
e.g.
```
steps:
    dependencies:
      INITIAL_AGENT_IMAGE: ocm2.5:<agentref>
      UPDATED_AGENT_IMAGE: pipeline:assisted-installer-agent
    env:
      BAD_AGENT_IMAGE: <not found image>

File: src/tests/test_e2e_install.py (hunk 1 above)
Comment (eliorerz) [reply src/tests/test_e2e_install.py]: @osherdp I think this test is more fit to be in "QE tests pool", WDYT?

**Review by eliorerz — CHANGES_REQUESTED** [review]

File: src/tests/test_upgrade_agent.py [hunk 17]
```diff
@@ -0,0 +1,201 @@
...
+        log.info(f"Previous image is '{previous_image}'")
+        log.info(f"Broken image is '{broken_image}'")
+        log.info(f"Current image is '{current_image}'")
+
+        try:
+            # Prepare the cluster for installation using the previous image:
+            log.info("Waiting cluster to move to 'ready' with previous image")
+            self._load_service_with_agent_image(k8s_client, namespace, previous_image)
+            cluster.prepare_for_installation()
+            cluster.wait_for_ready_to_install()
```
eliorerz, line 187: Redundant. Already done here https://github.com/openshift/assisted-test-infra/blob/master/src/assisted_test_infra/test_infra/helper_classes/cluster.py#L897

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,201 @@
...
+        result = json.loads(result)
+        return result["config"]["digest"]
+
+    def _get_repo_tags(self, repo: str) -> str:
+        """Gets the list of tags of the given image respository."""
+        result = self._skopeo(["list-tags", f"docker://{repo}"])
+        result = json.loads(result)
+        return result["Tags"]
+
+    def _parse_agent_tag(self, tag: str) -> List[int]:
```
eliorerz, line 57:
```suggestion
    @classmethod
    def _parse_agent_tag(cls, tag: str) -> Optional[List[int]]:
```

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,201 @@
...
+        """
+        match = re.match(r"^v(?P<x>\d+).(?P<y>\d+).(?P<z>\d+)(-(?P<r>\d+))?$", tag)
+        if match is None:
+            return None
+        x = int(match.group("x"))
+        y = int(match.group("y"))
+        z = int(match.group("z"))
+        r = match.group("r")
+        if r is not None:
+            r = int(r)
```
eliorerz, line 70: Rename to major, minor, patch ...

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,201 @@
...
+        else:
+            r = 0
+        return [x, y, z, r]
```
eliorerz, line 73:
```suggestion
        r = match.group("r")
        return [x, y, z, int(r) if r is not None else 0]
```

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,201 @@
...
+
+    def _get_current_agent_image(
+        self,
```
eliorerz, line 76:
```suggestion
    @classmethod
    def _get_current_agent_image(
        cls,
```

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,201 @@
...
+        client: kubernetes.dynamic.DynamicClient,
+        namespace: str,
+    ) -> str:
+        """Returns the agent image the is currently used by the service."""
+        configmaps_api = client.resources.get(api_version="v1", kind="ConfigMap")
+        config = configmaps_api.get(namespace=namespace, name="assisted-service-config")
+        return config.data["AGENT_DOCKER_IMAGE"]
+
+    def _load_service_with_agent_image(
+        self,
```
eliorerz, line 86:
```suggestion
    @classmethod
    def _load_service_with_agent_image(
        cls,
```

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,201 @@
...
+        client: kubernetes.dynamic.DynamicClient,
+        namespace: str,
+        image: str,
+    ) -> None:
+        """
+        Checks if the service is already using the given agent image. If it isn't using it then it changes the configmap
+        to use it and restarts the deployment.
+        """
+        # Check if the service is already using the given agent image:
+        current = self._get_current_agent_image(client, namespace)
```
eliorerz, line 96:
```suggestion
        current = cls._get_current_agent_image(client, namespace)
```

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,201 @@
...
+        # Update the configuration:
+        configmaps_api = client.resources.get(api_version="v1", kind="ConfigMap")
+        configmaps_api.patch(
+            namespace=namespace,
+            name="assisted-service-config",
+            body={
+                "data": {
+                    "AGENT_DOCKER_IMAGE": image,
+                },
+            },
```
eliorerz, line 110:
```suggestion
            body={"data": {"AGENT_DOCKER_IMAGE": image}},
```

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,201 @@
...
+                "spec": {
+                    "template": {
+                        "metadata": {
+                            "annotations": {
+                                "kubectl.kubernetes.io/restartedAt": date,
+                            },
+                        },
+                    },
+                },
+            },
```
eliorerz, line 130:
```suggestion
            body={"spec": {"template": {"metadata": {"annotations": {"kubectl.kubernetes.io/restartedAt": date}}}}},
```

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,201 @@
...
+        )
+
+    def _skopeo(self, args: List[str]) -> str:
+        """
+        Runs the 'skopeo' command with the given command line arguments. Raises an exception if the command finishes
+        with a non zero exit code. Returns a string containing whatever the command wrote to the standard output stream.
+        """
+        return subprocess.check_output(["skopeo"] + args).decode()
+
+    def _get_image_repo(self, image: str) -> str:
```
eliorerz, line 38:
```suggestion
    @classmethod
    def _get_image_repo(cls, image: str) -> str:
```

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,201 @@
...
+
+    @pytest.fixture(scope="session")
+    def k8s_client(self) -> kubernetes.dynamic.DynamicClient:
+        return kubernetes.dynamic.DynamicClient(
+            kubernetes.client.api_client.ApiClient(
+                configuration=kubernetes.config.load_kube_config(),
+            ),
+        )
+
+    def _skopeo(self, args: List[str]) -> str:
```
eliorerz, line 31:
```suggestion
    @classmethod
    def _skopeo(cls, args: List[str]) -> str:
```

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,201 @@
...
+    def _get_image_repo(self, image: str) -> str:
+        """
+        Calculates the image repository from the given image reference. For example, if the image is
+        'quay.io/my/image:v1.23' it will return 'quay.io/my/image'.
+        """
+        return re.sub(r":.*$", "", image)
+
+    def _get_image_digest(self, image: str) -> str:
+        """Gets the digest of the given image."""
+        result = self._skopeo(["inspect", "--raw", f"docker://{image}"])
```
eliorerz, line 47:
```suggestion
    @classmethod
    def _get_image_digest(cls, image: str) -> str:
        """Gets the digest of the given image."""
        result = cls._skopeo(["inspect", "--raw", f"docker://{image}"])
```

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,201 @@
...
+        result = json.loads(result)
+        return result["config"]["digest"]
+
+    def _get_repo_tags(self, repo: str) -> str:
+        """Gets the list of tags of the given image respository."""
+        result = self._skopeo(["list-tags", f"docker://{repo}"])
```
eliorerz, line 53:
```suggestion
    @classmethod
    def _get_repo_tags(cls, repo: str) -> str:
        """Gets the list of tags of the given image respository."""
        result = cls._skopeo(["list-tags", f"docker://{repo}"])
```

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,201 @@
...
+                                "kubectl.kubernetes.io/restartedAt": date,
+                            },
+                        },
+                    },
+                },
+            },
+        )
+        log.info(f"Restarted deployment with agent image '{image}'")
+
+    def _get_previous_agent_image(self, current: str) -> str:
```
eliorerz, line 134:
```suggestion
    @classmethod
    def _get_previous_agent_image(cls, current: str) -> Optional[str]:
```

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,201 @@
...
+        that is actually different (different digest) than then given image, and returns the reference to that tag, or
+        else None if no such tag exists.
+        """
+        repo = self._get_image_repo(current)
+        all_tags = self._get_repo_tags(repo)
+        candidate_pairs = []
+        for tag in all_tags:
+            parsed = self._parse_agent_tag(tag)
+            if parsed is not None:
+                candidate_pairs.append((tag, parsed))
```
eliorerz, line 146:
```suggestion
            if (parsed := cls._parse_agent_tag(tag)) is not None:
                candidate_pairs.append((tag, parsed))
```

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,201 @@
...
+        previous_tag = None
+        current_digest = self._get_image_digest(current)
+        for candidate_tag in candidate_tags:
+            candidate_digest = self._get_image_digest(f"{repo}:{candidate_tag}")
+            if candidate_digest != current_digest:
+                previous_tag = candidate_tag
+                break
+        if previous_tag is None:
+            return None
+        return f"{repo}:{previous_tag}"
```
eliorerz, line 158:
```suggestion
        return None if previous_tag is None else f"{repo}:{previous_tag}" 
```

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,201 @@
...
+                candidate_pairs.append((tag, parsed))
+        candidate_pairs.sort(key=lambda pair: pair[1], reverse=True)
+        candidate_tags = list(map(lambda pair: pair[0], candidate_pairs))
+        previous_tag = None
+        current_digest = self._get_image_digest(current)
+        for candidate_tag in candidate_tags:
+            candidate_digest = self._get_image_digest(f"{repo}:{candidate_tag}")
+            if candidate_digest != current_digest:
+                previous_tag = candidate_tag
+                break
```
eliorerz, line 155:
```suggestion
        for candidate_tag in candidate_tags:
            if cls._get_image_digest(f"{repo}:{candidate_tag}") != current_digest:
                previous_tag = candidate_tag
                break
```

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,201 @@
...
+        if previous_tag is None:
+            return None
+        return f"{repo}:{previous_tag}"
+
+    def _get_broken_agent_image(self, current: str) -> str:
+        """Returns an image reference that doesn't exist."""
+        repo = self._get_image_repo(current)
+        return f"{repo}:broken"
```
eliorerz, line 163:
```suggestion
    @classmethod
    def _get_broken_agent_image(cls, current: str) -> str:
        """Returns an image reference that doesn't exist."""
        return f"{cls._get_image_repo(current)}:broken"
```

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,201 @@
...
+        """
+        current_image = self._get_current_agent_image(k8s_client, namespace)
+        previous_image = self._get_previous_agent_image(current_image)
+        broken_image = self._get_broken_agent_image(current_image)
+        assert current_image
+        assert previous_image
+        assert broken_image
+        log.info(f"Previous image is '{previous_image}'")
+        log.info(f"Broken image is '{broken_image}'")
+        log.info(f"Current image is '{current_image}'")
```
eliorerz, line 180:
```suggestion
        assert (current_image := self._get_current_agent_image(k8s_client, namespace))
        assert (previous_image := self._get_previous_agent_image(current_image))
        assert (broken_image := self._get_broken_agent_image(current_image))

        log.info(f"Previous image is '{previous_image}'")
        log.info(f"Broken image is '{broken_image}'")
        log.info(f"Current image is '{current_image}'")
```

src/tests/test_upgrade_agent.py
```diff
@@ -0,0 +1,201 @@
...
+            log.info("Waiting for hosts to move to 'insufficient' with broken image")
+            self._load_service_with_agent_image(k8s_client, namespace, broken_image)
+            cluster.wait_until_hosts_are_insufficient()
+
+            # Restart the agent with the current agent image and wait for all hosts to be ready, which means that they
+            # upgraded correctly:
+            log.info("Waiting for hosts to move to 'ready'")
+            self._load_service_with_agent_image(k8s_client, namespace, current_image)
+            cluster.wait_for_ready_to_install()
+        finally:
```
eliorerz, line 200: Not sure I understood why the try-finally is here, van you elaborate?

src/tests/test_e2e_install.py
```diff
@@ -29,3 +37,86 @@ def test_networking(self, cluster, network_type):
...
+        configmaps_api = client.resources.get(api_version="v1", kind="ConfigMap")
+        deployments_api = client.resources.get(api_version="apps/v1", kind="Deployment")
+
+        def restart_service(agent_image):
+            """
+            Checks if the service is already using the given agent image. If it it isn't usisng it then it changes the
+            configmap to use it and restarts the deployment.
+            """
+            # Check if the service is already using the given agent image:
+            config = configmaps_api.get(namespace=namespace, name="assisted-service-config")
```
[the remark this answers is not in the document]
jhernand: Done, renamed to `configmap`.

src/tests/test_upgrade_agent.py
jhernand: Done.

src/tests/test_upgrade_agent.py
jhernand: Done.

src/tests/test_upgrade_agent.py
jhernand: Done.

src/tests/test_upgrade_agent.py
jhernand: Done.

src/tests/test_upgrade_agent.py
jhernand: Done.

src/tests/test_upgrade_agent.py
jhernand: Done.

src/tests/test_upgrade_agent.py
jhernand: Done.

src/tests/test_upgrade_agent.py
jhernand: Done.

src/tests/test_upgrade_agent.py
jhernand: Done.

src/tests/test_upgrade_agent.py
jhernand: Done.

src/tests/test_upgrade_agent.py
jhernand: Done.

src/tests/test_upgrade_agent.py
jhernand: Done.

src/tests/test_upgrade_agent.py
jhernand: The reason for this is to restore the service configuration if the test fails in the middle.

src/tests/test_upgrade_agent.py
jhernand: Replaced with `r"[@:].*$"`, that should address this issue.

jhernand:
@eranco74 I don't have any objection to use environment variables for the images. @osherdp @eliorerz Is it OK if I do that?

osherdp:
> @eranco74 I don't have any objection to use environment variables for the images. @osherdp @eliorerz Is it OK if I do that?

I'm in favor to have it as optional env-vars (so that default behavior will stay as it is if not supplied)

osherdp:
> > @eranco74 I don't have any objection to use environment variables for the images. @osherdp @eliorerz Is it OK if I do that?
> 
> I'm in favor to have it as optional env-vars (so that default behavior will stay as it is if not supplied)

okay, maybe a better suggestion:
just using some hardcoded old image (let's say v2.9.0) and remove all the skopeo logic and image-finding logic

jhernand:
> okay, maybe a better suggestion: just using some hardcoded old image (let's say v2.9.0) and remove all the skopeo logic and image-finding logic

Sounds reasonable to me, I am doing that.

src/tests/test_upgrade_agent.py
jhernand: No longer applies.

src/tests/test_upgrade_agent.py
jhernand: No longer applies.

jhernand:
@osherdp @eliorerz @eranco74 Is this what is required in the `openshift/release` repository? https://github.com/openshift/release/pull/32413

osherdp:
> @osherdp @eliorerz @eranco74 Is this what is required in the `openshift/release` repository? [openshift/release#32413](https://github.com/openshift/release/pull/32413)

Dropped just one comment for that PR
The code in this PR looks good to go, so we're just down to running it in the CI system 🙂

src/tests/test_upgrade_agent.py
jhernand: Done.

osherdp (approved):
Looking good!

jhernand:
@eliorerz do you have objections to have this merged?

PR #1847, opened 2022-09-15
MGMT-12026: do not reset trigger values

adriengentil:
When a trigger overrides an environment variable, it uses the [copy()](https://github.com/openshift/assisted-test-infra/blob/master/src/tests/global_variables/default_variables.py#L41) method which [sets](https://github.com/openshift/assisted-test-infra/blob/master/src/assisted_test_infra/test_infra/utils/env_var.py#L60) the new value for that environment variable.

But the call to [value()](https://github.com/openshift/assisted-test-infra/blob/master/src/assisted_test_infra/test_infra/utils/env_var.py#L46-L54) always re-evaluates against the current environment and does not take into account that the value was overridden by a trigger.


adriengentil:
/cc @osherdp 

e2e-metal-assisted-prod passed :partying_face:
It may be worth to refactor this part later in order to decouple the value a variable from its source.

osherdp (approved):

PR #1852, opened 2022-09-19
MGMT-12096: Fix errors accessing load-balancer http conf file

eliorerz:
On some cases the load balancer failed to start due to the following error

```
nginx: [emerg] bind() to 192.168.127.1:6443 failed (99: Cannot assign requested address)
```

This PR remove the conf file before each time we start it, also fixed the conf file path on the `stop_load_balancer` Makefile target 

See right path on https://github.com/openshift/assisted-test-infra/blob/master/terraform_files/none/main.tf#L170

/cc @osherdp @tsorya 

FYI @lalon4


osherdp (approved):

eliorerz:
Doesn't seems related

PR #1856, opened 2022-09-22
MGMT-11926: vsphere: create machines with HW 15

vrutkovs:
Hardware version 15 is required for CSI tests to pass

### DIFF
--- a/terraform_files/vsphere/main.tf
+++ b/terraform_files/vsphere/main.tf
@@ -91,6 +91,7 @@ resource "vsphere_virtual_machine" "master" {
   guest_id                    = "coreos64Guest"
   folder                      = var.vsphere_folder != "" ? "${var.vsphere_parent_folder}/${local.folder}" : vsphere_folder.folder[0].path
   enable_disk_uuid            = "true"
+  hardware_version            = 15
   # no network before booting from the ISO file, which isn't available until prepare_for_installation stage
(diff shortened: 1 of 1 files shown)


vrutkovs:
/cc @eliorerz 
/hold

for rehearsals

osherdp (approved):
:pizza:

terraform_files/vsphere/main.tf
```diff
@@ -91,6 +91,7 @@ resource "vsphere_virtual_machine" "master" {
   guest_id                    = "coreos64Guest"
   folder                      = var.vsphere_folder != "" ? "${var.vsphere_parent_folder}/${local.folder}" : vsphere_folder.folder[0].path
   enable_disk_uuid            = "true"
+  hardware_version            = 15
```
romfreiman, line 94: what's the default?

vrutkovs: That is baked in OVA which we use, so we start with HW 14, run 4.11 CSI tests and these complain that you need at least 15 - see `[sig-arch][Early] Managed cluster should start all core operators` test message in this [test run](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-assisted-test-infra-master-e2e-vsphere-assisted-periodic/1572012854410743808)

vrutkovs:
UMN is known to be broken, unholding

/hold cancel

PR #1849, opened 2022-09-18
NO-ISSUE: download kubeconfig in BaseKubeAPI test

danielerez:
On BaseKubeAPI test, download kubeconfig using secret ref in ACI.

PR #1864, opened 2022-09-27
NO-ISSUE: Fix late-binding job

adriengentil:
Two issues:

1/
Since https://github.com/openshift/assisted-test-infra/pull/1852 the
loadbalancer configuration is cleaned up before starting it.

This change does not cleanup the configuration when `TEST_TEARDOWN` is
set to `false`.

It should fix the late binding job, that configures the nginx server
before running the tests.

2/
Since https://github.com/openshift/assisted-test-infra/pull/1849/files#diff-21e688d6eb3b87727264f6934118765f035f5bf7504dff17b3ab9fb388f4d054L416

The kubeconfig used for the hub cluster is replaced by the kubeconfig that points
to the sno cluster which leads to fail the conformance test as the DNS for this cluster
 is not configured.

I made the parameter optional, as I guess the kubeconfig from the `AgentClusterInstall`
is not used in this case.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -183,7 +183,9 @@ start_load_balancer: stop_load_balancer
 stop_load_balancer:
 	@id=$(shell $(CONTAINER_COMMAND) ps --all --quiet --filter "name=load_balancer"); \
 	test ! -z "$$id"  && $(CONTAINER_COMMAND) rm -f load_balancer; \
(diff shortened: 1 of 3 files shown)


adriengentil:
ok, the late binding job failed on the conformance test

@eliorerz do you see a better way to fix this issue?

adriengentil:
no gather :unamused: 

osherdp (approved):

PR #1866, opened 2022-09-29
MGMT-11968: Use lock and tmp file when altering iptables rules

mkowalski:
In order to prevent potential races when multiple threads or processes try to modify iptables rules by using libvirt hook, we are introducing use of a temporary file as well as the lock. This should ensure that there are no concurrent attempts to -save and -restore the firewall configuration.

Co-authored-by: jhernand@redhat.com
Closes: [MGMT-11968](https://issues.redhat.com//browse/MGMT-11968)


jhernand (approved):
Looks good to me, thanks @mkowalski!

mkowalski:
/cc @osherdp 

Seems like I can't /approve and Juan can't /lgtm

osherdp (approved):

PR #1880, opened 2022-10-19
nutanix: don't allocation VIPs if DHCP allocation is not set

vrutkovs:
In CI we might want to allocate VIPs and create DNS records using existing IPI steps. Nutanix controller should not use DHCP to allocate VIPs if these are already set using environment variables.

Being tested in https://github.com/openshift/release/pull/31726

### DIFF
--- a/src/assisted_test_infra/test_infra/controllers/node_controllers/nutanix_controller.py
+++ b/src/assisted_test_infra/test_infra/controllers/node_controllers/nutanix_controller.py
@@ -125,6 +125,13 @@ class NutanixController(NodeController):
         vm.power_on()
 
     def get_ingress_and_api_vips(self, is_highly_available: bool = True):
+        if not self._entity_config.vip_dhcp_allocation:
+            if not self._entity_config.api_vip:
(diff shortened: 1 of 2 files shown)


src/assisted_test_infra/test_infra/controllers/node_controllers/nutanix_controller.py
```diff
@@ -125,6 +125,13 @@ def restart_node(self, node_name: str) -> None:
         vm.power_on()
 
     def get_ingress_and_api_vips(self, is_highly_available: bool = True):
+        if not self._entity_config.vip_dhcp_allocation:
+            if not self._entity_config.api_vip:
+                raise ConnectionError("API VIP is not set")
```
osherdp, line 130: shouldn't it be ``raise ValueError`` if it's a configuration mismatch made by the user?
or maybe i just don't fully understand the case here

vrutkovs: Right, `ValueError` fits better here

osherdp (approved):

PR #1888, opened 2022-10-26
NO-ISSUE: yaml-friendly pull secret removal

omertuc:
`*` has a special meaning in YAML and replacing this pull secret in this manner results
in an invalid YAML. changed to use `x`'s instead of `*`'s.


javipolo (approved):

osherdp (approved):

osherdp (approved):
sounds good as a quick patch
I'd like afterwards to use ruamel for loading and dumpint, to have an extra validation we're publishing a valid yaml file

PR #1889, opened 2022-10-27
NO-ISSUE: Upgrade Python 3.11 that is now officially available

eliorerz:
## [What’s New In Python 3.11](https://docs.python.org/3.11/whatsnew/3.11.html)

/cc @eliorerz


osherdp (approved):

osherdp:
/hold
Maybe kube-api job is broken

PR #1890, opened 2022-10-30
MGMT-12381: Remove VIP DHCP allocation defaults from test-infra

eliorerz:
Back on https://issues.redhat.com/browse/MGMT-10840 we changed the default value of vip_dhcp_allocation but left assisted-test-infra untouched causing vip_dhcp_allocation to stay `True` on some tests.

We need to remove vip_dhcp_allocation default from test-infra and allowing the service to set the default by itself.

/cc @osherdp

### DIFF
--- a/src/assisted_test_infra/test_infra/helper_classes/cluster.py
+++ b/src/assisted_test_infra/test_infra/helper_classes/cluster.py
@@ -106,15 +106,19 @@ class Cluster(Entity):
         return self._config.cluster_id
 
     def _create(self) -> str:
+        extra_vars = {}
+
         disk_encryption = models.DiskEncryption(
             enable_on=self._config.disk_encryption_roles,
             mode=self._config.disk_encryption_mode,
             tang_servers=self._config.tang_servers,
         )
 
-        platform_var = {}
         if self._config.platform:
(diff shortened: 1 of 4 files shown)


eliorerz:
FYI @lalon4

src/assisted_test_infra/test_infra/helper_classes/cluster.py
```diff
@@ -452,16 +455,18 @@ def set_advanced_networking(
         else:
             machine_networks = [models.MachineNetwork(cidr=cidr) for cidr in machine_networks]
 
+        vip_dhcp_allocation_var = {}
```
osherdp, line 458:
```suggestion
        extra_vars = {}
```
let's make this generic from start, I have a feeling we'll have a reason to append more to it in the future

osherdp (changes requested):

src/assisted_test_infra/test_infra/helper_classes/cluster.py
eliorerz: More generic than `extra_vars`?

osherdp (approved):

osherdp (approved):

PR #1891, opened 2022-10-30
Fix DNS for kube-api SNO installations

osherdp:


src/tests/test_kube_api.py
```diff
@@ -143,18 +143,25 @@ def kube_api_test(
...
             # wait till the ip is set for the node and read it from its inventory
-            self.set_single_node_ip(cluster_deployment, nodes)
+            single_node_ip = get_ip_for_single_node(cluster_deployment, nodes.is_ipv4)
+            nodes.controller.tf.change_variables(
+                {
+                    "single_node_ip": single_node_ip,
+                    "bootstrap_in_place": True,
+                }
+            )
+            api_vip = ingress_vip = single_node_ip
```
osherdp, line 158: this is the specifically important part. being able to choose api=ingress=sno-ip and use it afterwards in ``nodes.controller.set_dns(api_ip=api_vip, ingress_ip=ingress_vip)``

src/tests/base_kubeapi_test.py
```diff
@@ -197,18 +196,6 @@ def find_matching_node_and_return_name():
         log.info(f"patching agent {agent.ref} with hostname {hostname}")
         agent.patch(hostname=hostname)
 
-    @classmethod
```
paul-maidment, line 200: not sure if I wanted to make a <nit> critique of your decision to remove this 

as there are two places in base_test where it could be argued this is reused
https://github.com/openshift/assisted-test-infra/pull/1891/files#diff-21e688d6eb3b87727264f6934118765f035f5bf7504dff17b3ab9fb388f4d054R151

https://github.com/openshift/assisted-test-infra/pull/1891/files#diff-21e688d6eb3b87727264f6934118765f035f5bf7504dff17b3ab9fb388f4d054R423

but the added complexity of dealing with obtaining the returned single node IP justifies the duplication (I think?)

src/assisted_test_infra/test_infra/controllers/node_controllers/terraform_controller.py
```diff
@@ -219,27 +213,17 @@ def format_node_disk(self, node_name: str, disk_index: int = 0):
...
-        network_subnet_starting_ip = str(
-            ipaddress.ip_address(ipaddress.ip_network(self.get_primary_machine_cidr()).network_address) + 100
-        )
-        ips = utils.create_ip_address_list(2, starting_ip_addr=str(ipaddress.ip_address(network_subnet_starting_ip)))
-
-        if is_highly_available:
-            return {"api_vip": ips[0], "ingress_vip": ips[1]}
-
-        return {"api_vip": ips[0], "ingress_vip": ips[0]}
+        network_subnet_starting_ip = ip_address(ip_network(self.get_primary_machine_cidr()).network_address)
```
paul-maidment, line 224: Initially I found this change confusing, but then I realised that the +100 has been moved to the line below, I think it looks clearer that way.

We discussed this offline, just putting my thoughts here so we can record them.

IMO this looks good.

src/tests/base_kubeapi_test.py
osherdp: Right. In the next PR I intend to do a couple of more changes, specifically removing all those occurrences of dynamic changes of terraform object.

eliorerz (approved):
LGTM

PR #1898, opened 2022-11-04
CVE-2007-4559 Patch

TrellixVulnTeam:
# Patching CVE-2007-4559

Hi, we are security researchers from the Advanced Research Center at [Trellix](https://www.trellix.com). We have began a campaign to patch a widespread bug named CVE-2007-4559. CVE-2007-4559 is a 15 year old bug in the Python tarfile package. By using extract() or extractall() on a tarfile object without sanitizing input, a maliciously crafted .tar file could perform a directory path traversal attack. We found at least one unsantized extractall() in your codebase and are providing a patch for you via pull request. The patch essentially checks to see if all tarfile members will be extracted safely and throws an exception otherwise. We encourage you to use this patch or your own solution to secure against CVE-2007-4559. Further technical information about the vulnerability can be found in this [blog](https://www.trellix.com/en-us/about/newsroom/stories/research/tarfile-exploiting-the-world.html).

If you have further questions you may contact us through this projects lead researcher [Kasimir Schulz](mailto:kasimir.schulz@trellix.com).

PR #1900, opened 2022-11-04
sanitize tar archives due to CVE

osherdp:
Patching CVE-2007-4559
(https://github.com/advisories/GHSA-gw9q-c7gh-j9vm) which might allow attackers to overwrite arbitrary files which are not under destination path.

Implementation inspiration: https://github.com/openshift/assisted-test-infra/pull/1898

### DIFF
--- a/src/assisted_test_infra/test_infra/utils/logs_utils.py
+++ b/src/assisted_test_infra/test_infra/utils/logs_utils.py
@@ -1,6 +1,8 @@
+import contextlib
 import os
 import tarfile
 import time
+from pathlib import Path
 from tempfile import TemporaryDirectory
 
 import waiting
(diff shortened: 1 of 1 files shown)


src/assisted_test_infra/test_infra/utils/logs_utils.py
```diff
@@ -12,6 +15,25 @@
 OC_DOWNLOAD_LOGS_TIMEOUT = 60 * 60
 
 
+@contextlib.contextmanager
+def _safe_open_tar(tar_path: Union[str, Path], destination_path: Union[str, Path]) -> None:
```
eliorerz, line 19: You can now do:
```suggestion
def _safe_open_tar(tar_path: str | Path, destination_path: str | Path) -> None:
```

src/assisted_test_infra/test_infra/utils/logs_utils.py
```diff
@@ -140,39 +161,36 @@ def _verify_oc_logs_uploaded(cluster_tar_path):
 
 
 def _verify_node_logs_uploaded(dir_path, file_path):
-    gz = tarfile.open(os.path.join(dir_path, file_path))
-    logs = gz.getnames()
-    for logs_type in ["agent.logs", "installer.logs", "mount.logs"]:
-        assert any(logs_type in s for s in logs), f"{logs_type} isn't found in {logs}"
-    gz.close()
+    with tarfile.open(os.path.join(dir_path, file_path)) as gz:
```
eliorerz, line 164: Why not use `_safe_open_tar`?

osherdp: because it doesn't do extraction, only reads the names and asserts their values

eliorerz (approved):
LGTM

PR #1907, opened 2022-11-08
NO-ISSUE: Don't raise exception when VIPs DHCP allocation is disabled.

eliorerz:
/cc @vrutkovs

### DIFF
--- a/src/assisted_test_infra/test_infra/controllers/node_controllers/nutanix_controller.py
+++ b/src/assisted_test_infra/test_infra/controllers/node_controllers/nutanix_controller.py
@@ -126,12 +126,13 @@ class NutanixController(NodeController):
 
     def get_ingress_and_api_vips(self):
         if not self._entity_config.vip_dhcp_allocation:
-            if not self._entity_config.api_vip:
-                raise ValueError("API VIP is not set")
-            if not self._entity_config.ingress_vip:
-                raise ValueError("Ingress VIP is not set")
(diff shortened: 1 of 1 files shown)


src/assisted_test_infra/test_infra/controllers/node_controllers/nutanix_controller.py
```diff
@@ -126,12 +126,13 @@ def restart_node(self, node_name: str) -> None:
...
-            if not self._entity_config.api_vip:
-                raise ValueError("API VIP is not set")
-            if not self._entity_config.ingress_vip:
-                raise ValueError("Ingress VIP is not set")
-            return {"api_vip": self._entity_config.api_vip, "ingress_vip": self._entity_config.ingress_vip}
+            if self._entity_config.api_vip and self._entity_config.ingress_vip:
+                return {"api_vip": self._entity_config.api_vip, "ingress_vip": self._entity_config.ingress_vip}
+            return self._get_random_free_vips()
 
+        return {"api_vip": "", "ingress_vip": ""}
```
vrutkovs, line 133: Why would we want to return values which we know are incorrect?

vrutkovs:
/hold

Needs tests

src/assisted_test_infra/test_infra/controllers/node_controllers/nutanix_controller.py
eliorerz: The test will fail on VIPs validations, we can raise `NotImplementedError` if we want

PR #1909, opened 2022-11-08
vsphere: upload iso to test folder

vrutkovs:
Seems vSphere throws HTTP 500 when a file is uploaded. `test` folder is unaffected, so lets use that temporarily


eliorerz (approved):
LGTM

PR #1914, opened 2022-11-14
fix tput does not clear formatting

osherdp:
without it we're stuck with the red color in our shell even after 'make' returns.


eliorerz (approved):
LGTM

PR #1915, opened 2022-11-14
MGMT-8357: Support kind hub-cluster

osherdp:
This adds another option of setting up hub-cluster, this time using
'kind'. The cluster is being set-up the following way:
* It is constructed as a single-node cluster.
* The node mirrors its port to the outside host, meaning we have port mapping of 80->80.
* We install contour on the cluster, as our ingress controller. The three relevant services in our case (assisted-service, assisted-image-service and assisted-installer-ui) are creating ingress objects, to allow access to all of them via the open port.
* We are required to pass ``DEPLOY_TARGET`` with the relevant value, so that deployment scripts (and test-infra) know how to access the services.

Depends on https://github.com/openshift/assisted-service/pull/4613


romfreiman:
is this instead of minikube?

osherdp:
> is this instead of minikube?

yes, intends to replace minikube
but for now it won't be the default until we'll do a bit of beta testing

scripts/hub-cluster/delete.sh
```diff
@@ -0,0 +1,19 @@
+set -o nounset
```
adriengentil, line 1:
```suggestion
#!/usr/bin/env bash

set -o nounset
```

scripts/hub-cluster/create.sh
```diff
@@ -0,0 +1,19 @@
+set -o nounset
```
adriengentil, line 1:
```suggestion
#!/usr/bin/env bash

set -o nounset
```

scripts/hub-cluster/kind/kind.sh
```diff
@@ -0,0 +1,62 @@
...
+	echo "Installed successfully!"
+}
+
+function setup_contour() {
+	# based on https://kind.sigs.k8s.io/docs/user/ingress/#contour
+	# for more information about contour, see: https://projectcontour.io
+	kubectl apply -f https://projectcontour.io/quickstart/contour.yaml
+	kubectl patch daemonsets -n projectcontour envoy -p \
+		'{"spec":{"template":{"spec":{"nodeSelector":{"ingress-ready":"true"},"tolerations":[{"key":"node-role.kubernetes.io/control-plane","operator":"Equal","effect":"NoSchedule"},{"key":"node-role.kubernetes.io/master","operator":"Equal","effect":"NoSchedule"}]}}}}'
+	timeout 2m bash -c 'until kubectl wait --for=condition=ready pods -n projectcontour --selector=app=contour; do sleep 5s; done'
```
adriengentil, line 37: to check: maybe it would be more reliable to wait for the rollout for the deployment/daemonset as we did here:
https://github.com/openshift/assisted-service/pull/4489

Makefile
```diff
@@ -126,41 +118,49 @@ endif
...
-create_environment: image_build bring_assisted_service start_minikube
+create_environment: image_build bring_assisted_service create_hub_cluster
 
 image_build:
 	scripts/pull_dockerfile_images.sh
 	sed 's/^FROM .*assisted-service.*:latest/FROM $(subst /,\/,${SERVICE})/' Dockerfile.assisted-test-infra | \
 		$(CONTAINER_COMMAND) build --network=host -t $(IMAGE_NAME) -f- .
 	$(CONTAINER_COMMAND) tag $(IMAGE_NAME) test-infra:latest  # For backwards computability
 
+create_hub_cluster:
```
adriengentil, line 137: should probably added in the `.PHONY` (and for `delete_hub_cluster`,  `delete_kind` as well)

adriengentil (commented):
few nits, it looks good! :+1:

scripts/hub-cluster/kind/kind.sh
osherdp: This is a great idea!
And yes, it's working quite good

Makefile
```diff
@@ -5,12 +5,6 @@
 SHELL=/bin/sh
 CONTAINER_COMMAND := $(shell scripts/utils.sh get_container_runtime_command)
 
-ifneq (,$(findstring podman,$(CONTAINER_COMMAND)))
-    PULL_PARAM = --pull-always
-else
-    PULL_PARAM = --pull
-endif
```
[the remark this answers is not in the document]
osherdp: Yeah, sorry about it. It's not required for this change 
It actually seems like a redundant parameter (probably wasn't removed when its use got removed)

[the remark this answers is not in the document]
eliorerz: OK, seems fine to me, ping me when the PR will be ready

scripts/deploy_assisted_service.sh
```diff
@@ -140,7 +151,13 @@ else
     fi
 
     skipper run src/update_assisted_service_cm.py
-    (cd assisted-service/ && skipper --env-file ../skipper.env run "make deploy-all" ${SKIPPER_PARAMS} $ENABLE_KUBE_API_CMD DEPLOY_TAG=${DEPLOY_TAG} DEPLOY_MANIFEST_PATH=${DEPLOY_MANIFEST_PATH} DEPLOY_MANIFEST_TAG=${DEPLOY_MANIFEST_TAG} NAMESPACE=${NAMESPACE} AUTH_TYPE=${AUTH_TYPE} ${DEBUG_DEPLOY_AI_PARAMS:-})
+    if [ $DEPLOY_TARGET == minikube ]; then
+        export TARGET=local  # in assisted-service deployment scripts, minikube is annotated as 'local'
+    else
+        export TARGET=$DEPLOY_TARGET
+    fi
```
osherdp, line 158: I don't like those parts very particularly 
Maybe the idea of using ``local`` as code-name for ``minikube`` was making it more generalized, but now when adding additional local k8s option that has different networking arch it isn't completely interchangeable
 
I might be able to change occurrences of ``local`` in ``minikube`` in assisted-service, or maybe some other option I haven't thought of?

adriengentil: I guess `local` made sense when we only had one way to deploy locally.
Replacing it by `minikube` looks fine to me.

osherdp: Ack. Created https://github.com/openshift/assisted-service/pull/4642 to do the rename

osherdp:
Changed to draft as I'm waiting for https://github.com/openshift/assisted-service/pull/4642
(Tests won't pass until then, but you can still review the code)

eliorerz (commented):
LGTM

adriengentil (approved):

osherdp:
/test e2e-metal-assisted-none
we have a reason to suspect load-balancer is also using port 80, so testing it

osherdp:
/unhold
kind doesn't work right now with the load-balancer, but let's call it a known issue for now (it shouldn't affect minikube mode anyway)

PR #1916, opened 2022-11-14
MGMT-8707: Revert swtpm installation from source

osherdp:
Manual revert of
https://github.com/openshift/assisted-test-infra/pull/1748.

The fixed versions of swtpm packages has landed for RHEL 8.7+8.6 as well as latest Rocky Linux 8 version.

It means we should not have any kind of flaky behavior as we previously had in CI (might still happen locally, but either way disk-encryption is not heavily prioritized for testing and development).

/cc @nmagnezi @eliorerz

### DIFF
--- a/scripts/install_environment.sh
+++ b/scripts/install_environment.sh
@@ -41,50 +41,9 @@ function install_libvirt() {
         libvirt-devel \
         libvirt-daemon-kvm \
         qemu-kvm \
-        libgcrypt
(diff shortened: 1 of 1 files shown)


osherdp:
/test e2e-metal-assisted-tpmv2
just for good measure

nmagnezi (approved):
/lgtm

PR #1917, opened 2022-11-15
modifies update_oc_config for SNO cluster

eliajahshan:
@lalon4


osherdp:
/ok-to-test
@eliajahshan please give some meaningful description, even if it's a trivial change (what tests it affects, how the failures looks like, etc.)

osherdp:
/retitle Fix retrieving single-node clusters' API endpoint

PR #1918, opened 2022-11-15
add basic aws support

adriengentil:


terraform_files/aws/main.tf
```diff
@@ -0,0 +1,54 @@
...
+  user_data              = var.ipxe_script
+
+  tags = {
+    Name        = "${var.cluster_name}-worker-${count.index}"
+    ClusterName = "${var.cluster_name}"
+    NodeType    = "worker"
+  }
+
+  root_block_device {
+    volume_size = 100
```
osherdp, line 51: not sure if we can pick a smaller size, but workers can do with 20GB disks iirc

src/consts/consts.py
```diff
@@ -114,7 +114,7 @@ def all(cls):
 DEFAULT_LOAD_BALANCER_PORT = 6443
 DEFAULT_IPXE_SERVER_PORT = 8500
 DEFAULT_TANG_SERVER_PORT = 7500
-DEFAULT_IPXE_SERVER_IP = "192.168.122.1"
+DEFAULT_IPXE_SERVER_IP = "10.0.132.14"
```
osherdp, line 117: for afterwards: checking if the existing ipxe flow has been effected by it

adriengentil: it should take the private IP of the machine instead, but it's hardcoded at the moment :(

terraform_files/aws/main.tf
adriengentil: smaller, cheaper :)

terraform_files/aws/variables_aws.tf
```diff
@@ -0,0 +1,78 @@
...
+The resources should be associate with this name for easy recognition.
+EOF
+}
+
+
+// Inject iPXE script into user-data
+// #!ipxe
+// dhcp
+// chain http://boot.ipxe.org/demo/boot.php
+variable "ipxe_script" {
```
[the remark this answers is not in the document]
adriengentil: yeah, I did not know how to name it. In the context of the ipxe image, the data-user must contain an ipxe-script.

src/assisted_test_infra/test_infra/controllers/node_controllers/aws_controller.py
```diff
@@ -0,0 +1,155 @@
...
+        raise NotImplementedError
+
+    def shutdown_all_nodes(self) -> None:
+        raise NotImplementedError
+
+    def restart_node(self, node_name: str) -> None:
+        raise NotImplementedError
+
+    def get_ingress_and_api_vips(self):
+        return None
```
adriengentil, line 83: @osherdp FYI returning `None` here triggered UMN, I did not had to set the none platform.

adriengentil:
/close
Closing to avoid updated job to the triggered on it.
We'll need to follow-up on this PR once we decide to spend effort supporting AWS.

PR #1919, opened 2022-11-16
Fix network interface getting removed intermediately

osherdp:
It is possible to try reading a network interface that has been removed
while ``_fill_allocated_ips_and_bridges_by_interface`` traverses the
network interfaces. On those cases we get:
```
    def _fill_allocated_ips_and_bridges_by_interface(self):
        for interface in netifaces.interfaces():
            self._add_allocated_net_bridge(interface)
>           for ifaddresses in
netifaces.ifaddresses(interface).values():
E           ValueError: You must specify a valid interface name.
```

This changes the code to allow getting ``ValueError`` and continueing to
the next interface.

/cc @eliorerz @lalon4


lalon4:
/lgtm !
thanks for your help!

eliorerz (approved):
LGTM

PR #1925, opened 2022-11-22
Using the new base python image

osherdp:
Change the way we're installing the desired Python version. Instead of doing ``make install`` with provided code from the base image, this just uses a base image which has Python installed.
It reduces image size from 2.91GB to 2.54GB, as well as installation times (from 8m17 to 5m46).

### DIFF
--- a/Dockerfile.assisted-test-infra
+++ b/Dockerfile.assisted-test-infra
@@ -1,6 +1,6 @@
 FROM quay.io/edge-infrastructure/assisted-service:latest AS service
 
-FROM quay.io/centos/centos:stream9
+FROM quay.io/assisted-installer-ops/base-python:3.11.0
 
 # A directory in the path with write permission even for non-root users
 ENV TOOLS=/tools/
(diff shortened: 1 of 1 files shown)


Dockerfile.assisted-test-infra
```diff
@@ -1,6 +1,6 @@
 FROM quay.io/edge-infrastructure/assisted-service:latest AS service
 
-FROM quay.io/centos/centos:stream9
+FROM quay.io/assisted-installer-ops/base-python:3.11.0
```
adriengentil, line 3: should we move the dockerfile behind this image publicly in github ?

osherdp: You mean moving only the dockerfile, or moving the entire repo from gitlab to github and also using some other ci/cd env?

adriengentil: yeah I meant with the ci/cd, that's sad because we spend so much time to build it with gitlab

PR #1928, opened 2022-11-23
MGMT-12670: Set terraform parameters only on `create_nodes` and `destroy_nodes`

eliorerz:
This allowing us to update node configurations after the cluster was created but before creating the nodes
e.g.
```python
class TestInstall(BaseTest):

    def test_install(self, cluster, controller_configuration):
        min_host_ocp_requirements = cluster.get_preflight_requirements()
        controller_configuration.master_vcpu = min_host_ocp_requirements.ocp.master.quantitative.cpu_cores

```

/cc @osherdp

### DIFF
--- a/src/assisted_test_infra/test_infra/controllers/node_controllers/terraform_controller.py
+++ b/src/assisted_test_infra/test_infra/controllers/node_controllers/terraform_controller.py
@@ -27,9 +27,9 @@ class TerraformController(LibvirtController):
         self._suffix = self._entity_name.suffix or get_name_suffix()
         self.tf_folder = config.tf_folder or self._create_tf_folder(self._entity_name.get(), config.tf_platform)
         self.network_name = config.network_name + self._suffix
-        self.params = self._terraform_params(**config.get_all())
         self.tf = terraform_utils.TerraformUtils(working_dir=self.tf_folder)
         self.master_ips = None
+        self._params = Munch()
 
     @property
     def entity_name(self) -> BaseName:
@@ -39,6 +39,9 @@ class TerraformController(LibvirtController):
     def cluster_name(self) -> str:
         warnings.warn("cluster_name is deprecated. Use Controller.entity_name instead.", DeprecationWarning)
         return self._entity_name.get()
+
+    def _get_params_from_config(self) -> Munch:
+        return self._terraform_params(**self._config.get_all())
 
     def _create_tf_folder(self, name: str, platform: str):
         if isinstance(self._entity_config, BaseInfraEnvConfig):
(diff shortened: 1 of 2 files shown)


src/assisted_test_infra/test_infra/controllers/node_controllers/terraform_controller.py
```diff
@@ -316,12 +318,15 @@ def _delete_virsh_resources(cls, *filters):
         virsh_cleanup.clean_virsh_resources(skip_list=skip_list, resource_filter=filters)
 
     def prepare_nodes(self):
+        self._params = self._terraform_params(**self._config.get_all())
```
osherdp, line 321: if ``_params`` is something that should always dynamically infer its value from ``_config``, maybe we should make it a property:
```
@property
def _params(self):
    return self._terraform_params(**self._config.get_all())
```

Otherwise it's hard to remember having this line whenever we're using it (think of new methods added to the controller class)

osherdp (changes requested):

src/assisted_test_infra/test_infra/controllers/node_controllers/terraform_controller.py
eliorerz: I thought doing that, but we don't want to get it dynamically every time we want to access some variable, that is why we save params value on protected property and updating it only when destroying/creating nodes.
I'll add method for getting params from config. that might be better

src/assisted_test_infra/test_infra/controllers/node_controllers/terraform_controller.py
```diff
@@ -40,6 +40,9 @@ def cluster_name(self) -> str:
         warnings.warn("cluster_name is deprecated. Use Controller.entity_name instead.", DeprecationWarning)
         return self._entity_name.get()
 
+    def _get_params_from_config(self):
```
osherdp, line 43: add type annotation

osherdp (changes requested):

src/assisted_test_infra/test_infra/controllers/node_controllers/terraform_controller.py
eliorerz: OK, I'll wait that all of the jobs will pass and if so I'll update it and will override the jobs

eliorerz:
/override ci/prow/e2e-metal-assisted-ipv4v6
/override ci/prow/e2e-metal-assisted-ipv6
/override ci/prow/e2e-metal-assisted-none
/override ci/prow/e2e-metal-assisted-static-ip-suite 

osherdp (approved):

PR #1929, opened 2022-11-28
NO-ISSUE: Gather logs using sos report on nodes

adriengentil:
Use sos report along with toolbox to gather the logs from cluster
nodes

### DIFF
--- a/src/assisted_test_infra/download_logs/resources/man_sosreport.sh
+++ b/src/assisted_test_infra/download_logs/resources/man_sosreport.sh
@@ -1,135 +1,15 @@
 #!/bin/bash
-export LANG=C
 
-# If this script hangs, un-comment the below two entries and note the command that the script hangs on.  Then comment out that command and re-run the script.
-# set -x
-# set -o verbose
+SOS_TMPDIR="/var/tmp"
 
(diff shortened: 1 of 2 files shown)


src/assisted_test_infra/download_logs/resources/man_sosreport.sh
```diff
@@ -1,135 +1,15 @@
...
 
-echo -e "Gathering system information..."
-hostname &> hostname
-cp -a /etc/redhat-release  ./etc/ 2>> error_log
-uptime &> uptime
+sudo toolbox sos report --batch --tmp-dir "${SOS_TMPDIR}" --compression-type xz --all-logs \
+                   -o processor,memory,container_log,filesys,logs,crio,podman,openshift,openshift_ovn,networking,networkmanager,rhcos \
+                   -k crio.all=on -k crio.logs=on \
+                   -k podman.all=on -k podman.logs=on \
+                   -k openshift.with-api=on
```
osherdp, line 12: I saw there's an option to set a global timeout for all plugins
Maybe we should do so to make sure we always have as much data as we can?

adriengentil:
/close
in favour of https://github.com/openshift/assisted-test-infra/pull/1943

PR #1934, opened 2022-11-30
Add return to get_disks

bkopilov:
Current nody.py get_disks function returns none instead of disks Fixed and added return.

### DIFF
--- a/src/assisted_test_infra/test_infra/controllers/node_controllers/node.py
+++ b/src/assisted_test_infra/test_infra/controllers/node_controllers/node.py
@@ -172,7 +172,7 @@ class Node:
         self.set_ram_kib(self.original_ram_kib)
 
     def get_disks(self):
-        self.node_controller.list_disks(self.name)
+        return self.node_controller.list_disks(self.name)
 
     def attach_test_disk(self, disk_size, **kwargs):
         return self.node_controller.attach_test_disk(self.name, disk_size, **kwargs)


bkopilov:
@osherdp @eliorerz

src/assisted_test_infra/test_infra/controllers/node_controllers/node.py
```diff
@@ -172,7 +172,7 @@ def reset_ram_kib(self):
         self.set_ram_kib(self.original_ram_kib)
```
osherdp, line 173: Can you please add return type annotation to this function?

osherdp (changes requested):

src/assisted_test_infra/test_infra/controllers/node_controllers/node.py
osherdp: It's just something we're trying to improve over time, to help IDE do auto-completion in our code

bkopilov: Well maybe we can switch to properties ...  but not at this patch :)

osherdp: properties?
I mean:
```
def get_disks(self) -> List[<whatever it is>]:
```

and ``List`` imported via ``from typing import List``

bkopilov: I mean in python , do somthing like:

@property
def disks(self):
        self.node_controller.list_disks(self.name)

@disks.setter
def disk(self,....)

In case we have a simple get method and the name is in self we can use it ... 
about annotations ,
Need add anotations to all functions , can we do it in a seperate patch .. ?  i can do it if you like
.... 

or similar 
get_disks = property(get_disks)

It will allow us to call to the functions from instance like:
instance.disks

osherdp: yeah but you added this properties idea like it's connected to what I asked... it has almost nothing to do with it :) 

either way I'm okay this time if you'll do it in a separate patch
just bear in mind we prefer grouping relating changes in one PR instead of splitting it across multiple PRs. Especially when talking about such small change of adding a type annotation to a specific function

osherdp:
/retest
(just equinix missing machines, nothing relevant to those changes)

osherdp (approved):

src/assisted_test_infra/test_infra/controllers/node_controllers/node.py
bkopilov: yee agree , let me fix the while class in a single patch ...  a good start to learn this repo

bkopilov: And of course i will have to comment and fix by your requests to keep the review process as needed.

PR #1942, opened 2022-12-08
Allow 'onprem' deploy-target in hub-cluster setup

osherdp:
Back in https://github.com/openshift/assisted-test-infra/pull/1915 I forgot to allow setup for onprem/podman installation mode.

This should fix it.


eliorerz (approved):
LGTM

PR #1943, opened 2022-12-08
MGMT-11712: Gather logs using sos report on nodes

vrutkovs:
Use sos report along with toolbox to gather the logs from cluster
nodes.

https://github.com/openshift/assisted-test-infra/pull/1929 + feedback addressed


vrutkovs:
/hold

Not sure if this would reliably work with disconnected

vrutkovs:
/hold cancel

SOS report now has standard format

vrutkovs:
cc @eliorerz 

This also changes `is_active` implementation of vsphere/nutanix provider to assume all nodes are up (later on we'll need to use API to determine that, but for now we'll just not receive the sos report if the node is down)

adriengentil (approved):

vrutkovs:
Right, nutanix will be fixed in https://github.com/openshift/release/pull/35548
/skip

PR #1947, opened 2022-12-12
Add drop cache when change nodes state

bkopilov:
Current iptables fixture tears down nodes in order modify iptables. When the machines started again the cached info must be refreshed.

In case machines state changed like stop/reboot cashed info may hide bugs in the node controllers or fail tests.

### DIFF
--- a/src/tests/base_test.py
+++ b/src/tests/base_test.py
@@ -648,9 +648,19 @@ class BaseTest:
             log.info(f"Given node ips: {given_node_ips}")
 
             for _rule in iptables_rules:
-                _rule.add_sources(given_node_ips)
-                rules.append(_rule)
-                _rule.insert()
+                if not given_node_ips:
+                    # in case a rule without given_node_ips
+                    rules.append(_rule)
+                    _rule.insert()
+                    continue
+                for source in given_node_ips:
+                    # iptables check broken for list of sources
+                    # copy base rule and add a single source.
+                    copy_rule = deepcopy(_rule)
+                    copy_rule.add_sources([source])
+                    log.info(f"Adding rule with source: {copy_rule}")
+                    rules.append(copy_rule)
+                    copy_rule.insert()
 
         yield set_iptables_rules_for_nodes
         log.info("---TEARDOWN iptables ---")
--- a/src/assisted_test_infra/test_infra/controllers/iptables.py
+++ b/src/assisted_test_infra/test_infra/controllers/iptables.py
@@ -62,7 +62,8 @@ class IptableRule:
             utils.run_command(insert_rule, shell=True)
 
     def delete(self) -> None:
-        if self._does_rule_exist():
-            delete_rule = self._build_command_string(IpTableCommandOption.DELETE)
+        delete_rule = self._build_command_string(IpTableCommandOption.DELETE)
+        # delete duplicate rules - iptable deletes only first hit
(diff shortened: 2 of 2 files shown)


bkopilov:
@osherdp @lalon4 

Please review

adriengentil:
/ok-to-test

Would it make sense to drop the cache directly in `shutdown_given` and `shutdown_all`? So everytime shutdown is called, the cache would be dropped and potentially prevent potential discrepancies in other parts of test-infra.
https://github.com/openshift/assisted-test-infra/blob/12e7865c87c4b93e5ee26d9d7ad85ddeb1fcd330/src/assisted_test_infra/test_infra/helper_classes/nodes.py#L112

bkopilov:
> /ok-to-test
> 
> Would it make sense to drop the cache directly in `shutdown_given` and `shutdown_all`? So everytime shutdown is called, the cache would be dropped and potentially prevent potential discrepancies in other parts of test-infra.
> 
> https://github.com/openshift/assisted-test-infra/blob/12e7865c87c4b93e5ee26d9d7ad85ddeb1fcd330/src/assisted_test_infra/test_infra/helper_classes/nodes.py#L112

[bkopilov] Yes , the cache is reall risky if something goes wrong during tests or events.

src/tests/base_test.py
```diff
@@ -644,7 +644,8 @@ def set_iptables_rules_for_nodes(
             for node in given_nodes:
                 given_node_ips.append(node.ips[0])
             cluster.nodes.shutdown_given(given_nodes)
-
+            # Clear cached info about nodes after shutdown
+            cluster.nodes.drop_cache()
```
osherdp, line 648: Why we need to do the shutdown in the first place? Any chance you know?

bkopilov: The iptables fixture from base_test.py run shutdown, no idea if we really need shutdown.
We have failures in CI in tests that use this fixture and i suspect that when the libvirt comes back we dont have the latest ip.

osherdp: So let's try to remove the shutdown if it serves no purpose
There's no reason to just do more actions and increase our chances of failure

bkopilov: How do you recommend to fix all other cases when we stop/ reboot nodes during test lifetime ?

lalon4: No, lets not remove the shutdown, it was implemented long time ago, there was a reason - if i am not mistaken if for the rules to apply, proxy tests are highly depended on it

osherdp: Let me check this a bit
Either way Adrien's suggestion to embed the cache-drop in the shutdown function is a better way to construct this solution

bkopilov: Ack , yes Adrien's change will cover our needs but please note that we still may have an issue if user will try to call to Node().shutdown() .  but it good enough ,if no objection i am going to drop_cahce for: shutdown and reboot

osherdp: Seems like a good compromise

osherdp:
Code has changed into another change
For now I'm waiting for it to go back to original requested change
/hold

osherdp:
@bkopilov can we close this PR?
I see there's a replacement

bkopilov:
> 



> 



> @bkopilov can we close this PR? I see there's a replacement

Yes please .

PR #1949, opened 2022-12-12
MGMT-12439: Deploy heterogeneous infrastructure

adriengentil:
This PR contains:
- a terraform module able to provision multiple devices from equinix metal
- Two ansible playbooks: 
  -  create heterogeneous clusters, which is responsible to provision the devices from equinix using the terraform module above, and soon, configure both of the machines (ipip tunnel and libvirt)
  - destroy the machines created by the first playbook

Here is CI definition if you want to check how the playbooks are used: https://github.com/openshift/release/tree/master/ci-operator/step-registry/assisted/equinix

### DIFF
--- a/Dockerfile.assisted-test-infra
+++ b/Dockerfile.assisted-test-infra
@@ -7,6 +7,11 @@ ENV TOOLS=/tools/
 ENV PATH="$TOOLS:$PATH"
 RUN mkdir $TOOLS --mode g+xw
 
+# TODO: Remove once OpenShift CI supports it out of the box (see https://access.redhat.com/articles/4859371)
+RUN chmod g+w /etc/passwd && \
+    echo 'echo default:x:$(id -u):$(id -g):Default Application User:/alabama:/sbin/nologin\ >> /etc/passwd' > /usr/local/bin/fix_uid.sh && \
+    chmod g+rwx /usr/local/bin/fix_uid.sh
+
 # CRB repo is required for libvirt-devel
 RUN dnf -y install --enablerepo=crb \
   make \
@@ -53,7 +58,9 @@ ENV PATH=$PATH:/usr/local/go/bin:/go/bin
 
 COPY . .
 
-RUN chgrp -R 0 /home/assisted-test-infra && \
+# init terraform in order to cache the provider
+RUN terraform -chdir=/home/assisted-test-infra/terraform_files/equinix-ci-machine init && \
+    chgrp -R 0 /home/assisted-test-infra && \
     chmod -R g=u /home/assisted-test-infra
 
 # setting pre-commit env
--- a/ansible_files/equinix_heterogeneous_create_infra_playbook.yml
+++ b/ansible_files/equinix_heterogeneous_create_infra_playbook.yml
@@ -1,0 +1,38 @@
+- name: Create and configure the equinix infrastructure to run heterogneous cluster tests
+  hosts: localhost
+  vars_prompt:
+    - name: equinix_project_id
+      prompt: Equinix metal project ID in which the device(s) will be created
+      private: false
+    - name: equinix_auth_token
+      prompt: Equinix metal API token
+    - name: equinix_ssh_private_key_path
+      prompt: Path to an SSH private key allowed to connect on the created device(s)
+      private: false
+    - name: equinix_unique_id
+      prompt: Hostname prefix which will be used to identify the device(s)
+      private: false
+    - name: equinix_tf_vars_file
+      prompt: Place where the Terrafom variable file will be stored
+      private: false
+    - name: equinix_tf_state_file
+      prompt: Place where the Terrafom state file will be stored
+      private: false
+  vars:
+    terraform_equinix_workdir: "{{ [playbook_dir, '..', 'terraform_files', 'equinix-ci-machine'] | path_join | realpath }}"
+  tasks:
+    # Create the required devices to test heterogeneous scenarios
+    - ansible.builtin.import_role:
+        name: equinix/heterogeneous_create_infra
+    # Export the connection details, required for the common steps in Prow
+    - ansible.builtin.import_role:
+        name: equinix/export_primary_device_connection_details
+    # Build an IPIP tunnel between both of the machines in "heterogeneous" group
+    - ansible.builtin.import_role:
+        name: common/setup_ipip_tunnel
+      vars:
+        ipip_group: heterogeneous
+    # - ansible.builtin.import_role:
+    #     name: equinix/install_libvirt
+    # - ansible.builtin.import_role:
+    #     name: equinix/expose_libvirt
--- a/ansible_files/roles/common/setup_ipip_tunnel/tasks/main.yml
+++ b/ansible_files/roles/common/setup_ipip_tunnel/tasks/main.yml
@@ -1,0 +1,9 @@
+- name: Check if group "{{ ipip_group }}" exists and contains exactly 2 hosts
+  ansible.builtin.fail:
+    msg: Group "{{ ipip_group }}" must exist and contain exactly 2 hosts
+  when: groups[ipip_group] is not defined or (groups[ipip_group] | length != 2)
+
+- name: Ping hosts in group "{{ ipip_group }}"
+  ansible.builtin.ping:
+  delegate_to: "{{ item }}"
+  loop: "{{ groups[ipip_group] }}"
--- a/ansible_files/roles/equinix/export_primary_device_connection_details/tasks/main.yml
+++ b/ansible_files/roles/equinix/export_primary_device_connection_details/tasks/main.yml
@@ -1,0 +1,11 @@
+- name: Check if primary group exists and contains exactly 1 host
+  ansible.builtin.fail:
+    msg: Group primary must exits and contain exactly 1 host
+  when: groups[primary_device_group_name] is not defined or (groups[primary_device_group_name] | length != 1)
+
+- name: "Export connection details of primary device"
+  ansible.builtin.template:
+    src: "ci-machine-config.sh.j2"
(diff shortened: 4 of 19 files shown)


ansible_files/equinix_heterogeneous_create_infra_playbook.yml
```diff
@@ -0,0 +1,38 @@
...
+      prompt: Place where the Terrafom variable file will be stored
+      private: false
+    - name: equinix_tf_state_file
+      prompt: Place where the Terrafom state file will be stored
+      private: false
+  vars:
+    terraform_equinix_workdir: "{{ [playbook_dir, '..', 'terraform_files', 'equinix-ci-machine'] | path_join | realpath }}"
+  tasks:
+    # Create the required devices to test heterogeneous scenarios
+    - ansible.builtin.import_role:
```
osherdp, line 25: just something I don't understand,
what's the difference between this syntax (with tasks and role-importing) and running it like in the other playbook?
```
  roles:
    - equinix/generic_destroy_infra
```

ansible_files/roles/common/setup_ipip_tunnel/tasks/main.yml
```diff
@@ -0,0 +1,9 @@
+- name: Check if group "{{ ipip_group }}" exists and contains exactly 2 hosts
+  ansible.builtin.fail:
+    msg: Group "{{ ipip_group }}" must exits and contain exactly 2 hosts
```
osherdp, line 3:
```suggestion
    msg: Group "{{ ipip_group }}" must exist and contain exactly 2 hosts
```

ansible_files/roles/equinix/export_primary_device_connection_details/tasks/main.yml
```diff
@@ -0,0 +1,11 @@
+- name: Check if primary group contains exists and contains exactly 1 host
```
osherdp, line 1:
```suggestion
- name: Check if primary group exists and contains exactly 1 host
```

terraform_files/equinix-ci-machine/variables_equinix.tf
```diff
@@ -0,0 +1,45 @@
...
+}
+
+variable "devices" {
+  description = "The total configuration, List of Objects/Dictionary"
+  type = list(
+    object({
+      plan             = string
+      facilities       = optional(list(string), null)
+      operating_system = optional(string, null)
+      project_id       = optional(string, null)
```
osherdp, line 41: on what cases we'll want those to be optional/null?

osherdp (commented):
gave it a little swipe
mainly questions and nits

ansible_files/equinix_heterogeneous_create_infra_playbook.yml
```diff
@@ -0,0 +1,47 @@
...
+  hosts: localhost
+  vars_prompt:
+    - name: equinix_project_id
+      prompt: Equinix metal project ID in which the device(s) will be created
+      private: false
+    - name: equinix_auth_token
+      prompt: Equinix metal API token
+    - name: equinix_private_ssh_key_path
+      prompt: Path to an SSH private key allowed to connect on the created device(s)
+      private: false
```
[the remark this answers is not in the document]
adriengentil: if these variables are not defined on the command, the user will be prompted  to fill them, `private` tells if the user's input will be printed on the terminal or not while he is typing (like when SSH asks you for a password).

ansible_files/equinix_heterogeneous_create_infra_playbook.yml
```diff
@@ -0,0 +1,47 @@
...
+  vars:
+    terraform_equinix_workdir: "{{ (playbook_dir + '/../terraform_files/equinix-ci-machine') | realpath }}"
+  tasks:
+    - debug:
+        var: terraform_equinix_workdir
+
+    - debug:
+        var: playbook_dir
+
+    - ansible.builtin.pause:
```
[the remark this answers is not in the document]
adriengentil: yes, for PEBKAC bugs :see_no_evil:

ansible_files/equinix_heterogeneous_create_infra_playbook.yml
adriengentil: it's the same meaning, when using the `roles` it will runs the roles on the `hosts` specified at playbook level. The `tasks` directive gives more control on with hosts we want to run the role.

terraform_files/equinix-ci-machine/variables_equinix.tf
adriengentil: yeah, you're right, we can offload most of the templating to jinja/ansible, it would be more maintainable than mixing both.

ansible_files/equinix_heterogeneous_create_infra_playbook.yml
osherdp: and in this case it will allow using only hosts generated by ``equinix/heterogeneous_create_infra`` and it will ignore ``localhost``?

adriengentil: it'll be localhost exception if I specify a host, I'll do it with the libvirt role because I want to apply it on the arm machine only.

Also, I'm still discovering stuff, so it may change again :D

adriengentil:
We are almost good :partying_face: , the job provisioned the machines, ran the common workflow (failed because I'm using the cheapest machines available for testing), and destroyed the machines.

ansible_files/equinix_heterogeneous_create_infra_playbook.yml
adriengentil: ah also to pass variables, we need the `import_roles`, see below for the ipip setup.

terraform_files/equinix-ci-machine/variables_equinix.tf
adriengentil: ok I moved the heavy lifting in ansible, it's cleaner I think now, thanks!

Dockerfile.assisted-test-infra
```diff
@@ -53,7 +59,9 @@ ENV PATH=$PATH:/usr/local/go/bin:/go/bin
 
 COPY . .
 
-RUN chgrp -R 0 /home/assisted-test-infra && \
+# init terraform in order to cache the provider
+RUN terraform -chdir=/home/assisted-test-infra/terraform_files/equinix-ci-machine init && \
```
adriengentil, line 63: I don't know if there is a better solution, it's required to have the provider already in place when destroying the machines using ansible.

ansible_files/roles/common/setup_ipip_tunnel/tasks/main.yml
```diff
@@ -0,0 +1,9 @@
+- name: Check if group "{{ ipip_group }}" exists and contains exactly 2 hosts
```
adriengentil, line 1: It's just a placeholder at the moment, I'll push the full role with another PR

ansible_files/roles/equinix/heterogeneous_create_infra/defaults/main.yaml
```diff
@@ -0,0 +1,9 @@
+primary_device_group_name: "primary"
+secondary_device_group_name: "secondary"
+equinix_ssh_user: "root"
+equinix_plans: "{{
+      {
+       primary_device_group_name : 'c3.small.x86',
+       secondary_device_group_name : 'c3.small.x86'
```
adriengentil, line 7: The plans are not definitive, it's just they are cheap. I'll change them with another once the full playbook is ready.

adriengentil:
/hold
the update of terraform is not seamless, I'll update the version separately in another PR

ansible_files/equinix_heterogeneous_create_infra_playbook.yml
adriengentil: actually, no need for `import_task` to pass variables: https://github.com/openshift/assisted-test-infra/pull/1958/files#diff-5f56cc8fd10b4a77b4dc1326b5d54afde9b2c4b0ae285efab6a069c072505f73R23-R43

adriengentil:
/unhold
`e2e-metal-assisted-day2-arm-workers` is expected to fail but it correctly provisioned and destroyed the devices from equinix

ansible_files/equinix_heterogeneous_create_infra_playbook.yml
adriengentil: I'll tackle that in the coming IPIP PR

ansible_files/vars/ci_equinix_infrastucture.yml
```diff
@@ -0,0 +1,7 @@
+# Equinix related vars, built from CI vars
+equinix_project_id: "{{ lookup('file', cluster_profile_dir + '/packet-project-id') | trim }}"
+equinix_auth_token: "{{ lookup('file', cluster_profile_dir + '/packet-auth-token') | trim }}"
+equinix_ssh_private_key_path: "{{ cluster_profile_dir }}/packet-ssh-key"
```
osherdp, line 4: I'm a bit unsure about this file belonging here in test-infra. I think things like ``packet-ssh-key`` that are configured once in ``openshift/release`` should be mentioned only there

Either way, it's not a blocker. Just wanted to add a note about it

osherdp (approved):

ansible_files/vars/ci_equinix_infrastucture.yml
adriengentil: If needed, they can be overridden from the release config using the `ANSIBLE_EXTRA_VARS` environment variable: https://github.com/openshift/release/blob/master/ci-operator/step-registry/assisted/equinix/setup/assisted-equinix-setup-ref.yaml#L19-L21

It should help if we want to make consistent changes in the `release` repo without changing this one.

PR #1953, opened 2022-12-15
Fix iptables fixture and delete action

bkopilov:
Iptables check does not work when having a chain of sources
    When we insert a rule with multiple source, iptables split the list
    and adding each source with same conditions

    On check action iptables wont catch case when having a mix of source
    when 1 is not in the table. looks like iptables check only one.
    In case check action returned wrong result add/delete rule chnaged and
    may trigger failures due to unexpected results

    Also deletion , iptables delete only the first match when having duplicate
    rules.

    Example:
    iptables --insert FORWARD -p tcp -j DROP -s 192.168.128.10,192.168.128.13
    results with:
    -A FORWARD -s 192.168.128.13/32 -p tcp -m tcp --dport 80 -j DROP
    -A FORWARD -s 192.168.128.10/32 -p tcp -m tcp --dport 80 -j DROP

    We will get different result from iptables for:
    iptables --check FORWARD -p tcp -j DROP -s 7.7.7.7,192.168.128.10
    iptables --check FORWARD -p tcp -j DROP -s 192.168.128.10,7.7.7.7

    Means that only 1 is the best approch here.

### DIFF
--- a/src/assisted_test_infra/test_infra/controllers/iptables.py
+++ b/src/assisted_test_infra/test_infra/controllers/iptables.py
@@ -62,7 +62,8 @@ class IptableRule:
             utils.run_command(insert_rule, shell=True)
 
     def delete(self) -> None:
-        if self._does_rule_exist():
-            delete_rule = self._build_command_string(IpTableCommandOption.DELETE)
+        # delete duplicate rules when exists
+        delete_rule = self._build_command_string(IpTableCommandOption.DELETE)
+        while self._does_rule_exist():
             log.info(f"Removing iptable rule: {delete_rule}")
             utils.run_command(delete_rule, shell=True)
--- a/src/tests/base_test.py
+++ b/src/tests/base_test.py
@@ -2,6 +2,7 @@ import json
 import os
 import shutil
 from contextlib import suppress
+from copy import deepcopy
 from pathlib import Path
 from typing import Callable, Dict, List, Optional, Tuple, Type, Union
 
@@ -648,9 +649,18 @@ class BaseTest:
             log.info(f"Given node ips: {given_node_ips}")
 
             for _rule in iptables_rules:
-                _rule.add_sources(given_node_ips)
-                rules.append(_rule)
-                _rule.insert()
+                # Add rule without given sources
+                if not given_node_ips:
+                    rules.append(_rule)
+                    _rule.insert()
+                    continue
+                # iptables check does not work for chain of sources
+                for source in given_node_ips:
+                    copy_rule = deepcopy(_rule)
+                    copy_rule.add_sources([source])
+                    log.info(f"Adding rule with source: {copy_rule}")
+                    rules.append(copy_rule)
+                    copy_rule.insert()
 
         yield set_iptables_rules_for_nodes
         log.info("---TEARDOWN iptables ---")


lalon4:
/hold
Still need some testing

bkopilov:
> /hold Still need some testing

Just to sync , 
we has iptables versions issue between hypervisor 8.6 and centos 9,  problem addresses here :  https://github.com/openshift/assisted-test-infra/pull/1962

Let me know if we can move on with this patch

osherdp:
> > /hold Still need some testing
> 
> Just to sync , we has iptables versions issue between hypervisor 8.6 and centos 9, problem addresses here : #1962
> 
> Let me know if we can move on with this patch

I'm still waiting for the reproduction steps you've promised me :)

bkopilov:
> iptables --check FORWARD -p tcp -j DROP -s 7.7.7.7,192.168.128.10

POC why the iptables --check can not work on multiple sources !!!

# create two rules in iptables
iptables --insert FORWARD -p tcp -j DROP -s 1.1.1.1,2.2.2.2
# delete 1.1.1.1 
iptables --delete FORWARD -p tcp -j DROP -s 1.1.1.1

# here is the entry in iptable
-A FORWARD -s 2.2.2.2/32 -p tcp -j DROP
-----------------------------------
iptables --check FORWARD -p tcp -j DROP -s 1.1.1.1,2.2.2.2

iptables --check FORWARD -p tcp -j DROP -s 2.2.2.2,1.1.1.1
 
[root@ocp-edge82 ~]# iptables --check FORWARD -p tcp -j DROP -s 2.2.2.2,1.1.1.1
iptables: Bad rule (does a matching rule exist in that chain?).


The check command resturns different results when order changed means it supports only 1 source and we can not rely on it as condition to add/ remove rules.
Its going to be painfull in a corner cases.

As for automation testcase, i did not get to it but we need to close this hole in our code :)

bkopilov:
> > iptables --check FORWARD -p tcp -j DROP -s 7.7.7.7,192.168.128.10
> 
> POC why the iptables --check can not work on multiple sources !!!
> 
> create two rules in iptables
> iptables --insert FORWARD -p tcp -j DROP -s 1.1.1.1,2.2.2.2
> 
> delete 1.1.1.1
> iptables --delete FORWARD -p tcp -j DROP -s 1.1.1.1
> 
> here is the entry in iptable
> -A FORWARD -s 2.2.2.2/32 -p tcp -j DROP
> 
> iptables --check FORWARD -p tcp -j DROP -s 1.1.1.1,2.2.2.2
> 
> iptables --check FORWARD -p tcp -j DROP -s 2.2.2.2,1.1.1.1
> 
> [root@ocp-edge82 ~]# iptables --check FORWARD -p tcp -j DROP -s 2.2.2.2,1.1.1.1
> 
> iptables: Bad rule (does a matching rule exist in that chain?).
> 
> The check command resturns different results when order changed means it supports only 1 source and we can not rely on it as condition to add/ remove rules. Its going to be painfull in a corner cases.
> 
> As for automation testcase, i did not get to it but we need to close this hole in our code :)

osherdp:
@bkopilov those are not reproduction steps
reproduction steps should show how you can reach the problematic state using the code that you're changing. meaning it should either be:
```
make test <THIS> <THAT>
```
or:
```
        rule1 = IptableRule(...)
        rule2 = IptableRule(...)
        iptables(new_cluster, [rule1, rule2], None)
```

the reason for such requirement, is for others to get to the same state if in the future we're regressing 
please, do that thing for both me (so I can effectively review it) and for others in the future that will thank you for being so elaborating :)

bkopilov:
> @bkopilov those are not reproduction steps reproduction steps should show how you can reach the problematic state using the code that you're changing. meaning it should either be:
> 
> ```
> make test <THIS> <THAT>
> ```
> 
> or:
> 
> ```
>         rule1 = IptableRule(...)
>         rule2 = IptableRule(...)
>         iptables(new_cluster, [rule1, rule2], None)
> ```
> 
> the reason for such requirement, is for others to get to the same state if in the future we're regressing please, do that thing for both me (so I can effectively review it) and for others in the future that will thank you for being so elaborating :)

ack , 
we can close this issue , we just need to remeber that iptable --check does not work if multiple sources and one source is removed which wont happen the way the code is written.

osherdp:
As you wish
closing for now
/close

PR #1954, opened 2022-12-15
Current iptables fixture tears down nodes in order modify iptables.

bkopilov:
When the machines started again the cached info must be refreshed.

In case machines state changed like stop/reboot cashed info may hide bugs in the node controllers or fail tests.

This patch is continuation of https://github.com/openshift/assisted-test-infra/pull/1947 Due to some mismach i created new patch.


bkopilov:
@lalon4 , @osherdp @adriengentil

osherdp:
flake. retesting
/test e2e-metal-assisted

osherdp (approved):

PR #1956, opened 2022-12-18
Fix none platform domain name - unable to resolve redhat.com

bkopilov:
In current deployment nslookup to redhat.com will fail.
Change the network domain base on cluster name.
The network domain should not be redhat.com because we will not be able to pull images from libvirt nodes.

when we set domain name to redhat.com, dns resolving wont work  for the domain and libvirt nodes will not be able to pull the images due to resolving issue.


bkopilov:
@lalon4 , @osherdp 

once we fix it , all nodes will be discovered by the service but we still have an issue due to the loadbalancer we call to.

osherdp:
Change look good. Just two things:
* fix non -> none in commit and PR title
* fix merge-conflicts. seems like there's a recent change that just got in and conflicts with this commit

osherdp:
next time, please do ``git rebase master`` instead of using ``git merge`` to not create merge commits
for now it's fine
/ok-to-test

bkopilov:
> next time, please do `git rebase master` instead of using `git merge` to not create merge commits for now it's fine /ok-to-test

ack ok .

osherdp (approved):

PR #1958, opened 2022-12-19
MGMT-12440: Configure IPIP tunnel with heterogeneous setup

adriengentil:
Setup an IPIP tunnel between both devices.

Requires https://github.com/openshift/assisted-test-infra/pull/1949


adriengentil:
/override ci/prow/e2e-metal-assisted-day2-arm-workers
it did not fail :see_no_evil: (configuring the machine was ok)

osherdp (approved):

PR #1963, opened 2022-12-21
Allow overriding pytest flags

osherdp:
Back in https://github.com/openshift/assisted-test-infra/pull/1824 we prevented tests marked with skip to just be skipped (currently they are marked as failures).

The reason we did it is to prevent a situation where the list of openshift versions come off empty and the tests are considered skipped (and passing vacuously).

This enables emptying all pytest ``PYTEST_FLAGS`` to disable this kind of behavior.

/cc @lalon4 @eliorerz


osherdp:
/hold
to make sure it works

osherdp:
/hold
waiting for ack from qe

bkopilov:
Hi ,  
Tested this code, 
When the param is export PYTEST_FLAGS=""   the test fails... and not skipped
When the param is set to export PYTEST_FLAGS=" " , test skipped as expected

Benny

osherdp:
> Hi , Tested this code, When the param is export PYTEST_FLAGS="" the test fails... and not skipped When the param is set to export PYTEST_FLAGS=" " , test skipped as expected
> 
> Benny

ack. thank you!
/unhold

PR #1966, opened 2022-12-22
MGMT-12441: Install libvirtd on the ARM machine

adriengentil:
Install libvirtd and configure a throwable SSH key pair between both of
the machines, so the primary machine can control libvirtd on the ARM 
machine through SSH, e.g.:
```
virsh -c "qemu+ssh://secondary-host/system"
```


adriengentil:
/override e2e-metal-assisted-day2-arm-workers
as usual, the test failed but it's expected

osherdp (approved):

PR #1973, opened 2022-12-27
Cannot remove container as it is running issue

bkopilov:
Test failed when trying to remove proxy container squid Error message refers to running or paused containers cannot be removed The code stop && rm the podman but it might be that stop code not really blocked and waiting before we call remove.

Added sleep between stop to remove.

### DIFF
--- a/src/assisted_test_infra/test_infra/utils/utils.py
+++ b/src/assisted_test_infra/test_infra/utils/utils.py
@@ -428,7 +428,7 @@ def run_container(container_name, image, flags=None, command=""):
 
 def remove_running_container(container_name):
     log.info(f"Removing Container {container_name}")
-    container_rm_cmd = f"podman-remote stop {container_name} && podman-remote rm {container_name}"
+    container_rm_cmd = f"podman-remote rm --force {container_name}"
     run_command(container_rm_cmd, shell=True)
 
 


bkopilov:
@lalon4 , @osherdp 
We hit on error while trying to remove proxy container.

failed on teardown with "RuntimeError: command: podman-remote stop squid_https && podman-remote rm squid_https exited with an error: Error: cannot remove container ee03465cc2fcc81b85071ceff313fea5bace999d2982e1a4e4b82e3372bd08d8 as it is running - running or paused containers cannot be removed without force: container state improper code: 2"


Adding 5 seconds to stop action before trying to remove it, it might be that stop action not completly blocked before trying to remove.

Benny

src/assisted_test_infra/test_infra/utils/utils.py
```diff
@@ -428,7 +428,7 @@ def run_container(container_name, image, flags=None, command=""):
 
 def remove_running_container(container_name):
     log.info(f"Removing Container {container_name}")
-    container_rm_cmd = f"podman-remote stop {container_name} && podman-remote rm {container_name}"
+    container_rm_cmd = f"podman-remote stop {container_name} && sleep 5 && podman-remote rm {container_name}"
```
osherdp, line 431: this should ignore the running condition of a container and just do the two actions in one go:
```suggestion
    container_rm_cmd = f"podman-remote rm --force {container_name}"
```

osherdp: In general ``sleep``s still doesn't guarantee stability, so it should be avoided

osherdp (changes requested):

src/assisted_test_infra/test_infra/utils/utils.py
bkopilov: ack , ok . how can we wait for stop action to be completed before move to remove ?  anyway i am chainging as you suggested.

osherdp: Reading the documentation I think it should do so, not sure why not:
```
$ podman-remote stop --help
Stop one or more containers

Description:
  Stops one or more running containers.  The container name or ID can be used.

  A timeout to forcibly stop the container can also be set but defaults to 10 seconds otherwise.
```

In principle there's the ``podman-remote wait`` command that can wait for a specific condition

bkopilov: ack , so we know that 10 seconds its not enough base on the failure  and we can use force or use -t  flag for increasing timeout.
is it ok to add : f"podman-remote stop {container_name} && podman-remote rm --force {container_name}" ?

osherdp: It should work as well
Although redundant. Simply "podman rm --force" (without the stop command before) doesn't work?

bkopilov: > It should work as well Although redundant. Simply "podman rm --force" (without the stop command before) doesn't work?

Yes it should work i just tried on my setup and looks like 10 second is not enough or we blocked on something.

[root@ocp-edge82 assisted-test-infra]# podman rm --force 46d30a105f25
WARN[0010] StopSignal SIGTERM failed to stop container load_balancer in 10 seconds, resorting to SIGKILL 
46d30a105f253a26c4052fe17fa7e99a017630f0717136c4f366301816ad82e3


I see that kill terminate the container immediatly ...
podman-remote kill 0823ceb41f0e

Not sure if we force or kill, we may have leftovers or maybe i am wrong.

osherdp: so replace ``podman-remote stop`` with ``podman-remote kill``. I think ``stop`` sends sigterm instead of sigkill which is a gracefulness we don't really need in this case

src/assisted_test_infra/test_infra/utils/utils.py
```diff
@@ -428,7 +428,7 @@ def run_container(container_name, image, flags=None, command=""):
 
 def remove_running_container(container_name):
     log.info(f"Removing Container {container_name}")
-    container_rm_cmd = f"podman-remote stop {container_name} && podman-remote rm {container_name}"
+    container_rm_cmd = f"podman-remote kill {container_name}"
```
osherdp, line 431: wait, this is odd
we don't need to remove the container? (it's also in the function name and in the log, so we shouldn't do something different than that)

bkopilov: > wait, this is odd we don't need to remove the container? (it's also in the function name and in the log, so we shouldn't do something different than that)

From the trace in CI run , here is the fail point

 yield start_proxy_server
        if global_variables.test_teardown:
            log.info("--- TEARDOWN --- proxy controller")
            for server in proxy_servers:
>               server.remove()

src/tests/base_test.py:724: 

https://ci-jenkins-csb-kniqe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-assisted-installer-virt-tf/5136//testReport/junit/test_proxy/TestProxy/Run_assisted_installer_Functional_API_Tests___test_http_proxy_http_proxy_params1_https_proxy_params1_OCP_35314_/

osherdp: yeah but my argument here is that this code is misleading. method says ``remove_running_container`` and we're doing another thing

regardless, the error says ``cannot remove container ... as it is running - running or paused containers cannot be removed without force`` so I think we should use the force :smile: (going back to my suggestion to just do ``podman-remote rm --force ...`` and close the dill)

bkopilov: > yeah but my argument here is that this code is misleading. method says `remove_running_container` and we're doing another thing
> 
> regardless, the error says `cannot remove container ... as it is running - running or paused containers cannot be removed without force` so I think we should use the force smile (going back to my suggestion to just do `podman-remote rm --force ...` and close the dill)
sure no problem....

osherdp:
running some flows that should use this method
/test e2e-metal-assisted-none e2e-metal-assisted-ipv6

osherdp (approved):

PR #1976, opened 2022-12-29
remove redundant xsl files and use just one

osherdp:
All those files are indentical. this change uses just one instance of those files and reference it for each case.


eliorerz (approved):
LGTM
/hold wait for tests

PR #1977, opened 2023-01-01
Fix prepare network - get platform from configuration

bkopilov:
Current code is part of cluster fixture, configuration_controller already contains that updated platform.

This fix is going to fix test_none_platform test.

We will have to change the name of tf_platform name to common name that will be supported by other controllers , different patch

### DIFF
--- a/src/tests/base_test.py
+++ b/src/tests/base_test.py
@@ -294,7 +294,7 @@ class BaseTest:
 
     @classmethod
     def _prepare_nodes_network(cls, prepared_nodes: Nodes, controller_configuration: BaseNodesConfig) -> Nodes:
-        if global_variables.platform not in (consts.Platforms.BARE_METAL, consts.Platforms.NONE):
+        if controller_configuration.tf_platform not in (consts.Platforms.BARE_METAL, consts.Platforms.NONE):
             yield prepared_nodes
             return
 


bkopilov:
@osherdp , @lalon4 , @eliorerz

bkopilov:
More details about the issue :
in none platform terrafrom we have nat and route modes for non platform with condition.
mode = length(var.machine_cidr_addresses) == 1 && replace(var.machine_cidr_addresses[0], ":", "") != var.machine_cidr_addresses[0] ? "nat" : "route"


In case we hit on route we need nat by hypervisor which is added in current fixed code ,  in case we have nat mode , we will have to address another issue with routing between two differen network in cluster due to iptables rules.
libvirt on nat mode create isolatated network for the node and ping from node with network x to node with network y does not work ( different story - related to icmp-port-unreachable rules added by libvirt)

Current code was taking the param from DefaultVariables which is global directly but at the point we call to the code the global already use none as default. the cluster_configuration already contains the updated valude from tf_platofrm and can be used .

In the tescase we do use parameterized fixture. 
@pytest.mark.parametrize("platform", [consts.Platforms.NONE])

Benny

osherdp (approved):

PR #1980, opened 2023-01-09
Allow injecting additional files in BIP test

vrutkovs:
This would allow us to include promtail config to send bootstrap logs to Loki


src/assisted_test_infra/test_infra/helper_classes/config/base_cluster_config.py
```diff
@@ -10,7 +10,7 @@
 
 
 @dataclass
-class BaseClusterConfig(BaseEntityConfig, ABC):
+class wwBaseClusterConfig(BaseEntityConfig, ABC):
```
osherdp, line 13: a mistype of some sort?

src/tests/test_bootstrap_in_place.py
```diff
@@ -279,6 +279,19 @@ def waiting_for_installation_completion(
             if not skip_logs:
                 self.log_collection(master_ip)
 
+    def inject_bootstrap(self, ignition_filename: string, butane_config: string):
```
osherdp, line 282: afaik it needs to be:
```suggestion
    def inject_bootstrap(self, ignition_filename: str, butane_config: str):
```

src/tests/test_bootstrap_in_place.py
```diff
@@ -289,10 +302,14 @@ def prepare_installation(self, controller_configuration: BaseNodesConfig, cluste
         extract_installer(openshift_release_image, BUILD_DIR)
         self.installer_generate(openshift_release_image)
 
+        ignition_filename = "bootstrap-in-place-for-live-iso.ign"
+        if cluster_configuration.bip_butane_config:
```
osherdp, line 306: can you show an example of such config?

osherdp (changes requested):

src/tests/test_bootstrap_in_place.py
```diff
@@ -279,6 +279,19 @@ def waiting_for_installation_completion(
             if not skip_logs:
                 self.log_collection(master_ip)
 
+    def inject_bootstrap(self, ignition_filename: string, butane_config: string):
+        expected_ignition_name = "bootstrap_initial.ign"
+        shutil.move(ignition_filename, expected_ignition_name)
+        flags = shlex.split(f"--rm -v {os.getcwd()}:/data -w /data")
+        # retry to avoid occassional quay hiccups
+        self.retrying_run_container(
```
eranco74, line 287: This will merge the butane_config into the existing content of ignition_filename?
Seems to me that it will override it no?

src/assisted_test_infra/test_infra/helper_classes/config/base_cluster_config.py
vrutkovs: oops, sorry, fixed

osherdp (approved):

PR #1987, opened 2023-01-13
Update inject_bootstrap

vrutkovs:
Operate on files in IBIP_DIR


Makefile
```diff
@@ -107,6 +107,10 @@ ifeq ($(ENABLE_KUBE_API),true)
 	SERVICE_REPLICAS_COUNT=1
 	AUTH_TYPE=local
 endif
+ifdef BIP_BUTANE_CONFIG
+	BOOTSTRAP_INJECT_DIR=$(ROOT_DIR)/sno-bootstrap-manifests/
```
osherdp, line 111: is it "git ignored"?
if not, can we put it under a directory that is git ignored?

Makefile
```diff
@@ -107,6 +107,10 @@ ifeq ($(ENABLE_KUBE_API),true)
 	SERVICE_REPLICAS_COUNT=1
 	AUTH_TYPE=local
 endif
+ifdef BIP_BUTANE_CONFIG
```
osherdp, line 110: Something I just realized, it's missing in the documentation for env-vars
Can you add it in an appropriate place? (I think in https://github.com/openshift/assisted-test-infra/blob/009de89225024572fe99174beded1ff058fb0540/README.md#L127)

osherdp (changes requested):

Makefile
vrutkovs: Good idea, updated

Makefile
vrutkovs: Fixed

osherdp (approved):

osherdp:
still broken

PR #1988, opened 2023-01-15
Setting minikube registry addon to use images from quay.io

osherdp:
In local environments, we sometimes hit the rate-limitter of docker hub when doing a lot of ``make destroy run``. This is because the registry addon of minikube relies on images that are not in ghcr.io but from docker-hub.

This change uses images from quay.io for the registry, as a simple fix.


omertuc:
Should we also improve [this](https://github.com/openshift/assisted-service/blob/f08585e6ecc7867da5a3ebd6d2e9bef325f53946/docs/dev/running-test.md?plain=1#L36) document?

PR #1999, opened 2023-01-17
MGMT-13341: Use assisted service client wheel

adriengentil:
Currently, the sdist version fails to install because of an update of
setuptools: https://github.com/pypa/setuptools/issues/3772.

The wheel is not affected by this change.

Depending on the outcome of the setuptools ticket, we may have to
follow-up to provide a valid version scheme for the sdist package (or
just remove it?).


osherdp (approved):

adriengentil:
/override ci/prow/e2e-metal-assisted-kube-api-late-binding-single-node ci/prow/e2e-metal-assisted-kube-api-net-suite ci/prow/e2e-metal-single-node-live-iso

PR #2001, opened 2023-01-17
Fix podman container names on gather

osherdp:
For some reason, the names are not matching the actual container names used. It should be prefixed with ``assisted-installer`` (podman's pod name) and continue with the names ``db``, ``ui``, ``image-service`` and ``service``.
See:
https://github.com/openshift/assisted-service/blob/a1c6df56391b517f4bbf167ba841ffe040950ebd/deploy/podman/pod.yml#L8

### DIFF
--- a/scripts/download_logs.sh
+++ b/scripts/download_logs.sh
@@ -25,7 +25,7 @@ function download_service_logs() {
     podman ps -a || true
 
     for service in "assisted-installer-db" "assisted-installer-image-service" "assisted-installer-service" "assisted-installer-ui"; do
-      podman logs ${service} >${LOGS_DEST}/logs_${service}_${DEPLOY_TARGET}.log || true
+      podman logs ${service} >${LOGS_DEST}/logs_${service}_${DEPLOY_TARGET}.log 2>&1 || true
     done
   else
     CRS=node,pod,svc,deployment,pv,pvc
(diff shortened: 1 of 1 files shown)


adriengentil:
the logs were printed on the console instead of ending up in the logs file:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_assisted-test-infra/2001/pull-ci-openshift-assisted-test-infra-master-e2e-metal-assisted-onprem/1615338556648067072/artifacts/e2e-metal-assisted-onprem/assisted-common-gather/build-log.txt

scripts/download_logs.sh
```diff
@@ -24,7 +24,7 @@ function download_service_logs() {
   if [ "${DEPLOY_TARGET:-}" = "onprem" ]; then
     podman ps -a || true
 
-    for service in "assisted-service" "assisted-image-service" "assisted-installer-ui" "postgres"; do
+    for service in "assisted-installer-db" "assisted-installer-image-service" "assisted-installer-service" "assisted-installer-ui"; do
       podman logs ${service} >${LOGS_DEST}/logs_${service}_${DEPLOY_TARGET}.log || true
```
adriengentil, line 28:
```suggestion
      podman logs ${service} >${LOGS_DEST}/logs_${service}_${DEPLOY_TARGET}.log 2>&1 || true
```
would maybe do the trick?

osherdp: very possibly 
Done!

osherdp:
Looks much better now
Seems like UI doesn't output any logs (and I can confirm it's the same in my environment)

adriengentil (approved):

PR #2006, opened 2023-01-19
MGMT-12477: Validate that cluster platform is exists in `supported_platforms` API call result

eliorerz:
We do not call the `/v2/clusters/{cluster_id}/supported-platforms` during the installation and as a result we do not verify that the installed platform do exist in the `supported-platforms` list.

This PR adds 2 calls to that API with that expected results:

- When there are no hosts - platform should be none
- When hosts are exist, validate that the cluster.platform is exists in the supported platforms list.

/cc @osherdp @adriengentil


eliorerz:
FYI @lalon4

osherdp (approved):

PR #2008, opened 2023-01-23
MGMT-13438: Fix NetworkType trigger overrides parametrized values

eliorerz:
When running test with NetworkType set as parametrized and `OpenshiftVersion < 4.12, NetworkType` is set to None instead of the desired value 

e.g the following test will result creating cluster with NetowkType=None

```python
@pytest.mark.parametrize("network_type", [consts.NetworkType.OpenShiftSDN])
def test_install(self, cluster, network_type): 
```

/cc @osherdp @adriengentil

### DIFF
--- a/src/triggers/default_triggers.py
+++ b/src/triggers/default_triggers.py
@@ -1,7 +1,6 @@
 from typing import Dict
 
 from frozendict import frozendict
-from packaging import version
 
 from consts import consts, resources
 from triggers.env_trigger import Trigger
@@ -33,11 +32,6 @@ _default_triggers = frozendict(
             user_managed_networking=True,
             master_memory=resources.DEFAULT_MASTER_SNO_MEMORY,
             master_vcpu=resources.DEFAULT_MASTER_SNO_CPU,
-            network_type=None,
-        ),
-        "let_service_choose_network_type": Trigger(
-            condition=lambda config: version.parse(config.openshift_version)
-            >= version.parse(consts.OpenshiftVersion.VERSION_4_12.value),
             network_type=None,
         ),
         "ipv4": Trigger(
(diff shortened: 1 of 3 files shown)


eliorerz:
FYI @lalon4 @bkopilov

src/triggers/default_triggers.py
```diff
@@ -37,8 +37,9 @@
         ),
         "let_service_choose_network_type": Trigger(
```
osherdp, line 38: I think the name should be changed now. Maybe "default_to_sdn_for_ocp_under_4.12"

but another question, why do we need this trigger? if network-type not provided the service will probably choose sdn by itself

osherdp (changes requested):

src/triggers/default_triggers.py
eliorerz: Not on versions below 4.12.
I think that @danmanor added it a while back.

eliorerz: In that case the network type will be nil and will not be set

eliorerz: p.s I updated the trigger name

osherdp: but it gotta choose some network type eventually, no?
I mean, it will not get blocked on validations or something like it AFAIK

eliorerz: I think that openshift installer is choosing its default value

danmanor: IIRC we wanted to check if clusters with `openshift_version` >= 4.12 are set correctly with network type = `OVNKubernetes`, but test-infra was setting it to be `OpenshiftSDN` anyway, so we added this trigger to let the backend decide the network type when the version is >= 4.12. I am not sure I understand why the network type is set to None for versions < 4.12

osherdp (approved):

eliorerz:
I removed the trigger and verified that assisted-service is handling both cases: 

1. Openshift version >= 4.12: `OVNKubernetes`
2. Openshift version < 4.12:  `OpenShiftSDN`

PR #2009, opened 2023-01-23
MGMT-13379: dedicated terraform stack for day2 cluster

adriengentil:
In the scope of testing heterogeneous clusters, the day2 cluster must use
its own terraform stack. To implement that, I updated the code to create
day2 clusters in the same way we create day1 clusters, using the fixtures.

### DIFF
--- a/src/assisted_test_infra/test_infra/controllers/node_controllers/terraform_controller.py
+++ b/src/assisted_test_infra/test_infra/controllers/node_controllers/terraform_controller.py
@@ -174,8 +174,11 @@ class TerraformController(LibvirtController):
         tfvars["bootstrap_in_place"] = self._config.bootstrap_in_place
 
         vips = self.get_ingress_and_api_vips()
-        tfvars["api_vip"] = vips["api_vip"]
-        tfvars["ingress_vip"] = vips["ingress_vip"]
+        tfvars["api_vip"] = self._config.api_vip or vips["api_vip"]
+        tfvars["ingress_vip"] = self._config.ingress_vip or vips["ingress_vip"]
+        if self._config.base_cluster_domain:
+            tfvars["base_cluster_domain"] = self._config.base_cluster_domain
+
         tfvars["running"] = running
         tfvars["libvirt_master_macs"] = static_network.generate_macs(self._params.master_count)
         tfvars["libvirt_worker_macs"] = static_network.generate_macs(self._params.worker_count)
@@ -289,7 +292,7 @@ class TerraformController(LibvirtController):
 
     def _try_to_delete_nodes(self):
         log.info("Start running terraform delete")
-        self.tf.destroy()
+        self.tf.destroy(force=False)
 
     def destroy_all_nodes(self, delete_tf_folder=False):
         """Runs terraform destroy and then cleans it with virsh cleanup to delete
--- a/src/tests/base_test.py
+++ b/src/tests/base_test.py
@@ -96,12 +96,22 @@ class BaseTest:
         yield config
 
     @pytest.fixture
-    def infraenv_configuration(self) -> InfraEnvConfig:
-        yield InfraEnvConfig()
+    def new_day2_controller_configuration(self, request: FixtureRequest) -> BaseNodesConfig:
+        """
+        Creates the controller configuration object according to the platform.
+        Override this fixture in your test class to provide a custom configuration object
+        :rtype: new node controller configuration
+        """
+        assert global_variables.tf_platform == consts.Platforms.BARE_METAL
+
+        config = TerraformConfig()
+
+        self.update_parameterized(request, config)
+        yield config
 
     @pytest.fixture
-    def day2_cluster_configuration(self) -> Day2ClusterConfig:
-        yield Day2ClusterConfig()
+    def infraenv_configuration(self) -> InfraEnvConfig:
+        yield InfraEnvConfig()
 
     @pytest.fixture
     def prepared_controller_configuration(self, new_controller_configuration: BaseNodesConfig) -> BaseNodesConfig:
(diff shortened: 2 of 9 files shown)


adriengentil:
/cc @eliorerz @osherdp 
Can you have a first look please? :pray:

src/assisted_test_infra/test_infra/controllers/node_controllers/terraform_controller.py
```diff
@@ -174,8 +174,11 @@ def _fill_tfvars(self, running=True):
         tfvars["bootstrap_in_place"] = self._config.bootstrap_in_place
 
         vips = self.get_ingress_and_api_vips()
-        tfvars["api_vip"] = vips["api_vip"]
-        tfvars["ingress_vip"] = vips["ingress_vip"]
+        tfvars["api_vip"] = self._config.api_vip if self._config.api_vip else vips["api_vip"]
```
osherdp, line 177: isn't it equivalent to 
```suggestion
        tfvars["api_vip"] = self._config.api_vip or vips["api_vip"]
```

src/tests/base_test.py
```diff
@@ -141,6 +173,28 @@ def test_something(cluster):
...
+
+        For example:
+
+        @pytest.fixture
+        def FIXTURE_NAME(self, prepared_controller_configuration):
+            yield prepared_controller_configuration
+
+        @pytest.mark.override_controller_configuration(FIXTURE_NAME.__name__)
+        def test_something(cluster):
+            pass
```
osherdp, line 192: not an easy API
but at least there's an example usage now, so I'm grateful for that :)

src/tests/base_test.py
```diff
@@ -153,6 +207,42 @@ def new_cluster_configuration(self, request: FixtureRequest) -> ClusterConfig:
 
         return config
 
+    @pytest.fixture
+    def new_day2_cluster_configuration(
```
osherdp, line 211: (a general comment, not specific to this line)
make sure to check if this PR has any impact on qe (or maybe you already checked)

osherdp (commented):
looks good to me
only a few remarks

src/tests/base_test.py
adriengentil: it was already there in `controller_configuration()`  :)

src/tests/base_test.py
adriengentil: yes, once I get this PR working, I'll ping QE

src/tests/base_test.py
```diff
@@ -153,6 +207,42 @@ def new_cluster_configuration(self, request: FixtureRequest) -> ClusterConfig:
...
+    ) -> Day2ClusterConfig:
+        """
+        Creates new day2 cluster configuration object.
+        Override this fixture in your test class to provide a custom cluster configuration. (See TestInstall)
+        :rtype: new day2 cluster configuration object
+        """
+        config = Day2ClusterConfig()
+        self.update_parameterized(request, config)
+
+        if triggers_enabled:
```
adriengentil, line 222: I had to move the trigger for the day2 configuration here, otherwise it creates a dependency cycle between the day1 and day2 clusters

terraform_files/baremetal/main.tf
```diff
@@ -158,50 +158,55 @@ module "workers" {
...
-# TODO: leave only the wildcard cname entry defined and remove the other specific DNS assignments
+# TODO: leave only the wildcard address entry defined and remove the other specific DNS assignments
 # Read more at: https://bugzilla.redhat.com/show_bug.cgi?id=1532856
-data "libvirt_network_dnsmasq_options_template" "wildcard-ingress-cname" {
-  count        = var.master_count == 1 ? 1 : 0
-  option_name  = "cname"
-  option_value = "*.apps.${var.cluster_name}.${var.cluster_domain},${var.ingress_vip}"
+data "libvirt_network_dnsmasq_options_template" "wildcard-apps-ingress" {
+  # Enable "apps" wildcard in case of SNO and when we try to add day2 worker to SNO
+  count        = var.ingress_vip == var.api_vip ? 1 : 0
```
adriengentil, line 183: I take the assumption that the condition `var.ingress_vip == var.api_vip` is a SNO cluster, is that always true?

osherdp: in our infra it's probably always true
although I guess there are other ways to assert SNO cluster (num_masters == 1)

adriengentil: yeah, otherwise I need to pass an extra flag. What is this `single_node_ip` variable above, could it be reused here?

osherdp: yes I think so

osherdp: if anything I would prefer to move to more general values (api/ingress rather than having a separate var for SNO IP) but for now it wouldn't matter match. Only thing important here is not to add more flags as we already have too much IMHO :)

adriengentil: here I need to know if the day1 is an SNO in order to set the DNS or not in the day2 network, at terraform level I have almost no information about day1 cluster except api and ingress vips.

adriengentil: (I agree that we should not add more flags :) )

osherdp: ack. I missed this is in another context 
then the api==ingress is probably the best thing you can do ATM

adriengentil:
/cc @lalon4 @bkopilov 
Can you have a look? I refactored the Day2Cluster flow.

adriengentil:
seems to fail for an related reason, flake?

src/assisted_test_infra/test_infra/controllers/node_controllers/terraform_controller.py
```diff
@@ -174,8 +174,11 @@ def _fill_tfvars(self, running=True):
         tfvars["bootstrap_in_place"] = self._config.bootstrap_in_place
 
         vips = self.get_ingress_and_api_vips()
-        tfvars["api_vip"] = vips["api_vip"]
-        tfvars["ingress_vip"] = vips["ingress_vip"]
+        tfvars["api_vip"] = self._config.api_vip or vips["api_vip"]
```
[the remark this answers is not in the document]
adriengentil: In case of day2 cluster, we want to set the vips of the existing day1 cluster, which are set in `self._config.api_vip` and `self._config.ingress_vip`. 
If `self._config.ingress_vip` and `self._config.api_vip` are not set, then we let `get_ingress_and_api_vips()` select vips for us.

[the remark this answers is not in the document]
adriengentil: They are set [here](https://github.com/openshift/assisted-test-infra/pull/2009/files#diff-132040bf88e1eaf0881057dc6ba2099051c758b0c01cd7f2fe09d40eaeb20577R147-R148) while building the controller configuration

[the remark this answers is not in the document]
adriengentil: I could move it logic, in `get_ingress_and_api_vips()` though

osherdp (approved):

PR #2014, opened 2023-01-24
MGMT-12442: Upgrade to oc 4.12

adriengentil:
In the scope of the heterogeneous cluster story, we need to use a
version of oc that supports multi arch images.


adriengentil:
/cc @osherdp @eliorerz 
Do you think we should run other tests?

osherdp (approved):

PR #2015, opened 2023-01-26
MGMT-13476: Tune dnf to download packages faster

adriengentil:
Enable fastestmirror and increase max_parallel_downloads to 10. As a
side effect, we have a chance to select different mirrors from time to
time, which should mitigate issues with the default mirrors.


osherdp (approved):

adriengentil:
flake?

PR #2018, opened 2023-01-30
[release-ocm-2.7] Update default tag for index and service

adriengentil:
`e2e-metal-assisted-ha-kube-api-ipv6`, `e2e-metal-assisted-kube-api-late-binding-single-node`  and `e2e-metal-assisted-kube-api-late-unbinding-single-node` jobs are currently permafailing.

I suspect it's because the index not being aligned with branch version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,9 +42,9 @@ CONTAINER_RUNTIME_COMMAND := $(or ${CONTAINER_COMMAND}, ${CONTAINER_RUNTIME_COMM
 LINT_CODE_STYLING_DIRS := src/tests src/triggers src/assisted_test_infra/test_infra src/assisted_test_infra/download_logs src/service_client src/consts src/virsh_cleanup src/cli
 
 # assisted-service
-SERVICE := $(or $(SERVICE), quay.io/edge-infrastructure/assisted-service:latest)
+SERVICE := $(or $(SERVICE), quay.io/edge-infrastructure/assisted-service:ocm-2.7)
 SERVICE_NAME := $(or $(SERVICE_NAME),assisted-service)
-INDEX_IMAGE := $(or ${INDEX_IMAGE},quay.io/edge-infrastructure/assisted-service-index:latest)
+INDEX_IMAGE := $(or ${INDEX_IMAGE},quay.io/edge-infrastructure/assisted-service-index:ocm-2.7)
 REMOTE_SERVICE_URL := $(or $(REMOTE_SERVICE_URL), "")
 
 # terraform


Makefile
```diff
@@ -42,9 +42,9 @@ CONTAINER_RUNTIME_COMMAND := $(or ${CONTAINER_COMMAND}, ${CONTAINER_RUNTIME_COMM
 LINT_CODE_STYLING_DIRS := src/tests src/triggers src/assisted_test_infra/test_infra src/assisted_test_infra/download_logs src/service_client src/consts src/virsh_cleanup src/cli
 
 # assisted-service
-SERVICE := $(or $(SERVICE), quay.io/edge-infrastructure/assisted-service:latest)
+SERVICE := $(or $(SERVICE), quay.io/edge-infrastructure/assisted-service:ocm-2.7)
```
osherdp, line 45: I wonder if we can make some adjustments that don't demand us to handle all those tag changes for the branches
I mean, if anything I would assume we override ``SERVICE`` somewhere to use the right service image, and not use the default. Is it not the case?

adriengentil: You're right, I'll see if I can do it from release repo

adriengentil: and yes SERVICE shouldn't be required

osherdp: you know what, maybe this change is good for easing local testing of this version

adriengentil:
/test ci/prow/e2e-metal-assisted-kube-api-late-binding-single-node
looks like prow is flaky today...

adriengentil:
@osherdp @eliorerz `e2e-metal-assisted-ha-kube-api-ipv6` seems broken:
```
# kubectl get agentclusterinstall/test-infra-cluster-f883c57c-agent-cluster-install -n assisted-spoke-cluster-fd412a8f -o yaml
apiVersion: extensions.hive.openshift.io/v1beta1
kind: AgentClusterInstall
spec:
  apiVIP: 1001:db9::64
  ingressVIP: 1001:db9::65
  networking:
    clusterNetwork:
    - cidr: 172.30.0.0/16
      hostPrefix: 23
    serviceNetwork:
    - 10.128.0.0/14
status:
  apiVIP: 1001:db9::64
  apiVIPs:
  - 1001:db9::64
  ingressVIP: 1001:db9::65
  ingressVIPs:
  - 1001:db9::65
  machineNetwork:
  - cidr: 1001:db9::/120
  validationsInfo:
    network:
    - id: networks-same-address-families
      message: Address families of networks (MachineNetworks, ServiceNetworks, ClusterNetworks)
        are not the same.
      status: failure
```
it seems we provide IPv4 addresses to the configuration and as a consequence, the validation fails because we mix IPv4/6. Does it ring a bell to you?

adriengentil:
/override ci/prow/e2e-metal-assisted-ha-kube-api-ipv6
I'll open a bug for it

adriengentil:
@osherdp let's merge it?

osherdp (approved):

PR #2023, opened 2023-01-31
Use sudo for installation of kind

dantrainor:
use sudo for installation of kind to prevent permission denied errors


adriengentil:
Thanks!
/ok-to-test

@osherdp FYI

adriengentil (approved):

PR #2024, opened 2023-01-31
Use sudo when installing oc client

dantrainor:
use sudo for installation of oc clients to prevent permission denied errors


adriengentil:
Thanks!
/ok-to-test

adriengentil (approved):

PR #2025, opened 2023-01-31
NO-ISSUE: Fix parsing olm operator env var list

eliorerz:
Allow to pass multiple OLM operators using environment variables 
e.g.
```bash
OLM_OPERATORS=lvm,cnv make test
```
or
```bash
OLM_OPERATORS=lvm cnv make test
```

/cc @osherdp @adriengentil

### DIFF
--- a/src/tests/global_variables/env_variables_defaults.py
+++ b/src/tests/global_variables/env_variables_defaults.py
@@ -1,4 +1,5 @@
 import json
+import re
 from abc import ABC
 from dataclasses import dataclass
 from distutils.util import strtobool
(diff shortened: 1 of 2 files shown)


adriengentil:
It would be good to align the behaviour here as well: https://github.com/openshift/assisted-test-infra/blob/ea66466ed53859a0fe78de070bd2ed45ad860c2d/src/assisted_test_infra/test_infra/utils/operators_utils.py#L18-L19.

src/tests/global_variables/env_variables_defaults.py
```diff
@@ -60,7 +60,7 @@ class _EnvVariables(DataPool, ABC):
 
     namespace: EnvVar = EnvVar(["NAMESPACE"], default=consts.DEFAULT_NAMESPACE)
     spoke_namespace: str = EnvVar(["SPOKE_NAMESPACE"], default=consts.DEFAULT_SPOKE_NAMESPACE)
-    olm_operators: EnvVar = EnvVar(["OLM_OPERATORS"], loader=lambda operators: operators.lower().split(), default="")
+    olm_operators: EnvVar = EnvVar(["OLM_OPERATORS"], loader=lambda operators: operators.lower().split(","), default=[])
```
osherdp, line 63: makes much more sense. thank you!
leaving it to what @adriengentil said, and also let's document it in the README

osherdp (changes requested):

src/tests/global_variables/env_variables_defaults.py
eliorerz: I don't think this is the right approach.
We are holding almost all the env vars in the `_EnvVariables` object, the option to add lambada was added to prevent us from crating a lot of functions just for single call.
I don't see the benefit of creating a new method just for split

adriengentil: It's not about creating a new function, it's to make sure the existing one behaves in the same way.
Let's delete that function if it does not make sense (it doesn't seem to be used elsewhere).

eliorerz: The main reason that I don't think it should be a separate function is mainly because that I don't want to handle the env vars elsewhere.
It caused us a lot of issues before we had most of them in the same place.

osherdp: @eliorerz not sure I follow you
``parse_olm_operators_from_env`` isn't used anywhere, and it has the older format (``OLM_OPERATORS="odf cnv"``)
what do we gain from keeping it instead of just deleting it?

osherdp (approved):

PR #2026, opened 2023-02-01
Add waiter for ssh to nodes

bkopilov:
When starting a node and we need access to the node it may take some time till node will be ready and sshable from outside.
ping or verify ip is not enough in cases ssh service not ready.

Added a waiter in nodes.py

### DIFF
--- a/src/assisted_test_infra/test_infra/helper_classes/nodes.py
+++ b/src/assisted_test_infra/test_infra/helper_classes/nodes.py
@@ -4,6 +4,7 @@ from typing import Dict, Iterator, List
 
 import waiting
 from munch import Munch
+from scp import SCPException
 
 import consts
 from assisted_test_infra.test_infra.controllers.node_controllers import Node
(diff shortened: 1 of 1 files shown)


bkopilov:
@lalon4 , @osherdp

osherdp (approved):

osherdp:
/hold
lint fails

bkopilov:
Yes i see my fault

/src/assisted_test_infra/test_infra/helper_classes/nodes.py:237:25: B023 Function definition does not bind loop variable 'node'.  intersting ....

src/assisted_test_infra/test_infra/helper_classes/nodes.py
```diff
@@ -228,3 +229,24 @@ def _are_nodes_network_prepared(self):
...
+            except (TimeoutError, SCPException):
+                return False
+            else:
+                return True
+
+        waiting.wait(
+            lambda: _ssh_connection(),
+            timeout_seconds=180,
+            sleep_seconds=20,
+            waiting_for="Nodes to allow ssh connection",
```
osherdp, line 250: few fixes to make the code more clear:
```suggestion
        def _all_nodes_allow_ssh_connection():
            try:
                for node in self.nodes:
                    if node.ssh_connection is None:
                        return False
            except (TimeoutError, SCPException):
                return False

            return True

        waiting.wait(
            lambda: _all_nodes_allow_ssh_connection(),
            timeout_seconds=180,
            sleep_seconds=20,
            waiting_for="nodes to allow ssh connection",
```

osherdp (changes requested):

src/assisted_test_infra/test_infra/helper_classes/nodes.py
bkopilov: > few fixes to make the code more clear:

Ack no problem .

osherdp: hmm, you still didn't apply all those changes. please look carefully

bkopilov:
https://github.com/openshift/assisted-test-infra/pull/2026/files

sorry , what did i miss ?   by the way good catch which the bug that i
created with closures that was caught :)

osherdp (approved):

PR #2027, opened 2023-02-02
Makefile: _destroy_terraform should actually run terraform

vrutkovs:
For QEmu installs "_destroy_virsh" target is used


adriengentil (approved):
afaik, `virsh_cleanup` destroys all libvirt resources which may not what we want when destroying the terraform stack of one cluster. I think it would make sense to have two different targets instead.

Approving as it doesn't really change the current behaviour.

PR #2031, opened 2023-02-06
Add test which destroys targets via terraform

vrutkovs:
terraform.tfvars.json is generated on the flight and not stored on the file system.

After gather steps we call 'test_destroy_terraform' test to fetch active clusters, regenerate terraform.tfvars.json and run terraform destroy using python interface.

This fixes deprovision for terraform-controlled platforms - Nutanix and VSphere

### DIFF
--- a/src/tests/test_targets.py
+++ b/src/tests/test_targets.py
@@ -43,3 +43,22 @@ class TestMakefileTargets(BaseTest):
             cluster.delete()
 
         log.info(f"Successfully deleted {len(clusters)} clusters")
+
+    @JunitTestSuite()
+    def test_destroy_terraform(
+        self, api_client: InventoryClient, prepared_controller_configuration: BaseNodesConfig, cluster_configuration
+    ):
+        """Destroy cluster via terraform"""
+
+        cluster_id = cluster_configuration.cluster_id
+        clusters = api_client.clusters_list() if not cluster_id else [{"id": cluster_id}]
+
+        for cluster_info in clusters:
+            cluster = Cluster(api_client, ClusterConfig(cluster_id=cluster_info["id"]), InfraEnvConfig())
+            controller = self.get_terraform_controller(prepared_controller_configuration, cluster._config)
+            config_vars = controller.get_all_vars()
+            controller.tf.set_vars(**config_vars)
+            controller.tf.select_defined_variables()
+            controller.destroy_all_nodes()
+
+        log.info(f"Successfully destroyed {len(clusters)} clusters")
--- a/Makefile
+++ b/Makefile
@@ -217,10 +217,16 @@ _destroy_terraform:
 	terraform destroy -auto-approve -input=false -state=terraform.tfstate -state-out=terraform.tfstate -var-file=terraform.tfvars.json
 
 destroy_nodes:
-	skipper make $(SKIPPER_PARAMS) _destroy_virsh _destroy_terraform
+	skipper make $(SKIPPER_PARAMS) _destroy_virsh _destroy_terraform_controller
 
 _destroy_virsh:
 	python3 ${DEBUG_FLAGS} -m virsh_cleanup -f test-infra
+
+destroy_terraform_controller:
+	TEST=./src/tests/test_targets.py TEST_FUNC=test_destroy_terraform $(MAKE) test
+
+destroy_nutanix: destroy_terraform_controller
+destroy_vsphere: destroy_terraform_controller
 
 #######
 # Run #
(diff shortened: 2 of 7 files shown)


osherdp:
/assign @eliorerz 
PTAL

src/tests/test_targets.py
```diff
@@ -43,3 +45,27 @@ def test_delete_clusters(self, api_client: InventoryClient, cluster_configuratio
...
+        """Destroy cluster via terraform"""
+
+        cluster_id = cluster_configuration.cluster_id
+        clusters = api_client.clusters_list() if not cluster_id else [{"id": cluster_id}]
+
+        for cluster_info in clusters:
+            cluster = Cluster(api_client, ClusterConfig(cluster_id=cluster_info["id"]), InfraEnvConfig())
+            controller = None
+            if cluster._config.platform == consts.Platforms.NUTANIX:
+                controller = NutanixController(prepared_controller_configuration, cluster_configuration)
```
eliorerz, line 62: Don't we need
```suggestion
                controller = NutanixController(prepared_controller_configuration, cluster._config)
```

src/tests/test_targets.py
```diff
@@ -43,3 +45,27 @@ def test_delete_clusters(self, api_client: InventoryClient, cluster_configuratio
...
+            if cluster._config.platform == consts.Platforms.VSPHERE:
+                controller = VSphereController(prepared_controller_configuration, cluster_configuration)
```
eliorerz, line 64: Same here

eliorerz, line 64: Maybe create a method under `BaseTest` class that returns the controller by the `controller_configuration` and `cluster_configuration`

eliorerz (commented):
I just wanted to do it today :) 
But we do have the `destroy_nodes` target that needs to be replaced

src/tests/test_targets.py
eliorerz: Return it here also https://github.com/openshift/assisted-test-infra/blob/master/src/tests/base_test.py#L240-#L246

vrutkovs: Right, we have `NodeController` fixture

src/tests/test_targets.py
```diff
@@ -43,3 +44,15 @@ def test_delete_clusters(self, api_client: InventoryClient, cluster_configuratio
...
+
+    @JunitTestSuite()
+    def test_destroy_terraform(
+        self,
+        prepared_controller_configuration: BaseNodesConfig,
+        controller: NodeController,
+    ):
+        """Destroy cluster via terraform"""
+        kwargs = prepared_controller_configuration.get_all()
+        controller._tf.set_vars(**kwargs)
```
eliorerz, line 56: `_tf` is not NodeController Attribute 

```AttributeError: 'TerraformController' object has no attribute '_tf'```

eliorerz: `NutanixController`/`VSphereController` are different than the `TerraformController`.
There is a task for create some kind of interface for all of the terraform controllers but it's not prioritized

eliorerz (changes requested):

src/tests/test_targets.py
eliorerz: Also after removing the underscore it seems that it won't destroy the nodes

eliorerz: We need to somehow restore the controllers state and destroy all of the nodes

vrutkovs: Created `get_terraform_controller` which returns `TerraformController`. Later on the previous fixture can be replaced with it too, but I'd rather copy code instead of breaking things

eliorerz:
It seems that the new target is not working on `LibvirtController` and (by the code) it destroying only cluster attached nodes.
If this target is supposed to work only on cluster nodes of specific platform let's rename it to something like `destroy_cluster_nutanix_vsphere_terraform`

vrutkovs:
Lets rename it to `destroy_terraformcontroller` or similar - and add targets `destroy_nutanix` and `destroy_vsphere` running that? Later on Libvirtcontroller would be reworked and we can create `destroy_libvirt` to use it as well. WDYT?

eliorerz:
> Lets rename it to `destroy_terraformcontroller` or similar - and add targets `destroy_nutanix` and `destroy_vsphere` running that? Later on Libvirtcontroller would be reworked and we can create `destroy_libvirt` to use it as well. WDYT?

OK, let's change it to `destroy_terraform_controller`

eliorerz (approved):

PR #2038, opened 2023-02-09
MGMT-12442: Sync iso images between heterogeneous hosts

adriengentil:
Create a shared mount between the heterogenous hosts, so when the ISO image is downloaded on the primary, the secondary (ARM machine) hosts as also access to it automatically.

This changes creates a new role to setup the sshfs mount, fixes some linter issues in the ssh role and lower the ansible log level.

### DIFF
--- a/ansible_files/roles/common/setup_sftp_share/templates/rclone.service.j2
+++ b/ansible_files/roles/common/setup_sftp_share/templates/rclone.service.j2
@@ -1,0 +1,19 @@
+[Unit]
+Description=rclone: Remote FUSE filesystem for cloud storage config {{ rclone_config_file }}
+Documentation=man:rclone(1)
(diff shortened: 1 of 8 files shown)


adriengentil:
/test e2e-metal-assisted-day2-arm-workers
qemu failed to read the iso image, I fixed the permission on the mount

ansible_files/roles/common/setup_sftp_share/templates/rclone.service.j2
```diff
@@ -0,0 +1,19 @@
+[Unit]
```
osherdp, line 1: just to make sure I understand, this rclone service is being deployed on the primary provisioning host?

osherdp (approved):

PR #2039, opened 2023-02-10
MGMT-13247: Export day2 configuration

adriengentil:
Export day2 configuration into assisted-additional-config that will be
sourced by the CI.

Had to refactor the existing tasks to set the facts properly at host
level.


adriengentil:
/close
opened against master :man_facepalming:

PR #2053, opened 2023-02-14
Add minikube controller k8 client wrapper to allow upgrade and scaleup

bkopilov:
Added a base code for minikube controller and a k8_minikube fixture.

minikube_controller initialize instance with k8_client and allow access to resources and change configuration, scaleup/down simulate stop and start. in addition the code allows to change confiugration to any resource object plus waiters till stop/start completed.

Each minikube resource has a kind like ConfigMap, Pod .. we get a reference to the kind type and look for the instance object base on lookup name. Support for change configuration per resource kind type and verify that configuration really changed.

scaleup/down simulate stop/start . when stop we have a waiter and verify that Development and ReplicaSet counters set and till the pods are deleted. When start we verify that Development and ReplicaSet counter/ status set and Pod are up with valid status and expected number of replicas (default 1)

Added to BaseTest k8s_minikube fixture that passed to test functions Create object for minikuber_controller and store ConfigMap configuration We verify that all Developments are up and running. Cleanup fixture happens when test done, we compare ConfigMap after and restore configuration only when there was a patch for change configuration. cleanup has waiter in case configmap changed , we verify that configuration restored for the next testcase.

The test should run in serila mode only.

Planning to add more capablities to the minikube_controller
- Adding metrics for minikube and track kubectl top pods for cpu and mem


bkopilov:
@lalon4  , @eliorerz , @adriengentil

bkopilov:
@osherdp

src/tests/base_test.py
```diff
@@ -78,6 +82,41 @@ def update_parameterized(cls, request: pytest.FixtureRequest, config: BaseConfig
...
+        # Detect if configmap data was changed - need to restore configuration
+        if configmap_before != configmap_after:
+            log.debug(f"configmap {minikube.configmap_name} was changed during the test"
+                      f" {request.node.name}, restoring\n")
+            minikube.patch_resource("ConfigMap",
+                                    "assisted-service-config", {'data': configmap_before})
+            log.info(f"Set deployment scale assisted-service to 0 for test {request.node.name} -waiting")
+            minikube.change_deployment_scale("assisted-service", 0)
+            log.info(f"Set deployment scale assisted-service to 1 for test {request.node.name} -waiting")
+            minikube.change_deployment_scale("assisted-service", 1)
```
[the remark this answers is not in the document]
adriengentil: BTW, the clean way to rollout the deployment is to set an annotation: https://github.com/kubernetes-client/python/issues/1378

src/assisted_test_infra/test_infra/controllers/minikube.py
```diff
@@ -0,0 +1,222 @@
...
+
+    @property
+    def configmap_data(self):
+        config_map_store = self._get_obj_resources_by_name(self.CONFIGMAP, self.configmap_name)
+        assert len(config_map_store) == 1, "configmap kind name should be unique"
+        return config_map_store[0]
+
+    @property
+    def assisted_service_name(self):
+        return "assisted-service"
```
adriengentil, line 54: I think it should be a constant

src/assisted_test_infra/test_infra/controllers/minikube.py
```diff
@@ -0,0 +1,222 @@
...
+            self.POD: self._get_resources_by_kind(self.POD, "v1"),
+            self.SERVICE: self._get_resources_by_kind(self.SERVICE, "v1")}
+
+    @property
+    def namespace(self):
+        return os.environ.get("NAMESPACE", "assisted-installer")
+
+    @property
+    def configmap_name(self):
+        return "assisted-service-config"
```
adriengentil, line 44: I think it should be a constant

src/assisted_test_infra/test_infra/controllers/minikube.py
```diff
@@ -0,0 +1,222 @@
...
+        return "assisted-service"
+
+    def resource_kind(self, name):
+        """Get the resource kind object by name
+        :param name: ReplicaSet, Deployment, Pod, ConfigMap, Service Pod
+        :return: Resource object
+        """
+        return self.mapping_resources.get(name)
+
+    def _get_resources_by_kind(self, kind_name, api_version):
```
adriengentil, line 63:
```suggestion
    def _get_resources_by_kind(self, kind_name: str, api_version: str) -> SomeObject:
```
As the general comment, try to have type annotations, it makes clear what are the types manipulated by the methods

src/assisted_test_infra/test_infra/controllers/minikube.py
```diff
@@ -0,0 +1,222 @@
...
+        """Get resource by api_version and kind
+        :param kind_name: ReplicaSet, Deployment, Pod, ConfigMap, Service Pod
```
adriengentil, line 65:
```suggestion
        :param kind_name: ReplicaSet, Deployment, Pod, ConfigMap, Service, Pod
```

src/assisted_test_infra/test_infra/controllers/minikube.py
```diff
@@ -0,0 +1,222 @@
...
+        In case calling to pod resource we may have multiple pods beginning
+        the same name. assisted-service, assisted-service123 ...
+        :param resource_kind_name: ReplicaSet, Deployment, Pod, ConfigMap, Service Pod
+        :param name_begins_with: lookup name begins the same for all kinds
+        :return: List of all objects beginning the same name for pods
+        """
+        all_resources = self.resource_kind(resource_kind_name).get(
+            namespace=self.namespace, name=None)
+        if not exact_match:
+            replicas = [obj for obj in all_resources.items if name_begins_with in obj.metadata.name]
```
adriengentil, line 83:
```suggestion
            replicas = [obj for obj in all_resources.items if metadata.name.startswith(name_begins_with)]
```
to match the name of the variable `name_begins_with` with the behaviour

src/assisted_test_infra/test_infra/controllers/minikube.py
bkopilov: Ack, fixed. i am keeping the property func if some  actions will be needed before returning the configmap name

src/assisted_test_infra/test_infra/controllers/minikube.py
bkopilov: Ack i will change it  later during this patch ping pong . lets first see that we are aligned with the concept and function . - it will be fixed for sure.

src/tests/base_test.py
```diff
@@ -78,6 +82,41 @@ def update_parameterized(cls, request: pytest.FixtureRequest, config: BaseConfig
...
+        """k8 minikube wrapper client  object fixture passed to test function
+        This function fixture called by testcase, we initialize k8client
+        Storing configmap before test begins -> for later restoring
+        When test func done, check if configmap was changed and restore config.
+        :return:
+        """
+        log.debug(f"--- SETUP --- Creating k8s_minikube for test: {request.node.name}\n")
+        k8_client_ = kubernetes.dynamic.DynamicClient(
+            kubernetes.client.api_client.ApiClient(
+                configuration=kubernetes.config.load_kube_config(),
```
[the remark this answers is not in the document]
bkopilov: Sounds good to me , i see that we have a factory class to initilize client, i did not dig into it yet because i dont use it but i guess that the way we can do it is another fixture called from minikube ... 

Not changing it now , any way its may be extended IMO.

@pytest.fixture
    def k8s_minikube(self, k8s_client request: pytest.FixtureRequest)

src/tests/base_test.py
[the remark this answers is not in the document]
bkopilov: Ack the reason i used scale options because in resiliency testing i must to stop the service and have full controll when in scale up it again ... in addition it allow me to place with scale of assisted-pods if one day we will have more then one.  i implemented the waiters. if no difference in function we can use it unless scale down does not work ...  we can add for sure this option too.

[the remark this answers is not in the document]
bkopilov: > maybe you can factorize this code in `MinikubeController` into a method `rollout_assisted_service`

done

src/assisted_test_infra/test_infra/controllers/minikube.py
```diff
@@ -0,0 +1,242 @@
+import logging
```
eliorerz, line 1: Please rename file to match the class name: e.g. `minikube_controller.py`

src/assisted_test_infra/test_infra/controllers/minikube.py
```diff
@@ -0,0 +1,242 @@
...
+    _get_obj_resources_by_name -> get the obj resource name
+    """
+
+    CONFIGMAP = "ConfigMap"
+    DEPLOYMENT = "Deployment"
+    REPLICASET = "ReplicaSet"
+    POD = "Pod"
+    SERVICE = "Service"
+
+    def __init__(self, k8_client):
```
eliorerz, line 25: As @adriengentil mentioned, Please add type and return type annotations to all of the declared functions

src/assisted_test_infra/test_infra/controllers/minikube.py
```diff
@@ -0,0 +1,242 @@
...
+            self.CONFIGMAP: self._get_resources_by_kind(self.CONFIGMAP, "v1"),
+            self.DEPLOYMENT: self._get_resources_by_kind(self.DEPLOYMENT, "apps/v1"),
+            self.REPLICASET: self._get_resources_by_kind(self.REPLICASET, "apps/v1"),
+            self.POD: self._get_resources_by_kind(self.POD, "v1"),
+            self.SERVICE: self._get_resources_by_kind(self.SERVICE, "v1"),
+        }
+
+    @property
+    def namespace(self):
+        return os.environ.get("NAMESPACE", "assisted-installer")
```
eliorerz, line 42: Avoid getting variables from environment.
All environment variables can be defined [here](https://github.com/openshift/assisted-test-infra/blob/master/src/tests/global_variables/env_variables_defaults.py)

src/assisted_test_infra/test_infra/controllers/minikube.py
```diff
@@ -0,0 +1,242 @@
...
+
+    @property
+    def configmap_name(self):
+        return os.environ.get("CONFIGMAP", "assisted-service-config")
```
eliorerz, line 46: Same here

src/assisted_test_infra/test_infra/controllers/minikube.py
```diff
@@ -0,0 +1,242 @@
...
+        self.mapping_resources = {
+            self.CONFIGMAP: self._get_resources_by_kind(self.CONFIGMAP, "v1"),
+            self.DEPLOYMENT: self._get_resources_by_kind(self.DEPLOYMENT, "apps/v1"),
+            self.REPLICASET: self._get_resources_by_kind(self.REPLICASET, "apps/v1"),
+            self.POD: self._get_resources_by_kind(self.POD, "v1"),
+            self.SERVICE: self._get_resources_by_kind(self.SERVICE, "v1"),
+        }
+
+    @property
+    def namespace(self):
```
eliorerz, line 41: Pass those properties to the class constructor, same as for `configmap_name`, `assisted_service_name`  and every env var that is needed within the controller

src/assisted_test_infra/test_infra/controllers/minikube.py
```diff
@@ -0,0 +1,242 @@
...
+        :param resource_kind_name: ReplicaSet, Deployment, Pod, ConfigMap, Service Pod
+        :param name_begins_with: lookup name begins the same for all kinds
+        :return: List of all objects beginning the same name for pods
+        """
+        all_resources = self.resource_kind(resource_kind_name).get(namespace=self.namespace, name=None)
+        if not exact_match:
+            replicas = [obj for obj in all_resources.items if obj.metadata.name.startswith(name_begins_with)]
+        else:
+            replicas = [obj for obj in all_resources.items if name_begins_with == obj.metadata.name]
+        return replicas
```
eliorerz, line 86:
```suggestion
        if not exact_match:
            return [obj for obj in all_resources.items if obj.metadata.name.startswith(name_begins_with)]
        return [obj for obj in all_resources.items if name_begins_with == obj.metadata.name]
```

src/assisted_test_infra/test_infra/controllers/minikube.py
```diff
@@ -0,0 +1,242 @@
...
+        """Get object by resource obj by namespace and begins name.
+        In case calling to pod resource we may have multiple pods beginning
+        the same name. assisted-service, assisted-service123 ...
+        :param resource_kind_name: ReplicaSet, Deployment, Pod, ConfigMap, Service Pod
+        :param name_begins_with: lookup name begins the same for all kinds
+        :return: List of all objects beginning the same name for pods
+        """
+        all_resources = self.resource_kind(resource_kind_name).get(namespace=self.namespace, name=None)
+        if not exact_match:
+            replicas = [obj for obj in all_resources.items if obj.metadata.name.startswith(name_begins_with)]
```
eliorerz, line 83: Please add logs in this method for every case (if `exact_match` and `not exact_match`)

src/assisted_test_infra/test_infra/controllers/minikube.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def verify_deployments_are_ready(self):
+        """
+        Verify that all deployments are in ready status
+        :return:
+        """
+        deployments = self.resource_kind(self.DEPLOYMENT).get().items
+
+        def check_ready_replicas(x):
+            return x["status"].availableReplicas == 1 and x["status"].readyReplicas == 1
```
eliorerz, line 130: Please change `x` to something more clear

src/assisted_test_infra/test_infra/controllers/minikube.py
```diff
@@ -0,0 +1,242 @@
...
+                running = list(map(lambda pod: hasattr(pod.status, "phase") and pod.status.phase == "Running", pods))
+                return all(running)
+
+            waiting.wait(
+                lambda: pods_are_running() is True,
+                timeout_seconds=timeout,
+                sleep_seconds=sleep,
+                waiting_for="replicas to change successfully",
+            )
+        time.sleep(recover_time_pods)
```
eliorerz, line 185:
```suggestion
            )

        time.sleep(recover_time_pods)
```

src/assisted_test_infra/test_infra/controllers/minikube.py
```diff
@@ -0,0 +1,242 @@
...
+    In order to access to data resource we:
+    _get_resources_by_kind -> this is the kind of resource
+    _get_obj_resources_by_name -> get the obj resource name
+    """
+
+    CONFIGMAP = "ConfigMap"
+    DEPLOYMENT = "Deployment"
+    REPLICASET = "ReplicaSet"
+    POD = "Pod"
+    SERVICE = "Service"
```
eliorerz, line 23:
```suggestion
class Resources(Enum):
    CONFIGMAP = "ConfigMap"
    DEPLOYMENT = "Deployment"
    REPLICASET = "ReplicaSet"
    POD = "Pod"
    SERVICE = "Service"
```

src/assisted_test_infra/test_infra/controllers/minikube.py
```diff
@@ -0,0 +1,242 @@
...
+        name starts with same name with uuids like assisted-service-9fdd7f47-xd8vz
+        :param deployment_name: assisted-service
+        :param replicas_count: total number , 0 for stop 1 and more for start
+        :return:
+        """
+        deployment = self._get_obj_resources_by_name(self.DEPLOYMENT, deployment_name)
+        # verify single name for deployment replicaset exists
+        assert len(deployment) == 1, "deployment kind name should be unique"
+        replicaset = self._get_obj_resources_by_name(self.REPLICASET, deployment_name)
+        assert len(replicaset) == 1, "replicaset kind name should be unique"
```
eliorerz, line 199: Maybe create a function for getting deployment/replicaset/configmap..?
it will be clearer if we will not see the assert in the middle of the flow, e.g.
```python
deployment = self._get_resource(self.DEPLOYMENT, deployment_name)
replicaset = self._get_resource(self.REPLICASET, deployment_name)
```

While the method will look something like that:
```python
def _get_resource(resource_type: Resource, resource_name: str):
    resource = self._get_obj_resources_by_name(resource_type.value, resource_name)
    assert len(replicaset) == 1, f"{resource_type.value} kind name should be unique, name={resource_name}"
    return resource
```

src/assisted_test_infra/test_infra/controllers/minikube.py
```diff
@@ -0,0 +1,242 @@
...
+        self.patch_resource(self.DEPLOYMENT, deployment_name, body)
+        self.wait_for_scale_config_change(self.DEPLOYMENT, deployment_name, replicas_count)
+        self.wait_for_scale_config_change(self.REPLICASET, deployment_name, replicas_count)
+        # When changed to zero pods with this name should be deleted.
+        self.wait_for_pod_change(deployment_name, replicas_count)
+
+    def change_configmap_data(self, dict_data, restart_deployment=True, deployment_name="assisted-service"):
+        """Accept data we would like to change and patch it to configmap
+        Example:
+        {data: {'ENABLE_UPGRADE_AGENT': 'True', 'AGENT_DOCKER_IMAGE': agent_docker_image + tag_version}}
```
eliorerz, line 212:
```suggestion
            {data: {'ENABLE_UPGRADE_AGENT': 'True', 'AGENT_DOCKER_IMAGE': agent_docker_image + tag_version}}
```

src/assisted_test_infra/test_infra/controllers/minikube.py
```diff
@@ -0,0 +1,242 @@
...
+        :param deployment_name - the deployment name to scale down/up
+        :param dict_data:
+        :return:
+        """
+        body = dict_data
+        self.patch_resource(self.CONFIGMAP, self.configmap_name, body)
+        # stop /start development for service
+        if restart_deployment:
+            self.change_deployment_scale(deployment_name, 0)
+            self.change_deployment_scale(deployment_name, 1)
```
eliorerz, line 223: Maybe make a function out of it? You are doing the same logic also in `rollout_assisted_service`

src/tests/base_test.py
```diff
@@ -78,6 +82,31 @@ def update_parameterized(cls, request: pytest.FixtureRequest, config: BaseConfig
...
+        Storing configmap before test begins -> for later restoring
+        When test func done, check if configmap was changed and restore config.
+        :return:
+        """
+        log.debug(f"--- SETUP --- Creating k8s_minikube for test: {request.node.name}\n")
+        k8s_client_ = kubernetes.dynamic.DynamicClient(
+            kubernetes.client.api_client.ApiClient(
+                configuration=kubernetes.config.load_kube_config(),
+            ),
+        )
```
eliorerz, line 98: Maybe initialize it inside the controller constructor? It's more controller specific logic that is used only inside the controller.

src/tests/base_test.py
```diff
@@ -78,6 +82,31 @@ def update_parameterized(cls, request: pytest.FixtureRequest, config: BaseConfig
...
+        minikube = MinikubeController(k8s_client_)
+        # before test begins verify all developments are ready
+        if not minikube.verify_deployments_are_ready():
+            assert "Minikube deployments services not ready/active"
+        configmap_before = dict(minikube.configmap_data["data"])
```
eliorerz, line 103:
```suggestion
            assert "Minikube deployments services not ready/active"

        configmap_before = dict(minikube.configmap_data["data"])
```

src/tests/base_test.py
```diff
@@ -78,6 +82,31 @@ def update_parameterized(cls, request: pytest.FixtureRequest, config: BaseConfig
...
+        yield minikube
+        log.debug(f"--- TEARDOWN --- Deleting k8s_minikube resources for test: {request.node.name}\n")
```
eliorerz, line 105:
```suggestion
        yield minikube
        
        log.debug(f"--- TEARDOWN --- Deleting k8s_minikube resources for test: {request.node.name}\n")
```

src/tests/base_test.py
eliorerz, line 103: I'm not sure that is will do a deepcopy to the dict, maybe change `configmap_data` property` so that it will not be available from outside the class, only a deep copy of it.

eliorerz (changes requested):
Nice work, I really liked the implementation. I added a few remarks .

src/assisted_test_infra/test_infra/controllers/minikube.py
bkopilov: ack

src/assisted_test_infra/test_infra/controllers/minikube.py
bkopilov: ack , i created properies for each resource and update the mapping , hope that was the meaning.

src/assisted_test_infra/test_infra/controllers/minikube.py
bkopilov: ack

src/assisted_test_infra/test_infra/controllers/minikube.py
bkopilov: ack

src/assisted_test_infra/test_infra/controllers/minikube.py
bkopilov: ack

src/assisted_test_infra/test_infra/controllers/minikube.py
bkopilov: ack , done

src/tests/base_test.py
bkopilov: ack

src/tests/base_test.py
bkopilov: Yes , i missed this part , it will return a shallow copy and if someone will change it from the returned reference we will  get the same before and adter, i changed the property configmap_data to copy.deepcopy , Thanks !!!

src/assisted_test_infra/test_infra/controllers/minikube.py
bkopilov: no problem to change to enum but i prefer that the use that will run the function use strings .

src/tests/base_test.py
bkopilov: Hi , i think that that the configmap python code is protected from direct assignment
ResourceField
config_map_store[0]['data']['OS_IMAGES'] = 5
TypeError: 'ResourceField' object does not support item assignment

So if someone copy the configMap python we would not able to modify because all data fields are immutable (str and int) so no worries that someone snick into a list and add elements.....     the only way to hook it but i think its a rare case is by config_map_store[0]['data'].__dict__['OS_IMAGES']  .  looks like we do not need to change it . no real risk here

src/tests/base_test.py
```diff
@@ -78,6 +82,41 @@ def update_parameterized(cls, request: pytest.FixtureRequest, config: BaseConfig
...
+        When test func done, check if configmap was changed and restore config.
+        :return:
+        """
+        log.debug(f"--- SETUP --- Creating k8s_minikube for test: {request.node.name}\n")
+        k8_client_ = kubernetes.dynamic.DynamicClient(
+            kubernetes.client.api_client.ApiClient(
+                configuration=kubernetes.config.load_kube_config(),
+            ),
+        )
+        minikube = MinikubeController(k8_client_)
```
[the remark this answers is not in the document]
adriengentil: maybe something more like `AssistedInstallerInfraController`? I think `ServiceController` is a bit too generic.

src/assisted_test_infra/test_infra/controllers/minikube_controller.py
```diff
@@ -0,0 +1,289 @@
...
+from enum import Enum
+
+import kubernetes.client
+import kubernetes.config
+import kubernetes.dynamic
+import waiting
+
+from tests.global_variables.default_variables import DefaultVariables
+
+global_variables = DefaultVariables()
```
eliorerz, line 12: Remove this from here. There is only single instance of DefaultVariables.?
You can pass the vars you need from base_test

src/tests/base_test.py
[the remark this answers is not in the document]
osherdp: yes, changing the name is a must
especially that the used interface is not really minikube-specific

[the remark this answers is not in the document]
bkopilov: Ack , Do we agree on name ? AssistedInstallerInfraController  .

src/assisted_test_infra/test_infra/controllers/minikube.py
bkopilov: ack , starting to fix annotation

src/assisted_test_infra/test_infra/controllers/assisted_installer_infra_controller.py
```diff
@@ -0,0 +1,296 @@
...
+    @property
+    def configmap_resource(self) -> str:
+        return self._get_resources_by_kind(self.Resources.CONFIGMAP.value, "v1")
+
+    @property
+    def replicaset_resource(self) -> str:
+        return self._get_resources_by_kind(self.Resources.REPLICASET.value, "apps/v1")
+
+    @property
+    def deployment_resource(self) -> kubernetes.client:
```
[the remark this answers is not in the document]
bkopilov: yes, my mistake . i am going to annotate the exact resource return as possible. working on it.
Thanks!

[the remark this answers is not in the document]
bkopilov: as for return from _get_resources_by_kind , its kubernetes.dynamic.Resource . i am checking with debugger and this is the type(obj) for all resources.

src/assisted_test_infra/test_infra/controllers/assisted_installer_infra_controller.py
```diff
@@ -0,0 +1,296 @@
...
+    def configmap_data(self) -> dict:
+        config_map_store = self._get_obj_resources_by_name(self.Resources.CONFIGMAP.value, self.configmap_name)
+        assert len(config_map_store) == 1, "configmap kind name should be unique"
+        return dict(config_map_store[0])
+
+    @property
+    def assisted_service_name(self) -> str:
+        return self.global_variables.deployment_service
+
+    def resource_kind(self, name: str):
```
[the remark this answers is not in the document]
bkopilov: we have kubernetes.dynamic.Resource as a base.

src/assisted_test_infra/test_infra/controllers/assisted_installer_infra_controller.py
```diff
@@ -0,0 +1,296 @@
...
+from enum import Enum
+
+import kubernetes.client
+import kubernetes.config
+import kubernetes.dynamic
+import waiting
+
+
+class AssistedInstallerInfraController:
+    """Manage minikube or kind (TBD) commands / config changes via k8s_client
```
osherdp, line 12:
```suggestion
    """Manage hub cluster configuration changes via kubernetes commands.
```

osherdp: Rename any occurrence of "minikube" (in PR description, commit, code changes, etc.)

PR #2054, opened 2023-02-14
MGMT-12442: Use arm plan to run heterogeneous test

adriengentil:
Few tactical changes to made the e2e test to work:
- Add `packet-setup.sh` to make it compatible with `assisted-baremetal-*` steps (should be cleaned up once [MGMT-11816](https://issues.redhat.com/browse/MGMT-11816) is implemented)
- `cloud-init status --wait` should always pass because cloud-init always fails with `c3.large.arm64` (reported to Equinix)
- Add a check that pings the tunnel ips to ensure the tunnel is setup properly
- [workaround] Use the tunnel ip has much a possible (rclone, virsh) to allow the secondary machine to reach the primary machine, it's probably due to firewalld installed with test-infra, but I don't see any misconfiguration

### DIFF
--- a/ansible_files/equinix_heterogeneous_create_infra_playbook.yml
+++ b/ansible_files/equinix_heterogeneous_create_infra_playbook.yml
@@ -84,6 +84,17 @@
       ansible.builtin.import_role:
         name: common/setup_ipip_tunnel
 
+- name: Check tunnel connectivity
+  hosts: all
+  tasks:
+    - name: Ping peer
+      ansible.builtin.command:
+        cmd: "ping -c 1 -w 2 {{ ipip_tunnel_ipv4 | ansible.utils.ipaddr('peer') }}"
+      retries: 10
+      delay: 1
+      register: result
+      until: result.rc == 0
+
 - name: Share a unique SSH key pair among the created devices
   hosts: heterogeneous
   roles:
@@ -115,7 +126,7 @@
     - name: common/setup_sftp_share
       vars:
         rclone_remote_name: "iso-images"
-        remote_host_ip: "{{ hostvars[groups['secondary'][0]].access_private_ipv4 }}"
+        remote_host_ip: "{{ ipip_tunnel_ipv4 | ansible.utils.ipaddr('peer') }}"
         shared_directory: "{{ iso_images_shared_directory }}"
 
 - name: Export day2 heterogeneous configuration to assisted-test-infra
@@ -126,4 +137,5 @@
         name: test-infra/export-heterogeneous-day2-configuration
       delegate_to: localhost
       vars:
-        assisted_service_url: "{{ hostvars[groups['primary'][0]].access_private_ipv4 }}"
+        assisted_service_url: "{{ ipip_tunnel_ipv4 | ansible.utils.ipaddr('peer') }}"
+        access_libvirt_ip: "{{ ipip_tunnel_ipv4 | ansible.utils.ipv4('address') }}"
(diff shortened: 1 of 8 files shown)


ansible_files/roles/test-infra/export-heterogeneous-day2-configuration/templates/assisted-additional-config.j2
```diff
@@ -1,10 +1,10 @@
 export SERVICE_URL="{{ assisted_service_url }}"
 
-export DAY2_CPU_ARCHITECTURE="{{ ansible_architecture  }}"
+export TEST_FUNC=test_deploy_day2_nodes_cloud
```
adriengentil, line 3: will remove once https://github.com/openshift/release/pull/36638 is merged

terraform_files/equinix-ci-machine/main.tf
```diff
@@ -27,7 +27,10 @@ resource "equinix_metal_device" "ci_devices" {
 
   provisioner "remote-exec" {
     inline = [
-      "cloud-init status --wait"
+      # Wait for cloud-init to complete.
+      # Ignore any errors because some equinix machines fail in cloud-init but
+      # it is without consequenses.
```
osherdp, line 32: wdym about "without consequences"? 
does it mean the code can continue running, only without the intermediate check that makes sure cloud-init finished? iiuc in this case we reach the timeout (or not? if the failure for those machines is immediate) for ``cloud-init status --wait``, ignore it, and continue with the rest of the instructions

ansible_files/roles/equinix/heterogeneous_create_infra/defaults/main.yaml
```diff
@@ -3,7 +3,7 @@ secondary_device_group_name: "secondary"
 equinix_ssh_user: "root"
 equinix_plans: "{{
       {
-       primary_device_group_name : 'c3.small.x86',
-       secondary_device_group_name : 'c3.small.x86'
+       primary_device_group_name : 'c3.medium.x86',
+       secondary_device_group_name : 'c3.large.arm64'
```
osherdp, line 7: for the day2 nodes we need a larger machine? or it's just what equinix supports?

ansible_files/equinix_heterogeneous_create_infra_playbook.yml
```diff
@@ -84,6 +84,17 @@
       ansible.builtin.import_role:
         name: common/setup_ipip_tunnel
 
+- name: Check tunnel connectivity
+  hosts: all
+  tasks:
+    - name: Ping peer
+      ansible.builtin.command:
```
osherdp, line 91: not critical, but net-ping no good? https://docs.ansible.com/ansible/latest/collections/ansible/netcommon/net_ping_module.html#ansible-collections-ansible-netcommon-net-ping-module

adriengentil: `ansible.netcommon.net_ping` looks to be a bit more specialized, it requires to set `ansible_network_os` in which linux is not, so I followed the advice:
`For targets running Python, use the ansible.builtin.shell module along with ping command instead.` :D

ansible_files/roles/equinix/heterogeneous_create_infra/defaults/main.yaml
adriengentil: it's the smaller ARM machine available on equinix

terraform_files/equinix-ci-machine/main.tf
adriengentil: yeah, I mean it's not `cloud-init status --wait` can fail but we can continue nonetheless

osherdp: so it's fail after a timeout? (not sure if I fully understand)
either way, it's not critical

adriengentil: no, equinix metal injects invalid user-data
https://redhat-internal.slack.com/archives/C0NFH1Y1X/p1676456507662149

osherdp (approved):

osherdp (approved):

PR #2055, opened 2023-02-15
Mgmt 13607: Use latest hypershift operator and the payload hostedcontrolplane operator

eranco74:


adriengentil:
Looks like it was already inject something here: https://github.com/openshift/release/blob/master/ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-release-ocm-2.6.yaml#L177
Or is it something else?

eranco74:
> Looks like we already inject something here: https://github.com/openshift/release/blob/master/ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-release-ocm-2.6.yaml#L177 Or is it something else?

It's something else.
The one we inject is for taking the hosted control plane operator from the payload (not really required).
The one I added here is for using the latest hypershift operator image when installing hypershift.
reference for why this is required: https://redhat-internal.slack.com/archives/C01C8502FMM/p1675606561625359

eranco74:
The latest hypershift operator isn't working on  minikube.
I think we need to disable the `cluster-api-provider-agent-release-ocm-2.6-e2e-metal-assisted-capi-periodic`.
We can no longer test capi-agent-provider with assisted-test-infra

eranco74:
https://github.com/openshift/release/pull/36536

PR #2057, opened 2023-02-19
Fix ssh waiter - catch SSHException

bkopilov:
During test_proxy we hit on SSHException("Invalid key") parmiko tries to connect , looks like re-try when hit on failures with differnt method on failure, we should catch ssh exception and hanlde it to make the waiter return bool value


bkopilov:
@adriengentil , @lalon4 , @eliorerz , @osherdp

osherdp (approved):

PR #2061, opened 2023-02-20
MGMT-12442: Use SCSI drives instead of IDE

adriengentil:
ARM VM do not supports IDE/SATA drives, SCSI seems to be a common ground
between x86_64 and ARM machines.

Update bootstrap_in_place as the disk name changed.


adriengentil:
I did not make it configurable because I'm not sure that it is worth to have this granularity.

osherdp (approved):

adriengentil:
committed a change in this branch by mistake :man_facepalming:

osherdp (approved):

PR #2062, opened 2023-02-20
MGMT-12442: Fix DHCP limit range XSLT

adriengentil:
When we have a network `172.16.20.2`, the current transformation is
invalid and gives `172.16.128` in output.

This change fixes this problem by replacing only the last characters
of a given IP.


osherdp (approved):

adriengentil:
Checked dual stack, change behaves as expected:
```  
<ip family='ipv4' address='192.168.127.1' prefix='24'>
    <dhcp>
      <range start='192.168.127.128' end='192.168.127.254'/>
    </dhcp>
  </ip>
  <ip family='ipv6' address='1001:db9::1' prefix='120'>
    <dhcp>
      <range start='1001:db9::2' end='1001:db9::63'/>
    </dhcp>
  </ip>
```

PR #2066, opened 2023-02-22
Fix hardcoded _get_disk_scsi_identifier

bkopilov:
Current code assumes that disks begins with sdX .
CI fails because we are trying to access index 0 by default when lookup returns none . The disks that we have are vdX.

I created a lookup based on the target keys and sd as hardcoded.


bkopilov:
@lalon4 , @eliorerz , @osherdp , @adriengentil

osherdp (approved):

adriengentil (approved):

PR #2067, opened 2023-02-23
MGMT-13792: Setup proxy when downloading logs

adriengentil:
- Setup proxy, required at least for ipv6 job
- remove `podman pull ...` before calling `toolbox` because it seems to always fail with `Error: stat /run/user/0/containers/auth.json: no such file or directory`
- pipe `yes` into `toolbox` because `toolbox` is an interactive command, and it sometimes asks questions where we want to answer `y` all the time


adriengentil:
/cc @osherdp 
let's see...

adriengentil:
`e2e-metal-assisted-kube-api-net-suite` and `e2e-metal-assisted-ipv6` look to be fixed with this change

osherdp (approved):

PR #2072, opened 2023-02-28
MGMT-13792: fix download_logs for OCP <= 4.10

adriengentil:
On OCP<=4.10, toolbox is too old and runs a shell instead of running the
command specified in parameter.

This change detects if the version of toolbox is too old and downloads
a more recent version of it if needed.


osherdp (approved):

adriengentil:
/unhold
looks good!

PR #2080, opened 2023-03-08
Day2 post installation triggers node reboot while addding csr

bkopilov:
Current flow for day2 installation try to approve workers csr while the node is rebooting and we keep trying in the waiter.
Added sshable waiter after day2 installation to verify that node really returns from reboot and approve the csr when node is up and running.

We have failures in kni-auto with day2 and it might be a timeing issue depends on reboot time of the installed node.

### DIFF
--- a/src/assisted_test_infra/test_infra/helper_classes/day2_cluster.py
+++ b/src/assisted_test_infra/test_infra/helper_classes/day2_cluster.py
@@ -104,6 +104,8 @@ class Day2Cluster(BaseCluster):
     def start_install_and_wait_for_installed(self):
         ocp_ready_nodes = self.get_ocp_cluster_ready_nodes_num()
         self._install_day2_cluster()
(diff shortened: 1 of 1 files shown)


bkopilov:
@eliorerz , @osherdp

adriengentil:
> Current flow for day2 installation try to approve workers csr while the node is rebooting and we keep trying in the waiter.

After approving the CSR, we do a `oc get nodes`, and we keep retrying:
https://github.com/openshift/assisted-test-infra/pull/2080/files#diff-40ce21d21a6f351b07dcdaf726446921ffb33a0495f7b8d7845b7b64fd771567R114

it's like your day2 worker never join the cluster, no?
We can add the SSH check to ease the investigation of the issue, but I don't think it will solve it.

adriengentil (approved):

PR #2082, opened 2023-03-12
Revert "Allow reading unregistered cluster details"

osherdp:
Reverts openshift/assisted-test-infra#2081

Apparently only admin users can read unregistered clusters, which makes QE tests fail (as they also go into this path of getting cluster details).

Reverting for now. Later on I'll prepare a patch which only requests unregistered clusters when asked for (and not for all cases).


eliorerz (approved):
LGTM

bkopilov:
Tested the fix -> works for me , i was able to complete installation.

PR #2092, opened 2023-03-19
MGMT-14001: Generate events.html as standalone file

osherdp:
Due to the nature of Jira attachments, it's not possible to dynamically load JSON event files from the events.html based on relative placement of the files.

We can make the events.html file a standalone file (one that you can download by itself and will still show all relevant data) to cope with that issue.


src/assisted_test_infra/download_logs/download_logs.py
```diff
@@ -174,13 +183,22 @@ def download_logs(
...
-            client.download_cluster_events(cluster["id"], get_cluster_events_path(cluster, output_folder))
-            shutil.copy2(os.path.join(os.path.dirname(os.path.realpath(__file__)), "events.html"), output_folder)
+            cluster_events = get_cluster_events_path(cluster, output_folder)
+            client.download_cluster_events(cluster["id"], cluster_events)
+            event_files.append(cluster_events)
+
+            template = jinja2.Environment(
+                loader=jinja2.FileSystemLoader(searchpath=os.path.dirname(__file__))
+            ).get_template("events.html.j2")
+            template.stream(events=merge_events(event_files)).dump(os.path.join(output_folder, "events.html"))
```
eliorerz, line 201: Maybe move this logic to a separate function? This method is already way too long..

src/assisted_test_infra/download_logs/download_logs.py
```diff
@@ -123,6 +122,16 @@ def download_manifests(client: InventoryClient, cluster_id: str, output_folder:
     client.download_manifests(cluster_id, manifests_path)
 
 
+def merge_events(event_paths: list[str]) -> str:
+    events = []
+
+    for event_path in event_paths:
+        with open(event_path, "rt") as event_file:
+            events.extend(json.loads(event_file.read()))
```
eliorerz, line 130: Maybe `SuppressAndLog` here so we won't lose all the events due to single event failure? 
Note: not sure if this is a valid case

osherdp: It extends the list for each source (cluster and each infraenv object attached to it) and not for every event

Since test-infra created the events file a few lines before, I don't expect it to actually fail (if ``client.download_infraenv_events`` fails, then the addition to ``event_files`` wouldn't happen as it's one line after)

src/assisted_test_infra/download_logs/download_logs.py
osherdp: Done!

PR #2093, opened 2023-03-19
Improve usability of download-logs

osherdp:
Making it possible to run it as either:
* ``make download_cluster_logs`` - then it downloads logs for all available clusters.
* ``make download_cluster_logs CLUSTER_ID=...`` - then it downloads logs for the specified cluster ID.

Without this change ``make download_clsuter_logs`` just tries to do:
```
skipper run ... src.assisted_test_infra.download_logs ... --cluster-id # parameter value left empty
```

which doesn't download any log.


eliorerz (approved):
LGTM

PR #2094, opened 2023-03-20
Update skipper.env with AI_BASE_VERSION for agent upgrade versions.

bkopilov:
This param is used for agent upgrade and passed via env to users


bkopilov:
@osherdp , @adriengentil , @eliorerz

talhil-rh:
LGTM

PR #2100, opened 2023-03-22
Fix inspection of deleted clusters

osherdp:
Between listing all clusters (including deleted ones) and getting their information, quite some time can pass where they can get garbage collected.

This change does 2 things in that regard:
* Change the retry interval from 5 minutes (!) to 5 seconds. Reason it's fine is because Jenkins job itself is configured to run every 0.5 hours, so it's not necessary to wait for too long in the scope of a single run.
* If we're getting an API error for a deleted cluster, we can assume it got garbage collected. We only log it for tracking purposes.


adriengentil (approved):
:+1:

osherdp:
Job seems broken. Reported and override it for now

PR #2101, opened 2023-03-23
Fix stale references to infra-envs

osherdp:
Currently the method for fetching infra-env objects is as follows:
* Listing all existing clusters (including ones that are logically removed)
* Fetching cluster object from API (mainly to get the associated hosts which are not retrived in the first stage)
* Fetching infra-env objects from API based on IDs in the host objects

The problem is step 3, as infra-env objects might get deleted in the meantime, meaning those IDs no longer reference to existing infra-envs.

To minimize the problem, this change ignores infra-env IDs in the host objects, and just do another request for the infra-envs that are associated with a certain cluster ID.
It guarantees a more recent representation of current DB state, but more importantly we shouldn't get 404 for requesting those objects. Later on we download events for those infra-envs, but this action is suppressed in case of API errors.

### DIFF
--- a/src/service_client/assisted_service_api.py
+++ b/src/service_client/assisted_service_api.py
@@ -161,7 +161,7 @@ class InventoryClient(object):
     def cluster_get(self, cluster_id: str, get_unregistered_clusters: bool = False) -> models.cluster.Cluster:
         return self.client.v2_get_cluster(cluster_id=cluster_id, get_unregistered_clusters=get_unregistered_clusters)
 
-    def get_infra_env_by_cluster_id(self, cluster_id: str) -> List[Union[models.infra_env.InfraEnv, Dict[str, Any]]]:
+    def get_infra_envs_by_cluster_id(self, cluster_id: str) -> List[Union[models.infra_env.InfraEnv, Dict[str, Any]]]:
         infra_envs = self.infra_envs_list()
         return [infra_env for infra_env in infra_envs if infra_env["cluster_id"] == cluster_id]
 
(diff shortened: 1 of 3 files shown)


src/service_client/assisted_service_api.py
```diff
@@ -161,7 +161,7 @@ def get_all_clusters(self) -> List[Dict[str, Any]]:
     def cluster_get(self, cluster_id: str, get_unregistered_clusters: bool = False) -> models.cluster.Cluster:
         return self.client.v2_get_cluster(cluster_id=cluster_id, get_unregistered_clusters=get_unregistered_clusters)
 
-    def get_infra_env_by_cluster_id(self, cluster_id: str) -> List[Union[models.infra_env.InfraEnv, Dict[str, Any]]]:
+    def get_infra_envs_by_cluster_id(self, cluster_id: str) -> List[Union[models.infra_env.InfraEnv, Dict[str, Any]]]:
```
osherdp, line 164: QE doesn't seem to use ``get_infra_env_by_cluster_id``, so it should be ok to change

gamli75:
few questions:
1. Do we know how many clusters we have with the problem in step 3? 
2. Can we test the download logs job with this change before merging it?

osherdp:
> few questions:
> 
> 1. Do we know how many clusters we have with the problem in step 3?
> 2. Can we test the download logs job with this change before merging it?

1. not easy to estimate. it depends on the scheduling of the garbage-collection
2. yes, trying in http://assisted-jenkins.usersys.redhat.com/job/download_logs/49665/console

osherdp:
@gamli75 the jenkins job passed its data-collection stage, so I think we can get this in

gamli75 (approved):

PR #2115, opened 2023-04-05
NO-ISSUE: Use `.docker/config.json` when there is no `kubelet/config.json`

jhernand:
Currently the `toolbox` tool used by the `man_sosreport.sh` script to collect SOS reports from nodes expects the `/var/lib/kubelet/config.json` file to exist and fails if it doesn't. But in some situations that file will not exist because the cluster installation never started. For example, the upgrade agent test never starts the cluster installation. To avoid failures in this case this patch changes the script so that if the file does't exist it will use `/root/.docker/config.json` instead.

Related: https://github.com/coreos/toolbox/blob/934843e2440263989dd7c9e3f750e6106027165b/rhcos-toolbox


adriengentil (approved):
Looks good, thank you!

adriengentil:
/hold
to wait the result of the CI before merging

adriengentil:
@jhernand e2e-metal-assisted-upgrade-agent looks good!

adriengentil:
single node job should not be affected by this change

PR #2116, opened 2023-04-05
NO-ISSUE: Fix SOS report exception message format

jhernand:
The format used for exceptions that happen when fetching SOS reports fail is incorrect. This patch fixes that.


adriengentil (approved):

adriengentil:
single node job should not be affected by this change

adriengentil:

PR #2126, opened 2023-04-23
Add SSHException catch when ssh_connection

bkopilov:
In some cases when trying to init ssh connection we may get paramiko.ssh_exception.SSHException: Invalid key

While trying to connect to multiple ips we may catch it to prevent raising it back till we complete ssh tries


bkopilov:
@osherdp

osherdp (approved):

PR #2129, opened 2023-04-26
MGMT-14092: Add ansible and terraform to provison CI machine on OCI

adriengentil:
- add a terraform stack that
  - provision network, DNS and a machine that will be used by test-infra
- add an Ansible playbook that:
  - provision the infrastructure using the terraform stack above
  - export the connection details for the assisted common workflow
  - pass the created infrastructure to test-infra through the additional-assisted-config
  - tears down the CI machine infrastructure when the workflow finishes

New job `e2e-oci-assisted` is expected to fail at the test step (need https://github.com/openshift/assisted-test-infra/pull/2170), setup and teardown should work.


adriengentil:
/override ci/prow/e2e-oci-assisted
expected to fail on test step, all good

osherdp (approved):

PR #2136, opened 2023-05-03
MGMT-14411: Fix for OCM 2.8 test failures

paul-maidment:
During the tests `e2e-metal-assisted-kube-api-late-binding-single-node` and `e2e-metal-assisted-kube-api-late-unbinding-single-node`
The installation of the Assisted Service Operator fails on the hub due to a catalog source error

```
{
        "message": "constraints not satisfiable: no operators found in channel ocm-2.8 of package assisted-service-operator in the catalog referenced by subscription assisted-service-operator, subscription as>
        "reason": "ConstraintsNotSatisfiable",
        "status": "True",
        "type": "ResolutionFailed"
 }
```

This PR addresses that issue by making sure that the correct version of the assisted service and the correct version of assisted-service-index are used in the test.
Prior to this PR, the service was using the latest image, as was the index.


adriengentil:
@paul-maidment you need to open the PR against `release-ocm-2.8`, not `master`

PR #2137, opened 2023-05-04
MGMT-14518: Fix vSphere and Nutanix teardown

eliorerz:
- Fix vSphere and Nutanix teardown
- Making `destroy_terraform` target to work with all platforms.

/cc @eliorerz


eliorerz:
4.12 vSphere seems to work fine, Nutanix also

adriengentil (approved):

PR #2139, opened 2023-05-04
Extend and refactor verify_logs_uploaded

bkopilov:
Added more capabilites to the verify action on logs
- Repeated names under directory
- Repeated md5sum under directory - indicate if we have same content
- Included file name inside other files - sub names in other names
- Exceed limit size (1MB) under directory
- Minimal file size (17 Bytes) when '-- No entries --' in logs

Added support for md5sum calculation
Added extraction for tar.gz files

### DIFF
--- a/src/assisted_test_infra/test_infra/utils/logs_utils.py
+++ b/src/assisted_test_infra/test_infra/utils/logs_utils.py
@@ -1,5 +1,8 @@
 import contextlib
+import copy
+import hashlib
 import os
+import shutil
 import tarfile
 import time
 from pathlib import Path
@@ -12,6 +15,9 @@ from service_client import log
 
 OC_DOWNLOAD_LOGS_INTERVAL = 5 * 60
 OC_DOWNLOAD_LOGS_TIMEOUT = 60 * 60
+MAX_LOG_SIZE_BYTES = 1_000_000
+# based on '-- No entries --' size in file
+MIN_LOG_SIZE_BYTES = 17
 
 
 @contextlib.contextmanager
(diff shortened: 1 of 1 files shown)


bkopilov:
@lalon4 , @osherdp

eliorerz (changes requested):

src/assisted_test_infra/test_infra/utils/logs_utils.py
```diff
@@ -50,15 +133,15 @@ def verify_logs_uploaded(
...
-            for gz in os.listdir(tempdir):
-                if "bootstrap" in gz and verify_bootstrap_errors is True:
-                    _verify_node_logs_uploaded(tempdir, gz)
-                    _verify_bootstrap_logs_uploaded(tempdir, gz, installation_success, verify_control_plane)
-                elif "master" in gz or "worker" in gz:
-                    _verify_node_logs_uploaded(tempdir, gz)
-                elif "controller" in gz:
-                    if check_oc:
-                        _verify_oc_logs_uploaded(os.path.join(tempdir, gz))
+            _verify_duplicate_and_size_logs_file(tempdir)
```
adriengentil, line 136: shouldn't this function be called in every directories?

src/assisted_test_infra/test_infra/utils/logs_utils.py
```diff
@@ -33,6 +39,49 @@ def _safe_open_tar(tar_path: Path | str, destination_path: Path | str) -> None:
...
+    Repeated md5sum under directory
+    Included file name inside other files
+    Exceed limit size (1MB) under directory
+    Minimal file size (17 Bytes) when '-- No entries --'
+    :param dir_path:
+    :return: None
+    """
+    files = os.listdir(dir_path)
+    # assert if repeated file name exists
+    assert len(files) == len(set(files)), f"Repeated same file names {str(files)}"
```
adriengentil, line 56: how to files in the same directory can have the same name? You have an example?

bkopilov: Yes ... i know but see this bug : https://issues.redhat.com/browse/MGMT-13842

bkopilov: I dont know how it happended but i added coverage for it , its  strange

adriengentil: in the screenshot, the duplicated files are shown inside the tar? Once extracted, there will be only one file that "wins"? 
I think the duplicated names should be checked by inspecting the tar file, not after extracting it.

src/assisted_test_infra/test_infra/utils/logs_utils.py
```diff
@@ -33,6 +39,49 @@ def _safe_open_tar(tar_path: Path | str, destination_path: Path | str) -> None:
...
+    )
+    assert not any(
+        size_file > MAX_LOG_SIZE_BYTES for size_file in size_list
+    ), f"exceed size limit {MAX_LOG_SIZE_BYTES} from {str(files)}"
+    # check if file name exists in current list names, ex "m1" in [m1.copy, m1.1]
+    for f in files:
+        copy_files = copy.deepcopy(files)
+        copy_files.remove(f)
+        for copy_file in copy_files:
+            assert not (f in copy_file), f"{f} in {copy_files}"
```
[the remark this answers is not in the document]
bkopilov: sorry, what do you mean. in line 73 , we actually take a file name and remove it from the list and check if there is a file that duplicates or with extentions.  so this is also a duplicate test ... i guess we can remove the strange check in line 56

src/assisted_test_infra/test_infra/utils/logs_utils.py
bkopilov: ack

src/assisted_test_infra/test_infra/utils/logs_utils.py
bkopilov: Checking

bkopilov: When we call to this function here we accept
` ['test-infra-cluster-6c2d0ab7_master_test-infra-cluster-6c2d0ab7-master-2.tar', 'test-infra-cluster-6c2d0ab7_master_test-infra-cluster-6c2d0ab7-master-0.tar', 'test-infra-cluster-6c2d0ab7_bootstrap_test-infra-cluster-6c2d0ab7-master-1.tar', 'cluster_events.json', 'cluster_metadata.json', 'controller_logs.tar.gz']`

so we have only files tar that we can check base on the files that we dont have any duplicate content by md5 plus duplicate subset naming of files. we also here verify that tar files are not huge as we had in the bug you saw above (duplicate files + huge files)

bkopilov: Now for the deeper inspection inside the files , it depends on the stage and if its boot strap and it should be handled in the called function ->  line 138 if else section . i know its not optimal and there is a better way ... i want to at least have a test for the positive case that we install successfully and no issues .... 

I think this is the first patch for common usage and i will have to add mores ..... some results includes file and other do not ... so trying on the safe side

adriengentil: To avoid any confusion, can you just add a comment in the code stating what just said? (That basically, there only one level of files)

adriengentil (approved):

PR #2159, opened 2023-05-17
MGMT-14651: Add the option to create multiple disks on vSphere platform installation

eliorerz:
/cc @eliorerz

### DIFF
--- a/terraform_files/vsphere/main.tf
+++ b/terraform_files/vsphere/main.tf
@@ -82,16 +82,16 @@ resource "vsphere_file" "ISO_UPLOAD" {
 resource "vsphere_virtual_machine" "master" {
   count = var.masters_count
 
-  name                 = "${var.cluster_name}-master-${count.index}"
-  resource_pool_id     = data.vsphere_compute_cluster.cluster.resource_pool_id
-  datastore_id         = data.vsphere_datastore.datastore.id
-  num_cpus             = var.master_vcpu
-  num_cores_per_socket = var.vsphere_control_plane_cores_per_socket
-  memory               = var.master_memory
-  guest_id             = "coreos64Guest"
-  folder               = var.vsphere_folder != "" ? "${var.vsphere_parent_folder}/${local.folder}" : vsphere_folder.folder[0].path
-  enable_disk_uuid     = "true"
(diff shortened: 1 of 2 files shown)


eliorerz:
/override ci/prow/e2e-metal-assisted-ipv4v6
/override ci/prow/e2e-metal-assisted-ipv6
/override ci/prow/e2e-metal-assisted-none
/override ci/prow/e2e-metal-assisted-static-ip-suite

terraform_files/vsphere/main.tf
```diff
@@ -100,11 +100,15 @@ resource "vsphere_virtual_machine" "master" {
...
 
-  disk {
-    label            = "master-${count.index}-disk-0"
-    size             = var.master_disk_size_gib
-    eagerly_scrub    = false
-    thin_provisioned = true
+  dynamic "disk" {
+    for_each = range(var.master_disk_count)
+    content {
+      label            = "master-${count.index}-disk-${count.index + 1 + disk.key}"
```
osherdp, line 106: what's the idea of ``disk-${count.index + 1 + disk.key}``? why not say ``disk-${count.index + 1}-${disk.key}``?
also couldn't understand what's ``disk.key`` and where it's coming from

eliorerz: > what's disk.key and where it's coming from

https://registry.terraform.io/providers/hashicorp/vsphere/latest/docs/resources/virtual_machine#key

eliorerz: > why not say disk-${count.index + 1}-${disk.key}

I'll change it, that sounds more reasonable

eliorerz: OK, so the label for `disk-${count.index + 1}-${disk.key}` will be `master-0-disk-1-1` for disk number 2 while on `disk-${count.index + 1 + disk.key}` the label will be `master-0-disk-2`.

osherdp: yeah, I understand what's the difference between the options, but I think the addition of the numbers doesn't make sense and might lead to duplicated values (which is solved by having them separately)

osherdp (approved):

PR #2164, opened 2023-05-21
MGMT-14674: Add operators status log when waiting for operators to finish

eliorerz:
/cc @osherdp @adriengentil


eliorerz:

osherdp (approved):

PR #2165, opened 2023-05-22
MGMT-14518: Fix kubeapi vSphere and Nutanix teardown

eliorerz:
Supposed to fix the vsphere-kube-api tests 

/cc @eliorerz

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -223,10 +223,17 @@ _destroy_virsh:
 	python3 ${DEBUG_FLAGS} -m virsh_cleanup -f test-infra
 
 destroy_terraform_controller:
-	TEST=./src/tests/test_targets.py TEST_FUNC=test_destroy_terraform $(MAKE) test
+	@if [ "$(ENABLE_KUBE_API)" = "false"  ]; then \
+		TEST=./src/tests/test_targets.py TEST_FUNC=test_destroy_terraform $(MAKE) test; \
+	else \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -223,10 +223,17 @@ _destroy_virsh:
 	python3 ${DEBUG_FLAGS} -m virsh_cleanup -f test-infra
 
 destroy_terraform_controller:
-	TEST=./src/tests/test_targets.py TEST_FUNC=test_destroy_terraform $(MAKE) test
+	@if [ "$(ENABLE_KUBE_API)" = "false"  ]; then \
+		TEST=./src/tests/test_targets.py TEST_FUNC=test_destroy_terraform $(MAKE) test; \
+	else \
+	    TEST=./src/tests/test_targets.py TEST_FUNC=test_destroy_available_terraform $(MAKE) test; \
+	fi
```
adriengentil, line 230: isn't `test_destroy_available_terraform` more generic/reliable and should deprecate `test_destroy_terraform` ?

adriengentil (approved):
Just one question, otherwise LGTM

Makefile
eliorerz: Yes, And in future we should probably replace it. I just don't want to transfer it immediately as we did in the previous PR and later had to revert it (due to unhanded resources)

PR #2171, opened 2023-05-29
Update old image used by upgrade agent test

jhernand:
The upgrade agent prepares a cluster for installation with an old agent image and then checks that it can be upgraded to the latest one. But that old image is no longer capable to move the cluster to ready because of changes to the DNS resolution validation. To address that issue this patch changes the test to use a newer old image.

Related: https://github.com/openshift/assisted-service/pull/5208
Related: https://github.com/openshift/assisted-installer-agent/pull/542

### DIFF
--- a/src/tests/test_upgrade_agent.py
+++ b/src/tests/test_upgrade_agent.py
@@ -27,7 +27,7 @@ class TestUpgradeAgent(BaseTest):
     @classmethod
     def _get_previous_agent_image(cls) -> str:
         """Returns the reference to the previous agent image."""
-        return os.environ.get("PREVIOUS_AGENT_IMAGE", "quay.io/edge-infrastructure/assisted-installer-agent:v2.9.0")
+        return os.environ.get("PREVIOUS_AGENT_IMAGE", "quay.io/edge-infrastructure/assisted-installer-agent:v2.20.1")
 
     @classmethod
     def _get_broken_agent_image(cls) -> str:


eliorerz (approved):
LGTM

omertuc:
How can we prevent this test breaking in the future?

jhernand:
> How can we prevent this test breaking in the future?

When the initial test was written some other alternatives were tried, discussed and eventually rejected, see #1839 for details.

Basically we need two different image references that point to agent images that wok. The upgrade agent feature doesn't have a concept of order of versions, it just compares image references as strings. One possible way to do that is to publish a tag that the test can calculate or infer from the current image. For example, if the current image reference (by digest) is `quay.io/my/agent@123`  we could publish a tag `quay.io/my/agent:123`.

@eliorerz @osherdp should we try to do that? Should I open a ticket for it?

osherdp:
> > How can we prevent this test breaking in the future?
> 
> When the initial test was written some other alternatives were tried, discussed and eventually rejected, see #1839 for details.
> 
> Basically we need two different image references that point to agent images that wok. The upgrade agent feature doesn't have a concept of order of versions, it just compares image references as strings. One possible way to do that is to publish a tag that the test can calculate or infer from the current image. For example, if the current image reference (by digest) is `quay.io/my/agent@123` we could publish a tag `quay.io/my/agent:123`.
> 
> @eliorerz @osherdp should we try to do that? Should I open a ticket for it?

Maybe just upgrading between a tagged image and its exact same sha-digest address? this way it should work trivially and the essential behavior is still being tested

jhernand:
> Maybe just upgrading between a tagged image and its exact same sha-digest address? this way it should work trivially and the essential behavior is still being tested
> 
> edit: I missed that's what you're suggestion. Got confused because you wrote `quay.io/my/agent@123` and not `quay.io/my/agent@sha256:...`

Yes, that is what I meant, I forgot to add the `sha256:...` part, sorry.

The complication there is that the test will need to figure out what is the tag. Currently the test gets the image reference from the `AGENT_DOCKER_IMAGE` environment variable of the `assisted-service-config` configmap, and that happens to be a digest:

```
registry.build03.ci.openshift.org/ci-op-8q35nb4y/pipeline@sha256:d49ab91336bbd507ff54252dd7047b383de6d4e4019069cc449ca437512b6edd'
```

Is there a simple way to find the corresponding tag? What I was suggesting is to create a tag with the same text than the digest:

```
registry.build03.ci.openshift.org/ci-op-8q35nb4y/pipeline:d49ab91336bbd507ff54252dd7047b383de6d4e4019069cc449ca437512b6edd
```

That would simplify the code of the test. Without that the test can inspect the image repository (with skopeo, or something similar) and try to find the tag. I think I tried that in one of the first iterations of the original patch, but we considered it to be too complicated. Any other suggestion?

osherdp:
> > Maybe just upgrading between a tagged image and its exact same sha-digest address? this way it should work trivially and the essential behavior is still being tested
> > edit: I missed that's what you're suggestion. Got confused because you wrote `quay.io/my/agent@123` and not `quay.io/my/agent@sha256:...`
> 
> Yes, that is what I meant, I forgot to add the `sha256:...` part, sorry.
> 
> The complication there is that the test will need to figure out what is the tag. Currently the test gets the image reference from the `AGENT_DOCKER_IMAGE` environment variable of the `assisted-service-config` configmap, and that happens to be a digest:
> 
> ```
> registry.build03.ci.openshift.org/ci-op-8q35nb4y/pipeline@sha256:d49ab91336bbd507ff54252dd7047b383de6d4e4019069cc449ca437512b6edd'
> ```
> 
> Is there a simple way to find the corresponding tag? What I was suggesting is to create a tag with the same text than the digest:
> 
> ```
> registry.build03.ci.openshift.org/ci-op-8q35nb4y/pipeline:d49ab91336bbd507ff54252dd7047b383de6d4e4019069cc449ca437512b6edd
> ```
> 
> That would simplify the code of the test. Without that the test can inspect the image repository (with skopeo, or something similar) and try to find the tag. I think I tried that in one of the first iterations of the original patch, but we considered it to be too complicated. Any other suggestion?

maybe a couple of more options:
* making use of the ``quay.io/edge-infrastructure/assisted-service:latest`` image which should always get updated
* query ``v<x>.<y>.<z>``` tags under ``quay.io/edge-infrastructure/assisted-service`` and taking the latest sem-ver available. Try ``curl -L "https://quay.io/api/v1/repository/edge-infrastructure/assisted-service?onlyActiveTags=true" | jq -r '.tags[].name'`` to understand what I mean

jhernand:
Thanks @osherdp, using the `latest` tag is simple and reasonable: https://github.com/openshift/assisted-test-infra/pull/2180.

PR #2172, opened 2023-05-29
Add fall_on_pending_status for waiters

bkopilov:


bkopilov:
@adriengentil , @osherdp , @eliorerz , @lalon4

gamli75:
@bkopilov can you please add a description of why we need this change? what was the problem?

bkopilov:
> @bkopilov can you please add a description of why we need this change? what was the problem?

Its related to https://github.com/openshift/assisted-test-infra/pull/2161/files
We should add more places to modify the param : fall_on_pending_status

src/assisted_test_infra/test_infra/helper_classes/cluster.py
```diff
@@ -512,13 +512,20 @@ def wait_for_write_image_to_disk(self, nodes_count: int = MINIMUM_NODES_TO_WAIT)
             nodes_count=nodes_count,
         )
 
-    def wait_for_host_status(self, statuses, fall_on_error_status=True, nodes_count: int = MINIMUM_NODES_TO_WAIT):
+    def wait_for_host_status(
+        self,
+        statuses,
+        fall_on_error_status=True,
+        nodes_count: int = MINIMUM_NODES_TO_WAIT,
+        fall_on_pending_status=False,
```
adriengentil, line 520: @eliorerz "fall" was an intended wording or is it a typo for "fail"?

bkopilov: its fall_ in all places  ... because i guess pending is not a failure

adriengentil (approved):

src/assisted_test_infra/test_infra/helper_classes/cluster.py
adriengentil: Yeah, it's a wierd name to me, the goal is to raise en error when the boolean is set https://github.com/openshift/assisted-test-infra/blob/e8af7bad44d692ff65818e40eaaa4408fbefdeaa/src/assisted_test_infra/test_infra/utils/waiting.py#L33-L40

Anyway, I approve as it is already made like that.

bkopilov: ack Thanks !

PR #2180, opened 2023-05-30
Use `latest` agent image in upgrade test

jhernand:
This patch changes the upgrade agent test so that it uses the `latest` tag instead of a fixed version number.

Related: https://github.com/openshift/assisted-test-infra/pull/2171

### DIFF
--- a/src/tests/test_upgrade_agent.py
+++ b/src/tests/test_upgrade_agent.py
@@ -27,7 +27,7 @@ class TestUpgradeAgent(BaseTest):
     @classmethod
     def _get_other_agent_image(cls) -> str:
         """Returns the reference to the other agent image."""
-        return os.environ.get("OTHER_AGENT_IMAGE", "quay.io/edge-infrastructure/assisted-installer-agent:v2.20.1")
+        return os.environ.get("OTHER_AGENT_IMAGE", "quay.io/edge-infrastructure/assisted-installer-agent:latest")
 
     @classmethod
     def _get_current_agent_image(


omertuc:
This might cause issues when we run tests on release branches (ocm-release-2.7, ocm-release-2.6), etc

osherdp:
> This might cause issues when we run tests on release branches (ocm-release-2.7, ocm-release-2.6), etc

iiuc this functionality is only enabled for saas environments

omertuc:
>iiuc this functionality is only enabled for saas environments

Why? it's really needed in ACM where they have long running agents on unbound hosts

osherdp:
> > iiuc this functionality is only enabled for saas environments
> 
> Why? it's really needed in ACM where they have long running agents on unbound hosts

No idea, I'm just the messenger 
It's disabled by default (without an option to enable it in the operator) so maybe we just didn't want to tackle it ATM

jhernand:
> This might cause issues when we run tests on release branches (ocm-release-2.7, ocm-release-2.6), etc

This change only replaces `vX.Y.Z` with `latest`. If there were potential issues with testing release branches then they already existed; the change only makes things a little better.

> > iiuc this functionality is only enabled for saas environments
> 
> Why? it's really needed in ACM where they have long running agents on unbound hosts

We never had a explicit ticket asking to enable it by default for ACM. I opened it now: https://issues.redhat.com/browse/MGMT-14769. Please comment there.

osherdp:
@jhernand can you check the failed job?

jhernand:
> @jhernand can you check the failed job?

@osherdp I took a look at that, and to other recent failures of the job. Initially I thought this was related to workers registering too late, but it wasn't the case, it is also happening with master nodes. I decided that it is better to just avoid using the non-existing image: https://github.com/openshift/assisted-test-infra/pull/2182

PR #2182, opened 2023-05-31
Don't use non existing image for agent upgrade test

jhernand:
Currently the agent upgrade test uses a non existing image to ensure that all the hosts have stopped using the previous image. This causes timeouts because when hosts try to register using that non existing image they fail and then wait for one hour. To avoid that this patch changes the test so that it doesn't use that non existing image, instead of that it explicitly waits till all the hosts have upgraded to the new image.


eliorerz (approved):
LGTM

PR #2185, opened 2023-06-04
MGMT-14823: Allow installing MetalLB operator using test-infra

eliorerz:
- Add support with MetalLB operator.
- Allow setting operator properties 
- Add missing methods for MCE operator 

/cc @osherdp @gamli75 @adriengentil

### DIFF
--- a/src/assisted_test_infra/test_infra/helper_classes/cluster.py
+++ b/src/assisted_test_infra/test_infra/helper_classes/cluster.py
@@ -118,6 +118,11 @@ class Cluster(BaseCluster):
         if self._config.cpu_architecture:
             extra_vars["cpu_architecture"] = self._config.cpu_architecture
 
+        if len(self._config.olm_operators) > 0:
+            olm_operators = self.get_olm_operators()
+            if olm_operators:
+                extra_vars["olm_operators"] = olm_operators
+
         cluster = self.api_client.create_cluster(
             self._config.cluster_name.get(),
             ssh_public_key=self._config.ssh_public_key,
@@ -127,7 +132,6 @@ class Cluster(BaseCluster):
             additional_ntp_source=self._config.additional_ntp_source,
             user_managed_networking=self._config.user_managed_networking,
             high_availability_mode=self._config.high_availability_mode,
-            olm_operators=[{"name": name} for name in self._config.olm_operators],
             disk_encryption=disk_encryption,
             tags=self._config.cluster_tags or None,
             **extra_vars,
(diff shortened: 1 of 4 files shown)


src/consts/olm_operators.py
```diff
@@ -152,4 +158,15 @@ def get_exception_factory(operator: str):
...
+
     return OperatorFailedError
+
+
+def get_operator_properties(operator: str, **kwargs) -> str:
+    if operator == OperatorType.METALLB:
+        if (api_vip := kwargs.get("api_vip")) and (ingress_vip := kwargs.get("ingress_vip")):
+            return f'{{"api_ip": "{api_vip}", "ingress_ip": "{ingress_vip}"}}'
+
+    return ""
```
adriengentil, line 172:
```suggestion
def get_operator_properties(operator: str, **kwargs) -> Optional[str]:
    if operator == OperatorType.METALLB:
        if (api_vip := kwargs.get("api_vip")) and (ingress_vip := kwargs.get("ingress_vip")):
            return f'{{"api_ip": "{api_vip}", "ingress_ip": "{ingress_vip}"}}'

    return None
```
would be more expressive I think

src/assisted_test_infra/test_infra/helper_classes/cluster.py
```diff
@@ -235,14 +252,25 @@ def set_ignored_validations(
...
-        self.set_olm_operator(consts.OperatorType.CNV, properties=properties)
+    def set_lvm(self, properties: str = None, update: bool = False):
+        self.set_olm_operator(consts.OperatorType.LVM, properties=properties, update=update)
 
-    def set_lvm(self, properties=None):
-        self.set_olm_operator(consts.OperatorType.LVM, properties=properties)
+    def set_mce(self, properties: str = None, update: bool = False):
+        self.set_olm_operator(consts.OperatorType.MCE, properties=properties, update=update)
+
+    def set_metallb(self, properties: str = None, update: bool = False):
```
adriengentil, line 267:
```suggestion
    def set_metallb(self, properties: Optional[str] = None, update: bool = False) -> None:
```
- If the string can be `None` then the type should be `Optional`
- Let's use typing also for the return type

src/assisted_test_infra/test_infra/helper_classes/cluster.py
```diff
@@ -136,6 +140,19 @@ def _create(self) -> str:
         self._config.cluster_id = cluster.id
         return cluster.id
 
+    def get_olm_operators(self) -> List:
```
adriengentil, line 143:
```suggestion
    def get_olm_operators(self) -> List[str]:
```
`List` isn't complete without the type held inside

adriengentil (commented):
Looks good overall! Few nits around types

PR #2195, opened 2023-06-12
Enabling use all parameter of list events

iovadia:
Enabling use all parameter of list events
e.g. limit and offset in order test pagination

/cc @eliorerz @osherdp @lital @danmanor


danmanor:
Maybe we should use named query parameters for clarity sake ? @iovadia

osherdp (approved):

PR #2204, opened 2023-06-20
default_triggers skip when update_param called from tests

bkopilov:
Current code use default_triggers and when we set env param that appears in the triggers we skipp it in is_user_set function that check __is_user_set equal to False

Added support in update_parameterized to set __is_user_set to True. When triggered call it skip it.

The priority for params are:
global (env param from export)
update_parameterized --> called from testcases
default_triggers -> if variables not __is_user_set else skip

### DIFF
--- a/src/assisted_test_infra/test_infra/utils/env_var.py
+++ b/src/assisted_test_infra/test_infra/utils/env_var.py
@@ -50,6 +50,10 @@ class EnvVar:
     def is_user_set(self):
         return self.__is_user_set
 
+    @is_user_set.setter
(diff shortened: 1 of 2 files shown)


bkopilov:
@eliorerz , @lalon4

bkopilov:
@adriengentil

bkopilov:
Talked with Elior , understand that accessting to env param object is risky and better to handle it from baseConig.
Close this one and submit another patch

bkopilov:
closed , i can refactor this one but dont want to waist resources

PR #2206, opened 2023-06-20
update user_set when calls update_parameterized

bkopilov:
default_trigger checks if param already set by user in env variable.
Added support for update_parameterized from tests
The priority for params are:
global (env param from export)
update_parameterized --> called from testcases


bkopilov:
@eliorerz , @lalon4

eliorerz:
I just opened PR for that https://github.com/openshift/assisted-test-infra/pull/2207

PR #2207, opened 2023-06-20
MGMT-15018: Add option to mark variable origin

eliorerz:
We currently can only differ if variable is set as environment variable or not.
This PR add the option to know the origin of a variable - where it was set

### DIFF
--- a/src/assisted_test_infra/test_infra/helper_classes/config/base_config.py
+++ b/src/assisted_test_infra/test_infra/helper_classes/config/base_config.py
@@ -1,9 +1,10 @@
 from abc import ABC, abstractmethod
 from dataclasses import asdict, dataclass
 from types import NoneType
-from typing import Any, Union, get_args, get_origin
+from typing import Any, Optional, Union, get_args, get_origin
 
-from triggers.env_trigger import DataPool, Triggerable
+from service_client import log
+from triggers.env_trigger import DataPool, Triggerable, VariableOrigin
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


adriengentil (approved):
Looks good! Jous curious, what is the use case for that?

eliorerz:
> Hey suggestion , ( your code worked for me but may i suggesst something ? )

The main issue here is that triggerable set should be protected while config set should not, when we join those two we basically allowing anyone to update the global_config which we tried to prevent in the first place.

eliorerz:
> Looks good! Just curious, what is the use case for that?

@adriengentil QE want to prevent override when setting variable using `parametrized` the same as we are doing with environment variables

bkopilov:
> > Hey suggestion , ( your code worked for me but may i suggesst something ? )
> 
> The main issue here is that triggerable set should be protected while config set should not, when we join those two we basically allowing anyone to update the global_config which we tried to prevent in the first place. Also, the `_keys_origin` cannot be class member and need to be initialized as part of object initialization

Ack .
why `_keys_origin` cannot be class member   ??  i dont understand.  we hold  a trigger status for variable and we dont care what object is calling ??

eliorerz:
> why _keys_origin cannot be class member ?? i dont understand. we hold a trigger status for variable and we dont care what object is calling ?? It will allow us to mainain _keys_origin for all object created and it wont depends of its type object.

Because `parameterized` is "working" on instance of a configuration and not all of the configurations the same. Assuming that can cause issues during tests (serial or parallel)

eliorerz:

PR #2209, opened 2023-06-22
MGMT-14240: Plug OCI platform e2e test into CI

adriengentil:
Fix several fixes in order to make the OCI e2e test work in Prow:
- Fix NSGs on clusters nodes
  - Allow communications between them
  - Allow communications from the nodes to the LB 
- Prevent the creation of DNS entries on the CI machine, it's not required since we setup DNS in OCI 
- Implemented various retries, I observed a lot of flakiness from OCI
- Force user managed networking to true when platform is OCI

### DIFF
--- a/src/assisted_test_infra/test_infra/controllers/node_controllers/tf_controller.py
+++ b/src/assisted_test_infra/test_infra/controllers/node_controllers/tf_controller.py
@@ -101,6 +101,7 @@ class TFController(NodeController, ABC):
         for node in nodes:
             self.shutdown_node(node.name)
 
+    @abstractmethod
     def set_dns(self, api_ip: str, ingress_ip: str) -> None:
         utils.add_dns_record(
             cluster_name=self.cluster_name,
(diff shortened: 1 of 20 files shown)


src/assisted_test_infra/test_infra/controllers/node_controllers/tf_controller.py
```diff
@@ -101,6 +101,7 @@ def shutdown_all_nodes(self) -> None:
         for node in nodes:
             self.shutdown_node(node.name)
 
+    @abstractmethod
```
eliorerz, line 104: Why annotate it with `abstractmethod`? Not sure it's needed here

adriengentil: ah ok, I thought it was required since we override it

eliorerz: Nope, `abstractmethod` raise an exception in the inheriting class if the abstract method is not implemented. It's basically force the inheriting classes to override this method

eliorerz (approved):
LGTM

PR #2210, opened 2023-06-22
Containerized contoller cleanup when raised exception

bkopilov:
Current code depends on _is_running for cleanup fixture. In case this param is False , remove function skipp cleanup

Moving the _is_running to the top of the run function to trigger deletion .

### DIFF
--- a/src/assisted_test_infra/test_infra/controllers/containerized_controller.py
+++ b/src/assisted_test_infra/test_infra/controllers/containerized_controller.py
@@ -19,6 +19,7 @@ class ContainerizedController(ABC):
             return
         log.info(f"Running {self.__class__.__name__} Server {self._name}")
 
+        self._is_running = True
         self._on_container_start(**kwargs)
         base_run_flags = [
             "-d",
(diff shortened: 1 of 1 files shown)


bkopilov:
@eliorerz , @adriengentil  , @lalon4

adriengentil:
instead of maintaining a boolean, can't we just ask podman if the container is running?

bkopilov:
> podman if the container is running?

Yes we can but there is another problem that we should address on self._on_container_start(**kwargs)
In this code user is free to override way on create and cleanup ... so you may into issue that controller is "down" but user allocates resoures (podman-remote command) on _on_container_start

adriengentil:
we could define `is_running` as a property, and the user would have to override it as well

adriengentil (approved):
approving to solve the issue, we could think at something better later

bkopilov:
> we could define `is_running` as a property, and the user would have to override it as well

Yes , also no need boolean at all ... we just need to verify on run if the container already up and on remove we always run delete... 

maybe decorate the function that test if running before .... something like:

@is_running_container
def run(self, **kwargs):
    .....

def is_running_container(self, func):
     def inner(**kwargs):
         if check_conainer_running:
             return func(**kwargs)
         retun None
      return inner
      
I did not try it but the idea is that we wont call to run if container is already running
if you agree i will work on it.

PR #2211, opened 2023-06-25
MGMT-14944: Implement custom manifests upload in test infra

eliorerz:
/cc @eliorerz


adriengentil (changes requested):
I think we should expect a directory with both `manifests` and `openshift` inside (both optional) and we upload the files is their respective folders using the API.

README.md
```diff
@@ -126,34 +126,37 @@ Check the [Installation Guide](GUIDE.md) for installation instructions.
...
+| `IPXE_BOOT`                  | Boots VMs using iPXE if set to `true`, default: `false`                                                                                                   |
+| `PLATFORM`                   | The openshift platform to integrate with, one of: `baremetal`, `none`,`vsphere`, default: `baremetal`                                                     |
 | `DISCOVERY_KERNEL_ARGUMENTS` | JSON formatted string containing array of dictionaries each having 2 attributes: `operation` and `value`. Currently, only `append` operation is supported. |
 | `CPU_ARCHITECTURE`           | CPU architecture of the nodes that will be part of the cluster, one of: `x86_64`, `arm64`, `s390x`, `ppc64le`, default: `x86_64`                           |
+| `CUSTOM_MANIFESTS_FILES`     | List of local manifest files separated by commas                                                                                                              |
+| `CUSTOM_MANIFESTS_FOLDER`    | The folder for the manifest to modify in assisted-service, default: `manifests`                                                                            |
```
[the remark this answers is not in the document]
eliorerz: You can do it, this is why we have this env var. The default value is `manifests` (the same as in the service) but we can set it to `openshift` if you want

adriengentil, line 158: Can be removed, no?

src/consts/consts.py
```diff
@@ -139,6 +139,7 @@ def all(cls):
 TEST_TARGET_INTERFACE = "vnet3"
 SUFFIX_LENGTH = 8
 DEFAULT_BOOT_DEVICES = ["hd", "cdrom"]
+DEFAULT_CUSTOM_MANIFESTS_FOLDER = "manifests"
```
adriengentil, line 142: Can be removed here as well

src/tests/global_variables/env_variables_defaults.py
```diff
@@ -70,6 +71,11 @@ class _EnvVariables(DataPool, ABC):
     olm_operators: EnvVar = EnvVar(
         ["OLM_OPERATORS"], loader=lambda operators: re.split(r"\s|,", operators.lower()), default=[]
     )
+    custom_manifests: EnvVar = EnvVar(
+        ["CUSTOM_MANIFESTS_FILES"],
+        loader=lambda files: [get_manifest_from_path(Path(f)) for f in re.split(r"\s|,", files)],
```
adriengentil, line 76: what about discovering the files if "CUSTOM_MANIFESTS_FILES" points to a directory? I think it would be easier from user point of view, instead of listing each file.

eliorerz: I wanted to mimic the service and UI behavior, but we can add a folder

adriengentil: I think it would be great to be able to add a folder, it's also what `aicli` does

adriengentil (approved):
Looks good! Thanks!

PR #2215, opened 2023-06-27
Remove OCS

iovadia:
OCS no longer supported, all OCS renamed to ODF


eliorerz:

PR #2220, opened 2023-07-05
NO-ISSUE: Add info to the raised exception instead of exception chaining

eliorerz:
/cc @osherdp


osherdp (approved):
PyCon already proofs useful :)

eliorerz:
Exception seems ok 
```
waiting.exceptions.TimeoutExpired: Timeout of 3600 seconds expired waiting for Monitored ['builtin'] operators to be in of the statuses ['available']
Failed to deploy the following operators ['cvo']
```

PR #2225, opened 2023-07-10
Group dependabot updates in one PR

osherdp:
Following instructions at
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups, this makes dependabot to only open one PR for updating all the Python third-party dependencies, which should save some costs for us (as we don't need to run several e2e tests for each dependency being changed).

/cc @adriengentil @eliorerz


eliorerz (approved):
LGTM

PR #2228, opened 2023-07-13
NO-ISSUE: Delete terraform folder after cleaning all resources, add logs to destroy_available_terraform.

eliorerz:
/cc @osherdp @adriengentil

### DIFF
--- a/src/assisted_test_infra/test_infra/tools/terraform_utils.py
+++ b/src/assisted_test_infra/test_infra/tools/terraform_utils.py
@@ -17,12 +17,14 @@ class _Terraform(Terraform):
     force flag only if it's set to true"""
 
     def destroy(self, dir_or_plan=None, force=False, **kwargs):
+        capture_output = os.getenv("DEBUG_TERRAFORM") is not None
+
         default = kwargs
         if force:
             default["force"] = force
         options = self._generate_default_options(default)
         args = self._generate_default_args(dir_or_plan)
-        return self.cmd("destroy", *args, **options)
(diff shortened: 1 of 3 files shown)


src/tests/test_targets.py
```diff
@@ -112,5 +117,10 @@ def test_destroy_available_terraform(
                     controller.tf.set_vars(**config_vars)
                     controller.destroy_all_nodes()
                     destroyed_clusters += 1
+                    resources_deleted = True
```
osherdp, line 120: Do we need this boolean? Or we can detect if deletion occurred by checking if the files exist or not

src/assisted_test_infra/test_infra/tools/terraform_utils.py
```diff
@@ -17,12 +17,16 @@ class _Terraform(Terraform):
     force flag only if it's set to true"""
 
     def destroy(self, dir_or_plan=None, force=False, **kwargs):
+        capture_output = True
+        if os.getenv("DEBUG_TERRAFORM") is not None:
+            capture_output = False
```
osherdp, line 22: nit:
```suggestion
        capture_output = os.getenv("DEBUG_TERRAFORM") is not None
```

src/tests/test_targets.py
eliorerz: We want tot delete the resource folder only if the resources were destroyed, because that this target destroy specific platform resources a folder could exist but resources wasn't destroyed yet

osherdp (approved):

PR #2230, opened 2023-07-23
Increase timeout waiting for SSH connection

danielerez:
In order to stabilize day2 periodic e2e test[1],
increase the timeout waiting for SSH connection on nodes.

This could mitigate the error raised on sporadically failed tests[2]:
```
Timeout of 180 seconds expired waiting for nodes to allow ssh connection
```

[1] https://prow.ci.openshift.org/job-history/gs/origin-ci-test/logs/periodic-ci-openshift-assisted-test-infra-master-e2e-metal-assisted-day2-periodic
[2] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-openshift-assisted-test-infra-master-e2e-metal-assisted-day2-periodic/1681865505863897088/artifacts/e2e-metal-assisted-day2-periodic/assisted-common-test/build-log.txt


src/assisted_test_infra/test_infra/helper_classes/nodes.py
```diff
@@ -245,7 +245,7 @@ def _all_nodes_allow_ssh_connection():
 
         waiting.wait(
             lambda: _all_nodes_allow_ssh_connection(),
-            timeout_seconds=180,
+            timeout_seconds=10 * consts.MINUTE,
```
lalon4, line 248: 10 minutes sounds a lot to wait for nodes to have the ssh connection. How did we got to this number?

lalon4: just noticed its a draft. ignore me

danielerez: Yeah, 3 mins seems to low according to failed tests, we can try also with 5 mins.

osherdp:
@danielerez just in general, I am not 100% sure it will solve the actual issue
We've started seeing problems SSHing to some nodes in some cases (I think mainly the bootstrap node), possibly due to a bug introduced introduced when we moved to rhel9:
https://redhat-internal.slack.com/archives/C02CZNQHGN8/p1683201039540049

src/assisted_test_infra/test_infra/helper_classes/nodes.py
lalon4: Note that on QE CI we run tests in parallel of 3 (meaning possibly 15+ nodes at a time), and we dont have ssh failures. We do have rare cases where some nodes fails to boot but mostly stable. So i would also suggest to check other possible reasons for your failures

danielerez:
> @danielerez just in general, I am not 100% sure it will solve the actual issue
> 
> We've started seeing problems SSHing to some nodes in some cases (I think mainly the bootstrap node), possibly due to a bug introduced introduced when we moved to rhel9:
> 
> https://redhat-internal.slack.com/archives/C02CZNQHGN8/p1683201039540049

ack, so we can hold this change then.

PR #2234, opened 2023-07-25
NO-ISSUE: Fix issue that prevent deploying nodes on platform different from cluster platform

eliorerz:
Sometimes for testing purposes we want to create nodes on one platform and set the cluster platform with another, e.g.
```bash
make test TF_PLATFORM=nutanix PLATFORM=baremetal
``` 

Also remove unused code that uses `get_terraform_controller` method

/cc @osherdp @adriengentil

### DIFF
--- a/src/tests/base_test.py
+++ b/src/tests/base_test.py
@@ -34,6 +34,7 @@ from assisted_test_infra.test_infra.controllers import (
     TerraformController,
     VSphereController,
 )
+from assisted_test_infra.test_infra.controllers.node_controllers.tf_controller import TFController
 from assisted_test_infra.test_infra.helper_classes import kube_helpers
 from assisted_test_infra.test_infra.helper_classes.cluster import Cluster
 from assisted_test_infra.test_infra.helper_classes.config import BaseConfig, BaseNodesConfig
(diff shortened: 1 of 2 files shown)


adriengentil (approved):

src/tests/base_test.py
```diff
@@ -383,27 +384,25 @@ def trigger_configurations(
...
-        return TerraformController(controller_configuration, entity_config=cluster_configuration)
+        return self.get_terraform_controller(controller_configuration, cluster_configuration)
 
+    @classmethod
     def get_terraform_controller(
-        self, controller_configuration: BaseNodesConfig, cluster_configuration: ClusterConfig
-    ) -> TerraformController:
-        if cluster_configuration.platform == consts.Platforms.VSPHERE:
+        cls, controller_configuration: BaseNodesConfig, cluster_configuration: ClusterConfig
+    ) -> TerraformController | TFController:
```
osherdp, line 392: wait, we have ``TerraformController`` and ``TFController``? what's the difference between them?

eliorerz: Yes, `TerraformController` is the one you familiar with.
A while back we added the OCI controller and I noticed that we duplicate a lot of code between the other node controllers (`VsphereController`, `NutanixController`, `OciController`), all of those with had a common base code that we moved to `TFController`.
The final goal is to create a `LibvirtController` that will also inherit from `TFController` and will do some order with the terraform/libvirt controllers.
You can see the description in this PR - https://github.com/openshift/assisted-test-infra/pull/2179

eliorerz: The main problem with `TerraformController` are its dependencies, and moving the new base code will definitely will break QE (and probably our) tests. Keep it with backward compatibility will require a lot of work because the none/baremetal terraform files are not compatible with the new controller.

osherdp: yeah, I'm just commenting about the semantics, because:
* maybe we should have named it ``BaseTerraformController``? Currently it's hard seeing the relationship between the controller when reading their names
* Instead of ``TerraformController | TFController`` we can say this function returns ``TFController``, unless their API is very different from each other

eliorerz: - I think we can rename it to `BaseTerraformController`, It's actually a better name I just didn't want for someone to think that `BaseTerraformController` is somehow related to `TerraformController`.
- I think they kind of different, the base interface (`NodeController`) is similar, but there are few other methods that are different. Furthermore it's actually a different object, so I don't think we should return `TFController` alone.

Anyway, Back then I successfully was able to move `baremetal` flow to the new base cluster but had some difficulties with `none` (uses different terraform files) so I left it aside for now

PR #2235, opened 2023-07-26
Fix assets.py listAllNetworks

bkopilov:
In parallel tests we may get a list of all networks but testcase cleanup may delete the netowrk object from the test. We get here all libvirt network and we must consider case when one of the network no more available while we in the loop.

### DIFF
--- a/src/assisted_test_infra/test_infra/tools/assets.py
+++ b/src/assisted_test_infra/test_infra/tools/assets.py
@@ -2,6 +2,7 @@ import json
 import os
 from typing import Any, Dict, List, Optional, Union
 
+import libvirt
 import netifaces
 from munch import Munch
 from netaddr import IPAddress, IPNetwork, IPRange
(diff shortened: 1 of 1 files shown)


bkopilov:
@eliorerz  , @lalon4

bkopilov:
@adriengentil

bkopilov:
The problem is that :
```
def DHCPLeases(self, mac=None, flags=0):
        """Returns a list of dhcp leases for interfaces connected to the given virtual network """
        ret = libvirtmod.virNetworkGetDHCPLeases(self._o, mac, flags)
        if ret is None:
            raise libvirtError('virNetworkGetDHCPLeases() failed')
        return ret
```

bkopilov:
@eliorerz , @lalon4  ,

bkopilov:
@osherdp

adriengentil:
/test e2e-metal-assisted-day2-arm-workers
this job does some tricks to get libvirt network configuration

adriengentil (approved):

PR #2233, opened 2023-07-25
MGMT-15312: Create  policy for OCI CCM in terraform

adriengentil:
- Tag cluster nodes with a specific tag
- Create a dynamic group for the master nodes based on tag
- Create a policy based on the dynamic group in order to allow the masters/CCM to be authorized to manage instances/networks/load-balancers in OCI

PR #2237, opened 2023-07-28
MGMT-15035: retrieve sosreport from oci nodes

adriengentil:
Add the controller class of OCI in order to retrieve the node list and
get the sosreports from them.

Specify cluster name to make sure `make download_cluster_logs` will retrieve the sos reports from it instead of creating a new cluster.


adriengentil:
will rebase against master once https://github.com/openshift/assisted-test-infra/pull/2233 is merged

src/assisted_test_infra/download_logs/download_logs.py
```diff
@@ -416,16 +418,16 @@ def gather_sosreport_data(output_dir: str):
 
     nodes = []
     # Find matching controller by listing nodes
-    for controller_class in [LibvirtController, VSphereController, NutanixController]:
-        log.debug(f"Looking up nodes using controller {controller_class.__name__}")
+    for controller_config_classes in [(LibvirtController, TerraformConfig), (VSphereController, VSphereConfig), (NutanixController, NutanixConfig), (OciController, OciConfig)]:
```
osherdp, line 421: looks a bit weird to see it in this scope. maybe we can have a dictionary in ``src/assisted_test_infra/test_infra/controllers/__init__.py`` for those mappings?

adriengentil: I don't know, I think we should find another way to get the nodes because it's not reliable (CLUSTER NAME is required to be set for nutanix/vsphere/oci, for libvirt it will create a new cluster in AI and get all the ips of the the VMs).

I was thinking about adding a terraform output with the ip of the nodes provisioned in libvirt/oci/nutanix/vsphere/oci, then we just have to look at build/terraform/* to get the IPs/

osherdp: > maybe the naming can be improved... I'm open to suggestions
maybe ``create_minimal`` can merge into the regular constructor? or we can have a factory class method ``create`` where you can either provide a config or have the empty one as the default value. something like:
```python
    @classmethod
    def create(cls, config):
        return cls(TerraformConfig(), config if config is not None else ClusterConfig())
```
(or just have ``ClusterConfig()`` as the default value of ``config`` if this object is immutable)

adriengentil: I would like to avoid to make `__init__` less restrictive as it is used in the dependency injection, I know that investigating an issue related to that can be hell.

I will update with the `create` class method instead, thanks.

PR #2241, opened 2023-08-07
[WIP] MGMT-15510: Initialize terraform controller after cluster registration

eliorerz:
Currently the controllers are initialized before the cluster creation causing the terraform resources names to be different than the given ones (only in case where the cluster has a different name than the generated one - e.g create cluster via the UI and pass CLUSTER_ID)

Currently, the node controllers are initialized before the cluster creation, causing the Terraform resources names to be different from the ones provided (but only in cases where the cluster has a different name than the generated one). For example, if you create a cluster using the UI and pass CLUSTER_ID, this issue occurs.

/cc @talhil-rh


adriengentil:
replaced by https://github.com/openshift/assisted-test-infra/pull/2353

PR #2243, opened 2023-08-14
[WIP] MGMT-14942: Remove user_managed_networking attribute

eliorerz:
Remove user-managed-networking from assisted service as a preparation to the removal on the service.

/cc @osherdp @adriengentil


eliorerz:
FYI @lalon4

osherdp:
> Remove user-managed-networking from assisted service as a preparation to the removal on the service.

*from assisted-test-infra

src/triggers/default_triggers.py
```diff
@@ -11,30 +11,14 @@
...
         ),
-        "none_platform": Trigger(
-            conditions=[lambda config: config.platform == consts.Platforms.NONE],
-            user_managed_networking=True,
-            tf_platform=consts.Platforms.NONE,
-        ),
-        "vsphere_platform": Trigger(
-            conditions=[lambda config: config.platform == consts.Platforms.VSPHERE],
-            user_managed_networking=False,
-            tf_platform=consts.Platforms.VSPHERE,
```
osherdp, line 22: I guess somewhere else we're matching the terraform platform with the ocp platform?
Or we still need it just like the nutanix trigger below?

src/service_client/assisted_service_api.py
```diff
@@ -470,11 +482,12 @@ def get_hosts_by_role(self, cluster_id: str, role, hosts=None):
         return nodes_by_role
 
     def get_api_vip(self, cluster_info: dict, cluster_id: str = None):
-        cluster = cluster_info or self.cluster_get(cluster_id)
+        cluster = cluster_info or self.cluster_get(cluster_id).to_dict()
         api_vip = cluster.get("api_vip")
-        user_managed_networking = cluster.get("user_managed_networking")
+        platform_type = cluster.get("platform", {}).get("type")
+        is_platform_network_managed = platform_type == consts.Platforms.OCI or platform_type == consts.Platforms.NONE
```
osherdp, line 488:
```suggestion
        is_platform_network_managed = platform_type in (consts.Platforms.OCI, consts.Platforms.NONE)
```

src/service_client/assisted_service_api.py
```diff
@@ -106,8 +106,13 @@ def wait_for_api_readiness(self, timeout: int) -> None:
     def create_cluster(
         self, name: str, ssh_public_key: Optional[str] = None, **cluster_params
     ) -> models.cluster.Cluster:
-        cluster = models.ClusterCreateParams(name=name, ssh_public_key=ssh_public_key, **cluster_params)
-        log.info("Creating cluster with params %s", cluster.__dict__)
+        cluster = models.ClusterCreateParams(name=name, ssh_public_key=ssh_public_key, **cluster_params).to_dict()
+
+        # TODO remove after completely removing `user_managed_networking` from cluster object
```
osherdp, line 111: how will the service act after the change? will it ignore this parameter or it won't even access this unrecognized parameter?
the first option might be a slightly better option because we at least allow a gradual transition iiuc (clients are transitioning from this parameter to just setting the platform)

eliorerz: We actually don't need this parameter at all. 
The service currently set it automatically by a given platform if provided, and if not baremetal platform is the default one.
The first step (after removing the dependencies - e.g. test-infra, UI) is to remove the logic but to keep the parameter passing UMN to any API call won't fail but will do nothing

src/triggers/default_triggers.py
eliorerz: There is an issue here that I'm trying to solve, the case where we pass `PLATFORM=...` as an env var is already solved on the `env_variables_defaults.py` file. But we are not handling the case where we pass platform as a parametrize parameter.
The problem there is that we triggers are called after the node controller configuration is created but the configuration is created by a given `tf_platform`.
So for conclusion we don't need the triggers set for `tf_platform` and I'm working on fixing the other issue (I missed nutanix for some reason and I'll remove it on the next push)

src/service_client/assisted_service_api.py
osherdp: So what's the value in removing it here? iiuc it doesn't do a whole lot rather than removing this parameter under the hood (and the only place that really matters is the outer scope of this function, e.g. QE tests, test-infra tests, etc.)

src/assisted_test_infra/test_infra/controllers/node_controllers/node_controller.py
```diff
@@ -206,7 +206,7 @@ def notify_iso_ready(self) -> None:
     def set_dns(self, api_ip: str, ingress_ip: str) -> None:
         pass
 
-    def set_dns_for_user_managed_network(self) -> None:
+    def set_dns_for_none_platform(self) -> None:
```
adriengentil, line 209:
```suggestion
    def set_dns_for_none_platform(self) -> None:
```
Even though we will remove the user managed networking option from service's interface, I think it still makes sense to talk about "user managed networking" in the implementation, that means we require the controller to setup the network before cluster creation (required for none + oci platforms today).

src/service_client/assisted_service_api.py
```diff
@@ -470,11 +482,12 @@ def get_hosts_by_role(self, cluster_id: str, role, hosts=None):
...
     def get_api_vip(self, cluster_info: dict, cluster_id: str = None):
-        cluster = cluster_info or self.cluster_get(cluster_id)
+        cluster = cluster_info or self.cluster_get(cluster_id).to_dict()
         api_vip = cluster.get("api_vip")
-        user_managed_networking = cluster.get("user_managed_networking")
+        platform_type = cluster.get("platform", {}).get("type")
+        is_platform_network_managed = platform_type == consts.Platforms.OCI or platform_type == consts.Platforms.NONE
 
-        if not api_vip and user_managed_networking:
+        if not api_vip and is_platform_network_managed:
```
adriengentil, line 490: In my opinion, `is_platform_network_managed` is misleading. It like it means the platform manages the network which is the opposite, we expect the network to be setup upfront.

eliorerz: @adriengentil any suggestions? We also have this in assisted-service feature-support API

adriengentil: `externaly_managed_networking` but I'm biased :sweat_smile: 
`infrastructure_managed_networking`

PR #2248, opened 2023-08-22
NO-ISSUE: Use a regex to select OS image in OCI

adriengentil:
OS images are updated regulary and old ones expire. In order to avoid
spending time to update OS images when job start to fail, use a regex to
select the right OS image.

I took the opportunity of this PR to update OCI provider version.


osherdp (approved):

adriengentil:
/unhold
conformance tests are looking flaky, but clusters were provisioned as expected

PR #2249, opened 2023-08-22
Day2 assets_file differ from day1 asset_file

bkopilov:
Looks like assets_file used as a lock when we call to get or to release_all In parallel tests while we may have creation for day1 and day2 clusters, We may consider to use the same lock because the resources are common.

Any reason to use different lock for day2 ?

### DIFF
--- a/src/tests/base_test.py
+++ b/src/tests/base_test.py
@@ -1,4 +1,3 @@
-import hashlib
 import json
 import os
 import shutil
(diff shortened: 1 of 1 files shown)


bkopilov:
@adriengentil  , @lalon4 

Current day2 use different lock for assets, not sure why but it make sense to use a single lock.
Am i missing somthnig ?

adriengentil:
it was introduced here, because we may target another libvirt URI for day2: https://github.com/openshift/assisted-test-infra/pull/2044/files

bkopilov:
> DAY2_LIBVIRT_URI

Hi , 
Ohh i see . i missed the if statment there...  i am trying to fix parallel issues we have but i see now that my patch is not relevant to the tests becasue we dont use this param.

Thanks , lets close it . no need noise

bkopilov:
My mistake ....

PR #2256, opened 2023-09-04
Add support for cpu fallback model when set customize

bkopilov:
Due to bug in RHEL9, vm unable to boot with qemu64 cpu model. BZ-2188896[RHEL9] when libvirt cpu mode=custom hit on kernel panic when trying to load the virtual machine


bkopilov:
@lalon4 , @eliorerz 
Added this patch due to failure in test_cnv_cpu_virtualization_flag_validation

PR #2262, opened 2023-09-13
Reduce stress when running start/stop in parallel runs

bkopilov:
We have issues related to nodes installation post node reboot or timeout for operators CVO , CNV and more.

When runnin the same test without parallel we have less issues and its hard to reproduce.

Adding support for serial node actions , means all actions related to node start, node shutdown, clean_virsh_resources and assetes actions will share the same lock.

The purpose is to reduce stress from libvirt when having parallel jobs. The node start/stop will be delayed to lock, we will try to see if the CI results changed or node.

The current code missing external param to controll the nodes actions , parallel support or not ... will do it later after POC .

Current code wont impact on installation process only on node creation and shutdown triggered by controllers.

### DIFF
--- a/src/assisted_test_infra/test_infra/controllers/node_controllers/node.py
+++ b/src/assisted_test_infra/test_infra/controllers/node_controllers/node.py
@@ -5,6 +5,7 @@ from paramiko import SSHException
 from scp import SCPException
 
 import consts
+from assisted_test_infra.test_infra import utils
 from assisted_test_infra.test_infra.controllers.node_controllers import ssh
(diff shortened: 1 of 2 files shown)


bkopilov:
@eliorerz , @lalon4 , @adriengentil 
We have many timeout with CI and its very strange . i would like to try to reduce stress and parallel nodes actions as possible from libvirt and see if any impact.
What do you think ?

bkopilov:
No difference with this patch . same issue we have timeouts

PR #2265, opened 2023-09-19
MGMT-15598: Bugfix - list all cluster manifests irrespective of type

paul-maidment:
A recent change to Assisted Service changed the default behaviour of the manifest list functionality. By default, the manifest list will only return manifests that were supplied by the user and will exclude system generated manifests.

This PR works in conjunction with a PR to the service https://github.com/openshift/assisted-service/pull/5498/files  (and corresponding Python client) to supply an additional parameter
"include_system_generated" and set this to "true" so that assisted-test-infra will fetch all manifests irrespective of type.

### DIFF
--- a/src/service_client/assisted_service_api.py
+++ b/src/service_client/assisted_service_api.py
@@ -260,7 +260,9 @@ class InventoryClient(object):
 
     def download_manifests(self, cluster_id: str, dir_path: str) -> None:
         log.info(f"Downloading manifests for cluster {cluster_id} into {dir_path}")
-        response = self.manifest.v2_list_cluster_manifests(cluster_id=cluster_id, _preload_content=False)
+        response = self.manifest.v2_list_cluster_manifests(
+            cluster_id=cluster_id, include_system_generated=True, _preload_content=False
+        )
         for record in json.loads(response.data):
             response = self.manifest.v2_download_cluster_manifest(
                 cluster_id=cluster_id, file_name=record["file_name"], folder=record["folder"], _preload_content=False


src/service_client/assisted_service_api.py
```diff
@@ -260,7 +260,7 @@ def download_and_save_infra_env_file(self, infra_env_id: str, file_name: str, fi
 
     def download_manifests(self, cluster_id: str, dir_path: str) -> None:
         log.info(f"Downloading manifests for cluster {cluster_id} into {dir_path}")
-        response = self.manifest.v2_list_cluster_manifests(cluster_id=cluster_id, _preload_content=False)
+        response = self.manifest.v2_list_cluster_manifests(cluster_id=cluster_id, include_system_generated="true", _preload_content=False)
```
adriengentil, line 263:
```suggestion
        response = self.manifest.v2_list_cluster_manifests(cluster_id=cluster_id, include_system_generated=True, _preload_content=False)
```

isn't it a boolean?

paul-maidment: It's a parameter used in an API call to the service, not sure how it will serialize, wanted to guarantee that it would send a string

adriengentil: the boolean should be marshaled as a string in the request send to the service

adriengentil: saying that because it's clear the API exposes a boolean: https://github.com/openshift/assisted-service/pull/5498/files#diff-8b1949772e223a1da6a2049ada2733fa506410975b241cf86cf44c7a8665bc62R2852

paul-maidment: OK, will change this to a boolean

adriengentil:
e2e-metal-single-node-live-iso is known to be flaky
Looks good to me, can we retest and check the manifests are gathered all once your fix is merged in assisted service?

paul-maidment:
> e2e-metal-single-node-live-iso is known to be flaky Looks good to me, can we retest and check the manifests area all gathered once your fix is merged in assisted service?

Just found the download_cluster_logs makefile target and ran it (Test of infra was done by manually editing the part that sends include_system_generated=True) mirroring the PR here https://github.com/openshift/assisted-test-infra/pull/2265

Running against a custom service image with the updated python client from https://github.com/openshift/assisted-service/pull/5498/files

```
/home/test/assisted-test-infra/build/2023-09-19_14-08-26_766b6f0e-f308-4dff-9032-24a0046a75ba/cluster_files/manifests
[root@rdu-infra-edge-18 manifests]# ls -latr
total 12
-rwxr-xr-x. 1 root root 1129 Sep 19 10:19 50-masters-chrony-configuration.yaml
-rwxr-xr-x. 1 root root 1129 Sep 19 10:19 50-workers-chrony-configuration.yaml
drwxr-xr-x. 2 root root   94 Sep 19 10:19 .
drwxr-xr-x. 3 root root 4096 Sep 19 10:19 ..
[root@rdu-infra-edge-18 manifests]# 
```
Looks like the integration works just fine and I think this proves that all is well. These files would not be showing up if the integration was not working.

adriengentil:
we still need to wait for your PR to be merged before moving forward with this one, the gather step fails with:
```
      File "/root/.pyenv/versions/3.11.0/lib/python3.11/site-packages/assisted_service_client/api/manifests_api.py", line 384, in v2_list_cluster_manifests
        (data) = self.v2_list_cluster_manifests_with_http_info(cluster_id, **kwargs)  # noqa: E501
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/root/.pyenv/versions/3.11.0/lib/python3.11/site-packages/assisted_service_client/api/manifests_api.py", line 412, in v2_list_cluster_manifests_with_http_info
        raise TypeError(
    TypeError: Got an unexpected keyword argument 'include_system_generated' to method v2_list_cluster_manifests
  stdout_lines: <omitted>
```

adriengentil (approved):

PR #2268, opened 2023-09-20
Run test-infra on ARM64

bkopilov:
Current test-infra code does not support running tests from ARM , make image_build fails. This patch allows to run remotely kni tests by using remote shell. All actions from utils.download_image or utils.run_command which running locally today will be run on the remote ARM machine. we are able to do it by:
- set libvirt remote host
- wrap with ssh any "local" command and run it on remote shell (ARM machine) Evrey run_command that local runs on remote ARM machine and allow us to
- configure iptables on remote with same configuration
- access to libvirt on remote due to api calls and wrap virsh command
- virsh cleanup runs all remote ARM (it was implemented with bash and not API) The cleanup on ARM is different and requires - undefine --nvram to remove node I did not fix it yet because cleanup differ from X86.

We still have some issues with runnning proxy on ARM machine because no routing.

The code is not completed but used as POC ... need to do a lot of testing and fixes I would like to know if you agree or maybe test infra will run on ARM ?

In order to run the tests need:
export LIBVIRT_URI="qemu+tcp://nvd-srv-12.nvidia.eng.rdu2.redhat.com/system" export REMOTE_SHELL_ADDRESS="10.8.231.11"
export REMOTE_SHELL_PRIVATE_KEY="/root/.ssh/id_rsa" export CPU_ARCHITECTURE=arm64

Need to copy-ssh-id from node to arm machine to login with key and not password.

Please let me know what do you think because in case we have ARM machine we will have to run full cycle of tests.

I have managed to install basic cluster with test-infra code - as POC but it will require more tests and code tunning / properly design


bkopilov:
@eliorerz   @adriengentil

src/assisted_test_infra/test_infra/helper_classes/config/base_terraform_config.py
```diff
@@ -19,7 +17,9 @@ class BaseTerraformConfig(BaseNodesConfig, ABC):
     single_node_ip: str = None
     dns_records: Dict[str, str] = field(default_factory=dict)
 
-    libvirt_uri: str = consts.DEFAULT_LIBVIRT_URI
+    remote_shell_address: str = None
+    remote_shell_private_key: str = None
```
adriengentil, line 21: you don't need that, right? terraform will reach the remote machine through the libvirt uri

adriengentil: oh sorry, it's the config, not the parameters passed to terraform :facepalm:

src/assisted_test_infra/test_infra/utils/utils.py
```diff
@@ -64,15 +79,32 @@ def scan_for_free_port(starting_port: int, step: int = 200):
...
+    return process
+
+
+def run_command(command, shell=False, raise_errors=True, env=None, cwd=None, run_in_background=False,
+                stderr=subprocess.PIPE, stdout=subprocess.PIPE, universal_newlines=True, check=False):
+
+    if global_variables().remote_shell_address:
+        remote_shell = _initialize_remote_shell()
+        return remote_shell.run_command(command, raise_errors=raise_errors, background=run_in_background)
+
```
bkopilov, line 100: I think we should seperate the run_command to to functions:
run_command
run_command _background ....
the reason is that subprocess.popen and subprocess.run use different params and it causes mess

what do you think ? who is running today with run_in_background=True ?

AmadeusPodvratnik:
@bkopilov: I'm just writing a zVM controller which requires a gateway host hosting the required cli tools authorized to issue SMAPI commands. It might be that an separate user is needed connecting to the gateway host. I see the option of an username but not reflecting on ENV level. Make it sense to create a env var from beginning otherwise I would add a variable for the new controller.

src/assisted_test_infra/test_infra/utils/utils.py
adriengentil: I see no usages for `run_in_background=True` could be worth to get rid of it

bkopilov:
> @bkopilov: I'm just writing a zVM controller which requires a gateway host hosting the required cli tools authorized to issue SMAPI commands. It might be that an separate user is needed connecting to the gateway host. I see the option of an username but not reflecting on ENV level. Make it sense to create a env var from beginning otherwise I would add a variable for the new controller.

Yes . i dont see the whole picture for the requirements. its better to allow remote_shell with external user and also support password .. not only key

bkopilov:
> > @bkopilov: I'm just writing a zVM controller which requires a gateway host hosting the required cli tools authorized to issue SMAPI commands. It might be that an separate user is needed connecting to the gateway host. I see the option of an username but not reflecting on ENV level. Make it sense to create a env var from beginning otherwise I would add a variable for the new controller.
> 
> Yes . i dont see the whole picture for the requirements. its better to allow remote_shell with external user and also support password .. not only key

Any maybe split between test-infra logic and connections to remote which should be transparrent by ip in ip tunnels

PR #2271, opened 2023-09-21
MGMT-15707: Add s390x trigger

AmadeusPodvratnik:
To enable s390x architecture a trigger is necessary to set the required values for user_managed_networking (true) and iso_image_type (full_iso).


adriengentil (approved):

adriengentil:
It's a known issue

PR #2285, opened 2023-10-06
NO-ISSUE: Support passing registry CA to spokes in kube API disconnected mode

vrutkovs:
This updates setting additional trust bundle via infraenv (instead of manually patching ignition overrides in ClusterDeployment). Registry CA file path is being set via "REGISTRY_CA_PATH" env var.

This PR updated kube API flow to use additionalTrustBundle in the infraenv instead of generating Ignition override in ClusterDeployment. This doesn't change Cluster API tests yet (this best be done separately to avoid unexpected failures).

Tested via rehearsals in https://github.com/openshift/release/pull/43476


src/tests/test_kube_api.py
```diff
@@ -107,19 +107,19 @@ def kube_api_test(
...
-            ignition_config_override = None
-
+        ignition_config_override = None
         infra_env = InfraEnv(api_client, f"{cluster_name}-infra-env", spoke_namespace)
         infraenv = infra_env.create(
             cluster_deployment, secret, proxy, ignition_config_override, ssh_pub_key=cluster_config.ssh_public_key
         )
+        if is_disconnected and cluster_config.registry_ca_path is not None:
+            log.info("setting additional trust bundle for disconnected install")
+            registry_ca = None
```
omertuc, line 117: We lost the call to `ca_bundle = self.get_ca_bundle_from_hub(spoke_namespace)`

omertuc (changes requested):

src/tests/test_kube_api.py
vrutkovs: This is no longer needed:
* `spoke_namespace` is generated dynamically, so user can't  create CA bundle beforehand
* `ca_bundle` is being injected as Ignition override by `infra_env.patch` too

omertuc: >spoke_namespace is generated dynamically, so user can't create CA bundle beforehand

Not sure I understand that. User might be a script / automation in assisted_test_infra that knows the dynamically generated spoke_namespace. Are you saying this code never worked?

>ca_bundle is being injected as Ignition override by infra_env.patch too

`ca_bundle` doesn't exist in your PR anymore

vrutkovs: >Are you saying this code never worked?

It worked -  if there was a way to create the necessary configmap via some other method, but there wasn't any.

>ca_bundle doesn't exist in your PR anymore

Patching infra env would automatically create an Ignition override, no other action required

vrutkovs:
@adriengentil @rccrdpccl PTAL

adriengentil (approved):

PR #2287, opened 2023-10-10
MGMT-15945: Pull service and index images from ocm-2.9 tag

adriengentil:
By default in release-ocm-2.9 branch, pull ocm-2.9 tag for service and index images.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,9 +35,9 @@ CONTAINER_RUNTIME_COMMAND := $(or ${CONTAINER_COMMAND}, ${CONTAINER_RUNTIME_COMM
 LINT_CODE_STYLING_DIRS := src/tests src/triggers src/assisted_test_infra/test_infra src/assisted_test_infra/download_logs src/service_client src/consts src/virsh_cleanup src/cli
 
 # assisted-service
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -35,9 +35,9 @@ CONTAINER_RUNTIME_COMMAND := $(or ${CONTAINER_COMMAND}, ${CONTAINER_RUNTIME_COMM
 LINT_CODE_STYLING_DIRS := src/tests src/triggers src/assisted_test_infra/test_infra src/assisted_test_infra/download_logs src/service_client src/consts src/virsh_cleanup src/cli
 
 # assisted-service
-SERVICE := $(or $(SERVICE), quay.io/edge-infrastructure/assisted-service:latest)
+SERVICE := $(or $(SERVICE), quay.io/edge-infrastructure/assisted-service:ocm-2.9)
```
osherdp, line 38: This fix should only be relevant for local testing, right?
Because in prow we should substitute the image

osherdp (approved):

Makefile
adriengentil: Right, https://github.com/openshift/assisted-test-infra/pull/2138 the issue is mostly with the index

PR #2288, opened 2023-10-11
NO-ISSUE: Update to python 3.12

adriengentil:


adriengentil:
Hi @bkopilov,  any concerns about this change?

bkopilov:
> Hi @bkopilov, any concerns about this change?

Why do we need to change python version to 3.12 ?  

> Hi @bkopilov, any concerns about this change?

adriengentil:
>Why do we need to change python version to 3.12 ?
to keep things updated and avoid to much gaps between versions upgrades

eifrach (approved):

PR #2294, opened 2023-10-17
MGMT-15088: Add multi VIP support

nmagnezi:


src/assisted_test_infra/test_infra/helper_classes/cluster.py
```diff
@@ -932,7 +932,9 @@ def prepare_networking(self):
         # in our case when nodes are ready, vips will be there for sure
         if self._ha_not_none():
             vips_info = self.api_client.get_vips_from_cluster(self.id)
-            self.nodes.controller.set_dns(api_ip=vips_info["api_vip"], ingress_ip=vips_info["ingress_vip"])
+            api_ip = vips_info["api_vips"][0].ip() if len(vips_info["api_vips"]) > 0 else ""
```
rccrdpccl, line 935: would empty string as IP work, or could this be causing validation issues? Same below

nmagnezi: that should not change the flow here. if there is a VIP value it that array, it will get passed as `api_ip` and `ingress_ip`.
If there isn't, that was already the case in the existing flow. I just pass an empty string as it would have done anyhow.

src/assisted_test_infra/test_infra/controllers/node_controllers/terraform_controller.py
```diff
@@ -184,8 +184,20 @@ def _fill_tfvars(self, running=True):
...
-        tfvars["ingress_vip"] = self._config.ingress_vip or vips["ingress_vip"]
+        if self._config.api_vips:
+            tfvars["api_vips"] = [i.get("ip") for i in self._config.api_vips]
+        else:
+            tfvars["api_vips"] = [i.get("ip") for i in vips["api_vips"]]
+
+        if self._config.ingress_vips:
+            tfvars["ingress_vips"] = [i.get("ip") for i in self._config.ingress_vips]
+        else:
+            tfvars["ingress_vips"] = [i.get("ip") for i in vips["ingress_vips"]]
```
rccrdpccl, line 195: nit: should we encapsulate this logic in some helper methods/functions? i.e. `tfavars["api_vips"] = get_api_vips(vips)`

nmagnezi: Done as a single function for both

rccrdpccl:
/lgtm
/approve

@adriengentil could you please take a look even if it's merged :pray: ?

PR #2300, opened 2023-11-05
MGMT-16153: update utc call

eifrach:


danmanor:
Not sure I understand why will this solve the deprecation issue, or why is this causing the failure of `e2e-metal-assisted-day2-single-node` @eifrach ?

eifrach:
> Not sure I understand why will this solve the deprecation issue, or why is this causing the failure of `e2e-metal-assisted-day2-single-node` @eifrach ?

This won't fix the failure.
The failure's are due to timeout on the day2 node

src/tests/test_upgrade_agent.py
```diff
@@ -72,7 +72,7 @@ def _load_service_with_agent_image(
 
         # Restart the deployment:
         deployments_api = client.resources.get(api_version="apps/v1", kind="Deployment")
-        date = datetime.datetime.now(datetime.timezone.utc).isoformat()
+        date = datetime.datetime.utcnow().isoformat()
```
gamli75, line 75: @eifrach did you check if other places are using this method?

eifrach: yes. that's the only place used in the code 

```console
$ grep -R timezone.utc * || echo "Nope" 
Nope
```

adriengentil (approved):

src/tests/test_upgrade_agent.py
danmanor: @eifrach Are you sure this will solve the issue? the deprecation warning says: `datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC).`  but the date you are deleting is time-aware, after the change it doesn't (utcnow down't return time-aware object, just a datetime corresponsing to the current utc time).

eifrach: hey @danmanor,
seems that both functions are returning a string value and not a datetime obj 
the ISO format is a bit change but I don't think its an issue here 
```python
>>> import datetime
>>> date = datetime.datetime.utcnow().isoformat()
>>> print (f"{date},  {type(date)}")
2023-11-06T10:20:58.471341,  <class 'str'>

>>> date = datetime.datetime.now(datetime.timezone.utc).isoformat()
>>> print (f"{date},  {type(date)}")
2023-11-06T10:21:28.480799+00:00,  <class 'str'>
```

danmanor: Yes I understand, what I meant was `datetime.datetime.now(datetime.timezone.utc)` returns a time-aware datetime object (which has tz=utc) as requested in the warning, compared to `datetime.datetime.utcnow()` which returns a datetime object which is not time-aware, and therefore I believe this is not the origin of the warning, but I night be wrong here.

PR #2303, opened 2023-11-07
NO-ISSUE: Upgrade to terraform 1.6.3

adriengentil:
Since we have issues with vsphere infrastructure, let's make sure to use the latest version of terraform.


adriengentil:
/hold
wait for e2e-vsphere-assisted to complete

adriengentil:
vsphere failed but but not on terraform operation

PR #2306, opened 2023-11-12
Fix api_vip for  update_oc_config function

bkopilov:
Added new structure return when SNO node, get_ip_for_single node returns string. In case exception bubbled up we handle it and set "" as address

Fix config_etc_hosts call , changed param to api_vip


bkopilov:
@lalon4 , @adriengentil

lalon4:
@nmagnezi

adriengentil (approved):
/ok-to-test

adriengentil (approved):

PR #2308, opened 2023-11-14
Allow to change iptables rules without changing nodes state

bkopilov:
Current code start/stop node when asking to add iptables rule. Added support for adding rules when nodes are running. There is no real need to stop nodes when handling iptables.


bkopilov:
@lalon4 , @adriengentil

src/tests/base_test.py
```diff
@@ -826,16 +826,18 @@ def set_iptables_rules_for_nodes(
...
-            cluster.nodes.start_given(given_nodes)
+            if start_stop_nodes:
+                if cluster.enable_image_download:
+                    cluster.generate_and_download_infra_env(iso_download_path=cluster.iso_download_path)
+                cluster.nodes.start_given(given_nodes)
             for node in given_nodes:
                 given_node_ips.append(node.ips[0])
-            cluster.nodes.shutdown_given(given_nodes)
+            if start_stop_nodes:
+                cluster.nodes.shutdown_given(given_nodes)
```
eifrach, line 840: we have two identical if statements 
L833  & L839

Maybe you meant 
```suggestion
            if not start_stop_nodes:
                cluster.nodes.shutdown_given(given_nodes)
```

it will be better this way 

```suggestion
            if start_stop_nodes:
                if cluster.enable_image_download:
                    cluster.generate_and_download_infra_env(iso_download_path=cluster.iso_download_path)
                cluster.nodes.start_given(given_nodes)
            else:
               cluster.nodes.shutdown_given(given_nodes)
            for node in given_nodes:
                given_node_ips.append(node.ips[0])
                
```

src/tests/base_test.py
```diff
@@ -826,16 +826,18 @@ def set_iptables_rules_for_nodes(
             cluster: Cluster,
             iptables_rules: List[IptableRule],
             given_nodes=None,
+            start_stop_nodes=True,
```
eifrach, line 829: when is this will ever be `False` ?

bkopilov: The problem is that iptables fixture has no single responsibility and it handles shutdown nodes . basically we dont need to shutdown nodes to configure iptables .
The idea here is to apply iptables rules when node are running before installation /during installation

Trying to keep "old style" and allow  use iptables without touching the nodes and gain the cleanup method

src/tests/base_test.py
bkopilov: ok , let me change the logic a little bit .  basically shutdown should not be here ...

adriengentil (approved):

PR #2317, opened 2023-11-27
MULTIARCH-3980: Add new Controller (as stub) for zVM

AmadeusPodvratnik:
Add new zVMController (as stub) to in later PRs the necessary code be able to create a zVM Cluster.
The code is "disabled" for the time being.

### DIFF
--- a/src/tests/base_test.py
+++ b/src/tests/base_test.py
@@ -35,6 +35,7 @@ from assisted_test_infra.test_infra.controllers import (
     VSphereController,
 )
 from assisted_test_infra.test_infra.controllers.node_controllers.tf_controller import TFController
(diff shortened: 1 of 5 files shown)


src/tests/base_test.py
```diff
@@ -402,6 +403,9 @@ def get_terraform_controller(
         if platform == consts.Platforms.OCI:
             return OciController(controller_configuration, cluster_configuration)
 
+        if platform == consts.CPUArchitecture.S390X:
```
adriengentil, line 406: the idea is to set `TF_PLATFORM=s390x`, right?

adriengentil (approved):

PR #2329, opened 2023-12-11
NO-ISSUE: add the ability to update installer args

adriengentil:


src/assisted_test_infra/test_infra/helper_classes/cluster.py
```diff
@@ -498,7 +498,7 @@ def set_additional_ntp_source(self, ntp_source: List[str]):
             ntp_source_string = ntp_source
         else:
             raise TypeError(
-                f"ntp_source must be a string or a list of strings, got: {ntp_source}, type: {type(ntp_source)}"
+                f"ntp_source must be a string or a list of strings, got: {ntp_source}," f" type: {type(ntp_source)}"
```
eifrach, line 501: it may be missing a comma here? 
```suggestion
                f"ntp_source must be a string or a list of strings, got: {ntp_source},",  f" type: {type(ntp_source)}"
```

or remove the second f string

```suggestion
                f"ntp_source must be a string or a list of strings, got: {ntp_source}, type: {type(ntp_source)}"
```

adriengentil: bad merge, I updated the commit

src/assisted_test_infra/test_infra/helper_classes/infra_env.py
```diff
@@ -100,6 +100,15 @@ def update_host(
             disks_skip_formatting=disks_skip_formatting,
         )
 
+    def update_host_installer_args(self, host_id: str):
+        if not self._config.host_installer_args:
```
bkopilov, line 104: Do you think that update_host_installer_args may be called during cluster lifetime ? i mean allow to user update host_installer_args ?  the way its written no way to update on the fly

src/assisted_test_infra/test_infra/helper_classes/base_cluster.py
```diff
@@ -153,6 +153,11 @@ def set_hostnames_and_roles(self):
             role = consts.NodeRoles.MASTER if consts.NodeRoles.MASTER in name else consts.NodeRoles.WORKER
             self._infra_env.update_host(host_id=host.get_id(), host_role=role, host_name=name)
 
+    def set_installer_args(self):
+        hosts = self.to_cluster_hosts(self.api_client.get_cluster_hosts(self.id))
+        for host in hosts:
```
bkopilov, line 158: The way the code written should be :
if not self._config.host_installer_args
   ---> running the loop ...

PR #2333, opened 2023-12-14
Add function name caller to test-infra logs

bkopilov:
With this change we will print to log the function called by the line. example:
Adding rule - [path]/nat_controller.py:98)->_insert_rule (/home/benny/assisted-test-infra/src/tests/base_test.py:82)->update_parameterize


bkopilov:
@lalon4 , @adriengentil , @talhil-rh

adriengentil (approved):

PR #2335, opened 2023-12-15
MGMT-15695: Use oci external integration

adriengentil:


adriengentil:
/hold
merge once https://github.com/openshift/assisted-service/pull/5787#issuecomment-1857930580 is merged

PR #2338, opened 2023-12-19
Fix thread id in test-infra.log file

bkopilov:
Current code shows thread id in log file, the problem in parallel run that we create multiple proccesses and each one create threads. Due to that we may see same thread_id when calling threading.get_ident() under different services.

solution is to add the process id to the thread and create combination thread_id:proccess_id . i think under the same proccess we may not see duplicate thread id.

by default python store the thread_id and proccess_id unless user disable it. logging.logThreads = True
logging.logProcesses = True


bkopilov:
@lalon4 , @adriengentil

adriengentil (approved):

PR #2340, opened 2023-12-21
Add cluster as callback to fixture cluster

bkopilov:
Current cluster fixture when called create and register the cluster, We can not modify / update params dynamically when creating cluster

In some cases we may want to get cluster object and before it registerd send params/verify as openshit_version , pull keys and more. example for use case: a test try to send to "register" cluster missing params or ask for custom ocp version based on dynamic configuration.

The idea here allow to user decide on cluster fixture if its a registered cluster object or callback_cluster.
In case its a callback cluster we call callback_cluster() and the object created and registered. The callback_cluster is object that has reference to param from cluster fixtube and will allow to change configuration .

Example for testcase:
class TestProxy(BaseTest):

    callback_cluster = True

    def test_benny(self):
    	supported_versions = cluster.api_client.versions.v2_list_supported_openshift_versions()
 	cluster1 = cluster()
        cluster1.prepare_for_installation()

with this support we will be able to gain the cleanup that cluster fixture has.


bkopilov:
@lalon4 , @adriengentil , @danmanor 

This patch to address custom ocp tests ... let me know what do you think.
I know its a little bit overhead but it will allow us to extend tests

rccrdpccl:
@bkopilov is this still needed?

/cc @adriengentil @eifrach

PR #2342, opened 2023-12-24
NO-ISSUE: add ability to configure minikube

eifrach:
* Add ability to configure minikube setup
* allows to change all minikube configuration using env values
*  default setup has not change

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ Check the [Installation Guide](GUIDE.md) for installation instructions.
 | `MINIKUBE_MEMORY`| set amount of memory, default = 8G|
 | `MINIKUBE_DISK_SIZE`| set disk size, default = 50G |
 | `MINIKUBE_HOME`| set default location for minikube, default = ~/.minikube |
-| `MINIKUBE_REGISTRY`| set registry image, default = "quay.io/libpod/registry:2.8" | 
+| `MINIKUBE_REGISTRY_IMAGE`| set registry image, default = "quay.io/libpod/registry:2.8" | 
 ### Cluster configmap
 
 |     |     |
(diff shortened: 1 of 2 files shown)


gamli75:
Few remarks:
1. can you check why you have all those space/ tab changes? it's complicated to follow the real changes 
2. can you add why we need to configure minikube? what do we want to solve?

eifrach:
>  can you check why you have all those space/ tab changes? it's complicated to follow the real changes

The tab are from my IDE linter. I can fix it.

> can you add why we need to configure minikube? what do we want to solve?

so this is mostly for beaker lab where you dont have enoguh space in the default location.
since I've  add this variable, I figured best to move all the param as well so we can change them if needed.

adriengentil:
No need to specify the variables [here](https://github.com/openshift/assisted-test-infra/blob/master/skipper.env) to pass them to skipper env?

README.md
```diff
@@ -96,7 +96,15 @@ Check the [Installation Guide](GUIDE.md) for installation instructions.
...
-
+### Minikube configuration
+|   |   |
+|---|---|
+| `MINIKUBE_DRIVER`| set minikube driver, default = kvm2 | 
+| `MINIKUBE_CPUS`| set amount of cpus, default = 4|
+| `MINIKUBE_MEMORY`| set amount of memory, default = 8G|
+| `MINIKUBE_DISK_SIZE`| set disk size, default = 50G |
+| `MINIKUBE_HOME`| set default location for minikube, default = ~/.minikube |
+| `MINIKUBE_REGISTRY`| set registry image, default = "quay.io/libpod/registry:2.8" |
```
osherdp, line 107: just for some extra clarity
```suggestion
| `MINIKUBE_REGISTRY_IMAGE`| set registry image, default = "quay.io/libpod/registry:2.8" | 
```

scripts/hub-cluster/minikube.sh
```diff
@@ -29,7 +33,7 @@ function _init_minikube() {
     for i in {1..5}
     do
         minikube delete
-        minikube start --driver=kvm2 --memory="${MINIKUBE_RAM_MB}" --cpus=4 --force --wait-timeout=15m0s --disk-size="${MINIKUBE_DISK_SIZE}" || true
+        minikube start --force --wait-timeout=15m0s  || true
```
osherdp, line 36: so where do we use all the `MINIKUBE_*` values?

osherdp (changes requested):

scripts/hub-cluster/minikube.sh
osherdp: Oh, missed that
Thanks for the explanation!

osherdp (approved):

PR #2343, opened 2023-12-24
Fix verify_logs_uploaded

bkopilov:
Latest cluster logs contain manifest yaml file.
Due to naming "master" or "worker"  in yaml file name it goes to wrong flow. manifest_system-generated_openshift_50-workers-chrony-configuration.yaml

log verify code will ignore yaml file because it expect for tar files on verify


bkopilov:
@lalon4 , @adriengentil

gamli75 (approved):

PR #2344, opened 2023-12-24
MGMT-16332: add multi-disk for none platform deployment

eifrach:


gamli75:
You need this change for testing LVM on multi-node cluster, right? if the answer is yes - please run LVM test to verify that it's still working with this change

eifrach:
I did test the first part, up until the hosts are starting the installation
but let run the full tests
/test e2e-metal-assisted-lvm

gamli75 (approved):

PR #2349, opened 2024-01-02
MGMT-15772: fix metro as srting

eifrach:


eifrach:
bug with terraform model 

/close

PR #2350, opened 2024-01-02
Add new log record for testcase id

bkopilov:
Current code in parallel may have duplicate threads id , when run in parallel multiple proccesses created and threads per proccess with duplicate id.

Current patch adding new record to logger names test_id set to "" In case we want to update the param we can set it in the begining of test.

When request fixture exists per cluster creation we take the request.node.name (testcase name) as test_id.

I tested it in parallel run and it works but i think we need more tests.

### DIFF
--- a/src/service_client/__init__.py
+++ b/src/service_client/__init__.py
@@ -1,5 +1,5 @@
 from .assisted_service_api import InventoryClient
 from .client_factory import ClientFactory
-from .logger import SuppressAndLog, log
+from .logger import SuppressAndLog, add_log_record, log
 
-__all__ = ["InventoryClient", "ClientFactory", "log", "SuppressAndLog"]
+__all__ = ["InventoryClient", "ClientFactory", "log", "add_log_record", "SuppressAndLog"]
--- a/src/service_client/logger.py
+++ b/src/service_client/logger.py
@@ -73,6 +73,22 @@ class ColorizingStreamHandler(logging.StreamHandler):
             self.handleError(record)
 
 
+def add_log_record(test_id):
+    # Adding log record for testcase id
+    _former_log_record_factory = logging.getLogRecordFactory()
+
+    def log_record_uuid_injector(*args, **kwargs):
+        record = _former_log_record_factory(*args, **kwargs)
+        record.test_id = test_id
+        return record
+
+    logging.setLogRecordFactory(log_record_uuid_injector)
+
+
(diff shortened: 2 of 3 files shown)


bkopilov:
I run kni CI test and looks ok .
Base on what i saw in logs , for each pytest thread-proccess created a new logger object with different id.
I run pytest parallel run and track when we enter to service_client/logger.py .
On each new thread we actually getting new log object id so i am not expecting for collision or override

```
(The right side is the log object id            [Thread id][proccess_id]  -  [id(log)]   
2024-01-03 08:29:47,716 - root - INFO - :140544961398592:63 - 140544942781168 
2024-01-03 08:29:51,239 - root - INFO - :140314124851008:65 - 140314029131968 
2024-01-03 08:29:51,284 - root - INFO - :140202115204928:68 - 140202019554992 
2024-01-03 08:29:51,316 - root - INFO - :139783385294656:71 - 139783356843504 

```

I run make destory_node muliple times and see different log id . means  "last" test_id will not move between threads or requests.

danmanor:
@adriengentil Do you see these logs in the tests that just passed ? I don't see them

bkopilov:
More info for tracking / changes

in my changes i updated the logRecord object with new param test_id and set it to ""
The only place you will see the change is in fileHandler format  and only when you override the function.
log = logging.getLogger(logger_name)

Also note that i modifed only the RootLogger (when getLogger(... ) empy logger_name,  the returned logger is RootLogger and only there you should see the difference.

In case we run logger_name with a name inside the get logger returns Logger class  and not RootLogger.
The place i changed it is related to rootLogger.

The  _logRecordFactory (LogRecord class)  is a global and rootLogger impacted by this one 
When  calling to make_record from logger object , it refers to 
 rv = _logRecordFactory(name, level, fn, lno, msg, args, exc_info, func, sinfo)
from global ... so adding this param test_id insert it to the log record and it passed to the hanler.

There is another option which i did not get to it by setting log hander per logger object ... but it a pain IMO.
something like  object_loggger.manager.logRecordFactory
And probably we can set per logger the format but it looks to me complicated to manager

```
>>> logging.getLogger()
<RootLogger root (WARNING)>
>>> logging._logRecordFactory
<class 'logging.LogRecord'>
>>> 
```
>>> logging.getLogger("fdsfsffd")
<Logger fdsfsffd (WARNING)>

PR #2352, opened 2024-01-03
MGMT-16492: Use combined OCI manifests

adriengentil:
Use the manifests published by Oracle in
https://github.com/oracle-quickstart/oci-openshift to install CCM and
CSI driver.


adriengentil:
/hold 
e2e-oci-assisted must pass

PR #2354, opened 2024-01-04
Add colorized file handler same as stream handler

bkopilov:
Moved colored to enum object
Added colorized support for file handler to distinguish between different logging levels.


bkopilov:
@adriengentil  , @danmanor , @lalon4

PR #2355, opened 2024-01-07
MGMT-16499: Allow using custom OCP release when we create cluster

danmanor:
Allow using custom `OCP` release when we create cluster - Within the CI environment, our aim is to load the fewest possible number of ISO images because the image-service takes a significant amount of time to load each one. Conversely, for scenarios outside of CI, we aim to facilitate the installation of each `OCP` release using test-infra, necessitating the loading of all ISO images.

### DIFF
--- a/scripts/deploy_assisted_service.sh
+++ b/scripts/deploy_assisted_service.sh
@@ -41,6 +41,7 @@ export REGISTRY_SERVICE_NAMESPACE=kube-system
 export REGISTRY_SERVICE_PORT=80
 export REGISTRY_SERVICE_HOST_PORT=5000
 export ENABLE_HOST_RECLAIM=${RECLAIM_HOSTS:-false}
+export OPENSHIFT_CI=${OPENSHIFT_CI:-false}
 
 
 if [[ "${ENABLE_KUBE_API}" == "true" || "${DEPLOY_TARGET}" == "operator" ]]; then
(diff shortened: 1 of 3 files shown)


scripts/deploy_assisted_service.sh
```diff
@@ -67,8 +68,10 @@ if [ "${OPENSHIFT_INSTALL_RELEASE_IMAGE}" != "" ]; then
     fi
 fi
 
-OS_IMAGES=$(skipper run ./scripts/override_os_images.py --src ./assisted-service/data/default_os_images.json)
-export OS_IMAGES
+if [ "${OPENSHIFT_CI}" == "true" ]; then
```
eifrach, line 71: This line compares variable is  string `true`  which fail if we set it to `True`

it's better this way this wil ensure that `true` or `True` even worse `TRUE`  will be correct 
```suggestion
if [ [ "${OPENSHIFT_CI}" ] ]; then
```

danmanor: This variable is set by the CI, so I guess there is no risk here. Also https://github.com/openshift/assisted-test-infra/blob/1209bd8ce092e3043e3b89fe06e1e58c0aa942ac/scripts/bring_assisted_service.sh#L42

scripts/deploy_assisted_service.sh
```diff
@@ -67,8 +68,10 @@ if [ "${OPENSHIFT_INSTALL_RELEASE_IMAGE}" != "" ]; then
...
+    OS_IMAGES=$(skipper run ./scripts/override_os_images.py --src ./assisted-service/data/default_os_images.json)
```
eifrach, line 72: just nit may this should be in a variable for changes or overwrites
```suggestion
    IMAGES_JSON=' ./assisted-service/data/default_os_images.json'
    OS_IMAGES=$(skipper run ./scripts/override_os_images.py --src ${IMAGES_JSON})
```

also can you add in `override_os_images.py` validattion check and error output before the file read
```python
"""  here  check if file readable """
   with open(args.src, "r") as f:
        os_images: list = json.load(f)

```

danmanor: Also I don't think this is necessary (at least to run again all of the tests for it), because default_os_images.json always exists

scripts/deploy_assisted_service.sh
eifrach: yes I dont like it either -- so for the CI is none issue -- but it still best practice 
again becuse we have it all over the place it OK. just saying it better to use it as `bool` check and not `str`

PR #2357, opened 2024-01-08
Add informative info from assisted-installer-controller log when time…

bkopilov:
Add controller later error when timeout during installation

During installation when waiter hit on timeout , raise timeout exception without information about the root cause and the installer timers wait longer time till it change state to fail / success.

Added support for last error in assisted-installer-controller logs when timeout exception raised by the waiter.
Added the info from controller logs to raises exception.

Created decorator for waiter functions, in case waiter func wrapped:
- No exception from waiter return func to caller as is
- Waiter bubble-up TimeoutExpired
	- veify the waiter called from cluster waiter
	- Try to download kubeconfig , success return true
	- Try to get assisted-installer-controller logs filtered by level=error
	- Add the last error to the exception , if no error nothing to append. Sometime logs call may return empty or fail due to vip network so retry
	- raise updated new TimeoutExpired exception to the caller

### DIFF
--- a/src/assisted_test_infra/test_infra/utils/utils.py
+++ b/src/assisted_test_infra/test_infra/utils/utils.py
@@ -24,6 +24,7 @@ from typing import List, Tuple, Union
 import filelock
 import requests
 import semver
+import waiting
 from requests import Session
 from requests.adapters import HTTPAdapter, Retry
 from requests.exceptions import RequestException
@@ -560,6 +561,37 @@ def get_assisted_controller_status(kubeconfig):
     return response.stdout
 
 
+def retry_success(repeat, sleep_between=30):
+    def retry_func(func):
+        def inner(*args, **kwargs):
+            start = 1
+            while start <= repeat:
+                log.info(f"retry_success decorator call func {func.__name__} {start} time")
+                res = func(*args, **kwargs)
+                if res:
+                    return res
+                else:
+                    start += 1
+                    time.sleep(sleep_between)
+            return ""
+
+        return inner
+
+    return retry_func
+
+
+@retry_success(5)
+def get_assisted_controller_log(kubeconfig: str, log_resource: str) -> str:
+    log.info(f"Getting controller logs {kubeconfig} from {log_resource}")
+    command = f"oc logs --kubeconfig={kubeconfig} --insecure-skip-tls-verify {log_resource} -n assisted-installer "
+    out, err, return_code = run_command(command, shell=True, raise_errors=False)
+    if return_code != 0 or (not out):
+        log.error(f"failed to get controller status: {err}")
+        return ""
+
+    return str(out)
+
+
 def fetch_url(url, timeout=60, max_retries=5):
     """
     Returns the response content for the specified URL.
@@ -617,3 +649,66 @@ def get_iso_download_path(entity_name: str):
 def get_major_minor_version(openshift_full_version: str) -> str:
     semantic_version = semver.VersionInfo.parse(openshift_full_version, optional_minor_and_patch=True)
     return f"{semantic_version.major}.{semantic_version.minor}"
+
+
+def waiter_decorator(func):
+    """decorate cluster waiters
+    When waiter called from cluster timeout exception may raise without information
+    Adding information about the timeout from job.batch/assisted-installer-controller,
+    Taking the last error log levels and append to the timeout message object.
+
+    We will be able to download kubeconfig in the next states:
+    installing, finalizing, installed, error, adding-hosts, cancelled, installing-pending-user-action
+
+    Sometimes logs from job.batch/assisted-installer-controller may return empty or failed when:
+    - In case pod/assisted-installer-controller-zzz in pending status (~10 minutes after installing phase)
+    - In case api vip address not accessible.
+
+    Current code will retry_success 3 times log assisted-installer-controller.
+    The code will ignore failures when trying to get assisted-installer-controller logs  and return
+    the timeout exception with pre-configured message
+    """
+
+    def update_kube_config(cluster_object):
+        # download kubeconfig and update /etc/hosts with api-vip host ip
+        try:
+            cluster_object.download_kubeconfig()
+            config_etc_hosts(
+                cluster_name=cluster_object.name,
+                base_dns_domain=cluster_object.get_details().base_dns_domain,
+                api_vip=cluster_object.get_details().api_vips[0].ip,
+            )
+            return True
+        except Exception:
+            log.info(f"Unable to download kubeconfig from cluster {cluster_object.name} - ignored")
+            return False
+
+    def filter_log_by(output, filter_by, repeat_count=2):
+        if output:
+            lines = output.splitlines()
+            reverse_list = [line + "\n" for line in lines[::-1] if filter_by in line]
+            str_represent = map(str, reverse_list[:repeat_count])
+            str_represent = "\n".join(str_represent)
+            log.info(f"assister-installer-controller filter: {str_represent}")
+            return str_represent
+        log.info("assister-installer-controller filter returned empty")
+        return ""
+
+    def inner(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except waiting.exceptions.TimeoutExpired as time_expired:
+            # Check if self from function is Cluster object (isinstanceof is better)
+            # to prevent cycle loops
+            if args and args[0].__class__.__name__ in ["Cluster", "Day2Cluster"]:
+                # if self from func method is type cluster( waiters from cluster class)
+                cluster_object = args[0]
+                if update_kube_config(cluster_object):
+                    res = get_assisted_controller_log(
+                        cluster_object.kubeconfig_path, "job.batch/assisted-installer-controller"
+                    )
+                    time_expired._what += "\n" + filter_log_by(res, filter_by="error=")
+
+            raise
+
+    return inner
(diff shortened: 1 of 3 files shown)


bkopilov:
@lalon4

bkopilov:
@eliajahshan , @talhil-rh

bkopilov:
I run CI regression3 with this fix and it worked , on timeout we got more info .

https://ci-jenkins-csb-kniqe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-assisted-installer-virt-tf/12560/testReport/

We see that on installation timeout we get more info:

```
waiting.exceptions.TimeoutExpired: Timeout of 3600 seconds expired waiting for Nodes to be in of the statuses ['installed']
time="2024-01-09T13:00:35Z" level=error msg="Failed to get list of nodes from k8s client" func="github.com/openshift/assisted-installer/src/assisted_installer_controller.(*controller).waitAndUpdateNodesStatus" file="/remote-source/app/src/assisted_installer_controller/assisted_installer_controller.go:248" error="Get \"https://10.128.0.1:443/api/v1/nodes\": dial tcp 10.128.0.1:443: connect: connection refused" request_id=73103b20-8a7a-4693-8b08-806a3b88c7dd

time="2024-01-09T13:00:35Z" level=error msg="Failed to get list of nodes from k8s client" func="github.com/openshift/assisted-installer/src/assisted_installer_controller.(*controller).waitAndUpdateNodesStatus" file="/remote-source/app/src/assisted_installer_controller/assisted_installer_controller.go:248" error="Get \"https://10.128.0.1:443/api/v1/nodes\": dial tcp 10.128.0.1:443: connect: connection refused" request_id=17f6a60a-7096-43aa-b5db-5b5fa5396b61
```



Another example:
```
waiting.exceptions.TimeoutExpired: Timeout of 3600 seconds expired waiting for Monitored ['builtin'] operators to be in of the statuses ['available']
time="2024-01-09T14:17:22Z" level=error msg="Failed to check if console is enabled" func=github.com/openshift/assisted-installer/src/assisted_installer_controller.controller.waitingForClusterOperators.func1 file="/remote-source/app/src/assisted_installer_controller/assisted_installer_controller.go:1003" error="Get \"https://localhost:6443/apis/config.openshift.io/v1/clusterversions/version\": dial tcp [::1]:6443: connect: connection refused"

time="2024-01-09T14:16:56Z" level=error msg="Failed to check if console is enabled" func=github.com/openshift/assisted-installer/src/assisted_installer_controller.controller.waitingForClusterOperators.func1 file="/remote-source/app/src/assisted_installer_controller/assisted_installer_controller.go:1003" error="Get \"https://localhost:6443/apis/config.openshift.io/v1/clusterversions/version\": net/http: TLS handshake timeout - error from a previous attempt: unexpected EOF"

```
Failed to deploy the following operators ['console']

bkopilov:
Updated the log filter to "error="  , i see that some errors in the log file exposed with level=info ...

src/assisted_test_infra/test_infra/utils/utils.py
```diff
@@ -560,6 +561,37 @@ def get_assisted_controller_status(kubeconfig):
...
+                    start += 1
+                    time.sleep(sleep_between)
+            return ""
+
+        return inner
+
+    return retry_func
+
+
+@retry_success(5)
```
adriengentil, line 583: don't we already use a library to perform the retry?

src/assisted_test_infra/test_infra/utils/utils.py
```diff
@@ -560,6 +561,37 @@ def get_assisted_controller_status(kubeconfig):
...
+def get_assisted_controller_log(kubeconfig, log_resource):
```
adriengentil, line 584:
```suggestion
def get_assisted_controller_log(kubeconfig: str, log_resource: str) -> str:
```

src/assisted_test_infra/test_infra/utils/utils.py
```diff
@@ -560,6 +561,37 @@ def get_assisted_controller_status(kubeconfig):
...
+    log.info("Getting controller logs")
```
adriengentil, line 585: maybe add more context to the log line with the `kubeconfig` and `log_resources` ?

src/assisted_test_infra/test_infra/utils/utils.py
```diff
@@ -560,6 +561,37 @@ def get_assisted_controller_status(kubeconfig):
...
+    command = f"oc logs --kubeconfig={kubeconfig} --insecure-skip-tls-verify {log_resource} -n assisted-installer "
+    out, err, return_code = run_command(command, shell=True, raise_errors=False)
+    if return_code != 0 or (not out):
+        log.error(f"failed to get controller status: {err}")
+        return str("")
```
adriengentil, line 590:
```suggestion
        return ""
```

src/assisted_test_infra/test_infra/utils/utils.py
```diff
@@ -617,3 +649,67 @@ def get_iso_download_path(entity_name: str):
...
+            lines = output.splitlines()
+            reverse_list = [line + "\n" for line in lines[::-1] if filter_by in line]
+            str_represent = map(str, reverse_list[:repeat_count])
+            str_represent = "\n".join(str_represent)
+            log.info(f"assister-installer-controller filter: {str_represent}")
+            return str_represent
+        log.info("assister-installer-controller filter returned empty")
+        return ""
+
+    @wraps(func)
```
adriengentil, line 697: do we need this decorator? From what I understand we don't manipulate the `func` directly, so their should be not problems

src/assisted_test_infra/test_infra/utils/utils.py
```diff
@@ -617,3 +649,67 @@ def get_iso_download_path(entity_name: str):
...
+    def inner(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except waiting.exceptions.TimeoutExpired as time_expired:
+            # Check if self from function is Cluster object (isinstanceof is better)
```
adriengentil, line 702: why not use `isintance` then ? :)

src/assisted_test_infra/test_infra/utils/utils.py
```diff
@@ -617,3 +649,67 @@ def get_iso_download_path(entity_name: str):
...
+    def filter_log_by(output, filter_by, repeat_count=2):
+        if output:
+            lines = output.splitlines()
+            reverse_list = [line + "\n" for line in lines[::-1] if filter_by in line]
+            str_represent = map(str, reverse_list[:repeat_count])
+            str_represent = "\n".join(str_represent)
+            log.info(f"assister-installer-controller filter: {str_represent}")
+            return str_represent
+        log.info("assister-installer-controller filter returned empty")
+        return ""
```
adriengentil, line 695: can you explain what this function does with an example?

src/assisted_test_infra/test_infra/utils/utils.py
bkopilov: The retry decorator we use in existing calls expect to get exception from caller func , the success here is a string result. we decorate function that returns a value and not exception. 
I did not find retry based on string result ... 
As long we get "" as result we repeat x times.... (the exception not raised here)

src/assisted_test_infra/test_infra/utils/utils.py
bkopilov: Once we get timeout exception from waiter, we pass output to  filter_log_by func.
 the output from the command (the logs) and extract filter_by string we are looking for . the result appended to the timeout exceptions.

Example:
In Ci runs we got back on timeout installation :
Error Message
`waiting.exceptions.TimeoutExpired: Timeout of 3600 seconds expired waiting for Nodes to be in of the statuses ['installed']`

This message has nothing about the issue , it might be operators timeout or any other . the idea is to append to this message the "error" from job.batch/assisted-installer-controller 

So the filtered called , 
getting log as a list , reverse it because we want the 2 last filter_by we found.  in case filter is "error=" 
split the lines by /n , 
get all lines with "error="  , reverse the list because we need only the last messages.
convert the list to str and return .

Example:
the output is :

```
> ............
> time="2024-01-14T03:28:31Z" level=error msg="Failed to get list of CSRs." func=github.com/openshift/assisted-installer/src/k8s_client.k8sClient.ListCsrs file="/remote-source/app/src/k8s_client/k8s_client.go:367" error="the server was unable to return a response in the time allotted, but may still be processing the request (get certificatesigningrequests.certificates.k8s.io)"
> time="2024-01-14T03:28:31Z" level=error msg="Failed to get list of nodes from k8s client" func="github.com/openshift/assisted-installer/src/assisted_installer_controller.(*controller).waitAndUpdateNodesStatus" file="/remote-source/app/src/assisted_installer_controller/assisted_installer_controller.go:248" error="the server was unable to return a response in the time allotted, but may still be processing the request (get nodes)" request_id=191ba4cd-051f-41c5-9969-d5f90367b11e
> ...................
```

All sent to filter_by as outpu , looking for  "error="  lines and stored the last 2  (reverst listed)
convert the list element to str.
append the messages to the waiter traceback .
 time_expired._what += "\n" + filter_log_by(res, filter_by="error=")
 
 In case not found , nothing added and raised exception raises as is.

src/assisted_test_infra/test_infra/utils/utils.py
bkopilov: Not trivial ... getting "cannot import name 'BaseClusterConfig' from partially initialized module 'assisted_test_infra.test_infra' (most likely due to a circular import)" 

Honestly , its a pain to resolve circular import... so same gained by __name__

PR #2358, opened 2024-01-09
MGMT-16330: disallow SDN network type testing for 4.15 and above

rccrdpccl:
We have removed support for SDN from OCP 4.15 onward. This will throw explicit error early in case of misusage of assisted-test-infra
/cc @eifrach @adriengentil


rccrdpccl:
waiting for https://github.com/openshift/release/pull/47431 to be merged

PR #2359, opened 2024-01-10
Fix manifest api - add folder param.

bkopilov:
Manifest support two folders :
param str file_name: The manifest file name to delete from the cluster. (required) param str folder: The folder that contains the files. Manifests can be placed in 'manifests' or 'openshift' directories.


bkopilov:
@adriengentil , @lalon4

bkopilov:
Lets see the result , maybe i will have to set folder default value instead None....  (lets see the results)

adriengentil:
> Lets see the result , maybe i will have to set folder default value instead None.... (lets see the results)

yes, would be best to set folder only if it's set, and let the API use the default value.

bkopilov:
> > Lets see the result , maybe i will have to set folder default value instead None.... (lets see the results)
> 
> yes, would be best to set folder only if it's set, and let the API use the default value.

ack done

src/service_client/assisted_service_api.py
```diff
@@ -575,5 +575,9 @@ def create_custom_manifest(self, cluster_id: str, folder: str, file_name: str, b
...
         return self.manifest.v2_list_cluster_manifests(cluster_id=cluster_id)
 
-    def delete_custom_manifest(self, cluster_id: str, file_name: str) -> None:
-        self.manifest.v2_delete_cluster_manifest(cluster_id=cluster_id, file_name=file_name)
+    def delete_custom_manifest(self, cluster_id: str, file_name: str, folder: str) -> None:
+        if not folder:
+            # if folder is none , use default folder
+            self.manifest.v2_delete_cluster_manifest(cluster_id=cluster_id, file_name=file_name)
+        else:
+            self.manifest.v2_delete_cluster_manifest(cluster_id=cluster_id, file_name=file_name, folder=folder)
```
adriengentil, line 583:
```suggestion
        if folder:
            self.manifest.v2_delete_cluster_manifest(cluster_id=cluster_id, file_name=file_name, folder=folder)
        else:
             # if folder is none , use default folder
             self.manifest.v2_delete_cluster_manifest(cluster_id=cluster_id, file_name=file_name)
```
let's make the condition straightforward

adriengentil (approved):

PR #2365, opened 2024-01-16
Add commands  for system load during cluster installation.

bkopilov:
This addon simulate load on nodes before and during installation. The command load allows to:
- Load / increase CPU usage for nodes
- Load / increase RAM usage for nodes
- Load / increase Disk IO for node
- Load / limit network bandwidth for nodes

Each command get node invoker to run the command and support two actions:
- execute
- undo

The load simulator will allow us to install cluster similar to customers env for example ,we can limit the traffic rate to 50kbps and the installation will be slow and we will hit on timeout while downloading / configuring / finalzing. In addition we will be able to simulate overload ram memory or busy disk with IO.

Example for options in a test:
	masters = cluster.nodes.get_masters()
        load_cpu = load_system.LoadCPU(masters, 6)
        load_ram = load_system.LoadRam(masters,2)
        load_disk = load_system.LoadDisk(masters, 2)
        load_network = load_system.LoadNetwork(masters, "50kbit", "ens3")
        load_cpu.execute()
        load_ram.execute()
        load_disk.execute()
        load_network.execute()
        load_cpu.undu()
        load_ram.undu()
        load_disk.undu()
        load_network.undu()

### DIFF
--- a/src/assisted_test_infra/test_infra/tools/load_system.py
+++ b/src/assisted_test_infra/test_infra/tools/load_system.py
@@ -1,0 +1,163 @@
+from abc import ABC, abstractmethod
+from string import Template
+
+from service_client import log
+
+
+class Command(ABC):
+    base_command = None
+    redirect = " > /dev/null &"
+
+    def __init__(self, nodes: list):
+        """Initialize node_undo dict, keys are node object
+        :param nodes: node is the command receiver  run_command
+        """
+        self.node_undo_commands = dict()
+        self._nodes = nodes
+
+    @property
+    def receivers(self):
+        # check if nodes exist and active before running command
+        active_nodes = []
+        for node in self._nodes:
+            if self._verify_node_active(node):
+                active_nodes.append(node)
+        return active_nodes
+
+    @staticmethod
+    def _verify_node_active(node):
+        if node and node.is_active:
+            return node
+        log.info(f"Node {node.name} not active anymore")
+        return None
+
+    @abstractmethod
+    def execute(self):
+        pass
+
+    def undo(self):
+        for node in self.receivers:
+            cmd = self.node_undo_commands[node]
+            log.info(f"{node.name}  background execute: {cmd}")
+            node.run_command(cmd, background=True)
+
+
+class LoadCPU(Command):
+    """Load cpu command
+    The command replicated per cpu core and causes to high cpu usage.
+    In order to increase cpu load we can increase the core_numbers
+    example: in case we have 4 core we can run pipes and use more
+    command :
+    # cat /dev/urandom | gzip -9 | gzip -d | gzip -9 | gzip -d > /dev/null
+    The base command can be sha512sum /dev/urandom
+    """
+
+    base_command = "while true ; do echo; done"
+
+    def __init__(self, nodes, core_numbers: int):
+        super().__init__(nodes)
+        self.core_numbers = core_numbers
+
+    def execute(self):
+        cores_dup = "| gzip -9" * self.core_numbers
+        command = f"{self.base_command} {cores_dup} {self.redirect}"
+        for node in self.receivers:
+            log.info(f"{node.name}  background execute: {command}")
+            node.run_command(command, background=True)
+            self.node_undo_commands[node] = "kill `pidof gzip`"
+
+
+class LoadRam(Command):
+    """Load increase RAM memory by giga param
+    mount to ramfs and copy data to the ram storage and increasing ram usage.
+    """
+
+    base_command = Template(
+        """mkdir $folder;
+        mount -t ramfs ramfs $folder/;
+        dd if=/dev/zero of=$folder/file bs=1G count=$ram_giga;
+        """
+    )
+
+    def __init__(self, nodes, ram_giga: int, folder="ram123"):
+        super().__init__(nodes)
+        self.ram_giga = ram_giga
+        self.folder = folder
+
+    def execute(self):
+        command = self.base_command.safe_substitute(ram_giga=self.ram_giga, folder=self.folder)
+        command = f"{command} {self.redirect}"
+        for node in self.receivers:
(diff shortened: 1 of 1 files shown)


src/assisted_test_infra/test_infra/tools/load_system.py
```diff
@@ -0,0 +1,124 @@
...
+        pass
+
+    def undu(self):
+        for node in self.nodes:
+            cmd = self.node_undu_commands[node]
+            log.info(f"{node.name}  background execute: {cmd}")
+            node.run_command(cmd, background=True)
+
+
+class LoadCPU(Command):
```
ori-amizur, line 30: Maybe a way to simulate cpu overload is to limit cpu usage by limiting cpu instead of creating load.  Cgroups can provide it,or `cpulimit`

bkopilov: cpulimit not installed by default in our  CoreOS
cgroups never tried...
Not sure . 
cat /dev/urandom | gzip -9 | gzip -d | gzip -9 | gzip -d > /dev/null
This command will create the process on every cpu core .  we install out automation with minimal cores allowed by preflight .

src/assisted_test_infra/test_infra/tools/load_system.py
```diff
@@ -0,0 +1,164 @@
...
+    def execute(self):
+        command = self.base_command.safe_substitute(
+            action=self.action, interface=self.interface, rate_limit=self.rate_limit
+        )
+        for node in self.receivers:
+            log.info(f"{node.name}  background execute: {command}")
+            node.run_command(command, background=True)
+            self.node_undo_commands[node] = command.replace("add", "del")
+
+            self.node_undo_commands[node] = f"sudo tc qdisc del dev {self.interface} ingress"
```
bkopilov, line 164: I think line 164 should be removed.

bkopilov:
@eifrach , @danmanor

eifrach:
1.  I'm not sure this should be part of test-infra.  Test infra is for deploying and validating installation this seems more performance / stability test 

2. running remote command to load Network / CPU can cause loss of connectivity - not sure if  the assisted service know how to recover  / handle connection loss and/or slow connection
Furthermore, some of the agent validation tests can be affected thus causing validation failure during the registration /setup.

don't think this should be during the installation process

@osherdp  can you share your input here as well ?

bkopilov:
> 1. I'm not sure this should be part of test-infra.  Test infra is for deploying and validating installation this seems more performance / stability test
> 2. running remote command to load Network / CPU can cause loss of connectivity - not sure if  the assisted service know how to recover  / handle connection loss and/or slow connection
>    Furthermore, some of the agent validation tests can be affected thus causing validation failure during the registration /setup.
> 
> don't think this should be during the installation process
> 
> @osherdp can you share your input here as well ?

Please note that the reason is :
https://issues.redhat.com/browse/MGMT-8115  

We need to simulate slow / failures during installation . there is a huge difference in timeout installation when network is 1Gbps or 10Mbps.

bkopilov:
> > 1. I'm not sure this should be part of test-infra.  Test infra is for deploying and validating installation this seems more performance / stability test
> > 2. running remote command to load Network / CPU can cause loss of connectivity - not sure if  the assisted service know how to recover  / handle connection loss and/or slow connection
> >    Furthermore, some of the agent validation tests can be affected thus causing validation failure during the registration /setup.
> > 
> > don't think this should be during the installation process
> > @osherdp can you share your input here as well ?
> 
> Please note that the reason is : https://issues.redhat.com/browse/MGMT-8115
> 
> We need to simulate slow / failures during installation . there is a huge difference in timeout installation when network is 1Gbps or 10Mbps.

Finnaly we will be able to take "installation proccess and split it to stages and on each one of them reproduce timeout due to slow network. 
for example: bootstrap waiting ~1 hour to master to join , with slow network we can hit on timeout and make it as a test which validate this step + expected result
As for disk IO and rate not sure if i will do it but its anohter option.

osherdp:
> > 1. I'm not sure this should be part of test-infra.  Test infra is for deploying and validating installation this seems more performance / stability test
> > 2. running remote command to load Network / CPU can cause loss of connectivity - not sure if  the assisted service know how to recover  / handle connection loss and/or slow connection
> >    Furthermore, some of the agent validation tests can be affected thus causing validation failure during the registration /setup.
> > 
> > don't think this should be during the installation process
> > @osherdp can you share your input here as well ?
> 
> Please note that the reason is : https://issues.redhat.com/browse/MGMT-8115
> 
> We need to simulate slow / failures during installation . there is a huge difference in timeout installation when network is 1Gbps or 10Mbps.

* @bkopilov part one is for whether or not it should be in test-infra. If the only tests we're aiming to add are downstream and not upstream, then this code should be in `kni-assisted-installer-auto` where you have more autonomy to change it when needed. We guarantee much less "quality" upstream for assisted-installer, so I think load-testing should be part of downstream code (aka `kni-assisted-installer-auto`)
* as for part two, what kind of load should pre-exist on hardware that is getting installed? Those HWs aren't yet a cluster... and even if they have something running in there then they should be accounted for more resources (because openshift installation requires a certain amount of resources to be dedicated to it)

src/assisted_test_infra/test_infra/tools/load_system.py
```diff
@@ -0,0 +1,163 @@
...
+        if node and node.is_active:
+            return node
+        log.info(f"Node {node.name} not active anymore")
+        return None
+
+    @abstractmethod
+    def execute(self):
+        pass
+
+    def undo(self):
```
osherdp, line 38: just in general, it's better to use two terms that are opposites of each other (to make the API clearer)
something like `start` and `stop` instead of `execute` and `undo`

bkopilov:
The code will be moved to downstream

PR #2368, opened 2024-01-18
NO-ISSUE: improve dockerfile

eifrach:


Dockerfile.assisted-test-infra
```diff
@@ -39,33 +48,20 @@ RUN dnf -y install --enablerepo=crb \
...
-      pip3 install --upgrade /build/pip/*
-
-RUN curl --retry 5 --connect-timeout 30 -s https://storage.googleapis.com/golang/go1.18.1.linux-amd64.tar.gz | tar -C /usr/local -xz
-ENV GOPATH=/go
-ENV GOCACHE=/go/.cache
-ENV PATH=$PATH:/usr/local/go/bin:/go/bin
-
-COPY . .
+		pip3 install --no-cache-dir  assisted-service-client && \
+    pip3 install --no-cache-dir -I -r ./requirements.txt -r ./requirements-dev.txt
```
rccrdpccl, line 64: This was before copy all so that we could leverage caching (no need to pip install if any file changes apart from requirements)

PR #2373, opened 2024-01-23
Add support for disk tunning - rate and iops

bkopilov:
Libvirt allow to limit disks io and rates per disk name Added support to controller node ,it will allow us to slow down installation by limit disks speed or io.
On every installation phase we may limit disk and undo.


bkopilov:
@eifrach  , @danmanor

eifrach:
releated to #2365
will be reviewed togther

bkopilov:
Code will be moved to downstream

PR #2377, opened 2024-02-04
Add status_info for waiting for specific host

bkopilov:
Getting host status is not enough , we would like to catch a host in detailed status info.
Example:
installing-in-progress Waiting for bootkube and others


bkopilov:
@eifrach , @osherdp , @lalon4

PR #2381, opened 2024-02-13
Add default model when libvirt add_intrerface

bkopilov:
Current add_interface set the interface speed to 100M and limit the internal traffic between nodes when set the interface as CIDR and VIP Added default to virtio, configured the attach interface speed as N/A


bkopilov:
@adriengentil , @lalon4 , @eifrach

PR #2398, opened 2024-03-13
NO-ISSUE: netaddr package rollback do to breaking changes

eifrach:
netaddr remove support for `is_private` function.
reveting version and blocking packge update.

will revert after fix in ansible utils
https://github.com/ansible-collections/ansible.utils/issues/331


adriengentil (approved):
/hold
waiting for arm job to complete

PR #2403, opened 2024-03-21
fixing disk size in kickstart

eifrach:
- fix kickstart file to auto adjust to whatever disk size is given.
- remove calculation from MB to GB as it is not needed
- change installation type to "Minimal" faster installation and smaller footprint 

TODO: 
- align configuration to vSphere setup
- combine kickstart files to one shared file
- change OS type to Rocky 9

### DIFF
--- a/packer_files/nutanix_centos_template/variables.pkr.hcl
+++ b/packer_files/nutanix_centos_template/variables.pkr.hcl
@@ -44,7 +44,7 @@ variable "image_name" {
 variable "disk_size" {
   type = number
   default = 100
-  description = "The VM disk size in MB. default 100G"
+  description = "The VM disk size in GB. default 100GB"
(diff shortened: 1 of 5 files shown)


packer_files/nutanix_centos_template/variables.pkr.hcl
```diff
@@ -43,7 +43,7 @@ variable "image_name" {
 
 variable "disk_size" {
   type = number
-  default = 102400 
+  default = 100
   description = "The VM disk size in MB. default 100G"
```
adriengentil, line 47:
```suggestion
  description = "The VM disk size in GB. default 100GB"
```

adriengentil (commented):
Just a nit, otherwise it looks good!

PR #2404, opened 2024-03-24
MGMT-15772: change facilities to metro

eifrach:
Move Equinx to use metro

1. updating provider version to 1.32
2. setting metro instead of facilities
3. default `metro` is DA ( dallas )

metro DA has all need metal servers including ARM based

* note: there is one more metro with ARM images which is DC

### DIFF
--- a/terraform_files/equinix-ci-machine/main.tf
+++ b/terraform_files/equinix-ci-machine/main.tf
@@ -2,7 +2,7 @@ terraform {
   required_providers {
     equinix = {
       source = "equinix/equinix"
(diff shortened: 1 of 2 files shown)


danmanor:
@eifrach Can you please provide some details on the purpose of this switch for documentation sake ?

eifrach:
> @eifrach Can you please provide some details on the purpose of this switch for documentation sake ?

facilities has been deprecated -- so we need to move to using metro
along the way also upgraded the plugin version cause it was very old

PR #2406, opened 2024-03-26
MGMT-17080: test mce and storage

rccrdpccl:
add a test for MCE + storage functionality
we need to make sure the right objects are deployed correctly and the defined storage classes are matching
This will be added to the current MCE job

### DIFF
--- a/src/tests/test_e2e_install.py
+++ b/src/tests/test_e2e_install.py
@@ -1,8 +1,10 @@
 import pytest
 import semver
 from junit_report import JunitTestSuite
+from kubernetes import client, config
 
 import consts
+from assisted_test_infra.test_infra.utils import wait_for_pod_ready
 from tests.base_test import BaseTest
 from tests.config import global_variables
 from tests.conftest import get_available_openshift_versions, get_supported_operators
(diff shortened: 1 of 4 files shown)


src/tests/test_e2e_install.py
```diff
@@ -37,6 +38,47 @@ def test_olm_operator(self, cluster, olm_operators):
...
+        # get storage classes
+        storage_classes_list = client.StorageV1Api().list_storage_class()
+        assert len(storage_classes_list.items) > 0, "Expected storage class list to have one item or more"
+        storage_classes = [storage_class.metadata.name for storage_class in storage_classes_list.items]
+
+        # each storage pvc should have an existing storage class
+        storageClaims = ["databaseStorage", "filesystemStorage", "imageStorage"]
+        for key in storageClaims:
+            assert (
+                agent_service_config["spec"][key]["storageClassName"] in storage_classes
```
danmanor, line 62: Is is possible that `key` will not appear in ` agent_service_config` ? if so we will just get a `ValueError` or something not descriptive right ?

danmanor: Same for the other fields accessed in objects

src/tests/test_e2e_install.py
```diff
@@ -37,6 +38,47 @@ def test_olm_operator(self, cluster, olm_operators):
...
+        # workloads should be running
+        selectors = ["app=assisted-service", "app=assisted-image-service"]
+        for selector in selectors:
+            pods = v1.list_namespaced_pod(mce_namespace, label_selector=selector)
+            pod = pods.items[0]
+            for status in pod.status.container_statuses:
+                assert status.ready, "Expected pod to be ready"
+                assert status.started, "Expected pod to be started"
+                assert status.state.running is not None, "Expected pod to be running"
+
```
danmanor, line 81: Does the pod have enough time to be ready ? maybe we need some kind of waiting ?

danmanor (commented):
Looks good to me, have you tested it ?

danmanor:
@rccrdpccl Maybe we can add the job in `openshift/release` first and then test it in this PR ?

src/tests/test_e2e_install.py
rccrdpccl: Good point, we might need some waiting here

rccrdpccl:
> @rccrdpccl Maybe we can add the job in `openshift/release` first and then test it in this PR ?

I am planning to substitute an already existing job, I thought to first merge this as it's tested locally, and then test through release PR (and in case fix stuff here).
I feel this is a safer approach as AFAIU this function wouldn't be used otherwise, but if it takes too long to merge this we would break the existing job.
WDYT?

src/tests/test_e2e_install.py
rccrdpccl: I have added some, but most I think are adding a lot of verbosity but we don't gain much in safety (i.e. <obj>.metadata.name)

rccrdpccl:
> @rccrdpccl Your choice, the question is whether the change (if necessary) will be required in this code or in `openshift/release`. BTW, what job are you planning on replacing ?

I am planning to replace [these](https://github.com/openshift/release/blob/master/ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml#L292-L312). Basically just adding a post installations commands to run the function added in this PR.

danmanor:
> I am planning to replace [these](https://github.com/openshift/release/blob/master/ci-operator/config/openshift/assisted-test-infra/openshift-assisted-test-infra-master.yaml#L292-L312). Basically just adding a post installations commands to run the function added in this PR.

@rccrdpccl But you can still have only MCE, wouldn't it better to have a different test ?

rccrdpccl:
> @rccrdpccl But you can still have only MCE, wouldn't it better to have a different test ?

Maybe in a world of infinite resources, it definitely would :)

I don't think by deploying lvm or odf we are influencing mce installation, and we'll save resources

PR #2417, opened 2024-05-01
NO-ISSUE: Add natsort package to requirements.txt

bkopilov:
The package allows us to sort complicated versions with numbers and naming. Using this package in custom ocp tests, getting a list of versions and sorting for last element .

Package info: https://pypi.org/project/natsort/

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,6 +10,7 @@ junit-report==0.2.7
 kubernetes==29.0.0
 libvirt-python==10.4.0
 munch==4.0.0
+natsort==8.4.0 # natsort used in kni-assisted-installer-auto
 netaddr==1.3.0
 netifaces==0.11.0
 openshift-client==2.0.4


bkopilov:
@adriengentil , @danmanor

danmanor:
@bkopilov Why do you add the package but not use it ?

requirements.txt
```diff
@@ -10,6 +10,7 @@ junit-report==0.2.7
 kubernetes==29.0.0
 libvirt-python==10.2.0
 munch==4.0.0
+natsort
```
danmanor, line 13: We need the exact version

bkopilov:
> @bkopilov Why do you add the package but not use it ?

I am using it in kni-assisted-installer-auto tests which run from podman containers.

danmanor:
@bkopilov I don't understand, it is not used in this repo, so why add it here ?

bkopilov:
> @bkopilov I don't understand, it is not used in this repo, so why add it here ?

Because here we controll what pip packages are installed on build image.
Dockerfile.assisted-test-infra

> RUN pip3 install --upgrade pip && \
>       pip3 install --no-cache-dir -I -r ./requirements.txt -r ./requirements-dev.txt && \
>       pip3 install --upgrade /build/pip/*

danmanor:
I mean why not add it to `kni-assisted-installer-auto`  ?

eliorerz:
> I mean why not add it to `kni-assisted-installer-auto` ?

I agree, if it's not used in test-infra I think the best it to add it as a dependency to `kni-assisted-installer-auto`

bkopilov:
> > I mean why not add it to `kni-assisted-installer-auto` ?
> 
> I agree, if it's not used in test-infra I think the best it to add it as a dependency to `kni-assisted-installer-auto`

Hi , 
the kni-assisted-installer run from the container (Centos9) the package is part of pip installation on the container.
The Dockerfile.assisted-test-infra running the requirements.txt installation .

How can i install the python package from KNI ? 
Do you mean 
import pip 
pip.main(['install', "natsort"])

bkopilov:
> > > I mean why not add it to `kni-assisted-installer-auto` ?
> > 
> > 
> > I agree, if it's not used in test-infra I think the best it to add it as a dependency to `kni-assisted-installer-auto`
> 
> Hi , the kni-assisted-installer run from the container (Centos9) the package is part of pip installation on the container. The Dockerfile.assisted-test-infra running the requirements.txt installation .
> 
> How can i install the python package from KNI ? Do you mean import pip pip.main(['install', "natsort"])

@eliorerz , @lalon4  , Not sure how can i force this package from kni tests , the image docker set by you before we run kni ... and i must update the container package where the code runs from ?

What do you suggest  ?  the make action is from test-infra...  should i call to pip from KNI ?

eliorerz:
> > > > I mean why not add it to `kni-assisted-installer-auto` ?
> > > 
> > > 
> > > I agree, if it's not used in test-infra I think the best it to add it as a dependency to `kni-assisted-installer-auto`
> > 
> > 
> > Hi , the kni-assisted-installer run from the container (Centos9) the package is part of pip installation on the container. The Dockerfile.assisted-test-infra running the requirements.txt installation .
> > How can i install the python package from KNI ? Do you mean import pip pip.main(['install', "natsort"])
> 
> @eliorerz , @lalon4 , Not sure how can i force this package from kni tests , the image docker set by you before we run kni ... and i must update the container package where the code runs from ?
> 
> What do you suggest ? the make action is from test-infra... should i call to pip from KNI ?

I think that the best approach here is to add another step during kni-auto test-infra image build process, then build a `qe-test-infra` image using a new Dockerfile with all the dependencies you need.

bkopilov:
@eliorerz , Ok . i will close it and move the logic to kni-test-infra

bkopilov:
Moving the logic to kni-test

requirements.txt
```diff
@@ -10,6 +10,8 @@ junit-report==0.2.7
 kubernetes==29.0.0
 libvirt-python==10.2.0
 munch==4.0.0
+# natsort used in kni-assisted-installer-auto
+natsort==8.4.0
```
eliorerz, line 14:
```suggestion
natsort==8.4.0  # natsort used in kni-assisted-installer-auto
```

eliorerz (changes requested):

eliorerz:
/retitle NO-ISSUE: Add natsort package to requirements.txt

eliorerz (approved):
LGTM

PR #2419, opened 2024-05-06
MGMT-17250: Support RHEL9

danmanor:
Currently, `assisted-test-infra` is only compatible with `Rocky8`, `RHEL8`, and `CentOS8`. As `RHEL8` approaches end-of-life (`EOL`) at the end of this month, we aim to also support `Rocky9`, `RHEL9`, and `CentOS9`. A primary challenge with `stream9` is the absence of `xinetd` in these releases. To address this, I have replaced the port forwarding mechanism with `systemd` socket to listen (with `socat`) and service to forward. This solution should work on all versions.

- [x] Test on RHEL8
- [x] Test on RHEL9
- [x]  Test on Rocky8 (in the CI)
- [x] Test on Rocky9 (in the CI)

### DIFF
--- a/scripts/create_full_environment.sh
+++ b/scripts/create_full_environment.sh
@@ -19,13 +19,13 @@ fi
 
 #Check CentOS version
 VER=$(awk -F= '/^VERSION_ID=/ { print $2 }' /etc/os-release | tr -d '"' | cut -f1 -d'.')
-VER_SUPPORTED=8
-
-if [[ ${OS} =~ ^(centos)$ && ${VER} -ne ${VER_SUPPORTED} ]]; then
-    error "CentOS version ${VER_SUPPORTED} is required."
-    exit 1
-elif [[ ${OS} =~ ^(rhel)$ && ${VER} -ne ${VER_SUPPORTED} ]]; then
-    error "RHEL version ${VER_SUPPORTED} is required."
+SUPPORTED_VERSIONS=( 8 9 )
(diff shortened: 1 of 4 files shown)


talhil-rh:
Please don't merge this until we test it in our env, as it might impact us

vrutkovs:
Still breaks [here](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-metal/1788496740777725952):
```
 Installing libvirt-related packages...
Last metadata expiration check: 0:00:09 ago on Thu 09 May 2024 09:27:10 AM UTC.
No match for argument: libvirt-devel
Package libgcrypt-1.10.0-10.el9_2.x86_64 is already installed.
Error: Unable to find a match: libvirt-devel 
```
Tested with `workflow-test assisted-baremetal 4.16,https://github.com/openshift/assisted-test-infra/pull/2419 PACKET_OS=rocky_9` in cluster-bot

danmanor:
Thank you @vrutkovs

danmanor:
@vrutkovs  We need to wait for https://github.com/openshift/installer/pull/8391

danmanor:
hold

danmanor:
waiting for the CI to stabilize

danmanor:
overriding e2e-metal-single-node-live-iso as it passed the test but fails in conformance test that still uses 4.16 as it is not our workflow

PR #2421, opened 2024-05-09
NO-ISSUE: Update `requirements.txt` to match the package naming conventions of the interpreter.

eliorerz:
Some package names were incorrectly written with erroneous characters, causing certain IDEs to mistakenly believe that the packages were not present.
Also update `pycharm-remote-debugger` to latest

/cc @danmanor @adriengentil


eliorerz:
The failures doesn't seem to be related to the change, I think we can override it

danmanor:
@eliorerz It might worth backport, WDYT?

PR #2423, opened 2024-05-15
NO-ISSUE: Fix auto VIPs allocation on local Nutanix flow

eliorerz:
This changes will only take place when running nutanix test locally - because that `vip_dhcp_allocation` is always set to false in nutanix CI.

/cc @adriengentil @danmanor

### DIFF
--- a/src/assisted_test_infra/test_infra/controllers/node_controllers/nutanix_controller.py
+++ b/src/assisted_test_infra/test_infra/controllers/node_controllers/nutanix_controller.py
@@ -56,7 +56,17 @@ class NutanixController(TFController):
         vm.power_on()
 
     def get_ingress_and_api_vips(self):
-        if not self._entity_config.vip_dhcp_allocation:
+        """
+        Need to distinguish between 3 cases:
+        1) vip_dhcp_allocation is set to False: Need to provide API and Ingress VIPs - raise an exception if
+        one or more are empty.
+        2) vip_dhcp_allocation is set to True: No need to provide API and Ingress VIP, return None.
+        3) vip_dhcp_allocation is not being set at all and its value is equal to None: In this case, search free IPs
+        and set them as VIPs. The behavior is the same as vip_dhcp_allocation = False but getting the IPs first.
+        Note that (3) is supposed to happen only locally due to the face that vip_dhcp_allocation is set in CI to some
+        value.
+        """
+        if self._entity_config.vip_dhcp_allocation is False:
             if self._entity_config.api_vips is None or len(self._entity_config.api_vips) == 0:
                 raise ValueError("API VIP is not set")
             if self._entity_config.ingress_vips is None or len(self._entity_config.ingress_vips) == 0:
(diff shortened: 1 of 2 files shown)


src/assisted_test_infra/test_infra/controllers/node_controllers/nutanix_controller.py
```diff
@@ -66,6 +66,10 @@ def get_ingress_and_api_vips(self):
                 "ingress_vips": self._entity_config.ingress_vips,
             }
 
+        elif self._entity_config.vip_dhcp_allocation is True:
```
adriengentil, line 69:
```suggestion
        else:
```

should be enough?

eliorerz: Nope, we need to differ True, from False from None.
`else` in this case will include the `None` case

adriengentil: can we add some doc at the beginning of te function to describe the 3 cases? I think it would help to understand what we try to achieve

eliorerz: There is https://github.com/openshift/assisted-test-infra/pull/2423/files#diff-6016d1ff333f74b2b2703cf120015d2daea4cd905295107122b40f81a7b7e435R72

eliorerz: I can elaborate more if needed

adriengentil: yeah, it's just because I find this structure a bit misleading to read:
```
def f(param):
  if param==true:
    return
  if param==false
    return
  
  ... more code here because param is none...

```

eliorerz: @adriengentil I added a function documentation, LMK if that is OK

src/assisted_test_infra/test_infra/controllers/node_controllers/nutanix_controller.py
```diff
@@ -56,7 +56,17 @@ def restart_node(self, node_name: str) -> None:
...
     def get_ingress_and_api_vips(self):
-        if not self._entity_config.vip_dhcp_allocation:
+        """
+        Need to distinguish between 3 cases:
+        1) vip_dhcp_allocation is set to False: Need to provide API and Ingress VIPs - raise an exception if
+        one or more are empty.
+        2) vip_dhcp_allocation is set to True: No need to provide API and Ingress VIP, return None.
+        3) vip_dhcp_allocation is not being set at all and its value is equal to None: In this case, search free IPs
+        and set them as VIPs. The behavior is the same as vip_dhcp_allocation = False but getting the IPs first.
+        Note that (3) is supposed to happen only locally due to the face that vip_dhcp_allocation is set in CI to some
```
adriengentil, line 66:
```suggestion
        Note that (3) is supposed to happen only locally due to the fact that vip_dhcp_allocation is set in CI to some
```

src/assisted_test_infra/test_infra/controllers/node_controllers/nutanix_controller.py
adriengentil: perfect, thanks! Just a typo.

PR #2424, opened 2024-05-16
NO-ISSUE: Fix Nutanix fail to login ssh key

eliorerz:
Update packer nutanix public ssh key

/cc @danmanor


eliorerz:

PR #2442, opened 2024-05-29
NO-ISSUE: Remove unused methods from `BaseTest` class and add missing type hint annotations

eliorerz:
In this PR:
- Changed in `BaseTest` methods:
    - `get_nodes` - Unused fixture - Removed
    - `get_cluster` - Unused fixture - Removed
    - `wait_for_controller` - Unused method - Removed
    - `verify_no_logs_uploaded` - Unused method - Removed
    - `_collect_virsh_logs` - Moved to `utils.logs_utils.py`

- Added some of missing hint annotations in `BaseTest`. 
- Add `get_base_asset` method to be under `Day2ClusterConfig` instead of const under `BaseTest.prepared_day2_controller_configuration`.

### DIFF
--- a/src/assisted_test_infra/test_infra/controllers/node_controllers/libvirt_controller.py
+++ b/src/assisted_test_infra/test_infra/controllers/node_controllers/libvirt_controller.py
@@ -2,6 +2,7 @@ import json
 import os
 import re
 import secrets
+import shutil
 import string
 import tempfile
 import xml.dom.minidom as md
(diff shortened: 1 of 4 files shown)


eliorerz:
FYI @talhil-rh

bkopilov:
@eliorerz  I think its safe to megre from kni-assisted-installer-auto view (none of the above fixture called directly)
/lgtm

src/assisted_test_infra/test_infra/controllers/node_controllers/libvirt_controller.py
```diff
@@ -675,3 +676,48 @@ def set_single_node_ip(self, ip):
...
         pass
+
+
+def collect_virsh_logs(nodes, log_dir_name):
+    log.info("Collecting virsh logs\n")
+    os.makedirs(log_dir_name, exist_ok=True)
+    virsh_log_path = os.path.join(log_dir_name, "libvirt_logs")
+    os.makedirs(virsh_log_path, exist_ok=False)
+
+    libvirt_list_path = os.path.join(virsh_log_path, "virsh_list")
```
bkopilov, line 687: Hi @eliorerz  , We use libvirt CLI directly instead python api in different places , maybe we can add the remote_libvirt host ... because we do support remore_libvirt
The assumption that we run it locally  (run_command or libvirt) may be changed in the future


> class LibvirtController(NodeController, ABC):
>     TEST_DISKS_PREFIX = "ua-TestInfraDisk"
> 
>     def __init__(
>         self,
>         config: BaseNodesConfig,
>         entity_config: Union[BaseClusterConfig, BaseInfraEnvConfig],
>         libvirt_uri: str = consts.DEFAULT_LIBVIRT_URI,
>     ):

eliorerz: Ok, you can create a ticket for that.
This PR just refactored and moved things from one place to another.

danmanor (approved):

PR #2446, opened 2024-06-02
Add support for libvirt removable disks and bus types

bkopilov:
Added support for test disks  different bus types: usb, virtio, sata, scsi In case usb we set the disk to removable

### DIFF
--- a/src/assisted_test_infra/test_infra/controllers/node_controllers/libvirt_controller.py
+++ b/src/assisted_test_infra/test_infra/controllers/node_controllers/libvirt_controller.py
@@ -13,6 +13,7 @@ from xml.dom import minidom
 from xml.etree.ElementTree import Element, SubElement, tostring
 
 import libvirt
+import random
 import waiting
 
 import consts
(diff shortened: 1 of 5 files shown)


bkopilov:
@eliorerz   @adriengentil  @lalon4

adriengentil:
@bkopilov we need to update the usb controller as well, right?

bkopilov:
> @bkopilov we need to update the usb controller as well, right?

yes , i think terraform

eliorerz:
Please change the title and commit message to contain the issue number or NO-ISSUE if there isn't any.

src/assisted_test_infra/test_infra/controllers/node_controllers/libvirt_controller.py
```diff
@@ -266,39 +267,40 @@ def format_disk(cls, disk_path):
...
+        identifiers_in_use = [self._get_disk_identifier(disk) for disk in self._get_all_disks(node, bus=bus)]
         try:
             result = next(candidate for candidate in string.ascii_lowercase if candidate not in identifiers_in_use)
         except StopIteration as e:
             raise ValueError(f"Couldn't find available scsi disk letter, all are taken: {identifiers_in_use}") from e
 
         return result
 
-    def attach_test_disk(self, node_name, disk_size, bootable=False, persistent=False, with_wwn=False):
+    def attach_test_disk(self, node_name, disk_size, bus="scsi", bootable=False, persistent=False, with_wwn=False):
```
eliorerz, line 303: This whole method is relevant only to libvirt, not sure we want it here

src/assisted_test_infra/test_infra/controllers/node_controllers/libvirt_controller.py
```diff
@@ -309,8 +311,10 @@ def attach_test_disk(self, node_name, disk_size, bootable=False, persistent=Fals
         # We don't use `vd` virtio disks because libvirt overwrites our aliases if we do so, coming up with
         # its own `virtio-<num>` aliases instead. Those aliases allow us to identify disks created by this
         # function when we perform `detach_all_test_disks` for cleanup.
-        target_dev = f"sd{self._get_available_scsi_identifier(node)}"
-        disk_alias = f"{self.TEST_DISKS_PREFIX}-{target_dev}"
+        dev = {"scsi": "sd", "sata": "sd", "usb": "sd", "virtio": "vd"}
+        target_dev = f"{dev[bus]}{self._get_available_identifier(node, None)}"
+        # Allow to create multiple devices when node shutdown required (sata)
+        disk_alias = f"{self.TEST_DISKS_PREFIX}{str(random.randint(0, 10000))}-{bus}-{target_dev}"
```
eliorerz, line 317: Please explain what is the random for

bkopilov: There are two ways to add: 
- VIR_DOMAIN_AFFECT_LIVE  when support hotplugin (apply when execute on running node )
- VIR_DOMAIN_AFFECT_CURRENT - for SATA we must shutdown /start the node.

Not mandatory but if  the change does not return new target_dev and we want to run the command multiple time it is going to prevent duplicate disk_alias which bubble libvirt exceptions...
for example:
attach_disk_persistent(master, ODF_DISK_SIZE, bus="sata")  --> success
attach_disk_persistent(master, ODF_DISK_SIZE, bus="sata")  --> fail  without restart but indicate that  target sdb already exists  ......

bkopilov: so we can try catch this node.attachDeviceFlags(attach_template, attach_flags).....
and if it raise  exception we can increment the target_dev .... not critical .
if you dont like it... i can remove

PR #2457, opened 2024-06-05
[release-ocm-2.7] NO-ISSUE: fix for missing distutils

eifrach:


danmanor:
I M not sure we realy want to run this jobs in this case @eifrach 
@adriengentil WDYT ?

adriengentil:
Looks like the jobs are failling at installing skipper, we can merge this PR as the issue doesn't seem to be related to this change, but the skipper issue needs to be fixed as well.

eifrach:
> Looks like the jobs are failling at installing skipper, we can merge this PR as the issue doesn't seem to be related to this change, but the skipper issue needs to be fixed as well.

so It seems that the skipper version wasn't merged to this branches

adriengentil (approved):

PR #2460, opened 2024-06-06
MGMT-17246: add support for ignoring cluster installations that already exist on jira

omer-vishlitzky:
As part of the triage process, we download logs for all the failed cluster installations and create jira tickets for each failed installation.
Currently, we download all the logs time and time again, and this PR adds additional optional argument, a file containing a list of cluster ids, which we can exclude from downloading.

Also, since SSO_URL is a required env var, I added an exception in case it is not set.


src/assisted_test_infra/test_infra/helper_classes/config/base_nodes_config.py
```diff
@@ -42,7 +42,9 @@ class BaseNodesConfig(BaseConfig, ABC):
     base_cluster_domain: Optional[str] = None
 
     network_mtu: int = None
-    tf_platform: str = None  # todo - make all tf dependent platforms (e.g. vsphere, nutanix) inherit from BaseTerraformConfig  # noqa E501
+    tf_platform: str = (
+        None  # todo - make all tf dependent platforms (e.g. vsphere, nutanix) inherit from BaseTerraformConfig  # noqa E501
+    )
```
omer-vishlitzky, line 47: done by `make reformat`

src/assisted_test_infra/download_logs/__main__.py
```diff
@@ -89,6 +99,11 @@ def handle_arguments():
         type=str,
         default=os.path.join(os.getenv("HOME"), ".kube/config"),
     )
+    parser.add_argument(
```
danmanor, line 102: I think we should make it optional to avoid errors as we can't merge this PR and the other one at the same time

src/service_client/assisted_service_api.py
```diff
@@ -76,7 +76,14 @@ def refresh_api_key(config: Configuration) -> None:
...
             log.info("Refreshing API key")
-            response = requests.post(os.environ.get("SSO_URL"), data=params)
+            try:
+                sso_url = os.environ["SSO_URL"]
+            except KeyError:
+                log.error("The environment variable SSO_URL is mandatory but was not supplied")
+                raise
+            else:
+                response = requests.post(sso_url, data=params)
+
```
danmanor, line 86: Can you extract this to another PR to keep things clear ?

src/assisted_test_infra/download_logs/__main__.py
```diff
@@ -89,6 +99,11 @@ def handle_arguments():
...
+        "--clusters-to-filter-out-file",
+        help="A file containing a list of clusters to ignore, delimeted by a new line",
```
rccrdpccl, line 104: Looks like this is expected to be tuple `<cluster_id>,<install_ts>`, isn't it?

src/assisted_test_infra/download_logs/__main__.py
```diff
@@ -89,6 +99,11 @@ def handle_arguments():
         type=str,
         default=os.path.join(os.getenv("HOME"), ".kube/config"),
     )
+    parser.add_argument(
+        "--clusters-to-filter-out-file",
```
rccrdpccl, line 103: maybe `--exclude-clusters` or `--exclude-installs` would be more descriptive

src/service_client/assisted_service_api.py
omer-vishlitzky: Actually I think it should be in this PR because the absence of this env var prohibits one from running the download logs flow

src/assisted_test_infra/download_logs/__main__.py
omer-vishlitzky: any argument with -- preceding it is optional by default

src/assisted_test_infra/download_logs/__main__.py
omer-vishlitzky: Done, thanks!

src/assisted_test_infra/download_logs/__main__.py
omer-vishlitzky: yes, changed the docs, thanks!

PR #2466, opened 2024-06-10
NO-ISSUE: Use centos 9 as default. Upgrade Nutanix packer plugin to latest

eliorerz:
/cc @eifrach


eliorerz:

PR #2469, opened 2024-06-11
NO-ISSUE: fixing minor syntax

eifrach:
- changeing `EXTERNAL_PORT` to boolean
-  removing redundant commands


scripts/utils.sh
```diff
@@ -141,7 +139,7 @@ function close_external_ports() {
 
 function add_firewalld_port() {
     port=$1
-    if [ "${EXTERNAL_PORT}" = "y" ]; then
+    if [ "${EXTERNAL_PORT}" ]; then
```
adriengentil, line 142:
```
$ EXTERNAL_PORT=false
$ if [ "$EXTERNAL_PORT" ]; then
∙ echo "foo"
∙ fi
foo
```
this condition evaluates to true even though EXTERNAL_PORT is set to false. If it intended?

eifrach: yes, you are correct it because of the brackets.
fixed

adriengentil (approved):

PR #2470, opened 2024-06-13
NO-ISSUE: Remove docker as container runtime command option

eliorerz:
Due to the lack of docker support in newer skipper versions

/cc @danmanor

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 #############
 
 SHELL=/bin/sh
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -3,7 +3,7 @@
 #############
 
 SHELL=/bin/sh
-CONTAINER_COMMAND := $(shell scripts/utils.sh get_container_runtime_command)
```
danmanor, line 6: I think skipper sets this var to podman-remote

danmanor:
@eliorerz can we close this one ?

PR #2471, opened 2024-06-13
NO-ISSUE: Add fallback installation with crb repository as it is included by default in some OS variants and causes an error when specified

danmanor:
Add fallback installation with `crb` repository as it is included by default in some `RHEL9` versions and causes an error when specified

### DIFF
--- a/scripts/install_environment.sh
+++ b/scripts/install_environment.sh
@@ -45,14 +45,11 @@ function config_additional_modules() {
 
 function install_libvirt() {
     echo "Installing libvirt-related packages..."
-    source "/etc/os-release" # This should set `PRETTY_NAME` as environment variable
-    if [[ "${PRETTY_NAME}" == "Rocky Linux 9"* ]]; then
-        sudo dnf install -y 'dnf-command(config-manager)'
-        sudo dnf config-manager --set-enabled crb
-    fi
 
-    # CRB repo is required for libvirt-devel
-    sudo dnf install -y --enablerepo=crb \
+    # CRB repo is required for libvirt-devel in some versions of RHEL
+    sudo dnf install -y 'dnf-command(config-manager)' || true
+    sudo dnf config-manager --set-enabled crb || true
+    sudo dnf install -y \
         libvirt \
         libvirt-devel \
         libvirt-daemon-kvm \


adriengentil:
why not just try to enable `crb` before installing packages? 
```
dnf config-manager --set-enabled cdb
```

scripts/install_environment.sh
```diff
@@ -63,6 +65,24 @@ function install_libvirt() {
...
+        sudo dnf install -y \
+            libvirt \
+            libvirt-devel \
+            libvirt-daemon-kvm \
+            qemu-kvm \
+            libgcrypt \
+            swtpm \
+            swtpm-tools \
+            socat \
+            tigervnc-server
```
adriengentil, line 81: for maintenance reasons, I think the list of the packages should be defined once

adriengentil:
BTW in dev-scripts, they just don't setup crb if running on rhel 9: https://github.com/openshift-metal3/dev-scripts/blob/master/01_install_requirements.sh#L81-L90

scripts/install_environment.sh
danmanor: Agree

danmanor:
@adriengentil It seemse like in RHEL8 it happens also, so I changed the flow to try enabling without fail

scripts/install_environment.sh
```diff
@@ -45,14 +45,11 @@ function config_additional_modules() {
...
-        sudo dnf install -y 'dnf-command(config-manager)'
-        sudo dnf config-manager --set-enabled crb
-    fi
 
-    # CRB repo is required for libvirt-devel
-    sudo dnf install -y --enablerepo=crb \
+    # CRB repo is required for libvirt-devel in some versions of RHEL
+    sudo dnf install -y 'dnf-command(config-manager)' || true
+    sudo dnf config-manager --set-enabled crb || true
+    sudo dnf install -y \
```
eifrach, line 52: just a recomandation when using dnf you can add the following
```console
sudo dnf install -y \
     --setopt=tsflags=nodocs --setopt=install_weak_deps=False \
     ...
 ```
 
 This will remove a lot of unneeded packages and docs from the installation - make it faster and smaller

danmanor: Thank you

PR #2472, opened 2024-06-16
[MGMT-18102] fix none / external platform on RHEL9

eifrach:


eifrach:
/retitle [MGMT-18102] fix none / external platform on RHEL9

PR #2473, opened 2024-06-16
MGMT-8357: Deploy assisted-service for e2e tests in Kind

danmanor:
After https://github.com/openshift/assisted-service/pull/6664 is merged, most of assisted installer ecosystem deployment scripts will be part of assisted-service repository. This PR uses these scripts instead of duplicating code. In addition, it enables `kind` deployment in this repository alongside existing `minikube` option


danmanor:

scripts/create_full_environment.sh
```diff
@@ -34,11 +34,12 @@ scripts/install_environment.sh
...
 
 echo "Installing kind"
-scripts/hub-cluster/kind/kind.sh install
+make bring_assisted_service
+assisted-service/hack/kind/kind.sh install
 echo "Done installing kind"
 
 echo "Installing minikube"
-scripts/hub-cluster/minikube.sh install
+assisted-service/hack/minikube/minikube.sh install
```
eifrach, line 42: we are installing both kind and minikube ?

danmanor: because we want to support both, and the deployment choice is happening during deployment stage using `DEPLOY_TARGET`, while the installations are happening before during setup

scripts/utils.sh
```diff
@@ -277,4 +277,20 @@ function retry() {
...
+    kubectl get pods -n $namespace -l $pod_label -o name
+}
+
+function is_pod_ready() {
+    local pod_name=$1
+    local namespace=$2
+
+    while [ $(kubectl get -n $namespace $pod_name -o jsonpath='{.status.conditions[?(@.type=="Ready")].status}') !=  "True" ]; do
+        sleep 3
+    done
```
eifrach, line 293: better to use builtin check instead ( also adding a timeout is good practice, we should set it as a variable) 
```suggestion
 kubectl wait -n $namespace --for=condition=Ready=false --timeout=60s  $pod_name
```

eifrach: also I perfer to check if it's `running` not `Ready`
```suggestion
  kubectl wait -n $namespace --for=jsonpath='{.status.phase}'=Running --timeout=60s  $pod_name
```

src/assisted_test_infra/test_infra/utils/utils.py
```diff
@@ -233,11 +233,17 @@ def get_remote_assisted_service_url(oc, namespace, service, scheme):
 
 
 def get_local_assisted_service_url(namespace, service, deploy_target):
+    res = subprocess.check_output("ip route get 1", shell=True).split()[6]
```
eifrach, line 236: please check the `res.returncode == 0`

eifrach: also if you want the host IP, it's better to use
```shell
> hostname -i
10.1.155.11
```

I think we should use python commands instead of bash

```python
import socket
print ( socket.gethostbyname(socket.gethostname()))
# output '10.1.155.11'
```

src/assisted_test_infra/test_infra/utils/utils.py
```diff
@@ -233,11 +233,17 @@ def get_remote_assisted_service_url(oc, namespace, service, scheme):
...
+    ip = str(res, "utf-8")
     if deploy_target == "onprem":
         assisted_hostname_or_ip = os.environ["ASSISTED_SERVICE_HOST"]
         return f"http://{assisted_hostname_or_ip}:8090"
```
eifrach, line 240: this check should come before checking the IP address line 236

src/assisted_test_infra/test_infra/utils/utils.py
```diff
@@ -233,11 +233,17 @@ def get_remote_assisted_service_url(oc, namespace, service, scheme):
...
     elif deploy_target == "kind":
-        return f"http://{socket.gethostname()}"
+        url = f"http://{ip}:8090"
+        if is_assisted_service_reachable(url):
+            return url
+
+        raise RuntimeError(f"The parsed url {url} to service {service} in {namespace} namespace was not reachable.]")
     elif deploy_target == "ocp":
         res = subprocess.check_output("ip route get 1", shell=True).split()[6]
         ip = str(res, "utf-8")
```
eifrach, line 249: this is  repeats the code from line 236

Makefile
```diff
@@ -147,10 +148,10 @@ image_build: bring_assisted_service generate_python_client
 	$(CONTAINER_COMMAND) tag $(IMAGE_NAME) test-infra:latest  # For backwards computability
 
 create_hub_cluster:
-	scripts/hub-cluster/create.sh
+	TARGET=${DEPLOY_TARGET} assisted-service/hack/hub_cluster.sh create
 
 delete_hub_cluster:
-	scripts/hub-cluster/delete.sh
+	(cd assisted-service && TARGET=${DEPLOY_TARGET} ROOT_DIR=${ROOT_DIR}/assisted-service hack/hub_cluster.sh delete)
```
eifrach, line 154: same refering to external resource which  we are testing

Makefile
```diff
@@ -147,10 +148,10 @@ image_build: bring_assisted_service generate_python_client
 	$(CONTAINER_COMMAND) tag $(IMAGE_NAME) test-infra:latest  # For backwards computability
 
 create_hub_cluster:
-	scripts/hub-cluster/create.sh
+	TARGET=${DEPLOY_TARGET} assisted-service/hack/hub_cluster.sh create
```
eifrach, line 151: same refering to external resource which  we are testing

danmanor:

PR #2486, opened 2024-07-09
MGMT-18333: Generate python client instead of using the one from assisted-service image

danmanor:
Currently, we generate python client for assisted-service inside its primary Dockerfile. In order to generate the python client, we need the repository git tags. This behavior collisions with Konflux behavior because Konflux builds the image, searching for the tags in the fork instead of the upstream repository. As a result developers currently need to periodically push git tags to their fork for konflux to build successfully. 

This PR generates generated assisted-service client during assisted-test-infra image build, so later we could follow with a PR to remove assisted-service's generation

reference - https://redhat-internal.slack.com/archives/C035X734RQB/p1719496182644409

### DIFF
--- a/Dockerfile.assisted-test-infra
+++ b/Dockerfile.assisted-test-infra
@@ -1,5 +1,3 @@
-FROM quay.io/edge-infrastructure/assisted-service:latest AS service
-
 FROM quay.io/assisted-installer-ops/base-python:3.12
 
 # A directory in the path with write permission even for non-root users
@@ -56,10 +54,8 @@ COPY --from=quay.io/ocp-splat/govc:v0.29.0 /govc /usr/local/bin
 WORKDIR /home/assisted-test-infra
 
 COPY requirements.txt requirements-dev.txt ./
-COPY --from=service /clients/assisted_service_client-*.whl /build/pip/
 RUN pip3 install --upgrade pip && \
-      pip3 install --no-cache-dir -I -r ./requirements.txt -r ./requirements-dev.txt && \
-      pip3 install --upgrade /build/pip/*
+      pip3 install --no-cache-dir -I -r ./requirements.txt -r ./requirements-dev.txt
 
 RUN curl --retry 5 --connect-timeout 30 -s https://storage.googleapis.com/golang/go1.18.1.linux-amd64.tar.gz | tar -C /usr/local -xz
 ENV GOPATH=/go
@@ -77,3 +73,8 @@ RUN source scripts/utils.sh && \
 
 # setting pre-commit env
 ENV PRE_COMMIT_HOME build
+
+# Install the assisted-service Python client if it exists locally. 
+# In the CI, it will be missing during the first build in "images" in prow,
+# but should be present after running "make image_build" during setup.
+RUN pip3 install --no-index ./.pip/* || true
--- a/Makefile
+++ b/Makefile
@@ -142,10 +142,8 @@ setup:
 
 create_environment: image_build bring_assisted_service create_hub_cluster
 
-image_build:
-	scripts/pull_dockerfile_images.sh
-	sed 's/^FROM .*assisted-service.*:latest/FROM $(subst /,\/,${SERVICE})/' Dockerfile.assisted-test-infra | \
-		$(CONTAINER_COMMAND) build --network=host -t $(IMAGE_NAME) -f- .
+image_build: bring_assisted_service generate_python_client
+	$(CONTAINER_COMMAND) build --network=host -t $(IMAGE_NAME) -f Dockerfile.assisted-test-infra .
 	$(CONTAINER_COMMAND) tag $(IMAGE_NAME) test-infra:latest  # For backwards computability
 
 create_hub_cluster:
@@ -410,3 +408,7 @@ cli:
 
 validate_client:
 	skipper run "python3 ${DEBUG_FLAGS} src/service_client/client_validator.py"
+
+generate_python_client: bring_assisted_service
+	cd assisted-service && skipper make generate-python-client
+	mkdir -p ./.pip && mv assisted-service/build/assisted-installer/assisted-service-client/dist/*.whl ./.pip/
--- a/scripts/bring_assisted_service.sh
+++ b/scripts/bring_assisted_service.sh
@@ -18,13 +18,16 @@ export REPO_NAME=${REPO_NAME:-assisted-service}
 export JOB_TYPE=${JOB_TYPE:-}
 export PULL_BASE_REF=${PULL_BASE_REF:-master}
 
+
 if ! [[ -d "assisted-service" ]]; then
   echo "Can't find assisted-service source locally, cloning ${SERVICE_REPO}"
   retry -- git clone "${SERVICE_REPO}"
+elif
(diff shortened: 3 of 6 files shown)


Dockerfile.assisted-test-infra
```diff
@@ -1,4 +1,15 @@
-FROM quay.io/edge-infrastructure/assisted-service:latest AS service
+# Generate python client
+FROM quay.io/edge-infrastructure/swagger-codegen-cli:2.4.18 AS swagger_py
+
+COPY assisted-service/swagger.yaml /
+COPY assisted-service/tools/generate_python_client.sh .
```
rccrdpccl, line 5: Any reason why we should leave this python generating script in assisted-repo?

Dockerfile.assisted-test-infra
```diff
@@ -1,4 +1,15 @@
-FROM quay.io/edge-infrastructure/assisted-service:latest AS service
+# Generate python client
+FROM quay.io/edge-infrastructure/swagger-codegen-cli:2.4.18 AS swagger_py
+
+COPY assisted-service/swagger.yaml /
```
rccrdpccl, line 4: Can we retrieve this file without relying on a checkout? Maybe we can pass branch name as build argument?

Dockerfile.assisted-test-infra
```diff
@@ -1,4 +1,15 @@
...
+COPY assisted-service/tools/generate_python_client.sh .
+RUN chmod +x ./generate_python_client.sh && SWAGGER_FILE=swagger.yaml OUTPUT=/build ./generate_python_client.sh
+
+FROM registry.access.redhat.com/ubi9/python-311:latest AS pybuilder
+COPY --chown=1001:0 ./assisted-service /assisted-service
+COPY --from=swagger_py --chown=1001:0 /build /assisted-service/build
+RUN python3 -m pip install -I pip && pip3 install -I -r /assisted-service/dev-requirements.txt
+RUN cd /assisted-service/build && python3 ../tools/client_package_initializer.py ./ https://github.com/openshift/assisted-service
```
rccrdpccl, line 12: nit: In general I think we should avoid consecutive RUN statements: we can squash them into one and reduce layers

Dockerfile.assisted-test-infra
danmanor: we need this to stay in assisted-service because we have a postsubmit job in assisted-service that generates python client and push it to pypi

Dockerfile.assisted-test-infra
danmanor: what checkout do you mean ?

rccrdpccl: `assisted-service` subfolder is being checked out by assisted-test-infra. We don't really need to rely on this for this dockerfile

Dockerfile.assisted-test-infra
rccrdpccl: the issue I see with this is that 5 lines of bash script create a dependency on having this checkout available. Which might be OK during normal usage as it normally it would be there, but it might not always be the case.

I know we agreed on this approach in general, but this is the reason why I don't like having 2 places where we generate the python client. Can this be simplified?

scripts/bring_assisted_service.sh
```diff
@@ -18,14 +18,17 @@ export REPO_NAME=${REPO_NAME:-assisted-service}
...
 
+
 if ! [[ -d "assisted-service" ]]; then
   echo "Can't find assisted-service source locally, cloning ${SERVICE_REPO}"
   retry -- git clone "${SERVICE_REPO}"
+elif
+  [[ ${SERVICE_REPO} != $(cd assisted-service && git remote get-url origin) ]]; then
+  echo "assisted-service repository found locally but with a different origin, replacing with ${SERVICE_REPO}"
+  rm -rf ./assisted-service
+  retry -- git clone "${SERVICE_REPO}"
```
adriengentil, line 29: What happens if as a developer, I want to test with my own branch of assisted service? I don't think we want to align in with test-infra :thinking:

skipper.yaml
```diff
@@ -1,5 +1,5 @@
 registry: quay.io
-build-container-image: assisted-test-infra
+build-container-image: assisted-test-infra-build
```
adriengentil, line 2: why, they are not the same? What if I want to work with terraform?

scripts/bring_assisted_service.sh
danmanor: You can still test your own branch as I specified in the doc

skipper.yaml
danmanor: `assisted-test-infra-build` is `assisted-test-infra` + another layer with assisted-service-client. We currently can't test the building of that image in the CI `images` part as we don't have assisted-service-code, so I splited the images

Dockerfile.assisted-test-infra-build
```diff
@@ -0,0 +1,18 @@
...
+COPY assisted-service/swagger.yaml /
+COPY assisted-service/tools/generate_python_client.sh .
+RUN chmod +x ./generate_python_client.sh && SWAGGER_FILE=swagger.yaml OUTPUT=/build ./generate_python_client.sh
+
+FROM registry.access.redhat.com/ubi9/python-311:latest AS pybuilder
+
+COPY --chown=1001:0 ./assisted-service /assisted-service
+COPY --from=swagger_py --chown=1001:0 /build /assisted-service/build
+RUN python3 -m pip install -I pip && pip3 install -I -r /assisted-service/dev-requirements.txt
+RUN cd /assisted-service/build && python3 ../tools/client_package_initializer.py ./ https://github.com/openshift/assisted-service
```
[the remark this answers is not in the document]
danmanor: Currently we use upstream assisted-service-client by default. If the `SERVICE` env var is set, then it substitutes the old image with the new one. This way we get the correct client. my change achieves the same thing in a different way, using assisted-service code instead of substituting images

Makefile
```diff
@@ -410,3 +408,7 @@ cli:
 
 validate_client:
 	skipper run "python3 ${DEBUG_FLAGS} src/service_client/client_validator.py"
+
+generate_python_client: bring_assisted_service
+	cd assisted-service && skipper make generate-python-client
+	mkdir -p ./.pip && mv assisted-service/build/assisted-installer/assisted-service-client/dist/*.whl ./.pip/
```
eliorerz, line 414: I would give it more indicative name. Maybe under `build/` or `dist/`

eliorerz (changes requested):

Makefile
danmanor: `build` is in `.dockerignore`. I can use `dist`, but why is it more fitting ?

eliorerz: pip is the package installer in python, if in the future someone will investigate something and see .pip he could think it is related to the package installer.
I prefer to give it a name that someone that reads it will understand immediately

eifrach:
I would add documentation for the new make option but, we can do it in a new PR

/lgtm
\

danmanor:
/cherry-pick release-ocm-2.11 release-ocm-2.10 release-ocm-2.9 release-ocm-2.8

PR #2490, opened 2024-07-14
[MGMT-18102] fix for platform none/external on RHEL9

eifrach:
1. improving port forward configuration ( socat)
2. adding  LB ports to firewall policy 
3. enabling forward for `libvirt-routed` zone
4. adding teardown config

### DIFF
--- a/scripts/deploy_assisted_service.sh
+++ b/scripts/deploy_assisted_service.sh
@@ -151,7 +151,7 @@ else
         # on RHEl9 we need to open ports in a new policy
         # between libvirt and HOST
         print_log "Opening additional ports for none/external"
-        firewall-cmd --policy=libvirt-to-host --add-port={22623/tcp,6443/tcp}
+        firewall-cmd --policy=libvirt-to-host --add-port={22623/tcp,22624/tcp,6443/tcp}
         firewall-cmd --policy=libvirt-to-host --add-service={http,https}
         firewall-cmd --zone=libvirt-routed  --add-forward
     fi
--- a/scripts/deploy_assisted_service.sh
+++ b/scripts/deploy_assisted_service.sh
@@ -45,6 +45,7 @@ export OPENSHIFT_CI=${OPENSHIFT_CI:-false}
 export ENABLE_SKIP_MCO_REBOOT=${ENABLE_SKIP_MCO_REBOOT:-true}
 export ENABLE_SOFT_TIMEOUTS=${ENABLE_SOFT_TIMEOUTS:-false}
 export IMAGES_FLAVOR=${IMAGES_FLAVOR:-}
+export PLATFORM=${PLATFORM:-}
 export ASSISTED_SERVICE_DATA_BASE_PATH="./assisted-service/data"
 export RELEASE_IMAGES_PATH="${ASSISTED_SERVICE_DATA_BASE_PATH}/default_release_images.json"
 export OS_IMAGES_PATH="${ASSISTED_SERVICE_DATA_BASE_PATH}/default_os_images.json"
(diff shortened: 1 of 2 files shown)


scripts/utils.sh
```diff
@@ -134,11 +127,23 @@ function close_external_ports() {
     ports=$1
     for p in $ports; do
         sudo firewall-cmd --zone=public --remove-port=$p/tcp
+        if [ ! -z ${PLATFORM} ]; then
```
danmanor, line 130: nit: why not ?
```
if [ ! -z "${PLATFORM}" ] && ([[ "${PLATFORM}" == "none" || "${PLATFORM}" == "external" ]]); then
    sudo firewall-cmd --policy=libvirt-to-host --remove-port=$p/tcp
fi
```

scripts/utils.sh
```diff
@@ -134,11 +127,23 @@ function close_external_ports() {
...
+                sudo firewall-cmd --policy=libvirt-to-host --remove-port=$p/tcp
+            fi
+        fi
     done
 }
 
 function add_firewalld_port() {
     port=$1
+
+    if [ ! -z ${PLATFORM} ]; then
```
danmanor, line 141: nit: why not ?
```
if [ ! -z "${PLATFORM}" ] && ([[ "${PLATFORM}" == "none" || "${PLATFORM}" == "external" ]]); then
    sudo firewall-cmd --policy=libvirt-to-host --remove-port=$p/tcp
fi
```

scripts/utils.sh
```diff
@@ -51,29 +50,24 @@ function spawn_port_forwarding_command() {
         fi
     fi
 
-    cat <<EOF | sudo tee "${socket_file}"
```
danmanor, line 54: We don't need to socket file now ?

eifrach: no it just a single systemd file with the socket config inside of it 
make the config and managing it more simple

scripts/utils.sh
eifrach: easier to read, it’s exactly the same

scripts/utils.sh
eifrach: easier to read, it’s exactly the same

scripts/utils.sh
eifrach: eitherway , is not going to work the file is configured with `nounset` so we have to set it

scripts/deploy_assisted_service.sh
```diff
@@ -146,6 +147,15 @@ EOF
 else
     print_log "Updating assisted_service params"
 
+    if [[ "${PLATFORM}" == "none"  || "${PLATFORM}" == "external" ]]; then
+        # on RHEl9 we need to open ports in a new policy
+        # between libvirt and HOST
+        print_log "Opening additional ports for none/external"
+        firewall-cmd --policy=libvirt-to-host --add-port={22623/tcp,6443/tcp}
```
adriengentil, line 154:
```suggestion
        firewall-cmd --policy=libvirt-to-host --add-port={22623/tcp,22624/tcp,6443/tcp}
```
could be worth t add the MCS/http port as well

adriengentil:
Do you understand what changed to RHEL 9, and why it worked with RHEL8/xinetd?

scripts/deploy_assisted_service.sh
danmanor: will it also work on rhel8 ? maybe add a condition to act only in case of rocky9/rhel9 ?

danmanor:
@eifrach holding for you to merge on your time

danmanor (approved):

danmanor:
unless @adriengentil has an objection

eifrach:
> Do you understand what changed to RHEL 9, and why it worked with RHEL8/xinetd?

yes, on the old version libvirt was using the IPTABLE rule to limit  / nat etc traffic between hosts / VM etc...
now it fully support firewalld. 
When creating  a route network usually there are no firewall rules -- seem like this policy has been created on RHEL9

not sure if RHEL or the terraform module creates it - we can investigate if it's important.

What is happening is the VM cannot reach the Load Balancer or any other resource on the host due to this policy blocking traffic 


here we see that the policy REJECT traffic from `librit-routed` zone to `HOST`
```console
>  firewall-cmd --info-policy=libvirt-to-host
libvirt-to-host (active)
  ...
  target: REJECT
  ingress-zones: libvirt-routed
  egress-zones: HOST
  ...
```

adriengentil:
Thanks! If we stop using iptables, I wonder if we can get rid of https://github.com/openshift/assisted-test-infra/blob/167a677d60babfd26cdee1903f7ef9b26d1a45ee/ansible_files/roles/common/setup_libvirtd/files/allow-cross-network-traffic.sh as well? (In another PR)

eifrach:
> Thanks! If we stop using iptables, I wonder if we can get rid of https://github.com/openshift/assisted-test-infra/blob/167a677d60babfd26cdee1903f7ef9b26d1a45ee/ansible_files/roles/common/setup_libvirtd/files/allow-cross-network-traffic.sh as well? (In another PR)

I think we should leave it for now.
Just to have backwards compatible with rhel8.

Bit we can create a future task to remove it.

PR #2501, opened 2024-07-20
iPXE assets download to container and modify iPxe script

bkopilov:
Curren iPXE script contains links redirected to https, PXE can not run the downloaded script directly from node.

Steps: (as descibied in UI)
Download the example iPXE script file
Download the assets in the script file to your iPXE server Inside the example script file, customize the asset URLs to your local server address

- Added support for libvirt undefine network
- Added new flag local_pxe_assets to keep support "other" without modifying the pxe content
- Download the ipxe , decode to utf-8 and download http links to the container directoy
- Modify the iPXE scrips link to download from container.

iPXE script should contain local address  http://192.168.XXX.1:8500/ for each image type  pxe-initrd, kernel, rootfs

### DIFF
--- a/src/assisted_test_infra/test_infra/controllers/ipxe_controller/ipxe_controller.py
+++ b/src/assisted_test_infra/test_infra/controllers/ipxe_controller/ipxe_controller.py
@@ -1,3 +1,4 @@
+import copy
 import os
 import shutil
 
@@ -14,10 +15,12 @@ class IPXEController(ContainerizedController):
         name: str = None,
         port: int = consts.DEFAULT_IPXE_SERVER_PORT,
         ip: str = consts.DEFAULT_IPXE_SERVER_IP,
+        local_pxe_assets: bool = False,
     ):
         super().__init__(name, port, name)
         self._ip = ip
         self._api_client = api_client
+        self._local_pxe_assets = local_pxe_assets
         self._dir = os.path.dirname(os.path.realpath(__file__))
         self._ipxe_scripts_folder = f"{self._dir}/server/ipxe_scripts"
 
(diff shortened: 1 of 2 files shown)


bkopilov:
@eliorerz  , @adriengentil , @mburman5

bkopilov:
@eifrach

src/assisted_test_infra/test_infra/controllers/ipxe_controller/ipxe_controller.py
```diff
@@ -37,9 +40,40 @@ def _build_server_image(self):
...
+
+        # New pxe content replace http links to local http server
+        new_pxe_content = copy.deepcopy(pxe_content)
+        new_asset = f"http://{self._ip}:{self._port}/"
+        http_to_download = [res for res in pxe_content.split() if "http:" in res]
+        for http in http_to_download:
+            http = http[http.index("http"):]  # in case http not at the beginning
+            for img in ["pxe-initrd", "kernel", "rootfs"]:
+                if img in http:
+                    utils.run_command(f"curl -L '{http}' -o {self._ipxe_scripts_folder}/{img}")
```
eifrach, line 74: please use `urllib.request`instead and also add try / except to vlidate the file has been downloaded

src/assisted_test_infra/test_infra/controllers/node_controllers/libvirt_controller.py
```diff
@@ -624,6 +631,7 @@ def set_ipxe_url(self, network_name: str, ipxe_url: str):
         bootp.setAttribute("file", ipxe_url)
         dhcp_element.appendChild(bootp)
         self.destroy_network(network)
+        self.undefine_network(network)
```
eifrach, line 634: where is this been used?

bkopilov: When we override the network config with "bootp file" , sometimes it did not work once creating new one.
The undefine removes it totally and we create new one.
The idea here is that we take existing network and adding bootp support (path to the iPXE) .

src/assisted_test_infra/test_infra/controllers/ipxe_controller/ipxe_controller.py
bkopilov: Ack , thanks.
Already implemented in utiil.download_file

src/assisted_test_infra/test_infra/controllers/node_controllers/libvirt_controller.py
bkopilov: if no objection i will mark it as resolved.

adriengentil (approved):

bkopilov:
@adriengentil  IS it related to my path ? the bond failure ?

adriengentil (approved):

PR #2504, opened 2024-07-24
    MGMT-18436: Enable override of assisted-service and assisted-image-service images in onprem deployments.

danmanor:
Enable override of assisted-service and assisted-image-service images in `onprem` deployments. This enables the CI to set this variables with the new built images in PRs


danmanor:
/override ci/prow/e2e-metal-assisted-kube-api-late-binding-single-node ci/prow/e2e-metal-assisted-kube-api-net-suite ci/prow/e2e-metal-assisted-kube-api-net-suite

adriengentil (approved):
/hold
Look good, unhold when we have checked that the job ran with the right images

danmanor:
We can see that the images which were used were CI builds 

image

danmanor:

PR #2513, opened 2024-08-08
OCI - add support for attach / detach disk

bkopilov:
Added volume_client for blockstorage to support create volume Added waiter before attaching volume to instance

implements attach_test_disk and detach_all_test_disks

It will allow us to add additional disk from api tests and test operators.

example from tests (API)
cluster.nodes.start_all()
cluster.wait_until_hosts_are_discovered()
cluster.nodes.controller.attach_test_disk(cluster.nodes[0].name, 60) cluster.nodes.controller.detach_all_test_disks(cluster.nodes[0].name)


bkopilov:
@adriengentil

bkopilov:
We are going to refactor OCI code  , not need this patch

PR #2534, opened 2024-09-18
MGMT-18618 Fix local rhsso deployment of assisted-installer

dudyas6:
This PR adds the option to pass the JWKS_URL environment variable to the assisted-service when deploying with rhsso authentication. Additionally, it includes a fix for a minor logic issue in the Makefile where we create_hub_cluster before cloning assisted-service repo locally.


src/update_assisted_service_cm.py
```diff
@@ -40,6 +40,7 @@
     ("HW_VALIDATOR_MIN_CPU_CORES_SNO", ""),
     ("HW_VALIDATOR_MIN_RAM_GIB_SNO", ""),
     ("ENABLE_UPGRADE_AGENT", "true"),
+    ("JWKS_URL", ""),
```
danmanor, line 43: This should be the default right ?
https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/certs

README.md
```diff
@@ -128,6 +128,7 @@ before you run start the deployment.
 | `OCM_BASE_URL`                          | OCM API URL used to communicate with OCM and AMS, default: https://api.integration.openshift.com/ |
 | `OCM_CLIENT_ID`                         | ID of Service Account used to communicate with OCM and AMS for Agent Auth and Authz |
 | `OCM_CLIENT_SECRET`                     | Password of Service Account used to communicate with OCM and AMS for Agent Auth and Authz |
+| `JWKS_URL`                              | URL for retrieving the JSON Web Key Set (JWKS) used for verifying JWT tokens in authentication. Defaults to https://api.openshift.com/.well-known/jwks.json |
```
danmanor, line 131: this is not the default right ?

danmanor (changes requested):

src/update_assisted_service_cm.py
dudyas6: It should be but i didn't want to change current configurations, only to add the option to insert it as env param, should change it to default?

README.md
dudyas6: It is currently, should we change it to be the working param? didnt want to change current configurations as mentioned above

danmanor: yes looks like it doesn't work

src/update_assisted_service_cm.py
danmanor: yes looks like it doesn't work

danmanor:
@dudyas6 ignore `ci/prow/e2e-metal-assisted-kube-api-late-binding-single-node`

danmanor:

danmanor (approved):

danmanor:

danmanor:
overriding e2e-metal-assisted (the current changes only affect the deployment, the failures are conformance tests)

PR #2539, opened 2024-09-28
Add enable redirect to console for classes inherit from BaseTest

bkopilov:
Current code does not support kernel_arguments cosole for node when running virsh console.

Added support for class decorator enabled by request on BaseTest childs The decorator check all test_ function and add parametrize kernel_arguments without change the functions signature.
Added hook to cluster fixture to enable kernel_arguments params.

### DIFF
--- a/src/assisted_test_infra/test_infra/utils/__init__.py
+++ b/src/assisted_test_infra/test_infra/utils/__init__.py
@@ -11,6 +11,7 @@ from .utils import (
     get_openshift_release_image,
     recreate_folder,
     run_command,
+    console_redirect_decorator,
 )
 
 __all__ = [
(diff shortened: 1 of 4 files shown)


bkopilov:
@eifrach , @adriengentil , @AlonaKaplan 
I am trying the fix on test_infra test , already tested on kni-test-infra.
It will allow us to enable console redirect on all class functions without code change.

eifrach:
hey @bkopilov thanks for the contribute,
I think that we shouldn’t pytest inside util’s - also we can just add a simple addition to the kernel args which I think will be much more easier to maintain and readable 

Here is an example of an API call
https://github.com/eifrach/ansible-nvidia-sno/blob/main/roles/assisted-sno/tasks/03_host_deployment.yml#L45-L59

which applies this manifest 
https://github.com/eifrach/ansible-nvidia-sno/blob/main/roles/assisted-sno/templates/99_master_kernel_arg.j2

bkopilov:
yes make sense

PR #2545, opened 2024-10-06
MGMT-18956: Set OCP 4.17 as default version

danmanor:


danmanor:

PR #2552, opened 2024-10-30
MGMT-18603: OCI Add wait_callback while applying stack configuration

bkopilov:
Added interval value to reduce the total log messages. The message in log will be:
counter and job status every minute

_apply_job_from_stack: 12 -> IN_PROGRESS

Ticket:  https://issues.redhat.com/browse/MGMT-18603


bkopilov:
@adriengentil

PR #2556, opened 2024-11-05
OCI: Add default tag_namespace_name for terraform variables.

bkopilov:
Sometimes during stack apply we hit on:
Error: 400-InvalidParameter, Failed to validate tags: TagNamespace openshift-test-infra-cluster-b1c19bff does not exists Suggestion: Please update the parameter(s) in the Terraform config as per error message Failed to validate tags: TagNamespace openshift-test-infra-cluster-b1c19bff does not exists

Set the param in terraform.

description = "Name of tag namespace to create or use for tagging OCI resources". Defaults to "openshift-{cluster_name}"


bkopilov:
@adriengentil

adriengentil (approved):

PR #2561, opened 2024-11-12
NO-ISSUE: better handle pull secret or file of pull secret

eifrach:


adriengentil:
looks good, can you add a word about this new env var in the README? https://github.com/openshift/assisted-test-infra?tab=readme-ov-file#prerequisites

eifrach:
> Looks good! can you add a word about this new env var in the README? https://github.com/openshift/assisted-test-infra?tab=readme-ov-file#prerequisites

you have it already - I just moved it to to the python code instead of the make file 
here
https://github.com/openshift/assisted-test-infra/blob/master/README.md?plain=1#L67-L68
and here
https://github.com/openshift/assisted-test-infra/blob/master/README.md?plain=1#L174

adriengentil (approved):

adriengentil:
> you have it already - I just moved it to the python code instead of the Makefile

I'm blind :sweat_smile:

eifrach:

PR #2569, opened 2024-11-18
NO-ISSUE: allow test infra to work with older assisted versions where the  field is missing 

eifrach:


eifrach:

PR #2577, opened 2024-11-25
NO-ISSUE: Use OCP release image payload in image override

danmanor:
Currently when `OPENSHIFT_INSTALL_RELEASE_IMAGE` is set, test-infra searches for an existing release image from the default list before creating one from the given reference. We want to use the reference to create release image in this case


danmanor:

PR #2582, opened 2024-11-28
MGMT-19248: Adjust OpenShift AI resources

jhernand:
This patch adjust the resources needed to run the _OpenShift AI_ operator. The previous values were too high, in particular the default resources for master nodes are enough, so there is no need to change them.

Note that in order to calculate these resources correctly it is also necessary to apply #2580.

Related: https://issues.redhat.com/browse/MGMT-19248
Related: https://issues.redhat.com/browse/MGMT-19056


eifrach:

PR #2583, opened 2024-12-01
MGMT-19336: Enable assisted-service fetching by specific commit reference

danmanor:
Enable assisted-service fetching by specific commit reference in order to match the python client version used by QE in integration / stage / prod environments to assisted-service image deployed there

### DIFF
--- a/scripts/bring_assisted_service.sh
+++ b/scripts/bring_assisted_service.sh
@@ -17,21 +17,25 @@ export OPENSHIFT_CI=${OPENSHIFT_CI:-false}
 export REPO_NAME=${REPO_NAME:-assisted-service}
 export JOB_TYPE=${JOB_TYPE:-}
 export PULL_BASE_REF=${PULL_BASE_REF:-master}
+export SERVICE_COMMIT_REF=${SERVICE_COMMIT_REF:-}
 
 
-if ! [[ -d "assisted-service" ]]; then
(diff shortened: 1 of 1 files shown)


danmanor:

scripts/bring_assisted_service.sh
```diff
@@ -64,7 +65,11 @@ if [[ "${OPENSHIFT_CI}" == "true" && "${REPO_NAME}" == "assisted-service" && "${
   git rebase -v "origin/${PULL_BASE_REF}"
 else
   cd assisted-service
-  retry -- git fetch --force origin "${SERVICE_BASE_REF}:FETCH_BASE" "${SERVICE_BRANCH}"
-  git reset --hard FETCH_HEAD
-  git rebase FETCH_BASE
+  if [ "${SERVICE_COMMIT_REF}" != "" ]; then
```
rccrdpccl, line 68: Can we exit early for each option? Maybe adding a comment of why each option is used?
Also I'd recommend not reusing just the `cd assisted-service` , but rather have each option on the same level for  clearity

danmanor: We can't exit early as it is called by other scripts as well

danmanor: Maybe I can wrap this in functions though

danmanor: Done

danmanor:

PR #2584, opened 2024-12-02
NO-ISSUE: Remove deletion of `control_plane_count` field during cluster creation

danmanor:
Remove deletion of `control_plane_count` field during cluster creation as it was replaced in https://github.com/openshift/assisted-test-infra/pull/2569 to conditionally adding this field

### DIFF
--- a/src/service_client/assisted_service_api.py
+++ b/src/service_client/assisted_service_api.py
@@ -2,7 +2,6 @@
 import base64
 import contextlib
 import enum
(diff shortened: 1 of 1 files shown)


danmanor:
/override  ci/prow/e2e-metal-assisted-kube-api-late-binding-single-node

adriengentil (approved):

danmanor:

PR #2587, opened 2024-12-03
NO-ISSUE: Add filtered controller logs when timeout during installation.

bkopilov:
During CI runs we got timeout failures without any root cause for failures. The default timeout is 3600 seconds.

We are going to increase total masters and operators installed and probably May hit on timeout and the current timers are not enough.

Added support for filtering last messages from assisted controller log when timeout occured. In case controller logs exists on timeout we extract last error/ warning and report them as part of the raise exception

We expect the see additional info when bubbled up timeout exception and get more details from the raised exception
.

### DIFF
--- a/src/assisted_test_infra/test_infra/helper_classes/cluster.py
+++ b/src/assisted_test_infra/test_infra/helper_classes/cluster.py
@@ -6,14 +6,18 @@ import time
 from collections import Counter
 from typing import List, Optional, Set
 
-import waiting
-import yaml
 from assisted_service_client import models
 from assisted_service_client.models.operator_type import OperatorType
+
+import consts
+
 from junit_report import JunitTestCase
 from netaddr import IPAddress, IPNetwork
 
-import consts
+import waiting
+
+import yaml
+
 from assisted_test_infra.test_infra import BaseClusterConfig, BaseInfraEnvConfig, ClusterName, exceptions, utils
 from assisted_test_infra.test_infra.controllers.load_balancer_controller import LoadBalancerController
 from assisted_test_infra.test_infra.helper_classes.base_cluster import BaseCluster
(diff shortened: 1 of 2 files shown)


bkopilov:
@adriengentil , @mburman5 , @AlonaKaplan , @eifrach

danmanor:
@bkopilov Do we have a way to reproduce the issue and check it works ?

bkopilov:
> @bkopilov Do we have a way to reproduce the issue and check it works ?

Yes , i changed the timeout for 10 minutes .

bkopilov:
> > @bkopilov Do we have a way to reproduce the issue and check it works ?
> 
> Yes , i changed the timeout for 10 minutes . We have flow -> on timeout exception or any other exception raised -> Try to get the assisted-controller.log :
> 
> * if no catch exception and return []
> * if yes extract / filter.  in case of error exception should be caught and set to []

Example from Run:

>           raise Exception(e.__class__.__name__, e.__dict__)
E           Exception: ('TimeoutExpired', {'_timeout_seconds': 60, '_what': "Monitored ['builtin'] operators to be in of the statuses ['available']", '__notes__': ["Failed to deploy the following operators ['console', 'cvo']"], 'filter_message': '[\'time="2024-12-04T10:36:39Z" level=error msg="Failed to get list of nodes from k8s client" func="github.com/openshift/assisted-installer/src/assisted_installer_controller.(*controller).waitAndUpdateNodesStatus" file="/remote-source/app/src/assisted_installer_controller/assisted_installer_controller.go:256" error="the server was unable to return a response in the time allotted, but may still be processing the request (get nodes)" request_id=a1b43260-4693-49f8-9c1f-ec8b1cd08044\\n\']'})

src/assisted_test_infra/test_infra/helper_classes/cluster.py:706: Exception

danmanor:
@bkopilov when the PR is ready, lets run the tests

bkopilov:
> @bkopilov when the PR is ready, lets run the tests

Its ready , i will start running full CI with it

bkopilov:
Hi ,  Example of  a test when the installation failed:
https://ci-jenkins-csb-kniqe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-assisted-installer-virt-tf/21297/

eifrach:
/retitle NO-ISSUE: Add filtered controller logs when timeout during installation.

PR #2590, opened 2024-12-09
NO-ISSUE: Set workers hostname in 'none' platform according to their network

danmanor:


terraform_files/none/main.tf
```diff
@@ -113,7 +113,7 @@ module "workers" {
   source = "../baremetal_host"
   count  = var.worker_count
 
-  name           = "${var.cluster_name}-worker-${count.index}"
+  name           = count.index % 2 == 0 ? "${var.cluster_name}-worker-${count.index}" : "${var.cluster_name}-worker-secondary-${count.index}"
```
eifrach, line 116: So if you have 2 workers the first worker will be worker-secondary-1 and the second will be worker-2. Why?

eifrach (changes requested):

terraform_files/none/main.tf
eifrach: btw this is the names from the CI run's 
```
"name": "test-infra-cluster-8b3106c1-worker-0",
"name": "test-infra-cluster-8b3106c1-worker-secondary-1",
```

danmanor: Yes, the idea is to label the worker based on its network for easier debugging. Notice that this is what we do in maters already

danmanor:
/override ci/prow/e2e-metal-assisted-kube-api-late-binding-sno-4-18

terraform_files/none/main.tf
eifrach: to this set the name if the host number is even or not.  
so I'm not sure if this is the result you wanted here

danmanor: Yes, to split the network hosts half-half. I just copied the code from masters

PR #2599, opened 2024-12-19
NO-ISSUE: improve cleanup script

eifrach:


src/cleanup.py
```diff
@@ -0,0 +1,16 @@
+import pathlib
+import shutil
+
+REMOVE_FILES = ["*.py[co]", "test_infra.log"]
+REMOVE_FOLDERS = ["__pycache__", "UNKNOWN.egg-info", "build", "assisted-service", "reports", ".mypy_cache"]
```
danmanor, line 5: I don't think we should remove assisted-service directory, we will have to clone it every time

eifrach: when you do clean, don't you think we should clean everything?

danmanor: We have logic that sets that just resets to master if the repository already exists, so I think i is redundant. I think whenever we can we should avoid cloning

eifrach: changed

danmanor (approved):

PR #2601, opened 2024-12-22
Fix collect log raise exception when network does not exist

bkopilov:
In case we have a failure in a test before the concrete resource created, We may return another failure in teardown when network does not exist

Cached exception in case the libvirt network not created.


bkopilov:
@adriengentil  , @eifrach

danmanor (approved):

PR #2602, opened 2024-12-24
MGMT-19509: Enable automatic testing of user-managed load balancer in a k8s api flow

danmanor:
Enable testing k8s API flow for - https://github.com/openshift/assisted-service/pull/7096


danmanor:
overriding due to network issues with the conformance tests

PR #2605, opened 2025-01-09
NO-ISSUE: Add cluster name to OCI bucket name

eliorerz:
- Set OCI bucket name to include cluster name
- Upload terraform output to the generated bucket

/cc @talhil-rh

### DIFF
--- a/src/assisted_test_infra/test_infra/controllers/node_controllers/oci_api_controller.py
+++ b/src/assisted_test_infra/test_infra/controllers/node_controllers/oci_api_controller.py
@@ -150,8 +150,16 @@ class OciApiController(NodeController):
         # Need the namespace and bucket name
         return namespace
 
-    def _upload_iso_to_bucket(self, file_path: str, namespace: str, bucket_name: str) -> None:
-        log.info(f"Upload iso file to bucket object storage {file_path}")
+    def _upload_data_to_bucket(self, data: str, filename: str, namespace: str, bucket_name: str):
+        file_path = f"/tmp/{filename}"
+
+        with os.path.open(file_path, "w") as f:
+            f.write(data)
+
+        return self._upload_file_to_bucket(file_path, namespace, bucket_name)
+
+    def _upload_file_to_bucket(self, file_path: str, namespace: str, bucket_name: str):
+        log.info(f"Upload file to bucket object storage {file_path}")
         if os.path.isfile(file_path):
             try:
                 self._object_storage_client.put_object(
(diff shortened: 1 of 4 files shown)


src/assisted_test_infra/test_infra/controllers/node_controllers/oci_api_controller.py
```diff
@@ -460,9 +468,16 @@ def prepare_nodes(self) -> None:
...
+            f"stack-{self._entity_config.cluster_id}",
+            namespace,
+            bucket_name,
+            self._config.oci_infrastructure_zip_file,
+            terraform_variables,
         )
         terraform_output = self._apply_job_from_stack(stack_id, random_name("job-"))
+        self._upload_data_to_bucket(
+            terraform_output, f"terraform-output-{self._entity_config.cluster_id}.yaml", namespace, bucket_name
+        )
```
adriengentil, line 480: why? the bucket will be cleaned up after the test.
The terraform output is available in the OCI stack, so it's can be extracted from there directly using the API.

src/assisted_test_infra/test_infra/controllers/node_controllers/oci_api_controller.py
```diff
@@ -460,9 +468,16 @@ def prepare_nodes(self) -> None:
...
         )
         stack_id = self._create_stack(
-            random_name("stack-"), namespace, bucket_name, self._config.oci_infrastructure_zip_file, terraform_variables
+            f"stack-{self._entity_config.cluster_id}",
+            namespace,
+            bucket_name,
+            self._config.oci_infrastructure_zip_file,
+            terraform_variables,
         )
         terraform_output = self._apply_job_from_stack(stack_id, random_name("job-"))
```
adriengentil, line 477:
```suggestion
        terraform_output = self._apply_job_from_stack(stack_id, "job-{self._entity_config.cluster_id}")
```
as we applied it only once, it should be enough.

eliorerz: For my understanding QE has sometimes more then one job per cluster, that is why left it as it is. 
@talhil-rh @dudyas6 Can you reffer to that?

src/assisted_test_infra/test_infra/controllers/node_controllers/oci_api_controller.py
eliorerz: QE n some cases are deploying the nodes and leave all the resources up

src/assisted_test_infra/test_infra/controllers/node_controllers/oci_api_controller.py
adriengentil: the job is attached to the stack

src/assisted_test_infra/test_infra/controllers/node_controllers/oci_api_controller.py
talhil-rh: We would like to use it with Cypress flow. In this case, we need the ability to read the yaml from outside test-infra. 
/cc @dudyas6

bkopilov: About the resource that not cleanup: we store a stack for it and it must run at the reverse order .
In case we abort test-infra . it would stay there.

in othercase i have a bug with cleanup , did we try this flow on OCI ?

adriengentil: I think there are several issues described in this discussion. My initial comment was discussing the necessity to store the terraform output in a bucket while it's available using the stack API.

bkopilov: Yes , we may return from stack creation.
_apply_job_from_stack--->

 ```
# on success, we return the jobs output - list
        items = self._resource_manager_client.list_job_outputs(job.data.id).data.items
        for item in items:
            if item.output_name == "oci_ccm_config":
                return item.output_value
```

Lets return here all items back to the caller  from oracle terrafrm . we can re-use it .
What do you think ?

adriengentil: yes, that would be an option

PR #2573, opened 2024-11-22
MGMT-19450: Update oci e2e test

adriengentil:
- refactor OCI ansible  playbooks and terraform templates to handle new OCI workflow
- remove old OCI workflow, the infrastructure and custom manifests are now provided by Oracle
- improve OciApiController to return the roles of the nodes, and set custom manifests dynamically

PR #2609, opened 2025-01-13
NO-ISSUE: Adjust OCI to work with UI flow

dudyas6:
Cannot push to https://github.com/openshift/assisted-test-infra/pull/2605 so created new PR

Work process of UI : 
use **make deploy_nodes_oci** to deploy oci resources and leave them up, once ui flow finishes run **make destroy_nodes_oci** which finds all the used resources and removes them


src/tests/test_targets.py
```diff
@@ -21,6 +21,15 @@ class TestMakefileTargets(BaseTest):
     def test_target_deploy_nodes(self, cluster):
         cluster.prepare_nodes()
 
+    @JunitTestSuite()
```
bkopilov, line 24: @eliorerz , @adriengentil So we assume that we may have external proccess (run the code of prepare_nodes from cypress or others) , we probably maintain the cleanup too (assume case when its not our local terraform)

I was thinking about "yield locked_file"  waiter in such case.
Example:
 def test_target_deploy_nodes(self, cluster):
        cluster.prepare_nodes()
        yield external_locked_file_waiter --> here we set 0 or 1 on /tmp/file.
        
The code begins to run , we set in file /tmp/test_bla_bla > 1
The waiter keep running , we yield with a timeout ( 2 hours)

In case the proccess ( cypress ) done and its set the value of     /tmp/test_bla_bla=0  the waiter continue the test with rest of the flow ... cleanup.

With this behaviour we test-infra keep the cleanup if user need and prevent calling destroy ( assume that we use remote infrastructure to bring up / down the resources.

What do you think ?

eliorerz:
@dudyas6 @bkopilov 
Lets keep using the first [PR](https://github.com/openshift/assisted-test-infra/pull/2605).
If there are things that we need to change or add please write your comments there.
Thanks

src/tests/test_targets.py
adriengentil: does this lock file logic has to be in test-infra? We could imagine a wrapper cli that does something like:
```
make deploy_nodes
touch /tmp/lock
while /tmp/lock exists
do 
  sleep 1
done 
make destroy_nodes
```

src/assisted_test_infra/test_infra/controllers/node_controllers/oci_api_controller.py
```diff
@@ -432,10 +452,69 @@ def get_ingress_and_api_vips(self) -> dict:
...
+            self._generate_cleanup_resources()
         for resource in self._cleanup_resources[::-1]:
-            resource()
+            try:
+                resource()
+            except Exception as e:
+                log.error(f"Error during cleanup resource execution: {e}")
         pass
 
+    def _generate_cleanup_resources(self, timeout_seconds: int = 2800) -> None:
```
adriengentil, line 464: @dudyas6 I'll pull your destroy logic in https://github.com/openshift/assisted-test-infra/pull/2573 as we need it for our CI running in Prow

src/tests/test_targets.py
bkopilov: IMO if we give to test-infra full controll on create and destroy from API -> yes
Unless we store it because when deploy_nodes test run it close the python proccess.
if we keep it open with a timer (waiter) and we rely on external file , we may re-use the python resources

PR #2624, opened 2025-01-29
NO-ISSUE: Adjust OCI to work with UI flow

dudyas6:
Re-opening https://github.com/openshift/assisted-test-infra/pull/2609 after discussing with Elior

Work process of UI :
use make deploy_nodes_oci to deploy oci resources and leave them up, once ui flow finishes run make destroy_nodes_oci which finds all the used resources and removes them


src/assisted_test_infra/test_infra/controllers/node_controllers/oci_api_controller.py
```diff
@@ -444,25 +519,34 @@ def setup_time(self) -> str:
...
+            f"stack-{self._entity_config.cluster_id}",
+            namespace,
+            bucket_name,
+            self._config.oci_infrastructure_zip_file,
+            terraform_variables,
+        )
+        terraform_output = self._apply_job_from_stack(stack_id, f"apply-job-{self._entity_config.cluster_id}")
+        self._upload_data_to_bucket(
+            terraform_output, f"terraform-output-{self._entity_config.cluster_id}.yml", namespace, bucket_name
         )
```
adriengentil, line 549: already tackled here: https://github.com/openshift/assisted-test-infra/pull/2605#discussion_r1909994140, can it be removed from this PR?

dudyas6: I actually need it to upload the manifests in the UI flow instead of fetching it from the stack in someway, so i temporarily save it in /tmp folder, i can skip the upload to bucket part, but i'll still need to save it somewhere unless u have other suggestion

adriengentil: well, we need to choose in which PR we want to talk about it, it's confusing at the moment

dudyas6: I remove the upload data logic, and am writing the data manually after apply job is done, is this ok?

adriengentil: Can you close #2605?

It ok for me now, it's very important to understand this kind of changes because uploading the file to the bucket may have a lot of sense of you because you will use it in your test framework, but outside of it, it doesn't make much sense as it not re-used inside test-infra itself. Because of that, it makes your change fragile, someone else working on these files can break or remove the upload without noticing that your need it for you specific use case.

src/assisted_test_infra/test_infra/controllers/node_controllers/oci_api_controller.py
```diff
@@ -348,9 +356,14 @@ def _apply_job_from_stack(
         # on success, we return the jobs output - list
         items = self._resource_manager_client.list_job_outputs(job.data.id).data.items
         for item in items:
-            if item.output_name == "oci_ccm_config":
+            if item.output_name == "dynamic_custom_manifest":
+                file_path = (
+                    f"/tmp/oci-{self._entity_config.cluster_id}/terraform-output-{self._entity_config.cluster_id}.yml"
```
adriengentil, line 361: will it be ok it I upload the manifest directly to the AI in a next change: https://github.com/openshift/assisted-test-infra/pull/2573/files#diff-901cf347567b426cc7b9d878e54d1f44424c1b84cd94c5c444adb51fa6d0eebcR359 ?

adriengentil (approved):

PR #2625, opened 2025-01-29
Add nopin_timeout to iSCSI installation

bkopilov:
Increase the keepalive time between client to target solves issue related to multinode installation Solve issue related to libvirt slow performance

Did not receive response to NOPIN on CID: 0, failing connection


bkopilov:
@adriengentil  , @linoyaslan

adriengentil (approved):

PR #2627, opened 2025-01-30
Terraform output shown in logs as a single string without new lines.

bkopilov:
Print the terraform output to log.info which print new lines and its clear to read


danmanor (approved):

src/assisted_test_infra/test_infra/tools/terraform_utils.py
```diff
@@ -74,6 +74,10 @@ def apply(
         return_value, output, err = self.tf.apply(
             no_color=IsFlagged, refresh=refresh, input=False, skip_plan=True, capture_output=capture_output
         )
+        # print terraform output line by line without \n
+        if output and type(output) is str:
+            log.info(output)
```
adriengentil, line 79: better to store a file somewhere, I fear it can expose credentials publicly in Prow in case of vsphere or nutatix.

adriengentil: it it really the "terraform" output, or the output log of the resulting of `terraform apply`?

bkopilov: The terraform outut printed anyway but  with \n and shown in single line .
It always in test_infra.log . i just take the same output and print it with a way that translate the \n to new line in output. 

it was better to fix it in terraform code ... but ...

adriengentil: can't we fix it? Logs are already hard to read, printing twice the output would be less friendly?

bkopilov: The terrfaform package ,  inside runs the cmd for apply and return the output .

```
Terraform.apply: (self, dir_or_plan=None, input=False, skip_plan=False, no_color=IsFlagged, **kwargs)                                                                                     │
│ apply                                                                                                                                                                                     │
│ refer to https://terraform.io/docs/commands/apply.html                                                                                                                                    │
│ no-color is flagged by default                                                                                                                                                            │
│ :param no_color: disable color of stdout                                                                                                                                                  │
│ :param input: disable prompt for a missing variable                                                                                                                                       │
│ :param dir_or_plan: folder relative to working folder                                                                                                                                     │
│ :param skip_plan: force apply without plan (default: false)                                                                                                                               │
│ :param kwargs: same as kwags in method 'cmd'                                                                                                                                              │
│ :returns return_code, stdout, stderr   
```

bkopilov: The terraform package use subproccess module and run:

>  p = subprocess.Popen(cmds, stdout=stdout, stderr=stderr,
> 293                              cwd=working_folder, env=environ_vars)
> 294  
> 295         synchronous = kwargs.pop('synchronous', True)
> 296         if not synchronous:
> 297             return p, None, None
> 298  
> 299         out, err = p.communicate()
> 300         ret_code = p.returncode
> 301         log.debug('output: {o}'.format(o=out))
> 302  
> 303         if ret_code == 0:
> 304             self.read_state_file()
> 305         else:
> 306             log.warn('error: {e}'.format(e=err))
> 307  
> 308         self.temp_var_files.clean_up()
> 309         if capture_output is True:
> 310             out = out.decode('utf-8')
> 311             err = err.decode('utf-8')
> 312         else:
> 313             out = None
> 314             err = None
> 315  
> 316         if ret_code != 0 and raise_on_error:
> 317             raise TerraformCommandError(
> 318                 ret_code, ' '.join(cmds), out=out, err=err)
> 319  
> 320         return ret_code, out, err
>

bkopilov:
Not critical .... we can skip it

src/assisted_test_infra/test_infra/tools/terraform_utils.py
adriengentil: `log.debug('output: {o}'.format(o=out))`

we can just disable log debug for this package?

PR #2632, opened 2025-02-14
Mgmt-19419: replace HighAvailabilityMode with ControlPlaneCount - phase 1

andrej1991:
[MGMT-19419](https://issues.redhat.com//browse/MGMT-19419): it is the first phase of replacing HighAvailabilityMode with ControlPlaneCount. At the moment because assisted-service still uses high-availability-mode some references needs to be kept so the deployment would work.
 It depends on https://github.com/openshift/assisted-installer-agent/pull/911.


src/assisted_test_infra/test_infra/helper_classes/cluster.py
```diff
@@ -45,7 +45,7 @@ def __init__(
 
         super().__init__(api_client, config, infra_env_config, nodes)
 
-        self._high_availability_mode = config.high_availability_mode
+        self._control_plane_count = config.control_plane_count
```
adriengentil, line 48: no need for this line if you set the property below. Then you just need to refer to `self.control_plane_count`.

src/assisted_test_infra/test_infra/helper_classes/cluster.py
```diff
@@ -132,8 +136,8 @@ def _create(self) -> str:
         if self._config.registry_ca_path:
             extra_vars["registry_ca_path"] = self._config.registry_ca_path
 
-        if self.nodes.masters_count and self.nodes.masters_count > 3:
-            extra_vars["control_plane_count"] = self.nodes.masters_count
+#        if self.nodes.masters_count and self.nodes.masters_count > 3:
+#            extra_vars["control_plane_count"] = self.nodes.masters_count
```
adriengentil, line 140: please remove if it's not needed

src/tests/global_variables/env_variables_defaults.py
```diff
@@ -95,7 +95,8 @@ class _EnvVariables(DataPool, ABC):
         loader=lambda x: bool(strtobool(x)),
         default=env_defaults.DEFAULT_USER_MANAGED_NETWORKING,
     )
-    high_availability_mode: EnvVar = EnvVar(default=env_defaults.DEFAULT_HIGH_AVAILABILITY_MODE)
+    control_plane_count: EnvVar = EnvVar(default=env_defaults.DEFAULT_CONTROL_PLANE_COUNT)
+    high_availability_mode: EnvVar = EnvVar(default=consts.HighAvailabilityMode.FULL)
```
adriengentil, line 99: don't we want to remove `high_availability_mode`, since you seem to use `control_plane_count` everywhere now?

Makefile
```diff
@@ -23,7 +23,7 @@ SKIPPER_PARAMS ?= -i
 
 ASSISTED_SERVICE_HOST := $(or ${ASSISTED_SERVICE_HOST},$(shell hostname))
 
-# Openshift CI params
+# Openshift CI params
```
adriengentil, line 26:
```suggestion
# Openshift CI params
```
nit

adriengentil (approved):

PR #2637, opened 2025-02-20
Fix redfish validation for disk inserted

bkopilov:


bkopilov:
@dudyas6  , @adriengentil  , @danmanor

PR #2641, opened 2025-03-03
Update control plane count from trigger

bkopilov:
master_count is used to set the total masters, same value as control_plane_count created auto trigger and updated the "enum" like class.

### DIFF
--- a/src/consts/consts.py
+++ b/src/consts/consts.py
@@ -330,8 +330,10 @@ class KubeAPIPlatforms:
 
 
 class ControlPlaneCount:
-    THREE = 3
     ONE = 1
+    THREE = 3
+    FOUR = 4
+    FIVE = 5
 
 
 class HighAvailabilityMode:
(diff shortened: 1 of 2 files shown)


bkopilov:
@adriengentil  , @andrej1991  , @danmanor

danmanor:
@bkopilov Is there any functional gain from this change ?

bkopilov:
> @bkopilov Is there any functional gain from this change ?

We use master_count in our automation as for count nodes ,  the new change break our code because it was setting 3 by default  ( the hook you added on cluster.py was remove)  i just added it to triggers .

danmanor:
@bkopilov Did you test it locally ?

bkopilov:
> @bkopilov Did you test it ?

Yes on my setup which running 4, 5 masters.
It looks like master_count and control_plane_count are duplicate (behaviour)

bkopilov:
> @bkopilov Did you test it ?

yes on my setup with 4 and 5 masters .  its only make sure that master_count updates controll_count_

danmanor (approved):

danmanor (approved):

PR #2646, opened 2025-03-06
MGMT-19831: Let test-infra provide the machine network for vSphere umlb dynamically

danmanor:
Let test-infra find the common network for all hosts in the cluster which should contain the load balancer IP as well in order to avoid providing additional configuration for the load balancer cidr


linoyaslan (approved):

danmanor:
/override ci/prow/e2e-metal-assisted-4-18 ci/prow/e2e-metal-assisted-ha-kube-api-4-18 ci/prow/e2e-metal-sno-live-iso

PR #2647, opened 2025-03-07
MULTIARCH-5357: Add libvirt and install on s390x KVM a five node clus…

AmadeusPodvratnik:
This implementation enables remote libvirt access and introduces the installation of a KVM 5 node cluster on s390x.
There are some remarks for the remote libvirt on s390x:
- DHCP only support (MAC prefix must be specified)
- NAT and MACVTAP is supported
- DNS and a loadbalancer (e.g. haproxy) must be preconfigured (e.g. by having a bastion host on the KVM server or somewhere in the network).
- virt-install is being used so the image will be copied to the KVM host (ssh key need to be exchanged upfront with the KVM server).
- Due to the noautoconsole option the VMs will not be automatically restarted (first reboot), therefore an additional wait loop is added.
- multi-node, 3-node and SNO cluster are supported.

Test: Successfully tested with 4.17.8 and 4.19.0-ec.2-multi


README.md
```diff
@@ -108,7 +108,10 @@ before you run start the deployment.
 | `OPENSHIFT_VERSION`                  | The OCP version which will be supported by the deployed components. Should be in `x.y` format |
 | `OPENSHIFT_INSTALL_RELEASE_IMAGE`    | The OCP release image reference which will be supported by the deployed components. For example - `quay.io/openshift-release-dev/ocp-release:4.16.0-x86_64` |
 | `INSTALL_WORKING_DIR`                | The path to a working directory where files like iPXE scripts, boot artefacts, etc are strored. For example `/tmp` |
-| `LIBVIRT_URI`                        | The connection URI to remote libvirt. For example `qemu+ssh://user@192.23.236.34/system`. Default is `qemu:///system` | 
+| `MASCHINE_CIDR_IPV4`                 | The machine cidr for e.g. remote libvirt. Default is `192.168.127.0/24` | 
+| `MASCHINE_CIDR_IPV6`                 | The machine cidr for e.g. remote libvirt. Default is `1001:db9::/120` |
```
adriengentil, line 112:
```suggestion
| `MACHINE_CIDR_IPV4`                 | The machine cidr for e.g. remote libvirt. Default is `192.168.127.0/24` | 
| `MACHINE_CIDR_IPV6`                 | The machine cidr for e.g. remote libvirt. Default is `1001:db9::/120` | 
```

typo

src/assisted_test_infra/test_infra/controllers/node_controllers/kvm_s390x_controller.py
```diff
@@ -0,0 +1,148 @@
...
+class KVMs390xController(LibvirtController):
+    def __init__(self, config: BaseNodesConfig, entity_config: Union[BaseClusterConfig, BaseInfraEnvConfig]):
+        log.debug("KVMs390xController: --- Init --- ")
+        super().__init__(config, entity_config)
+        self._entity_name = self._entity_config.entity_name
+        self._default_variables = DefaultVariables()
+        
+    @property
+    def entity_name(self) -> BaseName:
+        return self._entity_name
```
adriengentil, line 34:
```suggestion
        self._default_variables = DefaultVariables()
        
    @property
    def entity_name(self) -> BaseName:
        return self._entity_config.entity_name
```

scripts/install_environment.sh
```diff
@@ -43,6 +43,11 @@ function config_additional_modules() {
     esac
 }
 
+function install_virt_install() {
+    echo "Installing virt-install for remote libvirt access..." 
+    sudo dnf install -y virt-install
```
adriengentil, line 48: I would move it below in `install_libvirt`. WDYT?

src/assisted_test_infra/test_infra/helper_classes/cluster.py
```diff
@@ -1435,3 +1454,6 @@ def validate_static_ip(self) -> None:
 
         del host_failure, config_host_network, current_host_network
         log.info("Static IP validation passed")
+
+    def get_infra_env(self):
+        return self._infra_env
```
adriengentil, line 1459:
```suggestion
    @property
    def infra_env(self):
        return self._infra_env
```
is more idiomatic

adriengentil (commented):
just did a quick review, mostly looks good, let's see if the jobs are passing

README.md
AmadeusPodvratnik: Thx ... good catch

adriengentil: the typo is still there, may be you missed a push?

src/consts/consts.py
```diff
@@ -314,6 +314,7 @@ class Platforms:
     NUTANIX = "nutanix"
     OCI = "oci"
     EXTERNAL = "external"
+    S390X = "s390x"
```
adriengentil, line 317: This object should match platfroms in the sense of OCP platform, so I think we shouldn't have s390x in this class.

src/tests/base_test.py
```diff
@@ -472,10 +473,13 @@ def prepare_infraenv_nodes(self, infraenv_nodes: Nodes, infra_env_configuration:
 
     @classmethod
     def _prepare_nodes_network(cls, prepared_nodes: Nodes, controller_configuration: BaseNodesConfig) -> Nodes:
-        if (controller_configuration.tf_platform not in (consts.Platforms.BARE_METAL, consts.Platforms.NONE)) or (
-            hasattr(controller_configuration, "redfish_enabled") and controller_configuration.redfish_enabled
-        ):
+        log.info("--- _prepare_nodes_network\n")
+        if (
+            controller_configuration.tf_platform
+            not in (consts.Platforms.BARE_METAL, consts.Platforms.NONE, consts.Platforms.S390X)
```
adriengentil, line 479: `CPUArchitecture` should be checked instead of `consts.Platforms.S390X`

src/consts/consts.py
AmadeusPodvratnik: Removed from consts.py

src/tests/base_test.py
AmadeusPodvratnik: Check done by using cpu_arch out env config.

Dockerfile.assisted-test-infra
```diff
@@ -34,6 +34,9 @@ RUN dnf -y install --enablerepo=crb \
   iptables-nft \
    && dnf clean all
 
+RUN dnf -y install virt-install \
```
adriengentil, line 37: can you club it with the block above? So we avoid to create a new layer

AmadeusPodvratnik: Done.

PR #2650, opened 2025-03-16
Fix ip static validations.

bkopilov:
static configuration should support case when no address assigned.

example:
interfaces:
- ipv4: auto-dns: false auto-gateway: false auto-routes: false dhcp: true  -----------> we can set it to false without ip for testing enabled: true

In case we dont have any address assigned format_host_from_config_mapping_file returns empty dictionary and validation fails.

### DIFF
--- a/src/assisted_test_infra/test_infra/helper_classes/cluster.py
+++ b/src/assisted_test_infra/test_infra/helper_classes/cluster.py
@@ -1380,9 +1380,7 @@ class Cluster(BaseCluster):
                         for address_version in consts.IP_VERSIONS.keys():
                             if address_version not in current_interface.keys():
                                 continue
-                            address = current_interface[address_version]["address"]
-                            if len(address) == 0:
-                                continue
+                            address = current_interface[address_version].get("address", [])
                             host_network[expected_interface_mac] = {
                                 consts.IP_VERSIONS[address_version]: [
                                     f'{item["ip"]}/{item["prefix-length"]}' for item in address


bkopilov:
@eifrach , @adriengentil  , @danmanor

src/assisted_test_infra/test_infra/helper_classes/cluster.py
```diff
@@ -1380,9 +1380,7 @@ def format_host_from_config_mapping_file(config: str) -> dict[str, dict[str, Lis
                         for address_version in consts.IP_VERSIONS.keys():
                             if address_version not in current_interface.keys():
                                 continue
-                            address = current_interface[address_version]["address"]
-                            if len(address) == 0:
-                                continue
```
eifrach, line 1385: this is made for empty IP address in interface
make sure you test all possible deployment static IPv6 / bonding / static IPv4

bkopilov: Yes , we have configuration when the ip is empty , like static configuration with dhcp

eifrach: the bonding and IPv6 should also run on this PR - just to make sure
but, as far as I see `lgtm`

bkopilov: yes , in kni-test-infra we have bond test and it works with this change

PR #2668, opened 2025-04-07
MGMT-20334: Fix pull-secret censor

danmanor:
`instal-config.yaml` is `json` now, therefore regex doesn't catch

### DIFF
--- a/src/assisted_test_infra/download_logs/download_logs.py
+++ b/src/assisted_test_infra/download_logs/download_logs.py
@@ -3,7 +3,6 @@
 import filecmp
 import json
 import os
-import re
 import shutil
 import tarfile
 import tempfile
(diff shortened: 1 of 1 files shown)


danmanor:
/override ci/prow/e2e-metal-assisted-ha-kube-api-4-18 ci/prow/e2e-metal-sno-live-iso

src/assisted_test_infra/download_logs/download_logs.py
```diff
@@ -205,10 +205,16 @@ def download_logs(
         with SuppressAndLog(assisted_service_client.rest.ApiException, KeyboardInterrupt):
             install_config = Path(output_folder) / "cluster_files" / "install-config.yaml"
             client.download_and_save_file(cluster["id"], "install-config.yaml", str(install_config))
```
eliorerz, line 207: Don't we want to change `install-config.yaml` to `install-config.json`

danmanor: Lets handle it in assisted no ?

eliorerz: I think we should change both but we can approve this for now

eliorerz (approved):
LGTM

PR #2675, opened 2025-04-29
Add masters_count to oci api

bkopilov:
Added support to change masters count for 1,3,4,5 masters run on oracle.


adriengentil (approved):

adriengentil:
/test e2e-oci-assisted-4-19
/test e2e-oci-assisted-bm-iscsi-4-19

feel free to unhold when these 2 jobs are passing

PR #2680, opened 2025-05-11
add support for virt bundle operators

eliajahshan:
Adding support for virt bundle in test infra
will set up operators required for virt bundle

@bkopilov @danmanor @dudyas6

### DIFF
--- a/src/assisted_test_infra/test_infra/helper_classes/cluster.py
+++ b/src/assisted_test_infra/test_infra/helper_classes/cluster.py
@@ -358,6 +358,16 @@ class Cluster(BaseCluster):
         self._config.olm_operators = olm_operators
         self.api_client.update_cluster(self.id, {"olm_operators": olm_operators})
 
+    def set_virt_bundle(self):
+        log.info(f"Setting virtualization bundle for cluster: {self.id}")
+        self.set_olm_operator("mtv")
+        self.set_olm_operator("nmstate")
+        self.set_olm_operator("node-healthcheck")
+        self.set_olm_operator("fence-agents-remediation")
+        self.set_olm_operator("kube-descheduler")
+        self.set_olm_operator("self-node-remediation")
+        self.set_olm_operator("node-maintenance")
+
     def set_host_roles(self, num_masters: int = None, num_workers: int = None, requested_roles=None):
         if requested_roles is None:
             requested_roles = Counter(


src/assisted_test_infra/test_infra/helper_classes/cluster.py
```diff
@@ -358,6 +358,16 @@ def set_olm_operator(self, operator_name, properties=None, update=False):
...
+    def set_virt_bundle(self):
+        log.info(f"Setting virtualization bundle for cluster: {self.id}")
+        self.set_olm_operator("mtv")
+        self.set_olm_operator("nmstate")
+        self.set_olm_operator("node-healthcheck")
+        self.set_olm_operator("fence-agents-remediation")
+        self.set_olm_operator("kube-descheduler")
+        self.set_olm_operator("self-node-remediation")
+        self.set_olm_operator("node-maintenance")
+
```
bkopilov, line 370: @eliajahshan  , yes we can use it and it will be managed here when something change...  i think you can use const.OperatorType as naming.

@danmanor  Is it ok ? we are going to switch to bundles

danmanor: Yes, have to tested it @eliajahshan ?

danmanor:
@eliajahshan @bkopilov I think it would be better if we will expand this solution to be more generic, then we could use it in our CI as well. For example I think we should add 3 new constants:

```
virt_bundle_operators = ["cnv", ...]
ai_amd_bundle = [...]
ao_nvidia_bundle = [..]
```

and add a new installation parameter `CLUSTER_BUNDLE ` that could be set to either `ai-amd`, `ai-nvidia` or `virtualization`, then we will be able to just parameterize a stadard test: `@pytest.mark.parametrize("cluster_bundle", [virtualization])`.

WDYT ?

bkopilov:
> @eliajahshan @bkopilov I think it would be better if we will expand this solution to be more generic, then we could use it in our CI as well. For example I think we should add 3 new constants:
> 
> ```
> virt_bundle_operators = ["cnv", ...]
> ai_amd_bundle = [...]
> ao_nvidia_bundle = [..]
> ```
> 
> and add a new installation parameter `CLUSTER_BUNDLE ` that could be set to either `ai-amd`, `ai-nvidia` or `virtualization`, then we will be able to just parameterize a stadard test: `@pytest.mark.parametrize("cluster_bundle", [virtualization])`.
> 
> WDYT ?

Sounds good to me

PR #2698, opened 2025-06-25
MGMT-20671: Add AWS heterogeneous playbooks

danmanor:
We are currently running jobs that install OCP heterogeneous cluster on Equinix (they are currently broken). Since Equinix metal will be EOL soon, this PR implements the same logic for AWS


ansible_files/roles/aws/prepare_inventory/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
+# tasks/main.yml
+
+- name: Load EC2 instances definitions from file
+  ansible.builtin.include_vars:
+    file: "{{ role_path }}/files/ec2_instances.yml"
```
adriengentil, line 5: I think it would be more flexible if passed as an extra vars when running `ansible-playbook` (`ansible-playbook -e @path/to/var/file.yml`)

ansible_files/roles/aws/prepare_inventory/files/ec2_instances.yml
```diff
@@ -0,0 +1,30 @@
+# files/ec2_instances.yml
```
adriengentil, line 1: `vars` is more appropriate for such files, but as I said below, I think it should be passed as an extra var

ansible_files/aws_hetrogeneous_prepare_infra_playbook.yml
```diff
@@ -1,28 +1,13 @@
...
-    - name: equinix_tf_state_file
-      prompt: Place where the Terrafom state file will be stored
-      private: false
-  vars:
-    terraform_equinix_workdir: "{{ [playbook_dir, '..', 'terraform_files', 'equinix-ci-machine'] | path_join | realpath }}"
-  roles:
-    - equinix/heterogeneous_create_infra # Create the required devices to test heterogeneous scenarios
-    - equinix/export_primary_device_connection_details # Export the connection details, required for the common steps in Prow
+  tasks:
+    - name: "Prepare inventory"
```
adriengentil, line 4: to me, it fine to have ofcir calls as part of a role. If we want to test locally, we can add a variable to disable that call (and rely on the ex2_instances.yml var file)

ansible_files/aws_hetrogeneous_prepare_infra_playbook.yml
```diff
@@ -1,28 +1,13 @@
...
+      ansible.builtin.include_role:
+        name: "aws/prepare_inventory"
```
adriengentil, line 6: if we use ofcir, we are agnostic to the cloud provider? So maybe we can remove references to aws?

ansible_files/roles/aws/export_primary_device_connection_details/tasks/main.yml
```diff
@@ -10,11 +10,6 @@
...
       mode: 0644
 
-  - name: Write fix uid file
-    ansible.builtin.copy:
-      src: fix-uid.sh
-      dest: "{{ shared_dir }}/fix-uid.sh"
-
   - name: Write Packet common configuration file
     ansible.builtin.copy:
       src: packet-conf.sh
```
adriengentil, line 15: we still have `packet-conf.sh`? Shouldn't we rename it at some point?

adriengentil:
I think it was a mistake in equinix playbook to mix provisionning and configuration, I think it's way better to split them in 2 playbooks:
- provisionning playbook runs on localhost and outputs an ansible inventory
- configuration playbook takes the previous inventory and use it to configure each hosts

ansible_files/roles/aws/prepare_inventory/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
...
+  ansible.builtin.shell: |
+    curl -s http://169.254.169.254/latest/meta-data/local-ipv4
+  delegate_to: "{{ item.name }}"
+  register: private_ip_results
+  loop: "{{ instances }}"
+  loop_control:
+    label: "{{ item.name }}"
+
+- name: Re-add each host with its private IP
+  ansible.builtin.add_host:
```
adriengentil, line 40: `set_fact` should work here :thinking:

danmanor:
/override ci/prow/e2e-metal-assisted-4-19 ci/prow/e2e-metal-assisted-ha-kube-api-ipv4-4-19 ci/prow/e2e-metal-sno-live-iso

danmanor:
/override ci/prow/e2e-metal-assisted-4-19 ci/prow/e2e-metal-assisted-ha-kube-api-ipv4-4-19 ci/prow/e2e-metal-sno-live-iso

adriengentil (commented):
few comments but looks good overall!

adriengentil (approved):

PR #2779, opened 2025-12-17
parse ISO URL from streaming tool_result; refactor to pass C901

eliajahshan:
- Summary
  - Extract the Discovery ISO URL from streaming SSE tool_result events and include it in the returned text.
  - Refactor `send_query` into small helpers to reduce McCabe complexity (fix flake8 C901) without changing behavior.

- What changed
  - `assisted-test-infra/src/chatbot_client.py`
    - Added handling for `event == "tool_result"` with `tool_name == "cluster_iso_download_url"`:
      - Parses nested JSON in `data.token.response` and appends the `url` to the response.
    - Split `send_query` into helpers (no behavior change):
      - `_maybe_refresh_access_token`, `_build_payload`
      - `_send_streaming_query`, `_send_non_streaming_query`
      - `_parse_sse_line`, `_accumulate_event`
      - `_update_conversation_id`, `_append_token`, `_append_tool_result`
    - Kept time-based token refresh and conversation ID handling.
    - Continued support for base URL (e.g., `.../v1`) via `_resolve_endpoint`.
    - Streaming behavior remains controlled by `streaming_enabled` (default True).

- Why
  - The assistant may avoid printing the ISO URL directly, but exposes it via a tool result during streaming. Parsing this event lets tests retrieve the URL reliably.
  - Refactoring reduces complexity and satisfies flake8 C901 while keeping the public API stable.

- Behavior and compatibility
  - Public API unchanged: `ChatbotClient.send_query(question: str) -> str`.
  - Existing callers continue to receive aggregated streaming text; now it will also include the ISO URL when available.
  - Supports both streaming and non-streaming endpoints; base `/v1` URLs are resolved automatically.

- Environment variables
  - `OFFLINE_TOKEN` (required)
  - `ASSISTED_CHAT_API_URL` (base e.g., `https://assisted-chat.api.openshift.com/v1` or full endpoint)
  - `CHATBOT_STREAMING_ENABLED` (optional; default True)

- Validation
  - For a cluster, ask for the Discovery ISO link; response should contain the pre-signed `.iso` URL.
  - Lint: `flake8 src/chatbot_client.py` (C901 resolved).
  - Non-streaming path still returns the single JSON `response`.

- Notes
  - ISO URL parsing is defensive; ignores malformed tool payloads.
  - No changes to logging or request headers.

### DIFF
--- a/src/chatbot_client.py
+++ b/src/chatbot_client.py
@@ -1,7 +1,7 @@
 import json
 import os
 from datetime import datetime, timedelta
-from typing import Optional
+from typing import Any, Dict, Optional
 
 import requests
 
@@ -16,6 +16,7 @@ class ChatbotClient:
         self.conversation_id: Optional[str] = None
         self.access_token: Optional[str] = None
         self.last_auth_time: Optional[datetime] = None
+        self._tool_calls: Dict[str, Any] = {}
         # Determine streaming mode:
         # 1) explicit arg overrides env
         # 2) env overrides default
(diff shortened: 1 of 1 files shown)


eranco74 (commented):
/lgtm

PR #2803, opened 2026-05-04
MGMT-21834: removing the get_cluster_supported_platforms api

andrej1991:
The generated assisted-service installer client no longer exposes get_cluster_supported_platforms.
The test code now implements get_supported_cluster_platforms(client: InventoryClient, cluster_id: str) instead of the old api function.
The new function in case of HA returns ['baremetal', 'none'] and in case of SNO it returns ['none'].

### DIFF
--- a/src/assisted_test_infra/test_infra/helper_classes/cluster.py
+++ b/src/assisted_test_infra/test_infra/helper_classes/cluster.py
@@ -1,4 +1,5 @@
 import base64
+import enum
 import json
 import random
 import re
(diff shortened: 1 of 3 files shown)


danmanor (changes requested):
`self.api_client.get_cluster_supported_platforms` was returning the list of supported platforms for a given cluster. I don't think the solution you are proposing achieves this goal. because the other endpoint you are using doesn't take into consideration all parameters, like whether the cluster is SNO or not. I would instead just add the check if the cluster is SNO or not, and accordingly return `baremetal` or `none`. this was all the `get_cluster_supported_platforms` was doing anyway

danmanor (approved):

